PR #105, opened 2025-08-30
use /dev/shm instead of /tmp dir for model storage

sii-xinglong:
spot TPU has lower disk quota but large memory, so we use shm for model storage


test/srt/test_qwen3_models.py
```diff
@@ -27,12 +27,10 @@ def setUpClass(cls):
                 "--skip-server-warmup",
                 "--random-seed",
                 "3",
-                "--mem-fraction-static",
-                "0.1",
```
JamesBrianD, line 31: The default mem-fraction-static is 0.88. Deleting this args here may cause out of memory

test/srt/test_qwen3_moe_models.py
```diff
@@ -27,12 +27,10 @@ def setUpClass(cls):
                 "--skip-server-warmup",
                 "--random-seed",
                 "3",
-                "--mem-fraction-static",
-                "0.1",
                 "--max-prefill-tokens",
```
JamesBrianD, line 32: same as above

test/srt/test_qwen3_models.py
sii-xinglong: it seem like higher mem-fraction-static will increase performance and not break ci test, maybe i should adjust for other number like 0.4?

JamesBrianD: It seems that the default memory fraction of 0.88 is ok in the qwen3 test case. For other test cases, we can try to remove the mem-fraction-static setting and use the default value to simplify the test. Perhaps we should consider lowering the default fraction a little to ensure robustness.

PR #159, opened 2025-09-10
Feat/fused flashattention with packing

aolemila:
## Command

```bash
# start server
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache  python3 -u -m sgl_jax.launch_server --model-path Qwen/Qwen3-8B --trust-remote-code  --dist-init-addr=0.0.0.0:10011 --nnodes=1  --tp-size=4 --device=tpu --random-seed=3 --node-rank=0 --mem-fraction-static=0.8 --chunked-prefill-size=2048 --download-dir=/tmp --dtype=bfloat16 --max-running-requests 256 --skip-server-warmup --page-size=128  --disable-radix-cache

# eval(evalscope==0.17.1)
evalscope eval  --model Qwen-8B --api-url http://127.0.0.1:30000/v1/chat/completions --api-key EMPTY --eval-type service --datasets gsm8k --eval-batch-size 64

# benchmark
python3 -m sgl_jax.bench_serving --backend sgl-jax --dataset-name random --num-prompts 16 --random-input 1024 --random-output 1024 --random-range-ratio 1 --warmup-requests 0 --max-concurrency=48
```

## Accuracy

```bash
+---------+-----------+-----------------+----------+-------+---------+---------+
| Model   | Dataset   | Metric          | Subset   |   Num |   Score | Cat.0   |
+=========+===========+=================+==========+=======+=========+=========+
| Qwen-8B | gsm8k     | AverageAccuracy | main     |  1319 |  0.9098 | default |
+---------+-----------+-----------------+----------+-------+---------+---------+
```

### DIFF
--- a/python/sgl_jax/srt/managers/tp_worker.py
+++ b/python/sgl_jax/srt/managers/tp_worker.py
@@ -191,7 +191,7 @@ class ModelWorker:
     def precompile_extend(self):
         start_time = time.perf_counter()
         logger.info(
-            f"[EXTEND] begin to precompile bs_paddings={self.precompile_bs_paddings[-1:]} token_paddings={self.precompile_token_paddings}"
+            f"[EXTEND] Begin to precompile bs_paddings={self.precompile_bs_paddings[-1:]} token_paddings={self.precompile_token_paddings}"
         )
 
         bs, _ = self.get_max_padded_size()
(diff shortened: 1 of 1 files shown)


python/sgl_jax/test/test_flashattention.py
```diff
@@ -372,6 +372,7 @@ def run_test(self, mode, lens, mode_args):
             forward_batch.seq_lens,
             page_table,
             forward_batch.attn_backend.forward_metadata.cu_q_lens,
+            # forward_batch.attn_backend.forward_metadata.cu_kv_lens,
```
JamesBrianD, line 375: maybe you can remove this line

PR #202, opened 2025-09-19
[Docs] add integrate_with_tunix.md

aolemila:


docs/features/integrate_with_tunix.md
```diff
@@ -0,0 +1,236 @@
...
+    engine=Engine(model_path= 'Qwen/Qwen-7B-Chat', trust_remote_code=True, dist_init_addr='0.0.0.0:10011', nnodes=1 , tp_size=4, device='tpu' ,random_seed=3, node_rank=0, mem_fraction_static=0.4, chunked_prefill_size=8192, download_dir='/tmp', dtype='bfloat16', precompile_bs_paddings = [64], max_running_requests = 64, skip_server_warmup=True, attention_backend='fa',precompile_token_paddings=[8192], page_size=64 ,log_requests=True, log_requests_level=3)
+    output=engine.generate(prompt=['您好', "hello"], sampling_params={"n",2, "temperature": 0.7}, return_logprob=True)
+    print(len(list(output)), output)
+```
+
+#### `vllm Sample` vs `sgl_jax Sample`
+
+##### Fields Discussion
+
+- `seed`: It is set by tunix sampler but is not used when sampling. Will it be used in the future?
```
wang2yn84, line 154: Yes we'd like to control the backend sampling.

docs/features/integrate_with_tunix.md
```diff
@@ -0,0 +1,236 @@
...
+- `presence_penalty` & `frequency_penalty`: Will they be used in the future?
```
wang2yn84, line 156: It's up to the use case. vllm supports these options.

docs/features/integrate_with_tunix.md
```diff
@@ -0,0 +1,236 @@
...
+- `logit_bias`: Will it be used in the future?
+- `logprobs`: Does it mean logprobs of `top_number+1` for every output position? 1 means it include output_id's logprob.
```
wang2yn84, line 158: In the current use case, we only need the logprobs for the chosen token id.

docs/features/integrate_with_tunix.md
```diff
@@ -0,0 +1,236 @@
...
+  > From vLLM: Number of log probabilities to return per output token. When set to
+    `None`, no probability is returned. If set to a non-`None` value, the
+    result includes the log probabilities of the specified number of most
+    likely tokens, as well as the chosen tokens. Note that the implementation
+    follows the OpenAI API: The API will always return the log probability of
+    the sampled token, so there may be up to `logprobs+1` elements in the
+    response. When set to -1, return all `vocab_size` log probabilities.
+
+   - Confusion1: According to the usage of logprobs in the PPO codes, it looks like we need to get the logprob of every token_id in prompt position rather than the top_num logprob.
+- `prompt_logprobs`: Does it mean logprobs of `top_number` for every prompt position?
```
wang2yn84, line 169: Yea that's some legacy code. We don't need the prompt_logprobs.

docs/features/integrate_with_tunix.md
```diff
@@ -0,0 +1,236 @@
...
+| allowed_token_ids            | None                          |                               | support, set in payload body:<br>- return_logprob = true<br>- token_ids_logprob = None |
+| extra_args                   | None                          |                               |lack, not to support because tunix does not use it |                               |
+| bad_words                    | None                          |                               |lack, not to support because tunix does not use it |                               |
+| _bad_words_token_ids         | None                          |                               |lack, not to support because tunix does not use it |                               |
+
+
+
+## Discussion
+
+D1. Will beam search be required in the future?
```
wang2yn84, line 237: For now, we don't need beam search for our use cases.

docs/features/integrate_with_tunix.md
aolemila: Could you describe how to control the sampling with seed?

Here is the [usage](https://github.com/sgl-project/sglang/blob/89971c4c3cb1d326e52e306734f8e9d117449a3c/python/sglang/srt/layers/sampler.py#L268) which is implemented in [issue](https://github.com/sgl-project/sglang/pull/10678/files) from sgl-project/sglang.

Is it satisfied your requirements?

docs/features/integrate_with_tunix.md
aolemila: Got it!

docs/features/integrate_with_tunix.md
aolemila: Got it!

docs/features/integrate_with_tunix.md
aolemila: Got it!

docs/features/integrate_with_tunix.md
aolemila: Got it!

docs/features/integrate_with_tunix.md
```diff
@@ -0,0 +1,236 @@
...
+) -> Union[Dict, Iterator[Dict]]:
+    pass
+```
+
+- generate output format
+
+  ```json
+  {
+      "text":"the output of generate", ## output seq
+      "output_token_logprobs":[[
```
aolemila: We decouple ability to return output_ids with return_logprob. What about to add a new field called `output_ids`?@pathfinder-pf    I will implement it.

```json
{
    "text": ,
    "output_ids": ,
}
```

docs/features/integrate_with_tunix.md
wang2yn84: Yes, that's exactly what we need. Thank you!

docs/features/integrate_with_tunix.md
wang2yn84: That's nice to have! The Tunix integration point will be easier!

docs/features/integrate_with_tunix.md
```diff
@@ -0,0 +1,236 @@
...
+        filtered_updates = nnx.state(updated_weights, filter_types) if hasattr(updated_weights, '__call__') else {}
+
+        # Merge filtered updates with current state
+        for key, value in filtered_updates.items():
+            if key in current_state:
+                current_state[key] = value
+
+        nnx.update(self._model, current_state)
+    else:
+        nnx.update(self._model, updated_weights)
```
wang2yn84, line 52: Tunix implementation of the training model might be different from SGLang's version. So the weight name mapping, or potentially transposing or other operations might be necessary. Really depends on the model implementation on both sides.

wang2yn84 (commented):
Thank you!

wang2yn84:
> > Thank you very much for the proposal! Generally look good to me. There is one thing I'd like to pointed out, we have 2 modes to run Tunix: Multi Controller Jax or Pathways. In order to comply with Pathways single process requirement, we need to run sglang backend in the main process. Can you describe what changes are necessary in the proposal?
> 
> @wang2yn84 Currently we have relatively little understanding of Pathways, and we're not quite clear on how it implements single controller + distributed computing, as well as how to use and integrate it. It would be great if there were some more detailed documentation available.

Yea completely understand! Can you try https://arxiv.org/abs/2203.12533? I believe this paper can give you a high level view of the system. Let me know if it helps.

docs/features/integrate_with_tunix.md
aolemila: Got it!

docs/features/integrate_with_tunix.md
aolemila: Got it!

docs/features/integrate_with_tunix.md
aolemila: Got it! Thanks for your suggestion. cc @JamesBrianD

aolemila: @wang2yn84 Currently I find out model(like llama3) implements methods like `to_hf_mappings, lora_to_hf_mappings`, etc. These methods are used to `update_params`. Different inference engines have different weight names, will Tunix follow this solution to implement every mapping for different framework every model? cc @JamesBrianD 

Another question: Not all models implement all methods. For example, llama3 does not implement `to_hf_hook_fns`, but `model.to_hf_hook_fns()` will be called when initializing `VllmSampler` in rl/rollout/vllm_rollout.py. This confuses me. Could you explain it for me?

```python3
    self._sampler = vllm_sampler.VllmSampler(
        tokenizer=tokenizer,
        config=vllm_sampler.VllmConfig(
            max_model_len=cache_config_or_size,
            mesh=mesh,
            model_version=model_version,
            hbm_utilization=hbm_utilization,
            init_with_random_weights=init_with_random_weights,
            tpu_backend_type=tpu_backend_type,
            mapping_config=vllm_sampler.MappingConfig(
                to_hf_mappings=model.to_hf_mappings(),
                to_hf_transpose_keys=model.to_hf_transpose_keys(),
                to_hf_hook_fns=model.to_hf_hook_fns(),
                lora_to_hf_mappings=model.lora_to_hf_mappings(),
                lora_config=lora_config,
            ),
        ),
    )
```

JamesBrianD: @wang2yn84 Thank you for your suggestion. We will consider adding key_mappings and transpose_keys (following the example of utils.transfer_state_with_mappings) to handle the mapping and transformation between states.

pathfinder-pf:
can you explain why needs single process metained in "In order to comply with Pathways single process requirement, we need to run sglang backend in the main process", as we all know ,sglang run with multi process, can sglang run using a single process ? @wang2yn84

PR #212, opened 2025-09-23
add get default_sampling_params

pathfinder-pf:


python/sgl_jax/srt/entrypoints/engine.py
```diff
@@ -159,6 +166,10 @@ async def async_generate(
         The arguments of this function is the same as `sglang/srt/managers/io_struct.py::GenerateReqInput`.
         Please refer to `GenerateReqInput` for the documentation.
         """
+
+        if sampling_params is None:
+            sampling_params = self.get_default_sampling_params()
```
aolemila, line 171: It looks like it repeats with line123-125.

python/sgl_jax/srt/entrypoints/engine.py
```diff
@@ -343,6 +354,43 @@ async def async_score(
...
+                "min_p",
+                "max_new_tokens",
+            ]
+            if any(p in self.default_sampling_params for p in available_params):
+                diff_sampling_param = {
+                    p: self.default_sampling_params.get(p)
+                    for p in available_params
+                    if self.default_sampling_params.get(p) is not None
+                }
+                self.default_sampling_params = diff_sampling_param
```
aolemila, line 388: compared with vLLM implementation, it lacks of else logic, like the following:

```python3
else:
    diff_sampling_param = {}
```

pathfinder-pf: fix

python/sgl_jax/srt/entrypoints/engine.py
pathfinder-pf: because these are in different func.

PR #218, opened 2025-09-24
feat: custom model load layers

Prayer3th:
* custom model load layers
* fix kv cache create buffer unbalance


python/sgl_jax/srt/server_args.py
```diff
@@ -288,6 +289,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
             '* "transformers" will use the Transformers model '
             "implementation.\n",
         )
+        parser.add_argument(
+            "--model-layer-nums",
+            type=int,
+            default=ServerArgs.model_layer_nums,
+            help="Number of model layers to load and use for inference. If not specified, uses the value from model config.",
+        )
```
JamesBrianD, line 297: Using None as the default value feels better

Prayer3th: use the variable default, so that when modifying the default value, only one place needs to be changed

PR #228, opened 2025-10-09
feat: Add dummy loader

chhzh123:
This PR adds a dummy model loader for initializing random weights, which is used to compare the results on TPUs and GPUs. It uses numpy to create random weights to avoid the PyTorch dependency.

Reference PyTorch implementation:
https://github.com/sgl-project/sglang/blob/e22b13c56988f290660e8e59058b9ea06fe070e2/python/sglang/srt/model_loader/weight_utils.py#L910

### DIFF
--- a/python/sgl_jax/srt/model_loader/loader.py
+++ b/python/sgl_jax/srt/model_loader/loader.py
@@ -107,7 +107,8 @@ class JAXModelLoader(BaseModelLoader):
             model = create_model(self.rng)
 
         rng_key = self.rng.default.key.value
-        model.load_weights(rng_key)
+        # FIXME: Better interface not to pass in model_config again
+        model.load_weights(model_config, rng_key)
 
         return model
 
--- a/python/sgl_jax/srt/models/llama.py
+++ b/python/sgl_jax/srt/models/llama.py
@@ -24,6 +24,7 @@ import jax.numpy as jnp
 from flax import nnx
 from transformers import LlamaConfig, PretrainedConfig
 
+from sgl_jax.srt.configs.model_config import ModelConfig
 from sgl_jax.srt.layers.embeddings import (
     Embed,
     ParallelLMHead,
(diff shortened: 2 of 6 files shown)


python/sgl_jax/srt/model_loader/loader.py
```diff
@@ -54,12 +56,6 @@ def __init__(
         self, load_config: LoadConfig, rngs: jax.Array, mesh: jax.sharding.Mesh
     ):
         super().__init__(load_config)
-        if load_config.load_format != LoadFormat.JAX:
```
Prayer3th, line 57: Thank you for your contribution! 
Currently, the model loader is not correctly using the `get_model` method in `python/sgl_jax/srt/model_loader/loader.py` for loader initialization. Therefore, validation for `JaxFormat` within the `JAXModelLoader` class remains necessary. 
However, I believe using the `get_model` method is the more standard approach. Could you help complete this refactoring? This would allow us to safely remove the JaxFormat validation within the `JAXModelLoader` class.

python/sgl_jax/srt/model_loader/loader.py
```diff
@@ -160,6 +156,116 @@ def _prepare_weights(
...
+            state = nnx.state(model)
+            pspecs = nnx.get_partition_spec(state)
+            sharded_state = jax.lax.with_sharding_constraint(state, pspecs)
+            nnx.update(model, sharded_state)
+            return model
+
+        with self.mesh:
+            model = create_model(self.rng)
+            # Assign random weights deterministically
+            self._initialize_dummy_weights(model)
```
Prayer3th, line 269: It appears that sharding is being handled twice within the `create_model` and `_initialize_dummy_weights` functions. Could we try deferring the application of sharding until after `self._initialize_dummy_weights(model)` in the `load_model` function?

python/sgl_jax/srt/model_loader/loader.py
chhzh123: Updated

python/sgl_jax/srt/model_loader/loader.py
chhzh123: This will be done in the next PR when all the weight initialization is deferred

PR #229, opened 2025-10-09
fix: unify the embed class dtype and better interfacer

Prayer3th:


python/sgl_jax/srt/layers/embeddings.py
```diff
@@ -109,26 +113,50 @@ def attend(self, query: jax.Array) -> jax.Array:
...
+    """
+
     def __init__(
         self,
         num_embeddings: int,
         features: int,
-        dtype: jnp.dtype = jnp.bfloat16,
+        dtype: Optional[jnp.dtype] = None,
         param_dtype: jnp.dtype = jnp.bfloat16,
+        promote_dtype: PromoteDtypeFn = dtypes.promote_dtype,
```
chhzh123, line 129: Do we really need these three different types? I see upstream only has the `dtype` argument.

PR #242, opened 2025-10-15
Deferred weight loading

### LINKED ISSUE #241: [Feature] Deferred Weight Initialization
### Checklist

### Motivation

The current NN module implementation only annotates the parameters with `kernel_axes` metadata attached to them. This requires a full initialization of the memory buffer, which is not scalable for large models like Grok, because (1) it incurs a long initialization time for creating random weights at the beginning; (2) it may cause OOM since it always needs to create an entire buffer first.

An example of the [`LinearBase`](https://github.com/sgl-project/sglang-jax/blob/main/python/sgl_jax/srt/layers/linear.py#L19) initialization method is shown below:

```python
        self.weight = nnx.Param(
            nnx.with_partitioning(nnx.initializers.normal(), kernel_axes)(
                rngs.params(), (input_size, output_size), params_dtype
            )
        )
```

To resolve this issue, we need to use something similar to `meta_device` in PyTorch that defers the initialization. In JAX, this is [`jax.ShapeDtypeStruct`](https://docs.jax.dev/en/latest/_autosummary/jax.ShapeDtypeStruct.html). 

Upstream SGL does not have this issue as the model implementation is already sharded with [`Column/RowParallelLinear`](https://github.com/sgl-project/sglang/blob/98923880bcf9de03a9ea965f2ed3e569edc71834/python/sglang/srt/layers/linear.py#L257).

This is a must-have and high-priority feature. I'll modify the code and create a PR.

### Related resources

_No response_

chhzh123:
This PR fixes #241 by creating a lazy_init function to prevent initializing random weights when creating the model at the very beginning.

### DIFF
--- a/test/srt/test_lazy_init.py
+++ b/test/srt/test_lazy_init.py
@@ -194,6 +194,32 @@ def run_lazy_demo(mesh: Mesh):
     print(f"Final Sharding Status: {lazy_model.weight.value.sharding}")
 
 
+def run_only_lazy_init(mesh: Mesh):
+    rngs = nnx.Rngs(3)
+    print("--- Lazy Init Only (no materialize) ---")
+    print_all_devices_memory("Before LazyOnly (current)")
+    mem_before = print_memory_usage("Before LazyOnly (peak)")
+
+    def create_lazy_only(r: nnx.Rngs):
+        # Create the model lazily; do NOT call materialize
+        return LinearBase(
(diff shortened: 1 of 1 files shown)


chhzh123:
cc @Kipsora

chhzh123:
Memory usage and latency comparison with and without lazy init:

Screenshot 2025-10-16 at 12 16 21 AM

PR #247, opened 2025-10-16
feat/integrate tunix v1

pathfinder-pf:


python/sgl_jax/srt/managers/tp_worker_overlap_thread.py
```diff
@@ -51,6 +52,7 @@ def __init__(
         # JAX handles device execution automatically, no need for explicit streams
         self.forward_thread = threading.Thread(
             target=self.forward_thread_func,
+            daemon=True if pathways_available() else False,
```
aolemila: replace pathways with a parameter, such like single_process=True|False

pathfinder-pf: i will use a server's args to do this

PR #255, opened 2025-10-20
support tie_word_embeddings

pathfinder-pf:
- serving launch 
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache  python3 -u -m sgl_jax.launch_server \
--model-path Qwen/Qwen3-8B \
--trust-remote-code  \
--dist-init-addr=0.0.0.0:10011 \
--nnodes=1  \
--tp-size=4 \
--device=tpu \
--random-seed=3 \
--node-rank=0 \
--mem-fraction-static=0.8 \
--chunked-prefill-size=2048 \
--download-dir=/tmp \
--dtype=bfloat16 \
--max-running-requests 256 \
--skip-server-warmup \
--page-size=128  \
--disable-radix-cache

- accuracy
evalscope eval  --model Qwen-8B --api-url http://127.0.0.1:30000/v1/chat/completions --api-key EMPTY --eval-type service --datasets gsm8k --eval-batch-size 64

image

- performance
python3 -m sgl_jax.bench_serving --backend sgl-jax --dataset-name random --num-prompts 48 --random-input 1024 --random-output 1024 --random-range-ratio 1 --warmup-requests 0 --max-concurrency=16
image

### DIFF
--- a/python/sgl_jax/srt/models/llama.py
+++ b/python/sgl_jax/srt/models/llama.py
@@ -383,8 +383,9 @@ class LlamaForCausalLM(nnx.Module):
         self.dtype = dtype
         logger.info("LlamaForCausalLM config dtype: %s", self.dtype)
         self.transformer = LlamaModel(config, dtype=self.dtype, rngs=rngs)
-        self.lm_head = ParallelLMHead(config.vocab_size, config.hidden_size, rngs=rngs)
-        self.logits_processor = LogitsProcessor(config.vocab_size, self.lm_head, self.mesh)
+        if not getattr(self.config, "tie_word_embeddings", True):
+            self.lm_head = ParallelLMHead(config.vocab_size, config.hidden_size, rngs=rngs)
+        self.logits_processor = LogitsProcessor(config.vocab_size, self.mesh)
 
     def load_weights(self, model_config: ModelConfig, rng_key: jax.Array):
         self.rng = nnx.Rngs(rng_key)
@@ -527,7 +528,12 @@ class LlamaForCausalLM(nnx.Module):
         hidden_states, layers_kv_fused, layers_callback_flag = self.transformer(
             forward_batch, token_to_kv_pool
         )
-        output = self.logits_processor(hidden_states, logits_metadata)
+        if not getattr(self.config, "tie_word_embeddings", True):
+            output = self.logits_processor(hidden_states, self.lm_head, logits_metadata)
+        else:
+            output = self.logits_processor(
+                hidden_states, self.transformer.embed_tokens, logits_metadata
+            )
         return output, layers_kv_fused, layers_callback_flag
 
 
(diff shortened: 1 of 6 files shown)


python/sgl_jax/srt/models/llama.py
```diff
@@ -527,7 +528,12 @@ def __call__(
         hidden_states, layers_kv_fused, layers_callback_flag = self.transformer(
             forward_batch, token_to_kv_pool
         )
-        output = self.logits_processor(hidden_states, logits_metadata)
+        if not getattr(self.config, "tie_word_embeddings", True):
+            output = self.logits_processor(hidden_states, self.lm_head, logits_metadata)
+        else:
+            output = self.logits_processor(
+                hidden_states, self.transformer.embed_tokens, logits_metadata
+            )
```
JamesBrianD, line 536: When tie_word_embeddings == True is set, does self.transformer.embed_tokens need to undergo matrix transposition?

python/sgl_jax/srt/models/qwen.py
```diff
@@ -410,7 +411,12 @@ def __call__(
         logits_metadata: LogitsMetadata,
     ):
         hidden_states, layers_kv_fused = self.transformer(forward_batch, token_to_kv_pool)
-        output = self.logits_processor(hidden_states, logits_metadata)
+        if not getattr(self.config, "tie_word_embeddings", True):
+            output = self.logits_processor(hidden_states, self.lm_head, logits_metadata)
+        else:
+            output = self.logits_processor(
+                hidden_states, self.transformer.embed_tokens, logits_metadata
+            )
```
JamesBrianD, line 419: Same as the above.

python/sgl_jax/srt/models/qwen2.py
```diff
@@ -466,7 +467,12 @@ def __call__(
         hidden_states, layers_kv_fused, layers_callback_flag = self.transformer(
             forward_batch, token_to_kv_pool
         )
-        output = self.logits_processor(hidden_states, logits_metadata)
+        if not getattr(self.config, "tie_word_embeddings", True):
+            output = self.logits_processor(hidden_states, self.lm_head, logits_metadata)
+        else:
+            output = self.logits_processor(
+                hidden_states, self.transformer.embed_tokens, logits_metadata
+            )
```
JamesBrianD, line 475: Same as the above.

python/sgl_jax/srt/models/llama.py
pathfinder-pf: i think no need to do，the logits_processor will do 
image

JamesBrianD: Got it. It seems that the shape of the embedding layer (embed) must be vocab_size * hidden_size, so transposition is required when calculating logits.

PR #258, opened 2025-10-20
Feat: Add fused moe

Kipsora:
This PR fix issue #244, which adds FusedMoE (i.e. tensor parallel version of MoE). FusedMoE is useful for large MoE layers (e.g. Grok-2 model). It also includes a unittest that verifies the correctness of the implementation.

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -516,18 +516,15 @@ class FusedMoE(nnx.Module):
         self,
         config,
         num_experts: int,
-        num_experts_per_tok: int,
         intermediate_dim: int = 2048,
         weight_dtype: jnp.dtype = jnp.bfloat16,
         dtype: jnp.dtype = jnp.bfloat16,
         layer_id: int = 0,
-        activation: str = "gelu",
         *,
         mesh: Mesh,
     ):
         self.config = config
         self.num_experts = num_experts
-        self.num_experts_per_tok = num_experts_per_tok
         self.intermediate_dim = intermediate_dim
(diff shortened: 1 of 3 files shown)


python/sgl_jax/srt/layers/moe.py
```diff
@@ -492,3 +492,109 @@ def _unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+        self.num_experts_per_tok = num_experts_per_tok
+        self.intermediate_dim = intermediate_dim
+        self.weight_dtype = weight_dtype
+        self.dtype = dtype
+        self.layer_id = layer_id
+
+        self.tp_size = mesh.shape.get("data") * mesh.shape.get("tensor")
+
+        self.mesh = mesh
+        self.activation = activation
```
jimoosciuc: activation is not used？

python/sgl_jax/srt/layers/moe.py
```diff
@@ -492,3 +492,109 @@ def _unpermute(self, intermediate, sorted_selected_experts, weights, batch_size,
...
+            nnx.with_partitioning(nnx.initializers.normal(), self.wi_sharding)(
+                rngs.params(),
+                (num_experts, config.hidden_size, intermediate_dim),
+                weight_dtype,
+            )
+        )
+
+        self.wo = nnx.Param(
+            nnx.with_partitioning(nnx.initializers.normal(), self.wo_sharding)(
+                rngs.params(),
```
jimoosciuc: For compatibility with eval_shape, rngs.params() cannot be used here and should be changed to jax.random.PRNGKey(0)

PR #283, opened 2025-10-26
running sgl-jax on gpu sucessfully and add a doc for running

### LINKED ISSUE #282: [Feature] supoort sgl-jax running on gpu successfully
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

supoort sgl-jax running on gpu successfully

### Related resources

_No response_

zkkython:
Resolves #282 
1. fix running error: inputs, min_ps = operands
as in sample.py below:
```
min_p_operands = (probs_sort, min_ps, False)
    probs_sort = lax.cond(
        need_min_p_sampling,
        _apply_min_p_filter,
        lambda operands: operands[0],  # No min_p filtering, just return probs_sort
        min_p_operands,
    )
```
operands needs 3 parameters
2. add gpu install
by default, the device is tpu, when we run pip install pip install -e "python[all]", it only install jax[tpu]; when the device is gpu, we can use the command pip install -e “python[gpu]" to support sgl-jax on gpu
3. add a doc for using sgl-jax on gpu for users

### DIFF
--- a/python/sgl_jax/srt/layers/sampler.py
+++ b/python/sgl_jax/srt/layers/sampler.py
@@ -314,11 +314,8 @@ def top_p_normalize_probs_jax(
 
 def _apply_min_p_filter(operands):
     """Apply min_p filtering when need_min_p_sampling=True"""
-    # Handle both 2-tuple and 3-tuple cases for backward compatibility
-    if len(operands) == 3:
-        inputs, min_ps, _ = operands  # Ignore the third parameter
-    else:
-        inputs, min_ps = operands
+    inputs, min_ps = operands
+
     if is_tpu_runtime():
         max_per_bs = jnp.max(inputs, axis=1)
         min_p_thresholds = max_per_bs * min_ps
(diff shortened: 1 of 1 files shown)


python/sgl_jax/srt/layers/sampler.py
```diff
@@ -314,7 +314,11 @@ def top_p_normalize_probs_jax(
 
 def _apply_min_p_filter(operands):
     """Apply min_p filtering when need_min_p_sampling=True"""
-    inputs, min_ps = operands
+    # Handle both 2-tuple and 3-tuple cases for backward compatibility
+    if len(operands) == 3:
+        inputs, min_ps, _ = operands  # Ignore the third parameter
+    else:
+        inputs, min_ps = operands
```
aolemila: Could you help to remove the False in the following codes?

https://github.com/sgl-project/sglang-jax/blob/d18baf64c36a151d8fddf9c181ac6c440e234e29/python/sgl_jax/srt/layers/sampler.py#L406

aolemila (approved):

PR #324, opened 2025-11-07
add return tensor flag if want tensor

pathfinder-pf:
## Motivation
we should return the tensor's value because the statistic information is not enough in some case.

## Modifications
add a parameter in the interface start_trace

## Test plan
pass,set save_tensor = true when call the interface start_trace
pass,set save_tensor = false when call the interface start_trace
## Checklist

### DIFF
--- a/python/sgl_jax/srt/entrypoints/http_server.py
+++ b/python/sgl_jax/srt/entrypoints/http_server.py
@@ -409,13 +409,16 @@ async def start_trace_async(obj: StartTraceReqInput | None = None):
             unique_suffix = random.randint(1000, 9999)
             output_file = f"debug_outputs/request_traces_{timestamp}_{unique_suffix}.jsonl"
 
-        precision_tracer.start_trace(req_num=obj.req_num, output_file=output_file)
+        precision_tracer.start_trace(
+            req_num=obj.req_num, output_file=output_file, save_tensor=obj.save_tensor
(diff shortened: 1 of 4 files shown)


python/sgl_jax/srt/server_args.py
```diff
@@ -714,6 +715,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
             help="Enable precision tracer for debugging tensor values. May have performance impact.",
         )
 
+        parser.add_argument(
+            "--enable-return-tensor",
```
jimoosciuc: Is it possible to put the parameters in the start_trace interface?

pathfinder-pf: i can't understand, how to do this

jimoosciuc: Similar to:
  curl -X POST http://localhost:30000/start_trace \
      -d '{"req_num": 2, "enable_return_tensor": true}'

jimoosciuc: Currently, _verbose_logging and req_num are both passed in this way, and PrecisionTracer is a global variable

pathfinder-pf: i have changed my code as your advice.

jimoosciuc (dismissed):
LGTM

PR #414, opened 2025-11-17
[BUG] Different Results on Used HBM

### LINKED ISSUE #367: [Bug] SGLang-Jax does support proxy device in pathways
raise `Invalid device: proxy` in pathways.

codes: https://github.com/sgl-project/sglang-jax/blob/669b56905ac46856238bb38899013c4287c19b11/python/sgl_jax/srt/utils/jax_utils.py#L78-L79

xiuhu17:
Add `pathways_hbm_used_mem()` function for`pathway` backends in

https://github.com/sgl-project/sglang-jax/blob/main/python/sgl_jax/srt/utils/jax_utils.py#L47

The implementation `pathways_hbm_used_mem` is copied from 

https://github.com/vllm-project/tpu-inference/blob/b832b02020bd42f8c5f9f437ec390812451d50b4/tpu_inference/utils.py#L60-L61

Since we do not have the access to PathWay at this stage. I tested with **tpu-v6e-4**. 

The scripts for reproducing the errors is:

https://github.com/google/tunix/blob/main/scripts/grpo_demo_sglang_jax_rollout.py

The used HBM from `dev.memory_stats()` and `pathways_hbm_usage_gb` are different as following:

```
hbm_used_mem[1606439424, 1606437376, 1606437376, 1606437376]
pathways_hbm_used_mem[1606374936, 1606374912, 1606374912, 1606374912]
hbm_used_mem[1606439424, 1606437376, 1606437376, 1606437376]
pathways_hbm_used_mem[3212749872, 3212749824, 3212749824, 3212749824]
hbm_used_mem[3213076992, 3213074944, 3213074944, 3213074944]
pathways_hbm_used_mem[4819387440, 4819387392, 4819387392, 4819387392]
hbm_used_mem[3213076992, 3213074944, 3213074944, 3213074944]
pathways_hbm_used_mem[6426025008, 6426024960, 6426024960, 6426024960]

```
Fixes #367

### DIFF
--- a/python/sgl_jax/srt/utils/jax_utils.py
+++ b/python/sgl_jax/srt/utils/jax_utils.py
@@ -114,7 +114,7 @@ def get_available_device_memory(device, distributed=False, empty_cache=True):
             stats = dev.memory_stats()
             avail_mem.append(stats["bytes_limit"] - stats["bytes_in_use"])
         avail_mem = jnp.array([min(avail_mem) / (1 << 10)], dtype=jnp.float32)
-    elif device in ("proxy"):
+    elif device == "proxy":
         live_arrays = jax.live_arrays()
         pathways_hbm_used_mem = pathways_hbm_usage_gb(live_arrays, devices)
         avail_mem = jnp.array(


python/sgl_jax/srt/utils/jax_utils.py
```diff
@@ -54,9 +111,17 @@ def get_available_device_memory(device, distributed=False, empty_cache=True):
         if empty_cache:
             jax.clear_caches()
         avail_mem = []
+        hbm_used_mem = []
```
aolemila: According to error(Invalid device: proxy), device in pathways is named as proxy. Please add this in line109.

aolemila:
> Is there a minimum step to reproduce and debug the issue of inconsistent report between the two times? We could print out all the live arrays to verify.

Ok. Please wait for a moment!

lc5211:
It turns out that the call of jax.live_arrays() will create an extra copy of new arrays that jax captured.
In your repro example, first call of jax.live_arrays() will return 1 array as that's what we've created but in the meantime, it will create another reference to this array so the second call of jax.live_arrays() gives you 2x of hbm usage (as it can see two jax Arrays now).

If you change the implementation to something like following, the issue can be bypassed:
```
live_arrays = jax.live_arrays()
devices = jax.devices()
def pathways_hbm_usage_gb(live_arrays, devices):
  for array in live_arrays:
    for buffer in array.addressable_shards:
      hbm_used[buffer.data.device] += buffer.data.nbytes
  return [(hbm_used[device]) for device in devices]
 pathways_hbm_usage = pathways_hbm_usage_gb(live_arrays, devices)
 # second call should give the same results.
 pathways_hbm_usage = pathways_hbm_usage_gb(live_arrays, devices)
```
In the real deployment, there seems over estimation on the hbm usage due to this issue. I'm following up internally on how to make it more precise.

aolemila:
> It turns out that the call of jax.live_arrays() will create an extra copy of new arrays that jax captured. In your repro example, first call of jax.live_arrays() will return 1 array as that's what we've created but in the meantime, it will create another reference to this array so the second call of jax.live_arrays() gives you 2x of hbm usage (as it can see two jax Arrays now).
> 
> If you change the implementation to something like following, the issue can be bypassed:
> 
> ```
> live_arrays = jax.live_arrays()
> devices = jax.devices()
> def pathways_hbm_usage_gb(live_arrays, devices):
>   for array in live_arrays:
>     for buffer in array.addressable_shards:
>       hbm_used[buffer.data.device] += buffer.data.nbytes
>   return [(hbm_used[device]) for device in devices]
>  pathways_hbm_usage = pathways_hbm_usage_gb(live_arrays, devices)
>  # second call should give the same results.
>  pathways_hbm_usage = pathways_hbm_usage_gb(live_arrays, devices)
> ```
> 
> In the real deployment, there seems over estimation on the hbm usage due to this issue. I'm following up internally on how to make it more precise.

Thanks for your reply, I have tested and gotten the same result after twice calls. Could you modify the codes according to the above suggestions? @xiuhu17 

```bash
hbm_used_mem[16.000070571899414]
pathways_hbm_used_mem[16.0]
hbm_used_mem[16.000030994415283]
pathways_hbm_used_mem[16.0]
```

xiuhu17:
Thanks for providing the solution!

python/sgl_jax/srt/utils/jax_utils.py
```diff
@@ -49,7 +105,7 @@ def get_available_device_memory(device, distributed=False, empty_cache=True):
     Get available memory for device:device_id.
     When distributed is True, the available memory is the minimum available memory of all devices.
     """
-    if device == "tpu":
+    if device in ("tpu", "proxy"):
```
aolemila: Hi, @xiuhu17  please remove proxy here because proxy does not support to use memory_stats.

aolemila: Please use device=='tpu'.

aolemila (approved):

aolemila (approved):

Issue #444, opened 2025-11-19
[Bug] SWA model "google/gemma-2-9b-it" can't run


SiqiLi-Fighting:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

<img width="1916" height="2106" alt="Image" src="https://github.com/user-attachments/assets/08e2fd82-a595-4005-b741-2d74da0d4914" />

it report this bug when I run ` uv run  python3 -u -m sgl_jax.launch_server --model-path google/gemma-2-9b-it --trust-remote-code  --tp-size=4 --device=tpu --mem-fraction-static=0.8 --max-running-requests 256 --skip-server-warmup --page-size=1 --context-length=4096 --swa-full-tokens-ratio 0.2`

after that, I update transformers==4.57.1, it passed above bug, but crashed when precompile

<img width="1842" height="1188" alt="Image" src="https://github.com/user-attachments/assets/72a9b563-6503-4336-96bf-e2de7fa1725b" />

### Reproduction

` uv run  python3 -u -m sgl_jax.launch_server --model-path google/gemma-2-9b-it --trust-remote-code  --tp-size=4 --device=tpu --mem-fraction-static=0.8 --max-running-requests 256 --skip-server-warmup --page-size=1 --context-length=4096 --swa-full-tokens-ratio 0.2`

### Environment

tpu-v6e-4

PR #446, opened 2025-11-19
[BUG]fix sharding bug

pengchengneo:
fix bug in issue https://github.com/sgl-project/sglang-jax/issues/444


python/sgl_jax/srt/models/qwen2_moe.py
JamesBrianD: need to sharding.

PR #481, opened 2025-11-24
[BUG]fix cpu profiling trace bug

pengchengneo:
image
before this pr, profiling trace is error when capature more num_steps

after add trace annotation

image

modify requires-python = ">=3.12,<3.14" to avoid uv sync error


Iamleos (approved):
LGTM

Issue #449, opened 2025-11-19
[Feature] Interruptible Sampling For SGLang-Jax: Code


aolemila:
### **Checklist**

- [ ] If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [ ] Please use English, otherwise it will be closed.

### **Motivation**

Support features which are required from Interruptible Sampling in SGLang-Jax.

### **Related resources**

PR #515, opened 2025-12-04
[FEAT] Add features to support Partial Rollout

xiuhu17:
## Benchmarking and Profiling

#449 

## Note

Based on the [batch issue](https://github.com/jax-ml/jax/issues/34080), in order to ensure the effect of retract mode for pause_generation request, we'd better to configure the launch arguments to ensure the prefill and decode forwards have the same batch.

### DIFF
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -785,7 +785,12 @@ class Scheduler(
         ret["chunked_req_is_none"] = self.chunked_req is None
 
         # request cache stat
-        ret["tree_cache_size"] = self.tree_cache.total_size() if self.tree_cache is not None else 0
+        if isinstance(self.tree_cache, ChunkCache):
+            ret["tree_cache_size"] = 0
+        else:
+            ret["tree_cache_size"] = (
+                self.tree_cache.total_size() if self.tree_cache is not None else 0
+            )
         if self.req_to_token_pool is not None:
             ret["req_to_token_pool_total"] = self.req_to_token_pool.size
             ret["req_to_token_pool_available"] = self.req_to_token_pool.available_size()


python/sgl_jax/srt/managers/io_struct.py
```diff
@@ -574,6 +574,49 @@ class VertexGenerateReqInput(GenerateReqInput):
...
+
+    retract: Pause the scheduler's event loop from performing inference;
+            only non-inference requests will be handled, and all currently running
+            requests will be retracted back to the waiting_queue.
+            Note: The KV cache can be flushed in this mode and will be automatically
+            recomputed after continue_generation.
+    """
+
+    mode: str = "abort"
+    request_id: str = ""
```
aolemila: Please use `rid` and keep the same style to the other xxxReqs, like AbortReq.

python/sgl_jax/srt/managers/io_struct.py
```diff
@@ -574,6 +574,49 @@ class VertexGenerateReqInput(GenerateReqInput):
...
+            in the running_batch.
+
+    retract: Pause the scheduler's event loop from performing inference;
+            only non-inference requests will be handled, and all currently running
+            requests will be retracted back to the waiting_queue.
+            Note: The KV cache can be flushed in this mode and will be automatically
+            recomputed after continue_generation.
+    """
+
+    mode: str = "abort"
```
aolemila: mode: Literal["abort", "retract", "in_place"] = "abort" ?

python/sgl_jax/srt/managers/tokenizer_manager.py
```diff
@@ -594,14 +596,23 @@ async def _execute_profile(self, req: ProfileReq):
...
+                # use len(self.rid_to_state) == 0 to ensure all requests are aborted
+                while True:
+                    self.abort_request(abort_all=True)
+                    if len(self.rid_to_state) == 0:
+                        break
+                    await asyncio.sleep(0.1)
+
+    async def continue_generation(self, obj: ContinueGenerationReqInput):
         async with self._cond:
             self._updating = False
```
aolemila: Please replace `self._updating` with `self.is_pause`, replace `self._cond` with `self.is_pause_cond`. The fields including stop help readers to understand them.

python/sgl_jax/srt/entrypoints/engine.py
```diff
@@ -254,20 +255,44 @@ def __exit__(self, exc_type, exc_value, traceback):
         return False
 
     def flush_cache(self):
-        loop = asyncio.get_event_loop()
-        return loop.run_until_complete(self.tokenizer_manager.flush_cache())
+        """
+        Descriptioin: requests will be sent to tokenizer manager. It will flush all cache: tree_cache(radix cache), req_to_token_pool, token_to_kv_pool_allocator(free physical cache through allocator)
+        """
```
aolemila, line 317: Radix cache is one of tree_cache, so please remove radix cache in annotation.

python/sgl_jax/srt/entrypoints/http_server.py
```diff
@@ -348,7 +350,11 @@ async def classify_request(obj: EmbeddingReqInput, request: Request):
 
 @app.api_route("/flush_cache", methods=["GET", "POST"])
 async def flush_cache():
-    """Flush the radix cache."""
+    """
+    Descriptioin: requests will be sent to tokenizer manager. It will flush all cache: tree_cache(radix cache), req_to_token_pool, token_to_kv_pool_allocator(free physical cache through allocator)
+
```
aolemila, line 355: remove radix cache.

python/sgl_jax/srt/entrypoints/http_server.py
```diff
@@ -635,19 +641,31 @@ async def separate_reasoning_request(obj: SeparateReasoningReqInput, request: Re
 
 
 @app.post("/pause_generation")
-async def pause_generation(request: Request):
-    """Pause generation."""
-    await _global_state.tokenizer_manager.pause_generation()
+async def pause_generation(obj: PauseGenerationReqInput, request: Request):
+    """
+    Description: pause in-flight generation, requests will be sent to tokenizer manager
```
aolemila: Description: Deal with requests according to mode. Now support abort, in_place and retract.

python/sgl_jax/srt/entrypoints/engine.py
```diff
@@ -254,20 +255,44 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        return self.loop.run_until_complete(self.tokenizer_manager.flush_cache())
+
+    def pause_generation(self, mode: str = "retract"):
+        """
+        Input: the pause generation mode: ["abort", "retract", "in-place"]
+
+        Description: pause in-flight generation, requests will be sent to tokenizer manager
```
aolemila: Description: Deal with requests according to mode. Now support abort, in_place and retract.

python/sgl_jax/srt/managers/io_struct.py
```diff
@@ -112,12 +126,58 @@ class TokenizedGenerateReqInput:
 
 
 @dataclass
-class AbortReq:
-    # The request id
-    rid: str = ""
+class AbortReq(BaseReq):
     # Whether to abort all requests
     abort_all: bool = False
```
aolemila, line 138: Please add `finished_reason` and `abort_message`. The former may be used by RL framework to check whether continue to generate or not.

python/sgl_jax/srt/managers/tokenizer_manager.py
```diff
@@ -641,6 +652,7 @@ async def close_session(
         await self.send_to_scheduler.send_pyobj(obj)
 
     async def get_internal_state(self) -> list[dict[Any, Any]]:
+        self.auto_create_handle_loop()
```
aolemila, line 678: Hi @xiuhu17 , why add this line here?

test/srt/test_engine_pause_continue.py
```diff
@@ -0,0 +1,343 @@
+"""
+Test pause_generation and continue_generation using Engine API directly.
+
+This test uses async/await pattern to avoid multi-threading event loop conflicts.
+
+Usage:
+python3 -m unittest test_engine_pause_continue.TestEnginePauseContinue
```
aolemila: Please add this test file and the following test file under "e2e-test-tpu-v6e-4" in run_suite.py

test/srt/test_engine_pause_continue.py
```diff
@@ -0,0 +1,343 @@
...
+        running_after_pause = internal_after["running_batch_size"]
+        waiting_after_pause = internal_after["waiting_queue_size"]
+        self.assertEqual(
+            running_after_pause, 0, "Running batch should be empty after retract mode pause"
+        )
+
+        # Verify requests are in waiting queue
+        self.assertGreater(
+            waiting_after_pause, 0, "Requests should be moved to waiting queue after retract"
+        )
```
aolemila, line 147: For waiting_after_pause, it is equal to num_requests?

test/srt/test_engine_pause_continue.py
```diff
@@ -0,0 +1,343 @@
...
+    def test_2_pause_generation_in_place_mode(self):
+        """Test pause_generation with in_place mode."""
+        self.engine.loop.run_until_complete(self._run_test_in_place_mode())
+        print_test_passed("TestEnginePauseContinue.test_2_pause_generation_in_place_mode")
+
+    def test_3_pause_continue_multiple_cycles(self):
+        """Test multiple pause/continue cycles."""
+        self.engine.loop.run_until_complete(self._run_test_multiple_cycles())
+        print_test_passed("TestEnginePauseContinue.test_3_pause_continue_multiple_cycles")
+
```
aolemila, line 410: Hi @xiuhu17  could you add abort mode case and mix this mode into multi cycles case? I mean two test cases.

test/srt/test_pause_continue_generation.py
```diff
@@ -0,0 +1,665 @@
+"""
```
aolemila: Please rename this file with `test_server_pause_continue.py` to distinguish with the engine.

test/srt/test_engine_pause_continue.py
```diff
@@ -0,0 +1,343 @@
...
+
+if __name__ == "__main__":
+    unittest.main()
```
aolemila, line 418: Please add test case for flush_cache.

test/srt/test_pause_continue_generation.py
```diff
@@ -0,0 +1,665 @@
...
+Test pause_generation and continue_generation endpoints.
+
+Usage:
+python3 -m unittest test_pause_continue_generation.TestPauseContinueGeneration
+"""
+
```
aolemila: Please take care about the comments mentioned for engine test. Please ensure whether this test file needs to modify or not.

python/sgl_jax/srt/managers/tokenizer_manager.py
xiuhu17: Thanks for asking. Need to automatically creates an event loop if there is not one. It will hang if `get_internal_state` happens to be the first function invoked by `tokenizer_manager`. Hang was previously appeared in my test.

test/srt/test_engine_pause_continue.py
xiuhu17: Yes it should. In my mind, it should be sufficient if we check _the number of decrease in running batch_ = _the number of increase in waiting queue_. Since there might be cases that some requests have finished.

aolemila: ok

python/sgl_jax/srt/managers/io_struct.py
aolemila: hi,@xiuhu17 , this one

test/srt/test_server_pause_continue.py
```diff
@@ -0,0 +1,658 @@
...
+                except Exception as e:
+                    results.append({"error": str(e)})
+
+            # Verify all requests completed successfully
+            for result in results:
+                self.assertIn("text", result, f"Request should complete with text output: {result}")
+
+        print_test_passed("TestPauseContinueGeneration.test_pause_generation_in_place_mode")
+
+    def test_pause_continue_flush_cache_retract_mode(self):
```
aolemila: It looks like there is no flush_cache call in this case

python/sgl_jax/srt/managers/io_struct.py
xiuhu17: Sure, I am going to add the message. Thanks for pointing out!

python/sgl_jax/srt/entrypoints/engine.py
```diff
@@ -107,6 +109,12 @@ def __init__(self, **kwargs):
         context = zmq.Context(2)
         self.send_to_rpc = get_zmq_socket(context, zmq.DEALER, self.port_args.rpc_ipc_name, True)
 
+        try:
+            self.loop = asyncio.get_running_loop()
+        except RuntimeError:
+            self.loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(self.loop)
+
```
JamesBrianD, line 122: Could you share a bit more about the motivation behind moving the event loop into the Engine state?

aolemila: @JamesBrianD 
Latest main has set loop as field to Engine. In Pathways test, I find out if the loop gained in generate function would results in hang.

Hang: First call generate returns the response. In the second call, the tokenizer_manager    loop exits and no one receives request from detokenzier_manager.

xiuhu17: The main reason is that I refer the design of sglang repo.

Other reasons are: 1. Both @aolemila and I met unexpected hang when testing the code. To  @aolemila, he met this when running code on Pathway. For me, is when testing pause generation in my unit test. 2. It seems that we do not need to keep fetching/creating the event loop in every functions and just set it to `self.loop` in `__init__()`

JamesBrianD: So it sounds like the hang was caused by async components being created under one event loop
and then driven from another loop (often triggered in multi-threaded or Pathways environments).
Storing the loop on the Engine helps enforce loop ownership and avoids this class of issues.
Thanks a lot.

test/srt/test_engine_determine_generation.py
```diff
@@ -0,0 +1,586 @@
...
+        obj = GenerateReqInput(
+            text=prompt,
+            sampling_params={
+                "temperature": 0,
+                "max_new_tokens": max_new_tokens,
+                "ignore_eos": True,
+            },
+        )
+        generator = self.engine.tokenizer_manager.generate_request(obj, None)
+        return await generator.__anext__()
```
aolemila: Why not use engine.async_generate directly? I think it is better not to call tokenizer_manager because you want to test engine.

test/srt/test_engine_determine_generation.py
```diff
@@ -0,0 +1,586 @@
...
+
+        obj = PauseGenerationReqInput(mode=mode)
+        await self.engine.tokenizer_manager.pause_generation(obj)
+
+    async def _async_continue_generation(self):
+        """Continue generation using async method."""
+        from sgl_jax.srt.managers.io_struct import ContinueGenerationReqInput
+
+        obj = ContinueGenerationReqInput()
+        await self.engine.tokenizer_manager.continue_generation(obj)
```
aolemila: If you want to test async, please add async_continue_generation for engine. The same to pause_generation

test/srt/test_engine_determine_generation.py
```diff
@@ -0,0 +1,586 @@
...
+                print(f"  ✗ Requests NOT moved to waiting queue: {Colors.RED}FAIL{Colors.RESET}")
+            
+            # Verify KV cache cleared
+            if tokens_after > tokens_before:
+                print(f"  ✓ KV cache cleared (tokens freed): {Colors.GREEN}PASS{Colors.RESET}")
+            else:
+                print(f"  ✗ KV cache NOT cleared: {Colors.RED}FAIL{Colors.RESET}")
+            
+            print(f"{Colors.BOLD}{Colors.BLUE}=================================={Colors.RESET}\n")
+
```
aolemila, line 197: Hi, @xiuhu17 , I wonder if it is possible to add a check about time.

T1: Total time to generate the complete outputs.
T2: Time to generate the left outputs when you called continue_generation.

Expected: T1 is obviously bigger than T2. The factor is tricky. Maybe this is not a good solution. Because it is not stable.

Note: Only comparing the output ids may not prove that partial rollout with retract works.

aolemila: Another solution: you can check bid through internal_state. Bid is a global variable and it means the how many batches have been scheduled.

python/sgl_jax/srt/managers/io_struct.py
```diff
@@ -112,12 +126,56 @@ class TokenizedGenerateReqInput:
...
 
+    finished_reason: BaseFinishReason | None = None
```
aolemila: Please use `Dict[str: Any] | None` to replace BaseFinishReason. Because when you execute `python3 -u -m sgl_jax.launch_server ...` will the error:

```bash
fastapi.exceptions.FastAPIError: Invalid args for response field! Hint: check that <class 'sgl_jax.srt.managers.io_struct.AbortReq'> is a valid Pydantic field type. If you are using a return type annotation that is not a valid Pydantic field (e.g. Union[Response, dict, None]) you can disable generating the response model from the type annotation with the path operation decorator parameter response_model=None. Read more: https://fastapi.tiangolo.com/tutorial/response-model/
```

aolemila (approved):

PR #523, opened 2025-12-08
Slightly Simplify Engine constructor API -- Accepts only ServerArgs but not dict

wangkuiyi:
### Motivation

Following the philosophical principle Occam's Razor — *"entities should not be multiplied beyond necessity"* — which practically resolves design disputes efficiently and keeps codebases maintainable, this PR slightly simplifies the signature of `Engine`'s constructor, in hope that the simpler, explicit API is easier to understand, type-check (lint), and maintain.

### Changes

**Before:** The constructor accepted `**kwargs` with a hidden `server_args` option, requiring users to understand two different initialization patterns:
```python
# Pattern 1: kwargs (implicit)
Engine(model_path="...", dtype="...")

# Pattern 2: hidden server_args key (undiscoverable)
Engine(server_args=server_args)
```

**After:** A single, explicit pattern:
```python
Engine(ServerArgs(model_path="...", dtype="..."))
```

### Files Changed

- `python/sgl_jax/srt/entrypoints/engine.py` — Simplified `__init__(self, server_args: ServerArgs)`
- `python/sgl_jax/bench_offline_throughput.py` — Pass `server_args` directly
- `test/srt/test_logprobs.py` — Use `ServerArgs` for config constant
- `test/srt/test_srt_engine.py` — Use `ServerArgs` directly
- `docs/features/integrate_with_tunix.md` — Updated example

## Accuracy Tests

N/A - This PR does not change accuracy.

I've run the following command to start a server on a single TPU VM.

```shell
python3 -u -m sgl_jax.launch_server \
--model-path deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B \
--trust-remote-code \
--device=tpu \
--tp-size=4 \
--skip-server-warmup \
--host 0.0.0.0 \
--port 30000
```

and run the following command to check it returns reasonable results:

```shell
curl http://localhost:30000/v1/chat/completions \
-d '{"model": "default", "messages": [{"role": "user", "content": "Hello!"}]}' | jq .
```

## Benchmarking and Profiling

N/A - This PR does not change performance.

## Checklist

### DIFF
--- a/test/srt/test_logprobs.py
+++ b/test/srt/test_logprobs.py
@@ -121,27 +121,16 @@ class TestLogprobsDense(unittest.TestCase):
         ]
         for i, logprob in enumerate(output_meta["output_token_logprobs"]):
             self.assertEqual(
-                logprob[0],
-                expected_output_logprobs[i][0],
-                f"{logprob[0]} logprob is invalid",
+                logprob[0], expected_output_logprobs[i][0], f"{logprob[0]} logprob is invalid"
             )
             self.assertEqual(
-                logprob[1],
-                expected_output_logprobs[i][1],
-                f"{logprob[1]} output id is invalid",
+                logprob[1], expected_output_logprobs[i][1], f"{logprob[1]} output id is invalid"
             )
             self.assertEqual(
-                logprob[2],
-                expected_output_logprobs[i][2],
-                f"{logprob[2]} token is invalid",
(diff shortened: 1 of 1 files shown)


JamesBrianD:
This is really a question about our API design goals: do we want to support two initialization styles — dict/kwargs plus ServerArgs (inherited from the main sglang repo) — or do we prefer to keep the Engine constructor as minimal and clean as possible?
@sii-xinglong @jimoosciuc @aolemila do you have any thoughts on this?

Issue #270, opened 2025-10-23
Instructions for Qwen3-32B on TPU v6e-4


JamesBrianD:
This issue records how to reproduce the benchmark and profiling result.

# Preface

## Model

Qwen/Qwen3-32B.

## Metrics
TTFT, ITL, Input Throughput and Output Throughput.

## Benchmark Scenarios

Here provide three scenarios. In particular, we set output length to 1 for TTFT to get more accurate results.
- Scenario1: input_len = 1024, output_len = 1 for TTFT and Input Throughput , output_len = 1024 for left two
- Scenario2: input_len = 4096, keep the same to scenario1
- Scenario3: input_len = 8192, keep the same to scenario1

## TPU version

TPU v6e.

## Version

- SGLang-JAX version: main-af32f095880ff676ed23eec19bc79584b5e20717
 - vLLM: vllm-tpu==0.11.1

## Setup Environment

**SGLang-Jax**

```bash
conda create -n sgl python=3.12 -y

conda activate sgl

git clone git@github.com:sgl-project/sglang-jax.git && cd sglang-jax

pip3 install -e python/
```

**vLLM**

```bash
# Refer to https://blog.vllm.ai/2025/10/16/vllm-tpu.html
conda create -n vllm_tpu python=3.12 -y

conda activate vllm_tpu

pip install vllm-tpu

sudo mount -t tmpfs -o size=200G tmpfs /tmp
```

# SGLang-Jax

### Launch Server

**General**
```bash
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache \
uv run python -u -m sgl_jax.launch_server \
--model-path Qwen/Qwen3-32B \
--trust-remote-code  \
--tp-size=4 \
--device=tpu \
--mem-fraction-static=0.95 \
--chunked-prefill-size=2048 \
--download-dir=/tmp \
--dtype=bfloat16 \
--max-running-requests 256 \
--skip-server-warmup \
--page-size=128  \
--disable-radix-cache
```

### Benchmark

Run `./script.sh sglang-oai`.

```bash
#!/bin/bash

set -e

if [ -z "$1" ]; then
    echo "Usage: $0 <engine>"
    echo "engine: sgl-jax, sglang-oai or vllm"
    exit 1
fi

backend=${1}
num_prompts_per_concurrency=3
input_seq_lens=(1024 4096) # modify this according your requirements.
output_seq_lens=(1 1024) # 1 for TTFT
[...]

PR #550, opened 2025-12-15
[perf] update model_state_leaves in prepare_lora_batch

aolemila:
## Benchmarking and Profiling

✅For benchmark, this PR will include two cases:
- Case1: Compare the performance without lora with #297. The instructions refer to #270. Use Qwen/Qwen3-32B.
    - Note: This case is used to test whether the original performance is influenced by the LoRA codes. 
- Case2: Based on the same base model, compare LoRA with no LoRA.
    - Base model: Qwen/Qwen3-32B.
    - LoRA: [flyfishxu/DeepNews-LoRA-Qwen3-32B](https://huggingface.co/flyfishxu/DeepNews-LoRA-Qwen3-32B)

✅For profile, check whether the LoRA influences the effect of overlap.

#### Case1: blog vs current branch

Note: For input_seq_len = 8192, blog outperforms current branch in output throughput. The biggest gap is about 3%. So add the fourth figure which compares the `main-dd04bc8fffcbcfe233c151acffa7cdbbd68168c9` with current branch. They behave almost identically. According to log, the number of running requests is almost about 25 all the time for the pair of input_seq_len=8192 and max_concurrency=256. Track slight performance degradation on main with #555.

image

image

image

image

#### Case2: LoRA vs noLoRA

Data(decode): 
- input_seq_len is 4096 and max_concurrency is 128
- LoRA vs noLoRA:
    - ITL: 28.66ms vs 30.28ms
    - output throughput: 1050.62 vs 1257.86

Note: LoRA outperforms noLoRA in ITL but output throughput is worse. Here are the key factors:
> 
- LoRA:
    - mem-fraction-static = 0.9, max_tokens=180946, the number of running requests < 40
- noLoRA:
    -  mem-fraction-static = 0.95, max_tokens=236367, the number of running requests < 50

The lower the number of running_requests is, the lower the output throughput is.

Let's dive into the profile result. We will find out the kernel execution time of every layer for noLoRA is about 278us. But LoRA only costs about 232us. In order to explore the relationship between kernel execution cost and max_tokens, I profile the following case whose mem-fraction-static is 0.9(max_tokens=210770 from log), input_seq_len and max_concurrency keep the same to above Data, the kernel execution time is about 256us.

image

#### Instructions for Case2

```bash
# launch server
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache \
uv run python -u -m sgl_jax.launch_server \
--model-path Qwen/Qwen3-32B \
--trust-remote-code  \
--tp-size=4 \
--device=tpu \
--mem-fraction-static=0.9 \
--chunked-prefill-size=2048 \
--download-dir=/tmp \
--dtype=bfloat16 \
--max-running-requests 256 \
--skip-server-warmup \
--page-size=128  \
--disable-radix-cache \
--lora-paths flyfishxu/DeepNews-LoRA-Qwen3-32B
```

```bash
# do benchmark
#!/bin/bash

set -e

if [ -z "$1" ]; then
    echo "Usage: $0 <engine>"
    echo "engine: sgl-jax, sglang-oai or vllm"
    exit 1
fi

backend=${1}
num_prompts_per_concurrency=3
input_seq_lens=(1024 4096 8192) # modify this according your requirements.
output_seq_lens=(1 1024) # 1 for TTFT
max_concurrencies=(8 16 32 64 128 256)

for input_seq_len in "${input_seq_lens[@]}"; do
    for output_seq_len in "${output_seq_lens[@]}"; do
        echo "======================================="
        echo "Testing with ISL/OSL: $input_seq_len/$output_seq_len"
        echo "======================================="
        for max_concurrency in "${max_concurrencies[@]}"; do
            echo "benchmark on max_concurrency=$max_concurrency"
            num_prompts=$((num_prompts_per_concurrency * max_concurrency))
            uv run python -m sgl_jax.bench_serving \
              --backend ${backend} \
              --dataset-name random \
              --num-prompts ${num_prompts} \
              --random-input-len ${input_seq_len} \
              --random-output-len ${output_seq_len} \
              --max-concurrency ${max_concurrency} \
              --random-range-ratio 1 \
              --disable-ignore-eos \
              --warmup-requests 0 \
              --lora-name DeepNews-LoRA-Qwen3-32B
        done
    done
done
```

### Profile Results

Note: The profile decode results are gained with the following conditions:
- input_seq_len = 4096
- max_concurrency = 128
- output_seq_len = 1024

#### LoRA

According to the profile result,the gap between forward of LoRA is about 7.7us, and the noLoRA is about 7.5us. So LoRA almost does not influence the overlap.

**Decode**

image

image

image

#### noLoRA

**Decode**

mem-fraction-static = 0.95

image

image

mem-fraction-static = 0.9
image

## Checklist

### DIFF
--- a/test/srt/nightly-test/test_perf.py
+++ b/test/srt/nightly-test/test_perf.py
@@ -22,11 +22,11 @@ from sgl_jax.test.test_utils import (
     QWEN3_MOE_30B,
     QWEN_7B,
     CustomTestCase,
+    get_benchmark_args,
     is_in_ci,
     popen_launch_server,
     run_bench_serving,
     write_github_step_summary,
-    get_benchmark_args,
 )
 
 MOUNT_ROOT = os.getenv("CI_MOUNT_ROOT", "/models")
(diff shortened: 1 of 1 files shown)


python/sgl_jax/srt/entrypoints/openai/serving_completions.py
JamesBrianD: Maybe we can remove these codes

python/sgl_jax/srt/managers/tp_worker.py
```diff
@@ -394,6 +394,12 @@ def generate_model_worker_batch(
     def get_model_runner(self):
         return self.model_runner
 
+    def prepare_lora_batch(self, model_worker_batch: ModelWorkerBatch):
+        has_new_weights = self.model_runner.lora_manager.prepare_lora_batch(model_worker_batch)
+        if has_new_weights:
+            _, model_state = nnx.split(self.model_runner.model)
+            self.model_runner.model_state_leaves, _ = jax.tree_util.tree_flatten(model_state)
```
JamesBrianD, line 401: Having to thread `has_new_weights` through the code path feels a bit inelegant.

python/sgl_jax/srt/entrypoints/openai/serving_completions.py
aolemila: done

python/sgl_jax/srt/managers/tp_worker.py
aolemila: set `has_new_weights` as the field for lora_manager.

jimoosciuc (approved):
LGTM

PR #585, opened 2025-12-22
Feat: Add weight quantization support for dense models

Bob-Chen222:
## Motivation
add quantization support and now qwen can run with int8 quantization tpuv6e

## Modifications
added PTQ quantization method support. more details can be find in quantization.md

Launching server:

```
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache uv run python -u -m sgl_jax.launch_server
--model-path Qwen/Qwen3-32B --trust-remote-code --dist-init-addr=10.128.0.19:10011 
--nnodes=1 --tp-size=4 --device=tpu  --random-seed=3 --node-rank=0 --mem-fraction-static=0.9
--max-prefill-tokens=8192 --download-dir=/tmp --dtype=bfloat16 --skip-server-warmup
--max-running-requests 256 --chunked-prefill-size=2048
--page-size=128 --quantization-config-path=int8_all_modules_w_only.yaml
```

## Accuracy Tests
Qwen3-32B shows no performance degradation on GSM8K, maintaining ~96% accuracy.
On AIME, Qwen3-32B quantized is 73% accurate while fp version is 76%

eval command:
```
evalscope eval  --model  Qwen/Qwen3-32B 
--api-url http://127.0.0.1:30000/v1/chat/completions 
--api-key EMPTY --eval-type openai_api 
--datasets gsm8k --eval-batch-size 64 
--generation-config '{"temperature": 0.7,"top_p":0.8,"top_k":20,"min_p":0.0,"presence_penalty":0.5}' 
```

## Benchmarking and Profiling
Under the benchmark bench_one_batch_server configuration (batch_size=1024, input_len=8192, output_len=1024), INT8-quantized Qwen3-32B delivers ~2.5× higher throughput than the FP baseline.

benchmark command:
```
python3 -m sgl_jax.bench_one_batch_server     --base-url http://127.0.0.1:30000  
--model-path Qwen/Qwen3-32B   --batch-size 1024  --input-len 8192   --output-len 1024
```

## Checklist

### DIFF
--- a/python/sgl_jax/srt/utils/quantization/quantization_utils.py
+++ b/python/sgl_jax/srt/utils/quantization/quantization_utils.py
@@ -1,5 +1,5 @@
-# SPDX-License-Identifier: Apache-2.0
-# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# Adapted from https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/models/jax/utils/qwix/qwix_utils.py
+
 import functools
 import os
 


docs/features/quantization.md
```diff
@@ -0,0 +1,66 @@
...
+  --mem-fraction-static=0.8 \
+  --max-prefill-tokens=8192 \
+  --download-dir=/tmp \
+  --dtype=bfloat16 \
+  --skip-server-warmup \
+  --quantization-config-path=sgl_jax/srt/utils/quantization/configs/int8_all_modules_w_only.yaml
+```
+
+# Supported models (To be updated)
+- QWenLMHeadModel series
```
JamesBrianD: Based on the code, it appears that all current dense models already support quantization.

python/sgl_jax/srt/model_executor/forward_batch_info.py
```diff
@@ -289,3 +289,71 @@ def init_new(
...
 
         return obj
+
+    @classmethod
+    def init_from_batch(
+        cls,
+        batch: ModelWorkerBatch,
+        mesh: jax.sharding.Mesh,
+        attn_backend: AttentionBackend,
+    ):
```
JamesBrianD: It might make sense to merge this with `init_new` into a single method. One option is to rename `init_new` to `init_from_batch`, or alternatively use a mock model_runner in combination with `init_new`.

python/sgl_jax/srt/utils/quantization/quantization_utils.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    token_to_kv_pool = MHATokenToKVPool(
+        size=DEFAULT_MAX_NUM_TOKENS,
+        page_size=DEFAULT_PAGE_SIZE,
+        dtype=jnp.bfloat16,
+        head_num=kv_head_num,
+        head_dim=head_dim,
+        layer_num=layer_num,
+        mesh=mesh,
+    )
```
JamesBrianD: Hard-coding `dtype = jnp.bfloat16` is not ideal. If the model activations use a different dtype, this could cause issues.

test/srt/quantization/test_w8_quantization.py
```diff
@@ -0,0 +1,100 @@
+import time
+import unittest
+from types import SimpleNamespace
+
+import requests
+from run_eval import run_eval
```
JamesBrianD, line 10: Should add e2e test file into `run_suite.py`

test/srt/quantization/test_w8_quantization.py
```diff
@@ -0,0 +1,102 @@
...
+
```
JamesBrianD:
```
  File "/home/runner/_work/sglang-jax/sglang-jax/test/srt/quantization/test_w8_quantization.py", line 6, in <module>
    from run_eval import run_eval
ModuleNotFoundError: No module named 'run_eval'
```
CI failed, please fix it.

JamesBrianD:
Could we add logs for the available KV cache size? After quantization, weights use much less memory, so the KV cache capacity should increase significantly. It would be great to see a before/after comparison using Qwen3-32B on TPU v6e-4.

Bob-Chen222:
> Could we add logs for the available KV cache size? After quantization, weights use much less memory, so the KV cache capacity should increase significantly. It would be great to see a before/after comparison using Qwen3-32B on TPU v6e-4.

Do you mean profiling number of tokens before and after quantization and logging the saved memory to the stdout?

JamesBrianD:
Yes, the number should be printed when server was launched. So, we’d like to capture logs before and after quantization to better understand and quantify its effects.

JamesBrianD (approved):
/LGTM

PR #589, opened 2025-12-22
[WIP]Feat: Support Data Parallelism

### LINKED ISSUE #3: [Feature] Data-Parallel(DP)  Support
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.

### Motivation

Support data-parallel, support mixture of tp and dp. Not consider Hybrid Parallel Attention DP+TP with Moe EP+TP.

### RoadMap
- [ ] Minimal DP (Single Machine, Single Processor)

### Related resources
- https://docs.google.com/document/d/17YpMpmpOqgCJ07Ti7i1_wg7bKPU0r0yM0iZrbc_TV0c/edit?tab=t.n2a6x3o6uiqc

JamesBrianD:
## Modifications

1.  Enable Data Parallelism (DP) for the prefix cache by modifying the radix key and KV-Cache allocator, while preserving compatibility with the existing scheduling logic.
2. `Schedule batch (breaking change)`: Modified the schedule batch logic by separating part of the DP-related scheduling information and storing it in `List[ScheduleReqsInfo]`. The scheduling logic in both scheduler and scheduler batch has been refactored to be DP-oriented.
3. Introduce `ModelWorkerSamplingInfo`: Added ModelWorkerSamplingInfo, which corresponds to ModelWorkerBatch, as the batch information used in the model execution layer (stored on CPU), instead of using SamplingBatchInfo as before.

## TODO
1. Compatibility for spec decode, penalty and returning logprobs is not finished yet.

#### DP = 1
```
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache uv run python -u -m sgl_jax.launch_server     --model-path Qwen/Qwen3-8B     --trust-remote-code     --dist-init-addr=0.0.0.0:10011     --nnodes=1     --tp-size=4     --device=tpu     --random-seed=3     --node-rank=0     --mem-fraction-static=0.8     --max-prefill-tokens=8192     --download-dir=/tmp     --dtype=bfloat16     --skip-server-warmup --page-size 128

evalscope eval  --model Qwen/Qwen3-8B  --api-url http://127.0.0.1:30000/v1 --api-key EMPTY --eval-type server --datasets gsm8k --eval-batch-size 64

+----------+-----------+----------+----------+-------+---------+---------+
| Model    | Dataset   | Metric   | Subset   |   Num |   Score | Cat.0   |
+==========+===========+==========+==========+=======+=========+=========+
| Qwen3-8B | gsm8k     | mean_acc | main     |  1319 |  0.9583 | default |
+----------+-----------+----------+----------+-------+---------+---------+
```

## Checklist


python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -1104,31 +1200,52 @@ def get_new_batch_prefill(self) -> ScheduleBatch | None:
             self.max_prefill_tokens,
             self.chunked_prefill_size,
             running_bs if self.is_mixed_chunk else 0,
```
aolemila, line 1262: `running_bs` sums all reqs across DP ranks. I wonder is it possible to only take care of the individual requests which has the same rank with the current scheduler?

aolemila: Current sum results in overcalculation?

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -1567,98 +2017,326 @@ def get_spec_model_worker_batch(
...
-            return self.token_to_kv_pool_allocator.available_size() >= num_tokens
+                if (full_available < num_tokens or swa_available < num_tokens) and self.tree_cache:
+                    full_num = max(0, num_tokens - full_available)
+                    swa_num = max(0, num_tokens - swa_available)
+                    self.tree_cache.evict(full_num, swa_num)
+            else:
+                available = self.token_to_kv_pool_allocator.available_size(dp_rank=dp_rank)
+                if available < num_tokens and self.tree_cache:
+                    self.tree_cache.evict(num_tokens)
+
```
aolemila, line 2204: `num_tokens` is required to evict for above mentioned `dp_rank`. I find the implementation of `radixCache.evict` will take `dp_rank` into consideration when calling `free` for allocator. My question is how to ensure the `dp_rank` as input for `free` is the expected `dp_rank` to evict?

python/sgl_jax/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -71,113 +71,133 @@ def process_batch_result_prefill(
...
-                        next_token_ids,
-                        num_input_logprobs,
-                        logits_output,
-                    )
-                    logprob_pt += num_input_logprobs
+                if (
+                    self.is_mixed_chunk and self.enable_overlap and req.finished()
+                ):  # TODO @Brian fix it
+                    j = len(info.out_cache_loc) - batch.batch_size() + i
+                    self.token_to_kv_pool_allocator.free(info.out_cache_loc[j : j + 1], req.dp_rank)
```
aolemila, line 114: Why use `[b_per_dp]`- `sum(len(info.reqs) if info.reqs else 0 for info in self.reqs_info)` + i?

can we use `j=i`?

python/sgl_jax/srt/model_executor/forward_batch_info.py
```diff
@@ -271,7 +265,6 @@ def init_new(
                 batch.seq_lens,
                 batch.out_cache_loc,
                 batch.positions,
-                batch.extend_start_loc,
                 batch.req_pool_indices,
                 batch.cache_loc,
                 batch.extend_prefix_lens,
```
aolemila, line 286: Due to no sharding in `device_array`, input_ids merged from different dp ranks will repeatly calculate for some modules like `q_proj`, `k_proj`, etc. But the attention kernel only calculates the coresponding dp rank. So for modules like `q_proj`, is it expectable?

aolemila: The same question to lora.

python/sgl_jax/srt/managers/schedule_batch.py
JamesBrianD: Yes, `dp_rank` does need to be taken into account. I’ll update the related `evict` logic accordingly.

python/sgl_jax/srt/managers/scheduler_output_processor_mixin.py
JamesBrianD: This needs to take mixed chunked prefill into account. In that case, the results for the decode requests are placed in the last few entries of out_cache_loc, so we compute a relative starting index j. (However, the current formula for j is likely incorrect — I haven’t validated it against the mixed chunked prefill scheduling logic yet. TODO.)

python/sgl_jax/srt/managers/scheduler.py
JamesBrianD: This value is related to mixed_with_decode_tokens. The computation here is probably incorrect, but since it only takes effect when mixed_chunk is enabled, I haven’t changed it yet.

python/sgl_jax/srt/model_executor/forward_batch_info.py
JamesBrianD: In the future, device_array is expected to be sharded. Currently it is not sharded, and this is intended to validate whether the control plane works correctly when dp = 1.

PR #349, opened 2025-11-10
feat: support MoE hybrid parallelism

Prayer3th:
## Motivation

* Implement MoE EP+TP hybrid parallelism to support more model implementations (e.g., Ling 1T/grok2)
* Achieve complete decoupling between attention parallelism and Moe parallelism
* Replacing the less efficient FusedMoE

## Modifications

* Modify some components in the layer (rmsnorm/embeddings/linearbase) to add manual sharding definition, enabling support for explicit mesh.

* Modify the layer's moe.py to support mesh switching under jax.jit, implementing ep+tp hybrid parallelism, and completing parallel decoupling between the attention layer and moe layer.

* New moe weight loading logic

* Address some compatibility issues related to cache misses. 

### baseline(main branch)
```
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache  python3 -u -m sgl_jax.launch_server --model-path Qwen/Qwen3-30B-A3B --trust-remote-code  --dist-init-addr=0.0.0.0:10011 --nnodes=1  --tp-size=4 --ep-size=4 --device=tpu --random-seed=3 --node-rank=0 --mem-fraction-static=0.85 --chunked-prefill-size=2048 --download-dir=/dev/shm --dtype=bfloat16 --max-running-requests 256 --skip-server-warmup --page-size=128
+---------------+-----------+-----------------+----------+-------+---------+---------+
| Model         | Dataset   | Metric          | Subset   |   Num |   Score | Cat.0   |
+===============+===========+=================+==========+=======+=========+=========+
| Qwen3-30B-A3B | gsm8k     | AverageAccuracy | main     |  1319 |  0.9378 | default |
+---------------+-----------+-----------------+----------+-------+---------+---------+
```
### pure EP
```
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache  python3 -u -m sgl_jax.launch_server --model-path Qwen/Qwen3-30B-A3B --trust-remote-code  --dist-init-addr=0.0.0.0:10011 --nnodes=1  --tp-size=4 --ep-size=4 --device=tpu --random-seed=3 --node-rank=0 --mem-fraction-static=0.85 --chunked-prefill-size=2048 --download-dir=/dev/shm --dtype=bfloat16 --max-running-requests 256 --skip-server-warmup --page-size=128

+---------------+-----------+-----------------+----------+-------+---------+---------+
| Model         | Dataset   | Metric          | Subset   |   Num |   Score | Cat.0   |
[...]

PR #621, opened 2025-12-31
integration with quantized gmm kernel

Bob-Chen222:
## Motivation
Integrate a new quantized GEMM kernel and support quantization for both MoE and standard layers.

## Modifications
This change refactors the quantization logic. Because qwix only supports weights used in `einsum` and `dot_product`, MoE kernel weights must be quantized manually. I added `quantize_tensor`, `dequantize_tensor`, and related utilities to implement a workaround enabling w8a8 quantization for MoE models. I also added support for custom quantization policies via a config file (e.g., int8.yaml) using a new `quantization_config` data class, and added `fp8_grok.yaml`, aligned with Grok-2 quantization in sglang-gpu.

## Accuracy Tests
I have tested grok-2 on gpqa for a few times and seems like that the performance is not decreased. I can have a rigorous test that test 8 or 16 times and get the average difference between quantized and unquantized version.

## Benchmarking and Profiling
The benchmarking and profiling

image
image

The quantized version (left) currently provides only a minimal throughput improvement over the unquantized version (right), indicating room for further optimization. I plan to profile this PR tomorrow. The implementation is functionally correct, and the overall design structure is in place.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -282,25 +282,28 @@ class EPMoE(nnx.Module):
         if self.quantized_dtype is None:
             return
 
+        # Replace original weights with quantized versions
         with jax.sharding.use_abstract_mesh(self.updated_mesh):
             # Quantize weights
             w0_value, w0_scale = quantize_tensor(
                 self.quantized_dtype,
                 self.wi_0.value,
                 axis=2,
+                pad_tensor=True,
             )
             w1_value, w1_scale = quantize_tensor(
                 self.quantized_dtype,
                 self.wi_1.value,
                 axis=2,
+                pad_tensor=True,
             )
             wo_value, wo_scale = quantize_tensor(
                 self.quantized_dtype,
                 self.wo.value,
                 axis=2,
+                pad_tensor=True,
             )
 
-            # Replace original weights with quantized versions
             self.wi_0 = nnx.Param(w0_value, out_sharding=P("expert", "tensor", None))
             self.wi_1 = nnx.Param(w1_value, out_sharding=P("expert", "tensor", None))
             self.wo = nnx.Param(wo_value, out_sharding=P("expert", None, "tensor"))
@@ -480,8 +483,8 @@ class EPMoE(nnx.Module):
             return empty_output
 
         m, k = x.shape[0], x.shape[1]
-        n_gate = w0_kernel.shape[2]
-        n_down = wo_kernel.shape[2]
+        n_gate = w0_kernel.shape[1]
+        n_down = wo_kernel.shape[1]
 
         default_tile_size = (512, 1024, 1024)
         tiling_gate = (
--- a/python/sgl_jax/srt/models/grok.py
+++ b/python/sgl_jax/srt/models/grok.py
@@ -912,7 +912,7 @@ class Grok1ForCausalLM(nnx.Module):
             prefix=prefix,
             target_prefix=target_prefix,
             num_experts=self.config.num_local_experts,
-            moe_path="block_sparse_moe",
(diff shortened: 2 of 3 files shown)


python/sgl_jax/srt/layers/moe.py
```diff
@@ -6,14 +6,18 @@
...
+)
 
 
 class GateLogit(nnx.Module):
     def __init__(
         self,
         input_size: int,
         num_experts: int = 0,
-        weight_dtype: jnp.dtype = jnp.bfloat16,
+        weight_dtype: jnp.dtype = jnp.float32,
```
JamesBrianD: default type changed

python/sgl_jax/srt/layers/moe.py
```diff
@@ -36,7 +40,7 @@ def __init__(
             self.bias = None
 
     def __call__(self, hidden_states: jax.Array) -> tuple[jax.Array, jax.Array | None]:
-        logits = hidden_states.astype(self.weight_dtype) @ self.kernel
+        logits = hidden_states.astype(jnp.float32) @ self.kernel
```
JamesBrianD, line 44: Is this change made to achieve higher accuracy?

python/sgl_jax/srt/utils/quantization/configs/fp8_w8a8.yaml
```diff
@@ -0,0 +1,15 @@
+# FP8 weight and activation quantization config
+# Works for both dense and MoE models
+
+quantization:
+  # Qwix rules for dense layers (attention, MLP, embeddings)
+  qwix:
+    rules:
+      - module_path: '.*'
+        weight_qtype: 'float8_e4m3fn'
+        act_qtype: 'float8_e4m3fn'
```
JamesBrianD, line 10: MoE-related layers were not excluded.

Bob-Chen222: This is a bit tricky. Native qwix only supports quantization for weights/activations used in three operator types: einsum, dot_general, and conv_general_dilate. As a result, the GEMM (Pallas op) is not quantized, and the moe section in the YAML specifically targets this GEMM path.

However, within the overall epmoe module, einsum is still used for the unpermute step and dot_general is used for the gating operation. These parts therefore continue to rely on qwix for quantization.

python/sgl_jax/srt/layers/moe.py
Bob-Chen222: This behavior matches what I observed in sglang-GPU, which is why I made the second change to explicitly cast to float32. I will revert the first change since it has no impact on Grok.

After retesting 10 times, I found that using bf16 vs. float32 for activations has a negligible effect that is difficult to measure, so I’m not sure converting activation to fp32 is actually necessary. That said, we could still apply it to align the implementation, as long as it does not negatively affect other models.
Screenshot 2026-01-09 at 12 02 29 PM

python/sgl_jax/srt/utils/quantization/configs/fp8_w8a8.yaml
JamesBrianD: So this part is for quantizing the gate, right?

Bob-Chen222: Yes the quantization of gate will be handled by qwix. And in the grok-2 specific setting I excluded the quantization of gate

JamesBrianD:
@Bob-Chen222 Since fused MoE was recently merged, could you help rebase the branch and test the baseline performance of Qwen3-MoE for both quantized and non-quantized setups? Compare with this baseline https://github.com/sgl-project/sglang-jax/pull/349

PR #680, opened 2026-01-16
fix: sharding not match when enable dp in Qwen3 Moe Model

Iamleos:
## Checklist

### DIFF
--- a/python/sgl_jax/srt/models/qwen3_moe.py
+++ b/python/sgl_jax/srt/models/qwen3_moe.py
@@ -2,8 +2,9 @@ import logging
 from typing import Any
 
 from flax import nnx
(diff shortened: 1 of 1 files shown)


python/sgl_jax/srt/models/qwen3_moe.py
```diff
@@ -120,9 +122,42 @@ def __call__(
...
+            self.q_head_num,
+            self.head_dim,
+            out_sharding=NamedSharding(
+                self.mesh,
+                P(
+                    "data",
+                    "tensor",
+                ),
+            ),
+        )
```
JamesBrianD, line 136: Please edit all the moe model files

PR #689, opened 2026-01-19
Feat/add abort req for mm

pengchengneo:
support abort_req when forward req

image

tested local , abort req at vae stage and diffusion stage

when abort req, the aborted req will send to Stage which running this req, and in diffusion stage, it will stop forward left steps if abort

### DIFF
--- a/python/sgl_jax/srt/multimodal/entrypoint/http_server.py
+++ b/python/sgl_jax/srt/multimodal/entrypoint/http_server.py
@@ -9,10 +9,10 @@ from fastapi import Request
 from fastapi.responses import ORJSONResponse, Response
 
 from sgl_jax.srt.entrypoints.http_server import _GlobalState, app, set_global_state
+from sgl_jax.srt.managers.io_struct import AbortReq
 from sgl_jax.srt.managers.template_manager import TemplateManager
 from sgl_jax.srt.multimodal.common.ServerArgs import MultimodalServerArgs
 from sgl_jax.srt.multimodal.manager.global_scheduler import run_global_scheduler_process
-from sgl_jax.srt.managers.io_struct import AbortReq
 from sgl_jax.srt.multimodal.manager.io_struct import (
     DataType,
     GenerateMMReqInput,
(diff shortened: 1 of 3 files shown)


python/sgl_jax/srt/multimodal/entrypoint/http_server.py
```diff
@@ -81,6 +82,24 @@ async def videos_generation(obj: VideoGenerationsRequest, request: Request):
         return _create_error_response(e)
 
 
+@app.post("/abort_request")
```
pathfinder-pf, line 85: Is this interface called when you press Ctrl+C, or is it called manually?

pengchengneo: it will be called automatically when press ctrl c

python/sgl_jax/srt/multimodal/manager/global_scheduler.py
```diff
@@ -114,20 +130,66 @@ def _build_stage(idx_cfg: tuple[int, Any]) -> tuple[int, Stage]:
...
+                    rid=rid,
+                    aborted_message="Aborted by client request",
+                )
+                try:
+                    self.in_queues[current_stage].put_nowait(stage_abort_req)
+                except Exception as e:
+                    logger.warning("Failed to send abort to stage %d queue: %s", current_stage, e)
+
+                # Send AbortReq to detokenizer -> tokenizer to notify client
+                self.send_to_detokenizer.send_pyobj(stage_abort_req)
```
pathfinder-pf, line 184: how to ensure the request is aborted successfully if we just send to detokenizer? can we get the result from stage, and then send to detokenizer?

pengchengneo: the aborted req will not be processed  by remain stages, that is the way we make sure it will be aborted.

I think we cannot make sure we will get result from stage, because we dont know whether the stage will be forward entirely , like diffusion, and the result from the aborted req is meaningless I think

pathfinder-pf: get

PR #699, opened 2026-01-20
[Bug] use float32 for router_logits and topk_weights in moe.py

### LINKED ISSUE #698: [Bug] use float32 for router_logits and topk_weights in MoE
### **Checklist**

- [ ]  I have searched related issues but cannot get the expected help.
- [ ] The bug has not been fixed in the latest version.
- [ ] Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.

### **Describe the bug**

It may result in precision problem. We keep the same to SGLang.

### **Environment**

aolemila:
## Motivation

Fixed #698.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -89,8 +89,7 @@ class TopK(nnx.Module):
             if self.routed_scaling_factor is not None:
                 topk_weights *= self.routed_scaling_factor
 
-        if topk_weights.dtype != jnp.float32:
-            topk_weights = topk_weights.astype(jnp.float32)
+        topk_weights = topk_weights.astype(jnp.float32)
 
         return topk_weights, topk_ids
 


python/sgl_jax/srt/layers/moe.py
```diff
@@ -87,7 +89,8 @@ def __call__(self, router_logits: jax.Array, correction_bias: jax.Array = None):
             if self.routed_scaling_factor is not None:
                 topk_weights *= self.routed_scaling_factor
 
-        topk_weights = topk_weights.astype(router_logits.dtype)
+        if topk_weights.dtype != jnp.float32:
+            topk_weights = topk_weights.astype(jnp.float32)
```
JamesBrianD: Don't need `if topk_weights.dtype != jnp.float32:`

aolemila: done

PR #722, opened 2026-01-23
Change proxy checker when Pathways on Cloud is enabled.

wang2yn84:
## Motivation

We set JAX_PLATFORMS=proxy,cpu when we run Pathways workloads, so the device is ('proxy', 'cpu'). In the current implementation, it checks devices == 'proxy' which is inaccurate. 

## Modifications

This PR fixes this issue by checking if 'proxy' is in the device list.

## Accuracy Tests

Run pw workload in local and it's passing.

## Benchmarking and Profiling

NA

## Checklist


aolemila:
Thanks for your fix!

aolemila (approved):

PR #767, opened 2026-02-05
[FIX] fix mrope_position for qwen2.5VL

pengchengneo:
set up env
`uv sync --extra multimodal,tpu`

launch server

`uv run  python3 -u -m sgl_jax.launch_server --multimodal --model-path /models/Qwen/Qwen2.5-VL-7B-Instruct  --log-requests --disable-precompile --page-size 64`

run evalscope

`evalscope eval   --model Qwen2.5-VL-7B-Instruct   --api-url http://127.0.0.1:30000/v1/chat/completions   --api-key EMPTY   --eval-type openai_api   --datasets math_vision  --eval-batch-size 16`

test result
image

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -1851,8 +1851,6 @@ class ModelWorkerBatch:
     sampling_info: SamplingBatchInfo
     # Position information [total_tokens]
     positions: np.ndarray
-    # MRoPE position information [3, total_tokens]
-    mrope_positions: np.ndarray | None
     # Start position for each sequence in extend mode [batch_size]
     extend_start_loc: np.ndarray
     # cache_loc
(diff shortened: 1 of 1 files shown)


python/sgl_jax/srt/managers/tp_worker.py
```diff
@@ -416,6 +416,7 @@ def generate_model_worker_batch(
             ),
             extend_input_logprob_token_ids=None,
             positions=np.concat([valid_positions, invalid_positions], axis=0),
+            mrope_positions=None,
```
pathfinder-pf, line 419: maybe cache miss

pengchengneo: it has been padding with _compute_mrope_positions_for_batch

pengchengneo: qwen VL precompile logic is not implemented here, I think positions, input_embeddings should be padded together

PR #840, opened 2026-02-26
feat: Add support for LTX-2 Video Diffusion Model

chandrasekhard2:
### Summary

This PR adds full support for **LTX-2** (Lightricks, 19B parameters), a state-of-the-art joint audio-video diffusion transformer, to sglang-jax. LTX-2 generates synchronized video and stereo audio from text prompts using a single unified DiT backbone with cross-modal attention. The integration spans the complete inference pipeline: text conditioning via Gemma-3 12B, a 48-layer dual-modality diffusion transformer with 4-pass guided denoising, video/audio VAE decoders, and a HiFi-GAN vocoder — all running on TPU v7 with 8-way tensor parallelism.

---

### Architecture Overview

LTX-2 Architecture Diagram

The pipeline is organized as a **3-stage system**, matching sglang-jax's existing multimodal stage abstraction (same pattern used by WAN 2.1):

| Stage | Component | TPU Allocation | Purpose |
|-------|-----------|---------------|---------|
| **Stage 0** | Text Encoder | 2 TPUs | Gemma-3 12B + feature extraction + dual connectors |
| **Stage 1** | DiT Denoiser | 1 TPU (8-way TP) | 48-layer transformer with Euler ODE solver |
| **Stage 2** | VAE Decode | 1 TPU | Video VAE + Audio VAE + HiFi-GAN vocoder |

---

#### Design Decision: Reuse Gemma-3 12B as Backbone

LTX-2 uses Gemma-3 12B as its text backbone, but in a non-standard way: instead of using only the final hidden state, it captures **all 49 intermediate representations** (embedding layer + 48 transformer layers) and projects them through a learned feature extractor. This required augmenting sglang-jax's existing `Gemma2Model` with:

- **QK normalization** (RMSNorm on queries and keys per attention head) to match Gemma-3's architecture
- **Per-layer RoPE** instead of shared RoPE across layers
- **Auxiliary hidden state capture** (`capture_aux_hidden_states=True`) to collect outputs from every layer

#### Design Decision: Causal (Not Bidirectional) Attention

Early iterations used bidirectional attention (encoder-style), which broke text conditioning entirely. The root cause: the LTX-2 connector was trained on hidden states produced by **causal** attention. Bidirectional attention produces different hidden states — the feature extractor then projects them to near-zero, making `v_cond ≈ v_uncond` and eliminating all text guidance.

**Fix**: Use standard causal (autoregressive) attention. This also unlocks a simplification — since RoPE is shift-invariant, raw unpadded tokens with causal attention produce **identical hidden states** to PyTorch's left-padded + attention_mask approach. We eliminate input padding entirely and use gather-based connector input construction for variable-length sequences.

#### Feature Extractor (`GemmaFeaturesExtractorProjLinear`)

1. Stack all 49 hidden states: `[T, 3840, 49]`
2. Per-request normalization: mean-center and scale to range 8.0, computed independently per request to prevent cross-contamination when positive/negative prompts are batched together
3. Linear projection: `3840 × 49 → 3840` via learned `aggregate_embed` weight

#### Dual Connectors (`Embeddings1DConnector`)

Two independent connector instances (video + audio) with identical architecture but separate weights:

- **2 transformer blocks** with self-attention (30 heads × 128 dim/head = 3840)
- **128 learnable register tokens** that replace zero-padded positions via tiling (8× tiles to fill 1024 positions)
- **Interleaved RoPE** positional embeddings (different from the DiT's split RoPE)
- QK normalization in attention
- Output: `[1024, 3840]` conditioning embeddings per prompt per modality

The register tokens are critical: the DiT was trained with fixed-length 1024-token text embeddings, so padding beyond real tokens must be semantically meaningful rather than zero-valued.

#### Weight Loading Strategy

- Gemma backbone weights loaded from `google/gemma-3-12b-it` HuggingFace snapshot
- Connector weights loaded from the LTX-2 checkpoint (`ltx-2-19b-dev.safetensors`)
- After `materialize_abstract_state()` zeros unfilled abstract params, connector norm scales are explicitly set to 1.0 (PyTorch default for `nn.LayerNorm`/`nn.RMSNorm` weight, but checkpoint omits them)
- Uses `setattr(current, key, nnx.Param(tensor))` for weight assignment rather than in-place mutation, which proved more reliable with Flax NNX's parameter tracking

---

#### Architecture: `LTX2Transformer3DModel`

A 48-layer dual-modality transformer with:

| Dimension | Video Path | Audio Path |
|-----------|-----------|------------|
| Hidden size | 4096 | 2048 |
| Attention heads | 32 | 32 |
| Head dim | 128 | 64 |
| FFN inner dim | 16,384 | 8,192 |
| Cross-attn dim | 4096 | 2048 |

Each `LTX2TransformerBlock` contains 7 sub-modules:

1. **Video self-attention** — with 3D interleaved RoPE and STG perturbation masking
2. **Video cross-attention** — text conditioning (no RoPE)
3. **Audio self-attention** — with 1D temporal RoPE and STG masking
4. **Audio cross-attention** — text conditioning
5. **Bidirectional audio-video cross-attention** — A2V + V2A with modality masking
6. **Video FFN** — GELU-tanh gated, 4096→16384→4096
7. **Audio FFN** — GELU-tanh gated, 2048→8192→2048

All sub-modules use **AdaLN** (Adaptive Layer Normalization) conditioning from a shared timestep embedding.

#### Design Decision: 3D Interleaved RoPE (Not Concatenated)

The RoPE implementation was one of the most critical correctness fixes. The naive approach concatenates per-axis frequencies: `[f_T₀..f_T_N, f_H₀..f_H_N, f_W₀..f_W_N]`. The correct approach **interleaves** them: `[f_T₀, f_H₀, f_W₀, f_T₁, f_H₁, f_W₁, ...]`.

This ensures each attention head receives a **mix of temporal, spatial-height, and spatial-width** frequency bands, rather than some heads seeing only temporal frequencies and others seeing only spatial. The fix reduced max activation difference from ~2.0 to ~0.01 versus the PyTorch reference.

**Implementation**: `jnp.swapaxes(freqs, -1, -2)` before the final reshape, converting `[num_tokens, 3_axes, 682_bands]` → `[num_tokens, 682_bands, 3_axes]` → flatten.

#### Design Decision: Split RoPE Format

LTX-2 uses **split** RoPE (not interleaved-pair RoPE): `cos/sin` tensors have shape `[1, num_heads, num_tokens, head_dim/2]`, applied via:

```
x_even, x_odd = split(x, 2, axis=-1)
rotated = concat([-x_odd, x_even], axis=-1)
output = x * cos + rotated * sin
```

This is applied on the **flat** representation `[B, T*H*W, 4096]` before per-head reshape, preserving the interleaved frequency assignment across heads.

#### 4-Pass Guided Denoising

LTX-2 uses up to **4 simultaneous forward passes** per denoising step for multi-signal guidance:

| Batch Index | Pass | Self-Attention | Cross-Modal Attention | Purpose |
|-------------|------|---------------|----------------------|---------|
| 0 | Conditional | Normal | Normal | Text-guided prediction |
| 1 | Unconditional | Normal | Normal | Null-text baseline (negative prompt) |
| 2 | Perturbed | **Zeroed in block 29** | Normal | STG: spatial-temporal deviation |
| 3 | Isolated modality | Normal | **Zeroed (A2V + V2A)** | Modality: cross-modal contribution |

The masks are implemented as multiplicative scalars in the attention output:
- **STG mask**: `[1, 1, 0, 1]` — zeros self-attention output for batch element 2 in STG blocks
- **Modality mask**: `[1, 1, 1, 0]` — zeros A2V/V2A cross-attention output for batch element 3

#### Design Decision: Guidance in x0 Space (Not Velocity Space)

This was the second most impactful correctness fix. LTX-2's DiT outputs **velocity** `v`, but guidance (CFG + STG + modality) must be applied in **predicted-x0 space**, not velocity space:

```
# Step 1: Convert velocity → x0
x0_cond  = x_t - v_cond  × σ
x0_uncond = x_t - v_uncond × σ
x0_ptb   = x_t - v_ptb   × σ
x0_mod   = x_t - v_mod   × σ

# Step 2: Apply guidance in x0 space
x0_guided = x0_cond
          + (cfg_scale - 1) × (x0_cond - x0_uncond)       # CFG
          + stg_scale       × (x0_cond - x0_ptb)           # STG
          + (mod_scale - 1) × (x0_cond - x0_mod)           # Modality

# Step 3: Variance rescaling (prevents saturation)
factor = 0.7 × std(x0_cond) / std(x0_guided) + 0.3
x0_guided *= factor

# Step 4: Convert back to velocity
v_guided = (x_t - x0_guided) / σ

# Step 5: Euler step
x_{t+1} = x_t + v_guided × (σ_{t+1} - σ_t)
```

**Why x0 space?** Velocity predictions at different guidance conditions have correlated noise components. Subtracting velocities directly amplifies noise rather than signal. Converting to x0 first denoises the predictions, making the guidance operate on clean signal differences. The WAN 2.1 model uses FlowUniPC which handles this conversion internally, so it correctly operates in velocity space — this difference is gated by `use_x0_guidance` based on scheduler type.

#### Euler Scheduler with Dynamic Sigma Shifting

The sigma schedule adapts to **latent token count** (resolution-dependent):

```
μ = base_shift + (tokens - 1024) × (max_shift - base_shift) / (4096 - 1024)
σ_shifted = exp(μ) / (exp(μ) + (1/σ - 1))
```

- `base_shift = 0.95` (for 1024 tokens ≈ low resolution)
- `max_shift = 2.05` (for 4096 tokens ≈ high resolution)
- Higher resolution → more aggressive noise schedule → better convergence

After shifting, sigmas are stretched to reach the terminal noise level (0.1) and the ODE is integrated with first-order Euler steps.

#### Per-Head Gating

Audio-video cross-attention uses **per-head sigmoid gating**: `output = 2 × sigmoid(gate_logits) × attn_output` where `gate_logits` are projected from `Linear(query_dim → num_heads)`. The 2× factor allows the gate to amplify (not just attenuate), matching the PyTorch reference. Early iterations had wrong projection dimensions and missing scaling.

---

#### Architecture

The video VAE decoder reconstructs RGB frames from latent codes with **8× temporal** and **32× spatial** upsampling:

```
Input: [B, T', H', W', 128] latents (e.g., [1, 11, 24, 40, 128])
  → Per-channel statistics denormalization
  → Input conv: 128 → feature_channels
  → 7 decoder blocks:
      Block 0: UNetMidBlock3D (5 ResBlock3D layers)
      Block 1: DepthToSpaceUpsample (2×2×2) ← temporal + spatial
      Block 2: UNetMidBlock3D (5 ResBlock3D layers)
      Block 3: DepthToSpaceUpsample (2×2×2)
      Block 4: UNetMidBlock3D (5 ResBlock3D layers)
      Block 5: DepthToSpaceUpsample (2×2×2)
      Block 6: UNetMidBlock3D (5 ResBlock3D layers)
  → Output norm + SiLU + conv: channels → 3 × 16
  → Unpatchify: move channels → 4×4 spatial patches
Output: [B, T, H, W, 3] RGB video
```

#### Design Decision: DepthToSpace (Not Interpolation) for Upsampling

LTX-2 uses **depth-to-space rearrangement** for upsampling: channels are reshaped into spatial/temporal dimensions. This avoids interpolation artifacts and is compute-efficient on TPU. Each `DepthToSpaceUpsample` block doubles T, H, and W simultaneously with a learned 3D convolution followed by channel rearrangement.

#### Design Decision: `jnp.tile` (Not `jnp.repeat`) for Residual Connections

The residual path in `DepthToSpaceUpsample` duplicates channels to match the upsampled output. PyTorch's `tensor.repeat()` **tiles** the entire tensor (`[A,B,C] → [A,B,C,A,B,C]`), while JAX's `jnp.repeat()` repeats each **element** (`[A,B,C] → [A,A,B,B,C,C]`). Using `jnp.repeat` produced a visible grid/checkerboard artifact in all generated videos. The fix: `jnp.tile(x_in, (1,) * (ndim-1) + (num_repeat,))`.

#### 3D Causal Convolutions

All convolutions in the video VAE use **causal temporal padding**: the first frame is duplicated as left-padding to prevent information leakage from future frames. Spatial padding uses reflect mode. This matches the encoder's causal convention.

#### Unpatchify

The final output projection produces `3 × patch_t × patch_h × patch_w` channels, which are rearranged back to spatial dimensions:
```
[B, T, H, W, 3×1×4×4] → reshape → transpose → [B, T×1, H×4, W×4, 3]
```

The transpose ordering `(0, 1, 5, 2, 7, 3, 6, 4)` was derived by matching PyTorch's einops pattern where `r→W, q→H` (H and W are swapped relative to the naive ordering).

---

#### Audio VAE Decoder

Decodes audio latents to mel spectrograms using **2D causal convolutions** where the height axis represents time (causality axis) and width represents frequency:

```
Input: [B, T_a, F, 8] latents
  → Denormalize via per-channel statistics
  → 3-level upsampling decoder (mirroring encoder)
  → Output: [B, T_audio, 64, 2] stereo mel spectrogram
```

The causality convention (height = time) is unconventional but matches the PyTorch reference. All padding is applied asymmetrically: full padding on the top (past), none on the bottom (future).

#### HiFi-GAN Vocoder

Converts mel spectrograms to waveforms with **240× total upsampling** (upsample rates: 6×5×2×2×2):

```
Input: [B, T, 64, 2] mel spectrogram
  → Reshape: [B, T, 128] (interleave stereo channels × mel bins)
  → conv_pre: 128 → 1024
  → 5 upsample stages:
      Each: ConvTranspose1d ↑ → 3× parallel ResBlock1 (dilations 1,3,5) → average
      Channels: 1024 → 512 → 256 → 128 → 64 → 32
  → conv_post: 32 → 2 (stereo) → tanh
Output: [B, audio_length, 2] stereo waveform @ 24kHz
```

**ConvTranspose1d kernel flipping**: JAX's `ConvTranspose` stores kernels in the opposite spatial order from PyTorch. After loading weights, we flip along axis 0 to correct this — without this fix, the vocoder produces white noise.

---

### 5. Latent Upsampler (Optional)

An optional 2× spatial upsampling module that operates in latent space before VAE decoding:

```
[B, 128, F, H, W] → Conv3d(128→1024) → GroupNorm → SiLU
  → 4× ResBlock3d → [per-frame Conv2d + PixelShuffle2D(2×)] → 4× ResBlock3d
  → Conv3d(1024→128) → [B, 128, F, 2H, 2W]
```

This enables generating at 2× spatial resolution without retraining the DiT, at the cost of additional compute in the upsampler.

---

#### Config Registry

LTX-2 is registered in the config registry with two variants:
- `Lightricks/LTX-2` → `LTX2ModelConfig` (audio-video, full model)
- `Lightricks/LTX-Video` → `LTX2VideoOnlyModelConfig` (video-only, no audio path)

#### Stage Configuration (YAML)

Stage configs define TPU allocation and model classes:
```yaml
- stage_id: 0  # Text encoder
  model_class: LTX2GemmaTextEncoder
  num_tpus: 2
  scheduler: text_encoder

- stage_id: 1  # DiT
  model_class: LTX2Transformer3DModel
  num_tpus: 1
  scheduler: diffusion

- stage_id: 2  # VAE
  model_class: AutoencoderKLLTX2Video
  num_tpus: 1
  scheduler: vae
```

#### Runner Registry

A new `LTX2DiffusionModelRunner` entry in the runner registry maps LTX-2 to its runner, which extends the base `DiffusionModelRunner` with LTX-2-specific guidance logic, Euler scheduling, and audio handling.

#### Checkpoint Isolation

The LTX-2 HuggingFace repo contains multiple safetensors variants (FP4, FP8, distilled). The `WeightLoader` scans all `*.safetensors` in a directory, causing key conflicts. Solution: `get_ltx2_checkpoint_dir()` creates a temp directory with a symlink to only `ltx-2-19b-dev.safetensors`, isolating the main checkpoint.

#### Modifications to Shared Code

Minimal changes to existing code (62 lines modified across shared files):
- `gemma2.py`: QK norm support, per-layer RoPE, auxiliary hidden state capture
- `hf_transformers_utils.py`: Gemma-3 tokenizer path handling
- `model_loader/loader.py`: LTX-2 checkpoint format detection
- `layers/attention/layer.py`: Attention mask support in `simple_attention`
- `layers/visual_embedding.py`: LTX-2 visual embedding dispatch
- `server_args.py`: Default tokenizer path for LTX-2

---

### 7. Key Design Decisions and Justifications

| Decision | Rationale |
|----------|-----------|
| **Causal attention in Gemma** (not bidirectional) | Connector was trained on causal hidden states; bidirectional produces different features → zero guidance signal |
| **x0-space guidance** (not velocity space) | Velocity subtraction amplifies noise; x0 subtraction operates on clean signal. WAN uses velocity space because FlowUniPC handles conversion internally |
| **Interleaved RoPE** (not concatenated) | Each head needs mixed T/H/W frequencies for proper 3D position encoding. Concatenation assigns each head to a single axis |
| **`jnp.tile` for residual channels** (not `jnp.repeat`) | PyTorch `repeat` = tile semantics; `jnp.repeat` = element-wise repeat. Wrong choice causes checkerboard artifacts |
| **Explicit norm initialization to 1.0** | `materialize_abstract_state` zeros all unfilled params; RMSNorm with scale=0 kills the signal. Checkpoint omits these weights |
| **4-pass batched guidance** (not sequential) | Single batched forward pass is more TPU-efficient than 4 sequential passes. STG/modality masks are cheap multiplicative ops |
| **Gather-based text input** (not `dynamic_slice`) | `dynamic_slice` requires compile-time-known sizes; gather handles variable-length sequences naturally in JAX |
| **Temp directory for checkpoint isolation** | Prevents WeightLoader from loading quantized variant weights alongside main checkpoint |
| **Audio FFN dim = 8192** (not 16384) | Audio hidden size is 2048, so FFN should be 4× = 8192. Using video's 16384 wastes compute and doesn't match trained weights |
| **Variance rescaling factor 0.7** | Blends 70% conditional std + 30% guided std to prevent saturation from aggressive guidance scales |

---

### 8. Testing

Two test suites are included:

- **`test_ltx2_e2e.py`**: Full end-to-end integration test that launches the server, sends a video generation request (`9 frames, 256×256, 4 steps`), and verifies successful completion
- **`test_ltx2_models.py`**: Model-level integration test with identical structure but different server parameters

Both tests use minimal frame counts and inference steps for fast CI validation while exercising the complete pipeline.

---

### 9. Server Launch

```bash
HF_HOME=/mnt/disks/persist/hf_cache PYTHONPATH=python python3 -m sgl_jax.launch_server \
    --model-path Lightricks/LTX-2 --host 127.0.0.1 --port 8100 --device tpu \
    --trust-remote-code --skip-server-warmup --multimodal \
    --tokenizer-path google/gemma-3-12b-it --download-dir /mnt/disks/persist/hf_cache \
    --tp-size 8 --enable-single-process --max-total-tokens 32768 --context-length 8192
```

### API

```bash
curl http://127.0.0.1:8100/api/v1/videos/generation \
  -d '{"prompt":"A dog running through snow","num_frames":81,"size":"1280*768",
       "num_inference_steps":50,"guidance_scale":3.0,"save_output":true}'
```

---

### New Files

<details>
<summary>Click to expand full file list (38 new files)</summary>

**DiT (Diffusion Transformer)**
- `models/ltx2/__init__.py`
- `models/ltx2/diffusion/__init__.py`
- `models/ltx2/diffusion/ltx2_dit.py` (1,244 lines) — Main transformer model
- `models/ltx2/diffusion/ltx2_dit_weights_mapping.py` (631 lines) — Weight mappings with sharding specs
- `models/ltx2/diffusion/euler_step.py` (101 lines) — Velocity/x0 conversion utilities
- `models/diffusion_solvers/euler_scheduler.py` (144 lines) — Euler ODE scheduler

**Text Encoder**
- `models/ltx2/text_encoders/__init__.py`
- `models/ltx2/text_encoders/ltx2_text_encoder.py` (721 lines) — Gemma + connector pipeline

**Video VAE**
- `models/ltx2/vaes/__init__.py`
- `models/ltx2/vaes/ltx2_vae_config.py` (45 lines)
- `models/ltx2/vaes/ltx2_vae_decoder.py` (948 lines) — 3D causal decoder
- `models/ltx2/vaes/weight_mappings.py` (100 lines)

**Audio VAE + Vocoder**
- `models/ltx2/audio_vae/__init__.py`
- `models/ltx2/audio_vae/ltx2_audio_vae.py` (700 lines) — 2D causal encoder/decoder
- `models/ltx2/audio_vae/ltx2_audio_vae_config.py` (48 lines)
- `models/ltx2/audio_vae/vocoder.py` (295 lines) — HiFi-GAN
- `models/ltx2/audio_vae/weight_mappings.py` (274 lines)

**Latent Upsampler**
- `models/ltx2/upsampler/__init__.py`
- `models/ltx2/upsampler/latent_upsampler.py` (234 lines)

**Utilities**
- `models/ltx2/utils.py` (80 lines) — HF cache discovery

**Configs**
- `configs/dits/ltx2_model_config.py` (111 lines)
- `models/static_configs/ltx2_audio_video_stage_config.yaml`
- `models/static_configs/ltx2_video_only_stage_config.yaml`

**Schedulers**
- `manager/scheduler/text_encoder_scheduler.py` (142 lines)
- `manager/scheduler/vae_scheduler.py` (130 lines)

**Runner**
- `model_executor/diffusion/runner_registry.py` (24 lines)

**Tests**
- `test/srt/multimodal/ltx2/test_ltx2_e2e.py`
- `test/srt/multimodal/test_ltx2_models.py`

</details>

---

### Test Plan

- [x] End-to-end server launch + video generation request completes successfully
- [x] Text conditioning verified: videos follow text prompts (tested with "dog running through snow" and "car driving into sunset")
- [x] Audio generation produces audible, prompt-correlated stereo output
- [x] Velocity magnitudes verified: `|v_cond - v_uncond|` mean ~0.28 at step 0 (was ~0.077 before text conditioning fix)
- [x] No checkerboard/grid artifacts in decoded video (tile vs repeat fix)
- [x] 8-way tensor parallelism on TPU v7 runs without sharding errors


pengchengneo:
@chandrasekhard2  Could you describe your code design and the results of the end-to-end verification in detail?

chandrasekhard2:
Hi @pengchengneo - Can you review the PR?

PR #846, opened 2026-02-27
gmm kernel update

Bob-Chen222:
## Motivation
update gmm kernel for better performance

## Modifications

- gmm_v1

Synced with the latest upstream changes from tpu-inference

Added a benchmark script to tune block sizes

Added Grok-2–specific tuned parameters

- gmm_v2

Integrated the latest tpu-inference implementation reference

Forced activation-quantization accumulation to FP32 to preserve accuracy

- Quantized matmul (general)

Set accumulation dtype to FP32, which resolves the accuracy regression

- Stability

Added safeguards in quantize_tensor_simple to prevent NaNs

And with the integration comes a side effect in transposing the weights being loaded and it seems to slow down the weight loading performance. May need to spend more time on to it.

Moreover, the original gmm_v2 kernel will generate NaN when block_abs_max is too small. https://github.com/vllm-project/tpu-inference/blob/4fb92e343a691d66fa18645307dc826979671382/tpu_inference/kernels/megablox/gmm_v2.py#L270-L279

And even with the guard of adding smallest_subnormal under some edge cases it will still fail. Will raise it to google.

## Accuracy Tests

Tested both gmm_v1 and gmm_v2 under w8a8 quantization and all of them achieved accuracy 0.5-0.55 on gpqa_diamond. Will add more quantization tests into CI with harder benchmark

## Benchmarking and Profiling
The performance using only gmm_v1 on grok-2
Screenshot 2026-03-04 at 11 20 21 PM

Due to some limitations, gmm_v2 currently can't be used with small batch sizes so the best is to mix the usage of gmm_v1 and gmm_v2
Image below shows the performance of mixing gmm_v1 and gmm_v2
Screenshot 2026-03-04 at 11 18 28 PM

## Checklist

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -12,13 +12,8 @@ from sgl_jax.srt.eplb.expert_location import (
 )
 from sgl_jax.srt.kernels.fused_moe.v1.kernel import FusedMoEBlockConfig, fused_ep_moe
 from sgl_jax.srt.kernels.gmm.megablox_gmm_backend import gmm
-from sgl_jax.srt.kernels.gmm.megablox_gmm_kernel.gmm_v2 import is_supported_by_gmm_v2
-from sgl_jax.srt.utils.jax_utils import is_tpu_runtime
 from sgl_jax.srt.utils.profiling_utils import named_scope
-from sgl_jax.srt.utils.quantization.quantization_utils import (
-    quantize_tensor,
-    quantize_tensor_simple,
-)
+from sgl_jax.srt.utils.quantization.quantization_utils import quantize_tensor
 from sgl_jax.srt.utils.weight_utils import WeightMapping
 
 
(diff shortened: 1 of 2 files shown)


python/sgl_jax/srt/layers/moe.py
```diff
@@ -1147,13 +1145,13 @@ def create_moe_weights_mapping(
...
-            # wo (Output projection) uses P("expert", None, "tensor")
+            # Weights are transposed from HF [n, k] to [k, n], stacked to [g, k, n].
+            # wi_0/wi_1: [g, hidden_size, intermediate_dim] -> P("expert", None, "tensor")
+            # wo:        [g, intermediate_dim, hidden_size] -> P("expert", "tensor", None)
             sharding = (
-                ("expert", None, "tensor") if target_name == "wo" else ("expert", "tensor", None)
+                ("expert", "tensor", None) if target_name == "wo" else ("expert", None, "tensor")
             )
-            transpose = False
+            transpose = True
```
JamesBrianD, line 1106: I think this might be related to the slowdown when loading weights. _create_stacked_split_moe_lazy_tensor needs to transpose the tensor first and then concatenate it to form the final tensor.

Bob-Chen222: Yeah I tested and the transpose is the major bottleneck. I am thinking about to add implementation to transpose the weights and store them onto the local storage for better performance. Would this be a good idea?

JamesBrianD: Move the transpose operation from _load_single_expert_slice to the end of _load_stacked_slice, so that:
- Each expert returns chunks in the original layout (no transpose).
- All expert chunks are first written into a temporary out_array in the original layout (contiguous writes).
- A single swapaxes(-1, -2) (or an explicit-axis transpose) is applied to the whole batch at the end.

PR #868, opened 2026-03-10
feat: support split KV cache for different K/V head dimensions

### LINKED ISSUE #876: [Feature] Support split KV cache for different K/V head dimensions
### Checklist

### Motivation

Models with MLA-style architectures (e.g. DeepSeek-V2, MiMo-V2-Flash) have different head dimensions for K and V (e.g. K=192, V=128). The existing fused KV cache interleaves K/V heads as `[K0,V0,K1,V1,...]`, which requires identical head_dim for K and V. This limitation prevents serving models that use heterogeneous head dimensions.

### Proposed Solution

Add a **split KV path** that stores K and V in separate buffers, enabling support for different K/V head dimensions:

1. **New split Pallas attention kernel** — separate VMEM scratch buffers and independent DMA for K/V, with heterogeneous QK matmul (`common_dim = min(head_dim, k_head_dim)`)
2. **Memory pool split mode** — `MHATokenToKVPool` with `v_head_dim` parameter; stores separate `k_buffer` / `v_buffer` when `head_dim != v_head_dim`
3. **FlashAttention backend split path** — routes to split kernel when `is_split` is True
4. **Model runner alignment** — pre-aligns head_dim to 128 at model_runner level before pool construction

### Related resources

- PR: #868

cjx0709:
## Motivation

Models (e.g. DeepSeek-V2, MiMo-V2-Flash) have different head dimensions for K and V (e.g. K=192, V=128). The existing fused KV cache interleaves K/V heads as `[K0,V0,K1,V1,...]`, which requires identical head_dim. This PR adds a split KV path that stores K and V in separate buffers, enabling support for heterogeneous head dimensions.

### New: Split Pallas Kernel (`ragged_paged_attention_split.py`)

New Pallas kernel for attention with separate K/V cache, adapted from the fused kernel with these key changes:

- **Separate VMEM scratch buffers**: `bk_double_buf` + `bv_double_buf` instead of one fused `bkv_double_buf`, allowing different K/V head_dim
- **Independent DMA transfers**: `_fetch_bkv` copies from two HBM arrays (k_cache, v_cache) into separate VMEM buffers
- **Heterogeneous QK matmul**: uses `common_dim = min(head_dim, k_head_dim)` to handle Q and K having different dimensions
- **PV result padding**: P@V output has `v_head_dim`, padded to `head_dim` (Q's dim) for accumulator compatibility since output aliases Q's HBM buffer
- **Separate KV head loading**: `strided_load_kv_separate` reads from independent K/V buffers instead of strided extraction from interleaved data
- **Separate cache writeback**: `_update_kv_cache` writes new tokens back to k_cache and v_cache independently

Input preparation utilities:
- `_prepare_single_kv`: pad heads and head_dim to alignment for a single K or V tensor
- `_prepare_single_kv_cache`: same for a single K or V cache tensor

### Modified: Dispatch Layer (`ragged_paged_attention.py`)

- Added dispatch: when `k_cache` and `v_cache` are both provided, routes to `ragged_paged_attention_split`
- Removed dead `prepare_kv_cache` function (replaced by `prepare_kv_cache_fused` and split-specific utilities)

### Modified: FlashAttention Backend (`flashattention_backend.py`)

- Added `_call_split` method and `v_head_dim` parameter
- Routes to split path when `token_to_kv_pool.is_split` is True
- Returns `(output, (updated_k, updated_v))` tuple instead of `(output, updated_kv_fused)`

### Modified: Memory Pool (`memory_pool.py`)

- `MHATokenToKVPool`: added `v_head_dim` parameter and `is_split` flag (`head_dim != v_head_dim`)
- Split mode stores separate `k_buffer` and `v_buffer` lists instead of fused `kv_buffer`
- `set_kv_buffer` / `set_kv_buffer_legacy`: pads incoming K/V tokens from model's actual dim to buffer's pre-aligned dim before writing
- `get_split_kv_buffer()`: returns separate K/V buffers for split path
- `SWAKVPool`: passes through `swa_head_dim` / `swa_v_head_dim` to sub-pools
- Head_dim 128-alignment is done at `model_runner` level (pool receives pre-aligned dims), keeping the pool layer simple

### Modified: Model Runner (`model_runner.py`)

- `init_memory_pool`: aligns `head_dim` / `v_head_dim` to 128 before passing to pool constructors (both `MHATokenToKVPool` and `SWAKVPool`)
- `profile_max_num_token`: accounts for split storage (K and V may have different aligned dims)
- `_set_kv_cache_after_forward`: handles `(k, v)` tuple for split path

### Kernel correctness (`TestSplitKernelAttention` — 13 tests)

All compared against a pure-JAX reference implementation (`ref_split_attention`).

| Test | head_dim | v_head_dim | mode | page_size | heads |
|------|----------|------------|------|-----------|-------|
| `test_ref_split_matches_ref_fused_when_same_dim` | 128 | 128 | — | 1 | MHA |
| `test_split_kernel_prefill_ps1` | 256 | 128 | prefill | 1 | MHA |
| `test_split_kernel_decode_ps1` | 256 | 128 | decode | 1 | MHA |
| `test_split_kernel_gqa_prefill_ps1` | 256 | 128 | prefill | 1 | GQA |
| `test_split_kernel_gqa_decode_ps1` | 256 | 128 | decode | 1 | GQA |
| `test_split_kernel_gqa_prefill_ps16` | 256 | 128 | prefill | 16 | GQA |
| `test_split_kernel_gqa_decode_ps16` | 256 | 128 | decode | 16 | GQA |
| `test_split_kernel_192_prefill_ps1` | 192 | 128 | prefill | 1 | MHA |
| `test_split_kernel_192_decode_ps1` | 192 | 128 | decode | 1 | MHA |
| `test_split_kernel_192_gqa_prefill_ps1` | 192 | 128 | prefill | 1 | GQA |
| `test_split_kernel_192_gqa_decode_ps1` | 192 | 128 | decode | 1 | GQA |
| `test_split_kernel_192_gqa_prefill_ps16` | 192 | 128 | prefill | 16 | GQA |
| `test_split_kernel_192_gqa_decode_ps16` | 192 | 128 | decode | 16 | GQA |

### Backend end-to-end (`TestSplitBackendAttention` — 16 tests)

FlashAttention backend + MHATokenToKVPool + RadixAttention full stack.

| Test | head_dim | v_head_dim | mode | page_size | notes |
|------|----------|------------|------|-----------|-------|
| `test_split_backend_prefill_ps1` | 256 | 128 | prefill | 1 | |
| `test_split_backend_decode_ps1` | 256 | 128 | decode | 1 | |
| `test_split_backend_gqa_prefill_ps1` | 256 | 128 | prefill | 1 | |
| `test_split_backend_gqa_decode_ps1` | 256 | 128 | decode | 1 | |
| `test_split_backend_gqa_prefill_ps16` | 256 | 128 | prefill | 16 | |
| `test_split_backend_gqa_decode_ps16` | 256 | 128 | decode | 16 | |
| `test_split_backend_cache_update` | 256 | 128 | prefill | 1 | verifies cache data |
| `test_split_backend_192_prefill_no_prefix` | 192 | 128 | prefill | 1 | no prefix tokens |
| `test_split_backend_192_prefill_no_prefix_mha` | 192 | 128 | prefill | 1 | MHA, no prefix |
| `test_split_backend_192_decode` | 192 | 128 | decode | 1 | |
| `test_split_backend_192_gqa_decode` | 192 | 128 | decode | 1 | |
| `test_split_backend_192_prefill_with_prefix` | 192 | 128 | prefill | 1 | with prefix tokens |
| `test_split_backend_192_gqa_prefill_ps16` | 192 | 128 | prefill | 16 | |
| `test_split_backend_192_gqa_decode_ps16` | 192 | 128 | decode | 16 | |
| `test_split_backend_192_cache_update` | 192 | 128 | prefill | 1 | verifies cache + padding zeros |

### KV cache pool (`TestSplitKVCacheUpdate` — 3 tests)

MHATokenToKVPool split mode set_kv_buffer correctness.

| Test | head_dim | v_head_dim | notes |
|------|----------|------------|-------|
| `test_split_pool_192_128_ps1` | 192 | 128 | basic write/read |
| `test_split_pool_256_128_ps1` | 256 | 128 | basic write/read |
| `test_split_pool_192_128_with_padding` | 192 | 128 | mixed valid/padding locs |

```bash
# Run all split KV tests
python -m pytest python/sgl_jax/test/test_split_kv_attention.py -v
python -m pytest python/sgl_jax/test/mem_cache/test_kv_cache.py::TestSplitKVCacheUpdate -v
```

---

Closes #876


python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention_split.py
```diff
@@ -0,0 +1,1054 @@
...
+            q.dtype,
+            k_cache_processed.dtype,
+            actual_num_q_heads,
+            actual_num_kv_heads,
+            head_dim,
+            page_size,
+            max_num_tokens,
+            pages_per_seq,
+            causal,
+        )
```
JamesBrianD, line 899: Does the block size for tuning need to be retested, specifically for the split kernel?

python/sgl_jax/srt/layers/attention/flashattention_backend.py
```diff
@@ -437,6 +445,13 @@ def __call__(
         Returns:
             Output tensor of shape [total_tokens, hidden_size]
         """
+        # Early branch: split KV path when head_dim != v_head_dim
+        is_split = (
+            getattr(token_to_kv_pool, "is_split", False) if token_to_kv_pool is not None else False
+        )
+        if is_split:
+            return self._call_split(q, k, v, layer, forward_batch, token_to_kv_pool, causal)
+
```
JamesBrianD: The fused and split here are two separate branches, but the early-branch approach is not a good pattern. Use this type:
```
    if isinstance(token_to_kv_pool, SplitKVCache):
        return self._call_split(q, k, v, layer, forward_batch, token_to_kv_pool, causal)
    elif isinstance(token_to_kv_pool, FusedKVCache):
        return self._call_fused(q, k, v, layer, forward_batch, token_to_kv_pool, causal)
```

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -246,32 +246,66 @@ def __init__(
...
+        self.v_head_dim = v_head_dim if v_head_dim is not None else head_dim
         self.kv_partition_axis = "tensor"
+        self.is_split = self.head_dim != self.v_head_dim
+
+        if self.is_split:
+            logger.info(
+                "[KVPool] split enabled: head_dim=%s v_head_dim=%s layer_num=%s size=%s page_size=%s",
+                self.head_dim,
+                self.v_head_dim,
+                self.layer_num,
```
JamesBrianD: Consider using separate classes (e.g., FusedKVBuffer and SplitKVBuffer) instead of the is_split if/else branching throughout MHATokenToKVPool. The current approach scatters conditional logic across every method (get_kv_buffer, set_kv_buffer, tree_flatten, clear_cache, etc.), which makes it easy to miss a branch when adding new methods. A strategy/subclass pattern would let each storage format be self-contained, eliminate the repeated branching, and make it straightforward to add new formats in the future.

cjx0709: good suggestion, now using SplitMHATokenToKVPool as the subclass of MHATokenToKVPool.

python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention_split.py
cjx0709: I tunned block config for v5e and v6e with mostly used qkv config setting. More tunned shape and config can add when someone else needed.

JamesBrianD: Ok, this way is solid.

JamesBrianD:
Great work. @Bob-Chen222 , please take a look at the memory pool changes — they might impact MLA development. I think it's good to go on my end.

JamesBrianD (dismissed):

python/sgl_jax/srt/layers/attention/flashattention_backend.py
```diff
@@ -437,6 +445,23 @@ def __call__(
         Returns:
             Output tensor of shape [total_tokens, hidden_size]
         """
+        if isinstance(token_to_kv_pool, SplitMHATokenToKVPool):
```
alexnails: won't this always default to call_fused? (re: `SWAKVPool` having not implemented get_fused_kv_buffer  and being unrelated to `SplitMHATokenToKVPool)

So if we try to support something like Deeseek V3.2(?) there will be crash. This is slightly out of scope for the PR or worth follow up PR, so totally fine to leave a TODO

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -469,6 +503,205 @@ def set_kv_buffer_legacy(
...
+            k=k,
+            v=v,
+            loc=loc,
+            k_cache=self.k_buffer[layer_idx],
+            v_cache=self.v_buffer[layer_idx],
+            page_size=page_size,
+            kv_partition_axis=self.kv_partition_axis,
+        )
+
+    def replace_kv_buffer(self, kv_buffer: list[tuple[jax.Array, jax.Array]]) -> None:
```
alexnails, line 659: no self.start_layer? (or is not necessary? edit: the other method uses it which is why I am asking)

python/sgl_jax/srt/kernels/ragged_paged_attention/tuned_block_sizes.py
alexnails, line 1768: 192 and 256 share the same entry from the above. I think this is non issue for now

Bob-Chen222 (approved):

python/sgl_jax/srt/layers/attention/flashattention_backend.py
cjx0709: This was indeed a bug while there wasn't an implementation for it at the moment. I fixed it and added documentation to explain the implementation of mempool.

python/sgl_jax/srt/mem_cache/memory_pool.py
cjx0709: It is necessary if we implement Pipeline Parallelism. Now not add self.start_layer because following the original fused case. I add a todo for this, do you think it is enough for now?

python/sgl_jax/srt/kernels/ragged_paged_attention/tuned_block_sizes.py
cjx0709: It has to be 128 align for pallas kernel limit, I add notes to describe this.

python/sgl_jax/srt/mem_cache/memory_pool.py
alexnails: yes TODO is fine!

Bob-Chen222 (approved):

PR #878, opened 2026-03-18
attention optimization by upgrading libtpu and jax

Bob-Chen222:
## Checklist


python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -557,6 +557,8 @@ def _get_attention_backend(self):
                 self.model_config.head_dim,
                 page_size=self.page_size,
                 mesh=self.mesh,
+                max_context_len=self.model_config.context_len,
+                # chunked_prefill_size=self.server_args.chunked_prefill_size,
```
JamesBrianD: There’s an issue here. TPU Inference pads all sequences to the context lengths to make KV indexing easier, whereas `sglang-jax` directly concatenates sequences of different lengths and uses `cu_kv` to index KV. So this context lengths approach may not be very suitable for `sglang-jax`.

Bob-Chen222: This makes sense. But may I ask what is the logistics of our design against theirs?

PR #884, opened 2026-03-30
feat: Implement GroupRMSNorm layer for Bailing Linear Attention support

### LINKED ISSUE #886: [RFC] GroupRMSNorm Layer Implementation
# GroupRMSNorm Layer Implementation

## Background

GroupRMSNorm is an internal component of Linear Attention, used to perform grouped normalization on attention outputs before output gating.

Linear Attention lacks the softmax constraint on numerical range, so output magnitudes can be unbounded. GroupRMSNorm normalizes before the gate to stabilize values. Within the same model, [MLA](https://huggingface.co/inclusionAI/Ling-2.5-1T/blob/main/modeling_bailing_moe_v2_5.py#L492) (which has softmax) does not need GroupRMSNorm, while [Linear Attention](https://huggingface.co/inclusionAI/Ling-2.5-1T/blob/main/modeling_bailing_moe_v2_5.py#L878-L883) (which has no softmax) does.

### Position in Linear Attention

Reference: [Ling-2.5-1T official implementation](https://huggingface.co/inclusionAI/Ling-2.5-1T/blob/main/modeling_bailing_moe_v2_5.py#L878-L883)

```
Q, K, V (projections)
  ↓
Linear Attention Kernel(Q, K, V)  →  produces o (shape: batch, seq, num_heads, head_v_dim)
  ↓
GroupRMSNorm(o)                   →  per-head group normalization (group_norm_size=8)
  ↓
× sigmoid(gate)                   →  gating
  ↓
Output Projection                 →  project back to hidden_size
```

Architecture

## Goal

Implement the GroupRMSNorm layer for [Ling-2.5-1T](https://huggingface.co/inclusionAI/Ling-2.5-1T) Linear Attention output gating.

## Implementation

JAX port of the following [PyTorch implementation](https://huggingface.co/inclusionAI/Ling-2.5-1T/blob/main/modeling_bailing_moe_v2_5.py#L185-L204):

```python
class BailingMoeV2_5GroupRMSNorm(nn.Module):
    def __init__(self, hidden_size, group_norm_size, eps=1e-6):
        """
        BailingMoeV2_5RMSNorm is equivalent to T5LayerNorm
        """
        super().__init__()
        self.weight = nn.Parameter(torch.ones(hidden_size))
        self.group_norm_size = group_norm_size
        assert hidden_size % group_norm_size == 0, "hidden_size must be divisible by group_norm_size"
        self.variance_epsilon = eps

    def forward(self, hidden_states):
        input_dtype = hidden_states.dtype
        input_shape = hidden_states.size()
        group_input_shape = input_shape[:-1] + (self.group_norm_size, input_shape[-1] // self.group_norm_size)
        hidden_states = hidden_states.view(group_input_shape)
        hidden_states = hidden_states.to(torch.float32)
        variance = hidden_states.pow(2).mean(-1, keepdim=True)
        hidden_states = hidden_states * torch.rsqrt(variance + self.variance_epsilon)
        return self.weight * hidden_states.to(input_dtype).view(input_shape)
```

## Testing

Test designed with reference to:
```python
@pytest.mark.skip(reason="group_rms_norm function removed from attention_gla")
class TestGroupRMSNorm:
  """GroupRMSNorm compared to numpy reference."""

  @pytest.mark.parametrize("group_size", [4, 8])
  def test_random(self, group_size: int) -> None:
    rng = np.random.default_rng(99)
    D = 128
    shape = (2, 16, D)
    x_np = rng.standard_normal(shape).astype(np.float32)
    w_np = rng.standard_normal(D).astype(np.float32)

    jax_out = group_rms_norm(jnp.array(x_np), jnp.array(w_np), group_size, eps=1e-6)  # pylint: disable=undefined-variable

    # Manual reference
    gs = group_size
    x_g = x_np.reshape(*shape[:-1], gs, D // gs)
    var = np.mean(x_g**2, axis=-1, keepdims=True)
    x_normed = x_g / np.sqrt(var + 1e-6)
    ref = w_np * x_normed.reshape(shape)
    _assert_close(jax_out, ref, msg=f"group_rms_norm gs={group_size}")
```

### Shape Preservation

Construct random input `(batch=2, seq=16, hidden=128)` with `num_groups=8`, assert output shape matches input.

### Group Independence

Construct two inputs that differ only in group 0 (first 16 dimensions), assert that the remaining 7 groups' outputs are unchanged and group 0's output has changed.

### Learnable Scale

Set random weights, compare JAX output against a NumPy reference implementation (`weight * normed_x`), assert agreement within `rtol=1e-6`.

### Cross-Framework Consistency

Compare JAX `GroupRMSNorm` against the official PyTorch `BailingMoeV2_5GroupRMSNorm` with identical inputs and weights:

- Output shapes match
- Numerical match with default weights (ones) at `rtol=1e-6`
- Numerical match with random weights
- 2D input `(batch, hidden)` compatibility
- Precision report: `max abs diff = 9.5e-7`, `mean abs diff = 1.8e-8`

Precision differences stem from different floating-point operation ordering between XLA and PyTorch backends — normal float32 rounding errors.

### Implementation Decisions

1. **JAX implementation aligned with official PyTorch**: reshape → per-group variance → rsqrt → reshape back → scale
2. **float32 upcast**: Consistent with the PyTorch implementation — upcast to float32 before computation, cast back to original dtype after normalization

zhengkezhou1:
## Motivation

fix #886

Implement `GroupRMSNorm` layer for [Bailing Linear Attention](https://huggingface.co/inclusionAI/Ling-2.5-1T/blob/main/modeling_bailing_moe_v2_5.py#L751-L752) support.

Used as a subcomponent of LinearAttention, called in the [forward](https://huggingface.co/inclusionAI/Ling-2.5-1T/blob/main/modeling_bailing_moe_v2_5.py#L878-L882) pass

This PR focuses on correctness only — faithfully porting the HuggingFace reference implementation to JAX. Performance optimizations (e.g., fusing gate computation via Pallas kernels) are out of scope and will be addressed in follow-up PRs

## Modifications

- Add `GroupRMSNorm` class in `python/sgl_jax/srt/layers/attention/fla/group_rmsnorm.py` that splits the hidden dimension into groups and applies RMSNorm independently within each group.

## Accuracy Tests

Cross-framework tests verify JAX output matches PyTorch reference (`BailingMoeV2_5GroupRMSNorm`)
```
==================================================
  Error bound verification (100 trials)
==================================================
  theoretical bound:   1.226664e-04
  JAX  max  abs err:   2.318312e-06
  JAX  mean abs err:   1.462548e-06
  JAX  max  rel err:   3.349566e-07
  JAX  mean rel err:   2.675635e-07
  Torch max  abs err:  1.998050e-06
  Torch mean abs err:  1.254086e-06
  Torch max  rel err:  2.699967e-07
  Torch mean rel err:  2.277907e-07
==================================================
```

## Checklist


python/sgl_jax/srt/layers/layernorm.py
```diff
@@ -217,6 +217,42 @@ def __call__(
         return x if residual is None else (x, residual)
 
 
+class GroupRMSNorm(nnx.Module):
```
alexnails: Grabbed this from JAX LLM, have you A/B tested against:

```
class GroupRMSNorm(nnx.Module):
    """Group RMS normalization.

    Splits the last dimension into `num_groups` groups and applies
    RMSNorm independently within each group.
    """

    def __init__(
        self,
        hidden_size: int,
        num_groups: int = 8,
        epsilon: float = 1e-6,
        param_dtype: Dtype = jnp.float32,
        scope_name: str = "group_rms_norm",
    ):
        if hidden_size % num_groups != 0:
            raise ValueError("hidden_size must be divisible by num_groups")

        self.hidden_size = hidden_size
        self.num_groups = num_groups
        self.group_size = hidden_size // num_groups
        self.epsilon = epsilon
        self.name = scope_name
        self.weight = nnx.Param(
            jnp.ones(hidden_size, dtype=param_dtype),
        )

    @named_scope
    def __call__(
        self, hidden_states: jax.Array, residual: jax.Array | None = None
    ) -> jax.Array | tuple[jax.Array, jax.Array]:
        orig_dtype = hidden_states.dtype
        orig_shape = hidden_states.shape

        if residual is not None:
            hidden_states = hidden_states + residual.astype(hidden_states.dtype)
            residual = hidden_states

        hidden_states = hidden_states.reshape(
            *orig_shape[:-1], self.num_groups, self.group_size
        ).astype(jnp.float32)
        variance = jnp.mean(lax.square(hidden_states), axis=-1, keepdims=True)
        weight = jnp.asarray(self.weight.value, jnp.float32).reshape(
            self.num_groups, self.group_size
        )
        hidden_states = (
            (weight * hidden_states * lax.rsqrt(variance + self.epsilon))
            .reshape(orig_shape)
            .astype(orig_dtype)
        )

        return hidden_states if residual is None else (hidden_states, residual)
```

alexnails: Even simpler optimization. This HLO should be better (tested on GPU):

```python
# Current — casts to bf16 BEFORE weight multiply:
hidden_states = hidden_states.astype(orig_dtype).reshape(orig_shape)
return self.weight[...] * hidden_states

# Fixed — casts to bf16 AFTER weight multiply in f32:
hidden_states = hidden_states.reshape(orig_shape)
return (weight.astype(jnp.float32) * hidden_states).astype(orig_dtype)
```

zhengkezhou1: I intend to follow the Ling 2.5 implementation

JamesBrianD: I agree with @alexnails  that the interface of this implementation is not general enough, especially the residual argument, which is not appropriate in a general-purpose layer.

I would also suggest referring to the approaches in the following PRs:

- [sglang PR #10239](https://github.com/sgl-project/sglang/pull/10239)
- [sglang PR #18598](https://github.com/sgl-project/sglang/pull/18598)

From the implementation perspective, this GroupRMSNorm appears to be specifically tailored for FLA (Bailing Linear Attention). Therefore, it would be more appropriate to place it in its own layers directory, rather than inside the general LayerNorm module. This keeps the module structure clean, maintainable, and avoids confusing the interface.

zhengkezhou1: OK. Moved GroupRMSNorm to `srt/layers/attention/fla/layernorm_gated.py` and removed the residual parameter to keep the interface clean.

JamesBrianD:
Please check these questions:
1. Add the reference file link for the standard PyTorch implementation in the test comments.
2. Confirm whether group norm in sglang-jax needs to fuse the gate computation.
3. In the future, CI for layer tests can add selective execution logic so they run only when certain layers or kernels folders are modified.

Issue #661, opened 2026-01-15
[Feature] Make submodules’ dtype be configurable


aolemila:
### **Checklist**

- [ ] If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [ ] Please use English, otherwise it will be closed.

### **Motivation**

Control the precision scope when using higher precision, like float32.

* softmax
* proj in attn and mlp

### **Related resources**

iwknow:
Can you please provide more details about this feature? maybe i can work on it.

iwknow:
let my summarize my understanding, @aolemila please let me know if it is correct.

Basically, this feature request is about allowing more precise control on the dtype on each submodule of a model. Use python/sgl_jax/srt/models/llama.py as an example, While `LlamaDecoderLayer` currently has a dtype parameter, the granularity of that configuration is limited: a single data type is applied to every component within a layer. In `LlamaAttention`, the `q_proj`, `k_proj`, `v_proj`, and `o_proj` are all initialized with the same dtype provided to the parent layer. You cannot currently specify, for example, that the output projection (o_proj) should use `float32` for while the others remain in `bfloat16`.  so we'd like to introduce a `precision_config` or `dtype_config` so that the dtype for each of the components can be specified.

iwknow:
i'd love to. i will provide my design and implementation plan shortly.

aolemila:
Hi @iwknow , thanks for your design. Here are some suggestions.

1. Please refer to [contribution document](https://github.com/sgl-project/sglang-jax/blob/main/docs/developer_guide/contribution_guide.md). I should provide you early. Move the design into a google document and create a subissue to put your document link. After our discussion, it is better to create a pr to put your document content under the docs/feature/xxx.md. Then you can create another subissue to do the development. You can refer to this #639.

2. In the design document, you can refer to [this one](https://docs.google.com/document/d/1ei4LWKQsQoQB87LCNSfLCMPEbnxPWrewdUVlWalO3rk/edit?tab=t.0#heading=h.gtpewejqczi). It is better to include the user story, design, test plan.

3. Take the `lllama.py` as an example, `rotary_emb` is missing in `self_attn`. Tips: qwen2 and qwen3 use o_proj and qwen uses c_proj. `c_fc` does not exist in `qwen.py` or `llama.py`.

[...]

PR #888, opened 2026-04-02
Feat/dtype config

iwknow:
## Motivation
See details at https://github.com/sgl-project/sglang-jax/issues/661

## Modifications
This PR 
- added the design doc of this feature (docs/features/dtype_config.md)
- implemented DtypeConfig class for carrying dtype configurations(python/sgl_jax/srt/configs/dtype_config.py)
- modified SRT models and layers to adapt DtypeConfig
- Use LLAMA model as the first example to support configurable dtype
- Add unit tests and consistency tests.

## Accuracy Tests
- test/srt/test_dtype_config_consistency.py
- test/srt/test_dtype_config_llama.py

## Benchmarking and Profiling
N/A

## Checklist


iwknow:
kindly ping @sii-xinglong

pengchengneo:
@iwknow  thanks for ur contribution, I will review it soon

and please fix e2e test for [PR Test / e2e-test-1-tpu (0) (pull_request)](https://github.com/sgl-project/sglang-jax/actions/runs/23932743064/job/69807401955?pr=888)

aolemila:
> @aolemila a kindly ping

Sorry for missing your message. I will review it soon.

python/sgl_jax/srt/models/llama.py
```diff
@@ -377,18 +413,35 @@ def __init__(
...
         mesh: jax.sharding.Mesh,
         dtype: jnp.dtype = jnp.bfloat16,
+        dtype_config: DtypeConfig | None = None,
     ):
         self.mesh = mesh
         self.config = config
         self.dtype = dtype
-        logger.info("LlamaForCausalLM config dtype: %s", self.dtype)
-        self.model = LlamaModel(config, dtype=self.dtype, mesh=mesh)
+        self.dtype_config = dtype_config
```
aolemila: remove it after the following if else.

python/sgl_jax/srt/models/llama.py
```diff
@@ -377,18 +413,35 @@ def __init__(
...
+            dtype_config = DtypeConfig(default_dtype=dtype)
+        elif dtype != dtype_config.default_dtype:
+            raise ValueError(
+                f"Global dtype ({dtype}) is not the same as the default dtype provided in dtype_config ({dtype_config.default_dtype})."
+            )
+
+        if dtype_config is not None:
+            logger.info("LlamaForCausalLM using dtype_config: %s", self.dtype_config)
+        else:
+            logger.info("LlamaForCausalLM config dtype: %s", self.dtype)
```
aolemila: It looks like that this case will never reach.

aolemila:
`test_logprobs.py` in CI fails. The following modification may be influence the result. 

```python3
  Before (main):                                                            
  p = jnp.exp(s - broadcast_minor(m_curr, s.shape))    # p is float32
  p_rowsum = jnp.sum(p, axis=1, keepdims=True)          # float32 accumulation
  return p, exp_m_diff                                                                                
  # step2: pv = einsum(p, v, preferred_element_type=float32)  ← p is float32                          
                                                                                                      
  After (PR #888):                                                                                    
  p = jnp.exp(s - broadcast_minor(m_curr, s.shape))    # p is float32                                 
                                                                                                      
  if p.dtype != v.dtype:                                 # float32 != bfloat16 → True                 
      p = p.astype(v.dtype)                              # p downcast to bfloat16!                    
                                                                                                      
  p_rowsum = jnp.sum(p, axis=1, keepdims=True)          # now bfloat16 accumulation                   
  return p, exp_m_diff                                                             
  # step2: pv = einsum(p, v, preferred_element_type=float32)  ← p is bfloat16  
```

python/sgl_jax/srt/models/llama.py
iwknow: i removed `self.dtype_config` from the class-level variable making it a local variable in the `__init__` method

python/sgl_jax/srt/models/llama.py
iwknow: removed.

iwknow:
> `test_logprobs.py` in CI fails. The following modification may be influence the result.
> 
> ```python
>   Before (main):                                                            
>   p = jnp.exp(s - broadcast_minor(m_curr, s.shape))    # p is float32
>   p_rowsum = jnp.sum(p, axis=1, keepdims=True)          # float32 accumulation
>   return p, exp_m_diff                                                                                
>   # step2: pv = einsum(p, v, preferred_element_type=float32)  ← p is float32                          
>                                                                                                       
>   After (PR #888):                                                                                    
>   p = jnp.exp(s - broadcast_minor(m_curr, s.shape))    # p is float32                                 
>                                                                                                       
>   if p.dtype != v.dtype:                                 # float32 != bfloat16 → True                 
>       p = p.astype(v.dtype)                              # p downcast to bfloat16!                    
>                                                                                                       
>   p_rowsum = jnp.sum(p, axis=1, keepdims=True)          # now bfloat16 accumulation                   
>   return p, exp_m_diff                                                             
>   # step2: pv = einsum(p, v, preferred_element_type=float32)  ← p is bfloat16  
> ```

i removed `v` dtype cast for the following reasons:
- If the user chooses Performance (softmax_dtype=jnp.bfloat16): s is cast to bfloat16 at line 460. The exp operation produces p in bfloat16. Since v is also bfloat16, no cast is needed at line 468 anyway. The subsequent einsum will run at full speed using the TPU's bfloat16 units.
- If the user chooses Precision (softmax_dtype=jnp.float32): s remains (or is cast to) float32. The exp operation produces p in float32. By not casting p back to bfloat16, the subsequent multiplication with v will happen with p in higher precision, correctly fulfilling the user's request for higher precision at the cost of performance.

As the result
- Scores are computed in float32 (via einsum preferred type).
- Softmax (exp, max) happens in softmax_dtype (User choice: float32 or bfloat16).
- Accumulation of $P @ V$ happens with p in softmax_dtype () and v in model dtype (bfloat16),
- Final Division happens in float32 for stability, and the result is cast to the requested output dtype.

iwknow:
@aolemila really appreciated for reviewing! sorry for the late response. i've update the implementation to address the feedbacks. please take a look again.

aolemila:
> @aolemila really appreciated for reviewing! sorry for the late response. i've update the implementation to address the feedbacks. please take a look again.

Thanks for your reply. These days I am very busy. I will review it in May.

iwknow:
@aolemila a kindly ping.

cjx0709:
@iwknow hello Frank, I'm sorry for that we missed your message. Our main branch had a lot of new updates since your pr, can you just rebase the newest main, and then I will make a review.

python/sgl_jax/srt/configs/dtype_config.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def get_config(self, key: str) -> "DtypeConfig":
+        """Returns a child config covering the sub-dictionary, preserving the default."""
+        return DtypeConfig(
+            config_dict=self.config_dict.get(key, {}), default_dtype=self.default_dtype
+        )
+
+    def get_dtype(self, key: str) -> jnp.dtype | None:
+        """Returns the specific dtype, or falls back to the default."""
+        val = self.config_dict.get(key, self.default_dtype)
```
cjx0709, line 66: This is the root cause of the CI failure, and it's a design issue rather than just a missing `static_argnames`.

`get_dtype` falls back to `default_dtype` for any key not in `config_dict`. That works fine for weight params like `gate_proj` or `q_proj` — if the user doesn't specify, you want the default dtype. But for operational params like `softmax`, `None` means "don't do any extra cast", and falling back to `default_dtype` changes that meaning.

What happens today even without `--dtype-config`:
1. `LlamaForCausalLM.__init__` creates `DtypeConfig(default_dtype=bfloat16)` (empty config_dict)
2. Passes down via `get_config` to `LlamaAttention`
3. `dtype_config.get_dtype("softmax")` → config_dict is empty → falls back to `default_dtype` → returns `bfloat16` (not `None`!)
4. `RadixAttention` gets `softmax_dtype=bfloat16` → passed to v3 kernel → JIT crash (because it's not in `static_argnames`)

The CI test that failed is Qwen3-32B + EAGLE3 — the EAGLE3 draft model inherits `LlamaDecoderLayer`, so it hits this path even though the user never configured any dtype_config.

Fixing the v3 `static_argnames` would stop the crash, but there's still a behavioral change: all Llama-based models would start passing `softmax_dtype=bfloat16` into the kernel instead of `None`. Numerically it's a no-op (casting bfloat16 to bfloat16), but it's an unintended semantic change.

I think `get_dtype` needs to distinguish between "user explicitly set this" vs "not configured, use fallback". For softmax specifically, you probably want something like checking `"softmax" in config_dict` before calling `get_dtype`, or having a separate method that returns `None` for unconfigured keys.

python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention_v3.py
```diff
@@ -1672,6 +1676,7 @@ def ragged_paged_attention(
     k_scale: float | None = None,
     v_scale: float | None = None,
     xai_temperature_len: float | None = None,
+    softmax_dtype: jnp.dtype | None = None,
```
cjx0709, line 1687: Immediate fix for the CI crash: `softmax_dtype` is in the function params but missing from the `static_argnames` tuple above (around line 1636). In v1's `ragged_paged_attention.py` it was correctly added to `_STATIC_ARGNAME_LIST`, but here it was missed. EAGLE3 speculative decoding goes through this v3 path, so JIT blows up with `TypeError` because `softmax_dtype` (a `numpy.dtype`) isn't marked as static.

But also see my comment on `dtype_config.py` — even after this fix, `softmax_dtype` will be `bfloat16` instead of `None` for all Llama-based models by default, which is an unintended behavioral change.

python/sgl_jax/srt/models/llama.py
```diff
@@ -47,15 +48,22 @@ def __init__(
         mesh: jax.sharding.Mesh,
         layer_id: int = 0,
         dtype: jnp.dtype = jnp.bfloat16,
+        dtype_config: DtypeConfig | None = None,
     ) -> None:
         self.layer_id = layer_id
+        if dtype_config is None:
```
cjx0709, line 54: This dtype vs `dtype_config.default_dtype` check is repeated 5 times — `LlamaMLP`, `LlamaAttention`, `LlamaDecoderLayer`, `LlamaModel`, and `LlamaForCausalLM` all have the same block. But `LlamaForCausalLM` already validates at the top and passes down via `get_config()`, so the child classes can never receive a mismatched pair. These are effectively dead code. I'd just keep the one in `LlamaForCausalLM` and drop the rest.

python/sgl_jax/srt/model_loader/loader.py
```diff
@@ -231,10 +231,14 @@ def _get_model(self, model_class: Any, model_config: ModelConfig) -> nnx.Module:
         else:
             config = model_config.hf_config
 
+        import inspect
```
cjx0709: `import inspect` shows up inside method bodies in two places (here and around line 302). Should probably just be a module-level import.

python/sgl_jax/srt/configs/dtype_config.py
```diff
@@ -0,0 +1,57 @@
...
+    "half": jnp.float16,
+    "float16": jnp.float16,
+    "float": jnp.float32,
+    "float32": jnp.float32,
+    "bfloat16": jnp.bfloat16,
+    "int8": jnp.int8,
+}
+
+
+class DtypeConfig:
```
cjx0709, line 15: `DtypeConfig` doesn't have `__repr__` or `__str__`, so the log line in `llama.py` (`logger.info("... using dtype_config: %s", dtype_config)`) will just print something like `<DtypeConfig object at 0x...>`. Would be nice to add a `__repr__` that shows `config_dict` and `default_dtype`.

test/srt/test_dtype_config_consistency.py
```diff
@@ -0,0 +1,78 @@
...
+
+class TestDtypeConfigConsistency(CustomTestCase):
+    def test_mlp_consistency(self):
+        mesh = create_device_mesh(ici_parallelism=[1, -1], dcn_parallelism=[1, 1])
+        jax.sharding.set_mesh(mesh)
+
+        hidden_size = 256
+        intermediate_size = 512
+
+        # Create Models (they will naturally share equivalent weights from the hardcoded seed!)
```
cjx0709: The comment says the three models "naturally share equivalent weights from the hardcoded seed", but initializing nnx modules with different dtypes (fp32 vs bf16) will give you different weight values due to dtype truncation. So the three models don't actually share the same weights, and the assertions (`diff_fp32_bf16 > diff_bf16_mixed`) might pass by coincidence. A more reliable way would be to init weights in fp32 once, then cast and assign to the other models so they truly start from the same values.

python/sgl_jax/srt/layers/attention/flashattention_backend.py
```diff
@@ -589,6 +589,7 @@ def _ragged_paged_attention_with_fused_kv(*args):
                 xai_temperature_len=(
                     layer.xai_temperature_len if layer.xai_temperature_len > 0 else None
                 ),
+                softmax_dtype=layer.softmax_dtype,
```
cjx0709, line 589: `softmax_dtype` is wired up here for the v3 kernel path, but other attention backends (e.g. `native_backend`, `mla_backend`) don't pass it through at all. If someone configures `softmax_dtype` but their model goes through a different backend, it would silently do nothing. Might want to wire it up in the other backends too, or at least log a warning.

cjx0709 (changes requested):

python/sgl_jax/srt/configs/dtype_config.py
```diff
@@ -0,0 +1,84 @@
...
+        self, config_dict: dict[str, Any] | None = None, default_dtype: jnp.dtype | None = None
+    ):
+        # Validate at least one of config_dict and default_dtype is provided
+        if config_dict is None and default_dtype is None:
+            raise ValueError("At least one of config_dict and default_dtype must be provided.")
+
+        self.config_dict = self._parse_dict(config_dict or {})
+
+        # Resolve the default dtype for this level
+        self.default_dtype = self.config_dict.get("default", jnp.dtype(default_dtype))
```
cjx0709: jnp.dtype(default_dtype) here is the default arg to dict.get(), so it's always evaluated even when "default" is in config_dict. When default_dtype=None, jnp.dtype(None) silently gives you float64 — not wrong today since the result is unused, but fragile. An explicit if "default" in self.config_dict branch would be cleaner.

python/sgl_jax/srt/layers/radix_attention.py
```diff
@@ -36,6 +37,7 @@ def __init__(
         sliding_window_size: int = 0,
         logit_cap: float = 0,
         attn_type: AttentionType = AttentionType.DECODER,
+        softmax_dtype: jnp.dtype | None = None,
```
cjx0709, line 40: Can we validate `softmax_dtype` here? Since `DtypeConfig` also accepts `"int8"`, users can currently set `"softmax": "int8"` and pass a non-floating dtype into softmax.

Probably worth rejecting non-floating dtypes at this common entry point.

python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention.py
```diff
@@ -105,7 +106,17 @@ def ref_ragged_paged_attention_fused(
             attn = attn * xai_temperature_reg[:, None]
 
         attn += jnp.where(mask, mask_value, 0.0)
-        attn = jax.nn.softmax(attn, axis=-1).astype(v.dtype)
+
+        # Cast to softmax_dtype if specified
+        if softmax_dtype is not None:
```
cjx0709, line 111: The v3 reference path does not take `softmax_dtype`, while the kernel does. Could we add it here too so future correctness tests compare the same behavior?

Related: the older ref in `ragged_paged_attention.py` still casts probs back to `v.dtype`, which differs from the kernel path.

cjx0709 (commented):
Thanks for your new fix commits, before merge, I think you can fix below these 3 questions.

iwknow:
@cjx0709 thanks for reviewing. updated accordingly.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -432,6 +432,7 @@ def _get_attention_backend(self):
         elif backend == "fa" and self.use_mla_backend:
             from sgl_jax.srt.layers.attention.mla_backend import MLAAttentionBackend
 
+            logger.warning("Softmax dtype is not configurable for MLA backend through DtypeConfig.")
```
cjx0709: this log should not appear in no dtyte config situation, maybe delete it or using debug level

cjx0709 (commented):
Hello, thanks for your update. I think this pr is almost ready to merge. May you can rebase latesd main and fix the small log.

iwknow:
really appreciated for reviewing! @cjx0709 
i've updated the code accordingly.

cjx0709 (approved):
LGTM

PR #905, opened 2026-04-10
fix: raise NotImplementedError for attention_sink in fused KV path + add CPU correctness tests

zorrofox:
## Motivation

While auditing the `attention_sink_bias` support for MiMo-V2-Flash, we found that the fused `ragged_paged_attention` path silently drops `attention_sink` without any error or warning.

**Root cause:** The fused kernel's `out_specs` only includes `[output, kv_cache_fused]` and does not output `logsumexp`. Without `logsumexp`, the post-kernel alpha correction used by the split path — `alpha = 1 / (1 + exp(sink_bias - logsumexp))` — cannot be applied. The split path (`ragged_paged_attention_split.py`) correctly implements this.

This means any model that enables `add_swa_attention_sink_bias=True` while using the fused KV path (i.e. `v_head_dim == head_dim`) silently gets no sink bias, degrading long-context performance with no diagnostic.

## Changes

**`python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention.py`**
- Raise `NotImplementedError` at the fused path entry when `attention_sink is not None`, with a message pointing to the split KV path as the workaround.

**`python/sgl_jax/test/test_mimo_attention_sink.py`** (new)
- 11 CPU-runnable unit tests (no TPU required):
  - Native backend numerical behavior: large positive bias reduces output norm; very negative bias recovers no-sink output
  - Fused kernel correctly rejects `attention_sink` with an informative error
  - MiMo `hybrid_layer_pattern` SWA/GA layer routing, including the 5:1 ratio
  - `MiMoMoeAttention` attention_sink_bias `nnx.Param` shape and zero-initialization

## Test Plan

- [x] All 11 CPU unit tests pass:
  ```bash
  cd python
  USE_DEVICE_TYPE=cpu python -m pytest sgl_jax/test/test_mimo_attention_sink.py -v
  ```
- [ ] TPU E2E validation of `attention_sink` through the split KV path (to be covered after #874 lands)

## Notes

Full support for `attention_sink` in the fused path requires the fused kernel to emit `logsumexp` as an additional output — a non-trivial kernel change tracked separately. This PR focuses on eliminating the silent failure.


zorrofox:
Closing in favour of a clean cherry-pick branch with no merge conflicts.

PR #913, opened 2026-04-14
Feat/integrate new rpav3 from tpu inference

### LINKED ISSUE #917: [Feature] Integrate and optimize RPA-v3 from tpu-inference
### Checklist

### Motivation

Integrate and optimize RPA-v3 from tpu-inference:

1. Integrate the newest rpav3 from tpu inference https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
2. Support attention sink feature
3. Refactor kv cache: use 5D shape for kv cache that is more friendly to rpav3
4. Wrap kv cache update functions to support dp later

### Related resources

- PR: https://github.com/sgl-project/sglang-jax/pull/913

Iamleos:
## Motivation

1. Integrate the newest rpav3 from tpu inference [https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py](url)
2. support atttention sink feature
3. refactor kv cache: use 5D shape for kv cache that is more friendly to rpav3
4. wrap kv cache update functions to support dp later

## Accuracy Tests
image

## Benchmarking and Profiling
Environment: 
TPU: 16x tpuv6e chips

`JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache python -u -m sgl_jax.launch_server     --model-path /model/Qwen3-30B-A3B-Instruct-2507     --trust-remote-code     --tp-size 16 --ep-size 16     --moe-backend epmoe     --nnodes 4 --node-rank "$RANK"     --dist-init-addr 10.31.179.10:30000     --host 0.0.0.0 --port 30271     --page-size 256     --context-length 262144     --disable-radix-cache     --chunked-prefill-size 2048     --dtype bfloat16     --mem-fraction-static 0.95      --skip-server-warmup     --log-level info     --max-running-requests 128`

Results:
image

## Checklist


python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -378,7 +403,14 @@ def get_fused_kv_buffer(self, layer_id: int) -> jax.Array:
...
         layer_idx = layer_id - self.start_layer
-        fused_kv = self.kv_buffer[layer_idx]  # [cache_size, num_kv_heads * 2, head_dim]
+        fused_kv = self.kv_buffer[
+            layer_idx
+        ]  # [num_pages, page_size, heads*2//packing, packing, head_dim]
+
+        # Flatten 5D -> 3D step by step to avoid ShardingTypeError
+        pages, ps, h_packed, packing, hdim = fused_kv.shape
+        fused_kv = fused_kv.reshape(pages * ps, h_packed, packing, hdim)
+        fused_kv = fused_kv.reshape(pages * ps, h_packed * packing, hdim)
```
JamesBrianD: Why reshape to 3d?

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -428,6 +462,10 @@ def get_cpu_copy(self, indices):
         kv_cache_host = []
         for layer_id in range(self.layer_num):
             fused_kv_host = jax.device_get(self.kv_buffer[layer_id][indices])
+            # Flatten 5D -> 3D: [num_indices, page_size, heads//packing, packing, hdim]
+            #                 -> [num_indices*page_size, heads, hdim]
+            s = fused_kv_host.shape
+            fused_kv_host = fused_kv_host.reshape(s[0] * s[1], s[2] * s[3], s[4])
```
JamesBrianD: Replace kv buffer ?

benchmark/kernels/flash_attention/utils.py
```diff
@@ -1,17 +1,21 @@
...
 ):
+    packing = get_dtype_packing(dtype)
     key = jax.random.PRNGKey(seed)
     keys = jax.random.split(key, 3)
-    total_num_pages = cdiv(max_kv_cache_tokens, page_size)
+    # total_num_pages = cdiv(max_kv_cache_tokens, page_size)
     kv_cache = jax.random.normal(
-        keys[1], (total_num_pages, page_size, head_num * 2, head_dim), dtype=dtype
+        keys[1],
+        (233, page_size, head_num * 2 // packing, packing, head_dim),
```
Bob-Chen222: is there any reason for this number?

JamesBrianD (dismissed):
Great works. I think the precision testing for the subsequent `return logprobs` needs to be redesigned.

python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention_v3.py
```diff
@@ -0,0 +1,2009 @@
...
+    # Decode-only
+    q, kv_cache_fused_processed = run_rpa_kernel(
+        q,
+        kv_cache_fused_processed,
+        **_prepare_block_sizes(d_block_sizes, RpaCase.DECODE),
+        static_q_len=1,
+        case=RpaCase.DECODE,
+    )
+
+    if chunk_prefill_size is not None:
```
Bob-Chen222, line 1986: Did you make any changes to the prefill kernel implementation? In its original design, this kernel seems a bit buggy to integrate directly into our repo. Specifically, I think the static_q_len needs to match the prefill bucket size we use, and there also seem to be a few guards that need to be added inside run_kernel for the prefill case. Could you clarify whether you have already made these changes? Also, do we have any performance comparison between using the prefill kernel and the mixed kernel? Thanks.

Bob-Chen222:
I think everything looks very good, but I still have a couple of questions about RPAv3.

First, have you tested the performance of the prefill kernel versus the mixed kernel specifically in prefill scenarios?

Second, could you add performance benchmarking on Grok-2 on TPU v7, since that is the model we are continuing to optimize? I’m asking because any data or improvement on Grok-2 from these changes would represent a state-of-the-art update, which would be a strong reference for the actual gains and a useful benchmark for future work.

Third, this is related to the prefill case of the RPA kernel, and I already left a comment there. We also have an implementation [here](https://github.com/sgl-project/sglang-jax/commits/attention-upgrade) that does not make too many changes to the original implementation of rpav3, while still adding the necessary guards to prevent index-out-of-bounds issues. I want to understand if you also see a similar issue and what changes you made. Thanks a lot!

benchmark/kernels/flash_attention/utils.py
Iamleos: Already fixed

python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention_v3.py
Iamleos: In the older version of RPAv3, we used different modes by setting different distributions in the attention metadata. Mixed mode for decode batch and chunked-prefill mode for extend batch. In old rpav3, although we use prefill mode for the extended batch, we do not set the chunk_prefill_size parameter, which means the computation logic remains the same as in the mixed mode. However, in new rpav3 kernel, if you set prefill distribution in attention metadata and don't set chunk_prefill_size parameter, prefill computation will be skipped. So, we set a guard here. In addition, we have already change the distribution in attention metadata. In new implementation, decode-only mode for decode batch and mixed mode for extend batch.

Iamleos:
> I think everything looks very good, but I still have a couple of questions about RPAv3.
> 
> First, have you tested the performance of the prefill kernel versus the mixed kernel specifically in prefill scenarios?
> 
> Second, could you add performance benchmarking on Grok-2 on TPU v7, since that is the model we are continuing to optimize? I’m asking because any data or improvement on Grok-2 from these changes would represent a state-of-the-art update, which would be a strong reference for the actual gains and a useful benchmark for future work.
> 
> Third, this is related to the prefill case of the RPA kernel, and I already left a comment there. We also have an implementation [here](https://github.com/sgl-project/sglang-jax/commits/attention-upgrade) that does not make too many changes to the original implementation of rpav3, while still adding the necessary guards to prevent index-out-of-bounds issues. I want to understand if you also see a similar issue and what changes you made. Thanks a lot!

1. I have tested the performance in swa scenario.
2. Do you mean adding performance benchmark tests for the Grok model on TPU v7 to CI?
3. I only migrated some attention-specific features from sglang-jax to the new RPAv3 (e.g. custom attention, xai temperature, attention sink) and did not encounter the issues you mentioned.

python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention_v3.py
Bob-Chen222: Ok got it so essentially we disabled the prefill kernel

Bob-Chen222 (approved):

PR #914, opened 2026-04-15
feat(moe): FP8 quantization, ablation support, and FusedEPMoE kernel optimizations

JamesBrianD:
## Summary

Cherry-pick MOE-related changes from two PRs authored by @Prayer3th , Full credit goes to him for the  implementation:

- sgl-project/sglang-jax#848: fused MoE kernel ablation support
- primatrix/sglang-jax#219: FP8 block-wise quantization, EPMoE kernel optimizations

## Motivation

Bring FP8 quantization and kernel optimization work from epic branches into main, enabling:
- FP8 sub-channel and block-wise quantization for fused MoE kernel
- Ablation benchmarking framework for systematic MoE kernel performance analysis
- Tuned block configurations for TPU
- EPMoE (GMM-based) performance optimizations

### FusedEPMoE kernel
- **`kernel.py`**: FP8 block-wise quantization (`quant_block_k`, `quant_block_n`), ablation flags (`disable_a2a`, `disable_permute`, etc.), variable renames (`subc_quant_wsz` → `quant_block_k`, `token_buf_seed` → `token_buf_offset`)
- **`fused_moe.py`**: FP8/block quant support in `FusedEPMoE` layer, scale expansion logic, fix output reshard to `P(None, None)` for TP-only compatibility
- **`weight_utils.py`**: `_expand_moe_block_scale()` for 2D→1D block scale pre-expansion at weight loading
- **`tuned_block_configs.py`** (new): TPU-tuned tile configurations

### EPMoE (GMM-based) optimizations
- **`moe.py`** — **Indexed GMM**: `_permute` no longer materializes the full `[M*top_k, D]` sorted_inputs tensor upfront. Instead, raw `inputs_2d` and `token_indices` are passed to `_gmm_compute`, where the gather happens just-in-time. This lets XLA fuse gather+matmul, reducing peak HBM usage and memory bandwidth pressure — especially beneficial for large `top_k` models (e.g. MiMo-V2-Flash with top_k=8, 256 experts).
- **`moe.py`** — **Adaptive tiling**: new `_adaptive_tile_fn` selects `tile_m=256` for large M (prefill, M≥8192) and keeps `tile_m=128` for small M (decode), giving +1-2% prefill throughput improvement.
- **`megablox_gmm_backend.py`**: Plumb `v2_tile_info` parameter through `gmm()` to support external tiling strategies for the v2 GMM kernel.

### Benchmarking & testing
- **`run_fused_moe_ablation.sh`** (new): Ablation benchmark runner script
- **`bench_fused_moe.py`** / **`utils.py`**: Expanded benchmarking with FP8 quant support
- **`fused_moe_v1_test.py`**: New FP8 quantization test cases
- **`moe_parity_repro.py`**: Updated parameter names

**Note**: `mimo_v2_flash.py` model integration excluded (file not yet on main).

## Accuracy Tests

| Model | Dataset | Backend | Metric | Num | Score |
|-------|---------|---------|--------|-----|-------|
| MiMo-V2-Flash | GSM8K | fused MoE | AverageAccuracy | 500 | **0.96** |
| Qwen3-30B-A8B | GSM8K | EPMoE | AverageAccuracy | 500 | **0.958** |

## Checklist


python/sgl_jax/srt/layers/moe.py
```diff
@@ -9,15 +9,36 @@
...
-from sgl_jax.srt.utils.quantization.quantization_utils import quantize_tensor
+from sgl_jax.srt.utils.quantization.quantization_utils import (
+    quantize_tensor,
+    quantize_tensor_simple,
+)
 from sgl_jax.srt.utils.weight_utils import WeightMapping
 
 
+def _adaptive_tile_fn(lhs_dtype, rhs_dtype, dims, vmem_limit_bytes):
+    """Adaptive TileFn: use tile_m=256 for large M (prefill), default for small M (decode).
```
Bob-Chen222: By adding this are you implying that the default size change can be applied to other moe models as well?

python/sgl_jax/srt/layers/fused_moe.py
```diff
@@ -166,37 +191,85 @@ def __init__(
...
+            del self.quant_block_k
+        self.quant_block_k = wsz
 
         with jax.set_mesh(self.mesh):
             if is_static:
                 ep_scale_sharding = P(("data", "tensor"), None, None, None)
 
+                # Compute the actual kernel-expected scale shapes so the weight
+                # loader can match dimensions and assign directly.
+                if self.quant_block_n is not None:
```
Bob-Chen222: are these two branches doing the same execution?

python/sgl_jax/srt/kernels/fused_moe/v1/kernel.py
```diff
@@ -221,6 +244,8 @@ def validate_fused_moe_block_config(
         raise ValueError(f"Expected {local_num_tokens=} to be aligned to {t_packing=}.")
     if bt % t_packing != 0:
         raise ValueError(f"Expected {bt=} to be aligned to {t_packing=}.")
+    if not (bt in (2, 4, 8) or bt % 8 == 0):
```
Bob-Chen222: Would it be better to add a comment explaining why 2 and 4 are included? And also can be shorten to bt in (2, 4) or bt % 8

JamesBrianD: Good catch! Simplified to `bt in (2, 4) or bt % 8 == 0` and added a comment:

- `bt=2,4` are for small-batch decode scenarios (e.g. `local_num_tokens=2` when `ep_size=32`)
- For prefill, `bt` must be a multiple of 8 to keep the tuned config search space manageable

Fixed in 4e425a3b.

python/sgl_jax/srt/layers/moe.py
JamesBrianD: Remove these hard code config.

python/sgl_jax/srt/layers/fused_moe.py
JamesBrianD: Yes, both branches produce identical shapes. The scale placeholder shape is always `(E, K//block_k, 1, N)` regardless of 1D or 2D quantization — in the 2D case, `_expand_moe_block_scale()` expands the compact `(E, K//bk, N//bn)` scales to this same unified layout at weight-loading time, so the kernel always sees the 1D shape.

Merged the two branches into one and added a comment explaining this. Fixed in f046a758.

python/sgl_jax/srt/layers/moe.py
JamesBrianD: Good point — this adaptive tiling was only benchmarked on MiMo-V2-Flash and isn't generalizable. Removed the entire `_adaptive_tile_fn` in commit 4426f529. The GMM layer now uses the default `calculate_tiling` which adapts tile sizes based on dtype and VMEM capacity automatically.

JamesBrianD: Done — removed entirely in commit 4426f529. The `_adaptive_tile_fn` and its hardcoded config have been deleted. EPMoE now relies on the default `calculate_tiling()` from gmm_v2.

Bob-Chen222 (approved):

PR #930, opened 2026-04-20
perf: optimize get_forward_metadata for SWA models

cjx0709:
## Summary

Optimize `get_forward_metadata` in FlashAttention backend for SWA models by reducing numpy array operations.

### Key changes

- **SWA mapping on page-level entries**: Apply `swa_index_mapping` on `~N/page_size` page entries instead of `~N` token entries, reducing fancy-index lookups by `page_size`x (e.g., 256x)
- **Simplify decode cu_q_lens**: Replace `concat([0], cumsum(ones(N)))` with `arange(N+1)`
- **Remove redundant copy**: Use `batch.seq_lens` directly instead of `np.copy`
- **Unify device_array call**: Merge conditional SWA/non-SWA branches into a single `device_array` call

Net diff: +31/-60 lines in `flashattention_backend.py`.

## Benchmarking

MiMo-V2-Flash on 4x TPU v6e-16, 256 prompts, input=16384, output=1024, concurrency=64, `--moe-backend fused`:

| Metric | Main (baseline) | Optimized | Change |
|:-------|:---:|:---:|:---:|
| Output throughput (tok/s) | 401.47 | 473.19 | **+17.9%** |
| Median ITL (ms) | 31.30 | 23.45 | **-25.1%** |
| P95 ITL (ms) | 40.49 | 25.04 | **-38.2%** |
| P99 ITL (ms) | 42.22 | 29.59 | **-29.9%** |
| Duration (s) | 652.96 | 553.99 | **-15.2%** |

256/256 requests completed successfully on both runs.

<details>
<summary>Full benchmark output (Main baseline)</summary>

```
============ Serving Benchmark Result ============
Backend:                                 sgl-jax   
Traffic request rate:                    inf       
Max request concurrency:                 64        
Successful requests:                     256       
Benchmark duration (s):                  652.96    
Total input tokens:                      4194304   
Total generated tokens:                  262144    
Total generated tokens (retokenized):    261176    
Request throughput (req/s):              0.39      
Input token throughput (tok/s):          6423.51   
Output token throughput (tok/s):         401.47    
Total token throughput (tok/s):          6824.97   
Concurrency:                             57.64     
----------------End-to-End Latency----------------
Mean E2E Latency (ms):                   147016.47 
Median E2E Latency (ms):                 134762.37 
---------------Time to First Token----------------
Mean TTFT (ms):                          98016.72  
Median TTFT (ms):                        93994.45  
P99 TTFT (ms):                           148019.67 
[...]

PR #922, opened 2026-04-16
feat: support MiMo-V2-Flash model

cjx0709:
## Summary

Add MiMo-V2-Flash model support for TPU v6e/v7x with 256-expert MoE, FP8 quantization, and hybrid attention (9 full-attention + 39 SWA layers).

### Key features

- **MiMo-V2-Flash model** (`mimo_v2_flash.py`, ~1017 lines): Full model implementation with split K/V head dimensions (q/k=192, v=128), sliding window attention, attention sink, and FP8 MoE
- **Split K/V head_dim support**: `v_head_dim` config parameter; FlashAttention and native backend handle `k_head_dim != v_head_dim`
- **Native backend enhancements**: Rewrote `forward_attention` with T-H-K einsum layout, Q padding for 128-aligned K cache, sliding window masking in both extend and decode, attention sink support in softmax
- **SWA bug fix**: Removed double-accounting in SWA eviction tracking that caused OOM under concurrency
- **Documentation**: `docs/basic_usage/mimo_v2_flash.md` with launch commands, config tips, and benchmark results

### Accuracy (GSM8K)

| Model | Dataset | Metric | Subset | Num | Score |
|:------|:--------|:-------|:-------|:----|:------|
| MiMo-V2-Flash | gsm8k | AverageAccuracy | main | 1319 | **0.9401** |

Evaluated on TPU v6e-16 (4 nodes) with `evalscope==0.17.1`, sampling params: `temperature=0.8, top_p=0.95, max_tokens=32768`.

### Throughput Benchmark

256 prompts, input=16384 tokens, output=1024 tokens, concurrency=64 on TPU v6e-16:
- 256/256 requests completed successfully (0 failures)
- No OOM crashes with SWA bug fix applied

### Environment Setup

```bash
# Clone and install
git clone https://github.com/sgl-project/sglang-jax.git
cd sglang-jax/python && pip install -e .
```

### Launch Server

**Option A: Multi-node TPU v6e-16 (4 nodes, Fused MoE)**

```bash
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache uv run python -u -m sgl_jax.launch_server \
    --model-path XiaomiMiMo/MiMo-V2-Flash \
    --trust-remote-code \
    --tp-size 16 --ep-size 16 \
    --moe-backend fused \
    --nnodes 4 --node-rank $RANK \
    --dist-init-addr $MASTER_IP:30000 \
    --host 0.0.0.0 --port 30271 \
[...]

PR #931, opened 2026-04-20
feat: add MiMo-V2-Flash tuned fused MoE configs on TPU v7x-8

jimoosciuc:
## Motivation

add MiMo-V2-Flash tuned fused MoE configs on TPU v7x-8

## Modifications

add MiMo-V2-Flash tuned fused MoE configs on TPU v7x-8

## Checklist


pengchengneo (approved):

zorrofox:
## Unexpected: `--moe-backend=epmoe` outperforms `--moe-backend=fused` by ~22% on TPU v7x for long-context workloads

Reporting a finding while reproducing the v7x results from #930. With identical hardware and server args (only `--moe-backend` differs), `epmoe` is **+21.7% Output throughput** and **−12% Median ITL** vs the tuned `fused` path landed in #931. Sharing the data in case it's useful for tuning coverage.

### Setup

- Hardware: **TPU v7x-4t** (single node, `2x2x1`, 8 chips), GKE `bodaborg-tpu7x-auto-nap2` (us-central1)
- Model: `XiaomiMiMo/MiMo-V2-Flash` (FP8, 256 experts)
- Framework: sglang-jax **main HEAD** (commit `febb350c`, includes #922 + #930 + #931 + #934)
- JAX 0.8.1 + libtpu 0.0.30
- Server args (only `--moe-backend` varies):
  ```
  --tp-size=8 --ep-size=8 --nnodes=1
  --context-length=262144 --page-size=256
  --mem-fraction-static=0.90 --swa-full-tokens-ratio=0.2
  --max-running-requests=64 --skip-server-warmup
  --chunked-prefill-size=2048 --disable-radix-cache
  --dtype=bfloat16
  ```
- Bench: `sgl_jax.bench_serving --dataset-name random --num-prompts 256 --random-input 16384 --random-output 1024 --max-concurrency 64 --random-range-ratio 1 --warmup-requests 0 --seed 12345`

### Results

| Metric | `fused` (this run) | `fused` (#930 baseline) | **`epmoe` (this run)** | epmoe vs fused |
|:-------|---:|---:|---:|---:|
| Successful requests | 256/256 | 256/256 | 256/256 | — |
| Output throughput (tok/s) | 387.73 | 387.93 | **471.86** | **+21.7%** |
| Input throughput (tok/s) | 6,203.68 | 6,206.91 | 7,549.79 | +21.7% |
| Median ITL (ms) | 41.48 | 39.60 | **36.39** | **−12.3%** |
| P95 ITL (ms) | 46.30 | 43.70 | 40.47 | −12.6% |
| P99 ITL (ms) | 49.99 | 45.28 | 42.37 | −15.2% |
| Mean TTFT (ms) | 64,275.76 | 65,120.61 | 51,267.00 | −20.2% |
| Concurrency | 63.99 | 63.96 | 63.99 | — |

`fused` reproduces #930's published v7x numbers within **0.05%**, so the bench harness is consistent. The delta is entirely the backend choice.

### Hypothesis

The tuned configs added in #931 likely cover a workload range (batch size × sequence length × expert count) that doesn't include this configuration: 16K-token prefill with `max-running-requests=64` on `tp=8 ep=8`. When the lookup misses the tuned grid, the GMM kernel falls back to default block sizes that under-utilize the v7x MXU, while `epmoe`'s dynamic dispatch path stays close to optimal.

Short-context numbers show a smaller but consistent gap (`epmoe` 799 vs `fused` 760 tok/s, +5.1%), supporting the "tuned config coverage" theory rather than a fundamental regression in the fused path.

### Questions / suggestions

1. Could the tuned config grid in #931 be extended to cover larger `max-running-requests` and longer prefill chunks?
2. Is there a recommended fallback policy when `(batch, seq)` is outside the tuned grid? (e.g., auto-switch to `epmoe` for unsupported shapes?)
3. Would a public benchmark sweep matrix (workloads × backends) be useful? Happy to contribute the v7x-4t data points.

### Reproduction artifacts

Full bench logs and deployment manifests (StatefulSet for v6e-16, Deployment for v7x-4t):
- `v7x4t-mainhead-20260423-014536.log` (epmoe)
- `v7x4t-mainhead-fused-20260423-031157.log` (fused)

Happy to share the manifests if helpful.

---

cc @cjx0709 @jimoosciuc @pengchengneo

zorrofox:
## Follow-up: epmoe vs fused on v7x-4t after #969 + cookbook params

Re-ran the comparison on main HEAD `b787fdef` (includes #969 FP8 dequant fix + #936) and also tested the cookbook params from `docs/basic_usage/mimo_v2.5_pro.md`. The epmoe advantage on **v7x-4t (8 devices)** persists and actually **widens** with cookbook params.

### Results (long context: 256×16K/1K, c=64)

| Config | moe | cps | swa | mr | mf | Output tok/s | Median ITL |
|:--|:--|--:|--:|--:|--:|--:|--:|
| baseline | epmoe | 2048 | 0.2 | 64 | 0.90 | 467.32 | 36.90 |
| baseline | **fused** | 2048 | 0.2 | 64 | 0.90 | 396.75 | 38.19 |
| cookbook | epmoe | 4096 | 0.25 | 512 | 0.95 | **480.04** | 37.84 |
| cookbook | **fused** | 4096 | 0.25 | 512 | 0.95 | 382.32 | 42.45 |

- epmoe advantage: **+17.8%** (baseline params) → **+25.6%** (cookbook params)
- #969 moved fused from 387.73→396.75 (+2.3%, likely from corrected gate routing); epmoe essentially unchanged (471.86→467.32, noise)
- `cps=4096` helps epmoe (+2.7%) but **hurts fused (−3.6%)** — server log shows fused GMM falls back to default block sizes `(512,2048,2048)` for the larger prefill shape, outside the #931 tuned grid

Also verified accuracy on the same setup: AIME2025-I 14/15 + AIME2025-II 15/15, matching the #969 reference.

### Suggestion

The cookbook currently states "epmoe is also supported but slower at this scale" for v7x-16. Might be worth a note that for **v7x single-host (≤8 devices, EP=8)** the recommendation inverts — `--moe-backend=epmoe` is ~18-26% faster. Happy to send a PR to the cookbook if useful.

cjx0709:
Thanks for the thorough benchmarking @zorrofox ! The data is very helpful.                                                                                                
Your hypothesis is correct — ep moe may do a better performance in some situation.
                                                                                                                                              
For your questions:                                                                                                                                                      
1. The tuned config not cover seems strange, if we use fused moe, we will not use the gmm tuner "tuned_block_sizes.py" but fused_moe "tuned_block_configs.py", maybe you can comment for more miss tuner config logs；
2. Auto-fallback to epmoe is a good idea, but may meet the trouble: epmoe and fused moe with different weight layout, if we support two style moe in the same server, it may cause the memory shortage；                                                                   
3. A public benchmark matrix would be great. It's great to send a PR with the v7x-4t data points and updated the cookbook(better pr body with the benchmark screenshot).

zorrofox:
Thanks for the pointer @cjx0709 — you're right, I was looking at the wrong file.

### Correction

The `(512,2048,2048)` I quoted came from `kernels/gmm/megablox_gmm_kernel/tuned_block_sizes.py` (the `get_default_gmm_block_sizes` heuristic), which is the **epmoe** GMM path, not fused. So that line of evidence is invalid — apologies for the noise.

For `--moe-backend=fused` the lookup is indeed `kernels/fused_moe/v1/tuned_block_configs.py::get_tuned_fused_moe_block_config`, and the silent fallback there is `DEFAULT_FUSED_MOE_BLOCK_CONFIG = FusedMoEBlockConfig(bt=32, bf=512, bd1=1024, bd2=1024, ...)`.

### Tuner logs (re-run 2026-04-30, v7x-4t @ b787fdef, cookbook-fused: ep=8 fused cps=4096)

Re-ran the cookbook-fused config with server log capture. The fused MoE tuned config **does hit** for all precompiled EXTEND shapes:

```
[2026-04-30 02:35:49] [EXTEND] Begin to precompile bs_paddings=[128] token_paddings=[128, 256, 512, 1024, 2048, 4096]
[2026-04-30 02:35:50] Using tuned block config: (16, 2048, 2048, 2048, 32, 32, 2048, 2048, 2048, 2048)
[2026-04-30 02:36:08] Using tuned block config: (32, 2048, 2048, 2048, 64, 64, 2048, 2048, 2048, 2048)
[2026-04-30 02:36:26] Using tuned block config: (64, 2048, 2048, 2048, 32, 32, 2048, 2048, 2048, 2048)
... (remaining shapes also hit)
```

So my earlier "tuned grid not covering ep=8" hypothesis was **wrong** — the `"TPU v7"` table in `tuned_block_configs.py` covers the MiMo-V2-Flash ep=8 shapes and the lookup succeeds. The +18~26% epmoe-over-fused gap on v7x-4t must come from something else. Candidates: per-expert load skew at EP=8 (256 experts / 8 = 32 experts/device, vs 256/32=8 on v7x-16), or fused dispatch overhead not amortizing at single-host scale. Happy to dig further if useful — let me know what diagnostics would help.

### Next steps from our side

- [x] Re-ran cookbook-fused on v7x-4t @ `b787fdef` with server log capture — tuned config hits, see above.
- [ ] Send a cookbook PR adding the v7x-4t (EP=8) data points + a note that `epmoe` is preferred for single-host v7x, with bench screenshots in the PR body as you suggested.

cjx0709:
Thanks for the thorough follow-up @zorrofox ! Really appreciate you going back and re-running with log capture to verify — that clears things up nicely.  Maybe latter we can figure out the root cause of why fused moe performs poorly in certain situations(eplb may be one reason), we already have a pr for this: https://github.com/sgl-project/sglang-jax/pull/818, you can find more information from this.
And I'm very glad to see your pr to fix our cookbook，I will latter approve and merge it.

PR #940, opened 2026-04-22
fix: advance sampler RNG each decode step

Prayer3th:
## Summary

- The sampler's `nnx.Rngs` state was captured once during `initialize_jit()` and reused unchanged for every JIT call. This caused every decode step to use the **identical RNG key**, making `temperature>0` sampling fully deterministic rather than random.
- Once the model entered a near-repetitive state (similar logits across steps), the fixed Gumbel noise guaranteed the same token was selected every time, locking the model into an inescapable repetition loop.
- Fix: maintain a `jax.random.PRNGKey` on `ModelRunner`, split it before each `sample()` call, and pass the fresh key into the JIT sampler as an explicit argument.

## Reproduction

With the bug present, sending the same prompt twice with `temperature=1.0` produces **byte-identical output** — confirming zero randomness. Long reasoning chains (~6000+ tokens) reliably collapse into exact repetition (e.g. repeating the same equation forever).

## Test plan

- [ ] Verify same prompt with `temperature=1.0` now produces **different** output across runs
- [ ] Verify greedy decoding (`temperature=0`) is unaffected
- [ ] Verify no JIT recompilation regression (single retrace expected on first call due to new arg)

### DIFF
--- a/python/sgl_jax/srt/layers/sampler.py
+++ b/python/sgl_jax/srt/layers/sampler.py
@@ -163,6 +163,7 @@ class Sampler(nnx.Module):
         logits_output: LogitsProcessorOutput,
         sampling_metadata: SamplingMetadata,
         use_sort_for_toppk_minp: bool,
+        rng_override: jax.Array | None = None,
     ):
         """Run a sampler & compute logprobs and update logits_output accordingly.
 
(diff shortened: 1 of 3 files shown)


jimoosciuc (dismissed):
LGTM

jimoosciuc (approved):
LGTM

PR #941, opened 2026-04-22
Feat/deepseek non absorption

Bob-Chen222:
## Motivation

Add DeepSeek V2 / V3 model support to sglang-jax. This lands the **non-absorbed MLA** path — a correctness-first baseline that materializes full K/V from the compressed latent via `kv_b_proj` and feeds it through the existing `RadixAttention` + `MHATokenToKVPool` infrastructure. It uses ~43× more KV-cache memory than the absorbed form; the absorbed variant will land in a follow-up once a production MLA Pallas kernel is ready (see `python/sgl_jax/srt/layers/attention/mla.py:18-22`).

Supported configurations tested here:
- `deepseek-ai/DeepSeek-V2-Lite` (16B, 64 routed + 2 shared experts, top-k=6, MLA with `q_lora_rank=None`, YaRN rope scaling).
- Weight-mapping covers both `q_proj` (no LoRA) and `q_a_proj`/`q_b_proj` (with LoRA) branches, so DeepSeek-V3 (`q_lora_rank=1536`) is supported out of the box; `fp8_deepseek_v3.yaml` provides the FP8 recipe.

## Modifications

1. **`python/sgl_jax/srt/models/deepseekv3.py`** (new, ~680 lines). `DeepseekV3ForCausalLM` + `DeepseekV2ForCausalLM` (subclass). Reuses `MLAAttention`, `GateLogit`, `TopK`, `EPMoE`/`FusedEPMoE`. Handles dense/MoE alternation (`first_k_dense_replace`/`moe_layer_freq`), shared experts (non-fused: separate `DeepseekV3MLP`; fused: inside `FusedEPMoE`), `noaux_tc` routing-bias load balancing, optional YaRN `mscale_all_dim` attention-scaling adjustment, and a `patch_model_config()` classmethod to override `attention_arch=MLA` and `head_dim = qk_nope + qk_rope` before the KV pool is allocated. Weight-loading `_create_layer_mappings` matches the flat-dict convention used by `qwen3_moe.py` / `grok.py` / `bailing_moe.py`, with a single nested `add_linear` closure for the static-FP8 branch.

2. **`python/sgl_jax/srt/layers/embeddings.py`** (+122). New `YarnRotaryEmbedding` class and `"yarn"` scaling branch in `get_rope()`. Blends original and scaled inverse frequencies using a correction range from `beta_fast`/`beta_slow` and applies the DeepSeek-style `mscale = 0.1 · mscale · log(scale) + 1` factor to `cos`/`sin`.

3. **`python/sgl_jax/srt/configs/model_config.py`** (+20/-2). New `_apply_model_specific_config()` hook invoked during `ModelConfig` construction — before `ModelRunner` reads `attention_arch` for backend selection — so each model class can own its own config patches (`attention_arch`, `head_dim`, …) via an optional `patch_model_config(cls, mc)` classmethod instead of a centralized if/elif chain.

4. **`python/sgl_jax/srt/layers/attention/mla.py`** (±2). `q_a_layernorm` / `kv_a_layernorm` param dtype switched from `jnp.float32` to the layer `dtype` (bf16), matching the reference DeepSeek implementation and avoiding an unnecessary fp32 upcast inside RMSNorm.

5. **`python/sgl_jax/srt/layers/moe.py`** (+13). Sublane-alignment row-padding in `EPMoE._gmm_compute`. DeepSeek-V2-Lite has `num_experts_per_tok=6`; with the default power-of-2 `bs_paddings`, `padded_bs × 6` can land on values that are `> 16` and not a multiple of 16 (e.g. `bs=4 → size_m=24`), which the megablox `gmm_v2` sublane-tiled reshape requires. The pad rounds `size_m` up to `sublane_tiling` (dtype-dependent via `pltpu.get_tpu_info().get_sublane_tiling(x.dtype)`) and attributes the extra rows to the last expert group; `_unpermute` already trims them. Models with power-of-2 top_k (Grok=2, DeepSeek-V3=8, Qwen3-MoE=8) are unaffected.

6. **`python/sgl_jax/srt/utils/quantization/configs/fp8_deepseek_v3.yaml`** (new, 25 lines). W8A8 FP8 quant config for DeepSeek-V3 — MoE gate and `logits_processor` excluded from quant for numerical stability; experts quantized (per-channel weight + dynamic per-token activation).

7. **`benchmark/gsm8k/bench_sglang_jax.py`** (new, 259 lines). Async GSM8K benchmark that hits a running server's `/generate` endpoint. Supports both the original `openai/grade-school-math` test set and `madrylab/gsm8k-platinum` (via `--platinum`). Optional chat-template wrapping with `--enable-thinking`.

## Accuracy Tests

Target: **DeepSeek-V2-Lite** (bf16, TP=8, TPU v7, non-absorbed MLA, 5-shot).

```bash
python benchmark/gsm8k/bench_sglang_jax.py --num-questions 1400
```

| metric | value |
|---|---|
| Accuracy | **0.394** |
| Invalid  | 0.004 |
| Latency  | 165.94 s |
| Output throughput | 958.23 tok/s |

(GSM8K test set has 1319 questions; `--num-questions 1400` caps at the full set.) Huggingface score is 41.1.

## Benchmarking and Profiling

Target: **DeepSeek-V2-Lite** (bf16, TP=8, TPU v7, non-absorbed MLA, `chunked-prefill-size=16384`, `page-size=128`, `mem-fraction-static=0.5`). Command run twice; first pass is JIT compile and is discarded, numbers below are from the second run.

```bash
python3 -m sgl_jax.bench_one_batch_server \
  --model None --tokenizer-path alvarobartt/grok-2-tokenizer \
  --base-url http://localhost:30000 \
  --batch-size 1 4 8 16 32 64 \
  --input-len 8192 --output-len 1024 --show-report
```

Input lens: 8192. Output lens: 1024.

| batch size | latency (s) | input throughput (tok/s)  | output throughput (tok/s) | acc length | ITL (ms) | input cost ($/1M) | output cost ($/1M) |
| ---------- | ----------- | ------------------------- | ------------------------- | ---------- | -------- | ----------------- | ------------------ |
| 1 | 8.31 | 25891.91 | 128.07 | n/a | 7.81 | 0.03 | 4.34 |
| 4 | 9.16 | 29012.52 | 509.76 | n/a | 7.85 | 0.03 | 1.09 |
| 8 | 11.95 | 29644.59 | 840.74 | n/a | 9.52 | 0.03 | 0.66 |
| 16 | 18.11 | 29939.02 | 1192.91 | n/a | 13.41 | 0.03 | 0.47 |
| 32 | 28.44 | 30088.16 | 1661.42 | n/a | 19.26 | 0.03 | 0.33 |
| 64 | 61.77 | 30170.70 | 1476.12 | n/a | 43.36 | 0.03 | 0.38 |

(`$/1M` columns are the bench script's built-in H100-based estimate — not applicable to TPU; ignore.)

## Checklist


python/sgl_jax/srt/configs/model_config.py
```diff
@@ -172,9 +172,8 @@ def __init__(
             "head_dim",
             self.hf_text_config.hidden_size // self.hf_text_config.num_attention_heads,
         )
-        self.v_head_dim = getattr(self.hf_text_config, "v_head_dim", self.head_dim)
-
```
JamesBrianD, line 176: Please not change this line, maybe effect `mimo_v2_flash`

python/sgl_jax/srt/configs/model_config.py
```diff
@@ -350,6 +349,25 @@ def is_dynamic_fp8_act(cfg):
...
+        """
+        from sgl_jax.srt.models.registry import ModelRegistry
+
+        try:
+            model_cls, _ = ModelRegistry.resolve_model_cls(self.hf_config.architectures)
+        except ValueError:
+            return
+        patch = getattr(model_cls, "patch_model_config", None)
+        if patch is not None:
+            patch(self)
```
JamesBrianD, line 370: What is the purpose of this function?

python/sgl_jax/srt/layers/embeddings.py
```diff
@@ -665,3 +688,102 @@ def _yarn_find_correction_range(
     low = math.floor(_yarn_find_correction_dim(low_rot, dim, base, max_position_embeddings))
     high = math.ceil(_yarn_find_correction_dim(high_rot, dim, base, max_position_embeddings))
     return max(low, 0), min(high, dim - 1)  # Clamp values just in case
+
+
+def _deepseek_yarn_get_mscale(scale: float, mscale: float) -> float:
+    """Compute mscale factor for DeepSeek-style YaRN (parameterized by mscale coefficient)."""
+    if scale <= 1:
+        return 1.0
+    return 0.1 * mscale * math.log(scale) + 1.0
```
JamesBrianD, line 697: `_yarn_get_mscale` This method was introduced during the adaptation to Grok 2. Please verify whether these two methods are consistent.

python/sgl_jax/srt/configs/model_config.py
Bob-Chen222: The attributes inside the model config will be changed based on model specifics before initalizing the model file. For example, for deepseek MLA, the head dim is not the config.head_dim but instead is head_dim = qk_nope + qk_rope. I think that these model specific changes will become messy and if each model file manages its own attribute change it would be cleaner, which is the reason why I made these changes. Open to discussion

python/sgl_jax/srt/layers/embeddings.py
Bob-Chen222: They are not the same. Grok-2 uses math.sqrt(scaling_factor) but deepseek uses 0.1 * mscale * math.log(scale) + 1.0. I changed the name to make them separate.

python/sgl_jax/srt/configs/model_config.py
JamesBrianD: Understood—that's very clear.

python/sgl_jax/srt/models/deepseek_v3.py
```diff
@@ -0,0 +1,855 @@
+"""DeepSeek V3 model implementation.
```
aolemila, line 1: file name : deepseek_v3.py?

Iamleos:
1. What are the differences in performance and accuracy between absorb MLA and non-absorb MLA?
2. Will they be implemented as separate backends?
3. In the future, will only one implementation be retained, or will both coexist?
4. If they do coexist, how can we specify which implementation to use?

aolemila:
`bench_sglang_jax.py` looks like to do not exist in repo and current PR.

aolemila:
Which accuracy reference does this repo use? Like [huggingface](https://huggingface.co/deepseek-ai/DeepSeek-V2-Lite)?

JamesBrianD:
> 1. What are the differences in performance and accuracy between absorb MLA and non-absorb MLA?
> 2. Will they be implemented as separate backends?
> 3. In the future, will only one implementation be retained, or will both coexist?
> 4. If they do coexist, how can we specify which implementation to use?

1. The implementation of `absorbed` is not covered here.
2. See the mla RFC
3. In the future,  two implmentations will be retained. MHA Backend and MLA Backend

JamesBrianD:
> `bench_sglang_jax.py` looks like to do not exist in repo and current PR.

https://github.com/sgl-project/sglang-jax/pull/941. To make reviewing this PR easier, I suggested to Qinghan that we remove it for now. Should we add it back in and review that file again?

aolemila (approved):

PR #942, opened 2026-04-23
perf: add new fused moe tunner config for v6e

cjx0709:
## Summary
- Add tuned fused MoE block configs for 384-expert FP8 models on TPU v6e-64 (8x8 topology)


jimoosciuc (approved):
LGTM

PR #944, opened 2026-04-23
fix(swa): graceful abort instead of crash when decode OOM after retract

cjx0709:
## Problem
When decode KV cache is exhausted in SWA hybrid mode, the server crashes instead of gracefully degrading.

### Env
Running 16K benchmark on v6e-64 with `swa_full_tokens_ratio=0.2` and `page_size=256`. KV cache is split into two independent pools:
- **full pool**: 569,344 tokens (80%) for full attention layers
- **swa pool**: 113,664 tokens (20%) for sliding window attention layers

When many 16K sequences are decoding simultaneously, full pool usage approaches 1.0.

### Crash path

The scheduler calls `check_decode_mem` before each decode step. When memory is insufficient, it calls `retract_decode` to remove requests one by one (sorted by output length), freeing their KV cache until memory is sufficient.

The crash occurs in the **corner case where only 1 request remains but memory is still insufficient** — the old code hits an `assert` crash:

```python
if len(sorted_indices) == 1:
    if self.is_hybrid:
        assert (full_available_size > 0 and swa_available_size > 0), ...
    else:
        assert (self.token_to_kv_pool_allocator.available_size() > 0), ...
    break
```

In SWA hybrid mode with 16K sequences, a single request's KV footprint can be large enough that even after freeing all other requests, the next decode page allocation cannot be satisfied. The assert fails and the server exits.

Since the retract log message is printed **after** `retract_decode` returns (`scheduler.py`), and the assert throws inside the function, no retract messages appear in the server log — only the crash.

### Dead code

The old code also contained unreachable dead code: `if len(retracted_reqs) == 0` checked after `retracted_reqs.append(req)`, which can never be True.

## Fix

Following sglang's approach (`schedule_batch.py:2100-2133`), replace the crash with a graceful abort:

### `schedule_batch.py` — `retract_decode`

1. Remove `assert` crash — `break` out of the loop when 1 request remains
2. After the loop, re-check `check_decode_mem`. If still insufficient, set `FINISH_ABORT` on the last request, release its KV cache, and return it via a new `reqs_to_abort` list
3. Add zero-division guard: `total_max_new_tokens + 1` (needed because the batch can now become empty after abort — this was impossible before since the old code either kept the last request or crashed)
4. Remove unreachable dead code

### `scheduler.py` — `update_running_batch`

1. Handle `reqs_to_abort`: send `AbortReq` to the tokenizer so clients receive an HTTP 500 error response instead of a hung connection
2. Add `batch.is_empty()` guard to skip `prepare_for_decode` when all requests have been retracted/aborted

## Verification

- All 41 SWA unit tests pass
- Re-ran the same 16K benchmark on v6e-64 (256 prompts, input=16384, output=1024, concurrency=128): 256/256 requests completed successfully, server did not crash

## Test plan

- [x] SWA unit tests pass (`test_swa_allocator.py`)
- [x] 16K benchmark on v6e-64: 256/256 successful, no crash

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -976,7 +976,9 @@ class ScheduleBatch:
         self._evict_tree_cache_if_needed(num_tokens)
         return self._is_available_size_sufficient(num_tokens)
 
-    def retract_decode(self, server_args: ServerArgs):
+    def retract_decode(
+        self, server_args: ServerArgs
+    ) -> tuple[list[Req], float, list[Req]]:
         """Retract the decoding requests when there is not enough memory."""
         sorted_indices = list(range(len(self.reqs)))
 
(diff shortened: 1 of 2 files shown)


JamesBrianD (approved):
LGTM

PR #952, opened 2026-04-23
feat: Allow sequence parallel and enable it for grok-2

hlFu:
## Motivation

For TP-sharded models on TPU (grok-2, TP=8), the reduce across the row-parallel of out-projections of attention, down-projections of MLP, weighted sum of expert activations of MoE was a plain `psum` that broadcasts the full hidden activation back to every device. And the following RMSNorm was observed to be very slow, mainly due to that the operations are done on VPU and grok-2 uses dual RMSNorm.
image

When enabling sequence parallel, each device only needs to do 1/TP compute for RMSNorm, this reduced the time cost linearly when the global tensor is large enough.
image

This PR introduces **sequence parallel** for grok-2: the row-parallel reductions become `psum_scatter` along the sequence dimension when the per-device slice is large enough to be worthwhile (default threshold ≥128 tokens/device), and all-gather activations back at the beginning of attention and MLP + MoE.

The feature is gated behind a new opt-in server flag and defaults to off, so existing behavior is unchanged.

## Other options
A fused dual RMSNorm kernel can also help with the RMSNorm slowness. This is an orthogonal solution and in the case of large seq_len the sequence parallel turns to have better performance.

## Modifications

- **New server arg `--enable-sequence-parallel`** (`server_args.py`), default `False`, plumbed through `model_runner.py` into `model_config.hf_config.enable_sequence_parallel`.
- **New helper `should_scatter(dim_size, num_devices)`** (`srt/utils/parallel_utils.py`) — returns `True` only when the local slice is ≥ `tpu_scatter_min_local_size` (default 128, overridable via `TPU_SCATTER_MIN_LOCAL_SIZE` env var) and the dim divides evenly across devices (hard requirement of `psum_scatter(tiled=True)`).
- **`QuantizedLinear` (and `xla_quantized_matmul_local`)** now accept an `output_scatter_dimension`. When set and `should_scatter(...)` is true, the post-matmul all-reduce is replaced with `psum_scatter(tiled=True)` along that dim, and `out_specs` is updated to reflect the sharded output.
- **`EPMoE`** wires `enable_sequence_parallel` down through the `shard_map`: when enabled and the token dimension is scatter-eligible, the internal `psum` over the `"tensor"` axis becomes a `psum_scatter(scatter_dimension=0, tiled=True)`, the shard_map `out_specs` switches from `P(None)` to `P("tensor", None)`, and the result is resharded back to the original mesh via the sharded pspec instead of a full replication.
- **`Grok1MLP` / grok.py**: threads `enable_sequence_parallel` through to the row-parallel down-projection and sets `output_scatter_dimension=0` on the affected `QuantizedLinear`s.
- **`global_config.py`**: adds `tpu_scatter_min_local_size` (default 128).
- **Unit test**: `test/srt/test_parallel_utils.py` covers the `should_scatter` predicate (num_devices ≤ 1, non-divisible dim, below-threshold, exactly-at-threshold).

### Unit tests
Ran unit tests in `python/sgl_jax/test/` folder
[unit-test-sp-0423.txt](https://github.com/user-attachments/files/27030239/unit-test-sp-0423.txt)
The only test failure also happens in main branch
[unit-test-sp-0423-gmm-main.txt](https://github.com/user-attachments/files/27030267/unit-test-sp-0423-gmm-main.txt)

### E2E accuracy tests
Benchmark: **GPQA-Diamond** (198 items), 5-shot (evalscope auto-forces 5-shot for `Idavidrein/gpqa`), sampling `temperature=0.5, top_p=0.8, top_k=40, max_tokens=4096`, run via `evalscope` against `sgl_jax.launch_server` serving `xai-org/grok-2` on TPU (`--tp-size=8 --dtype=bfloat16 --quantization-config-path=fp8_grok.yaml --disable-radix-cache --page-size=128 --random-seed=3`).

5 runs each on this branch (default-off: `enable_sequence_parallel=False`) and on `main` at commit `7907875a`:

| Branch                                 | n | Mean     | Min    | Max    | Stdev  | Runs                                           |
|----------------------------------------|--:|---------:|-------:|-------:|-------:|------------------------------------------------|
| main @ 7907875a                        | 5 | 0.4707   | 0.4495 | 0.4949 | 0.0165 | 0.4495, 0.4747, 0.4949, 0.4697, 0.4646         |
| allow-sequence-parallel-dev (flag=off) | 5 | 0.4697   | 0.4444 | 0.4848 | 0.0171 | 0.4798, 0.4848, 0.4444, 0.4798, 0.4596         |

- Δmean = +0.0010 (main higher), Welch's t = 0.094 — no detectable difference.
- The observed per-run spread is consistent with sampling noise (binomial 1σ on 198 items at ~47% is ≈3.5pp).
- This demonstrates **no regression on the default-off path**. The flag-on path (actual sequence-parallel execution) was not exercised here because `launch-grok.sh` does not pass `--enable-sequence-parallel`; a separate accuracy run with the flag set is recommended before enabling it by default.

### Summary

| Metric | Direction | sp_fixed wins / total | Mean Δ | Median Δ | Range |
|---|---|---|---|---|---|
| Total token throughput | higher is better | 15 / 18 | +24.63% | +3.16% | -36.7% … +396.1% |
| Output token throughput | higher is better | 15 / 18 | +24.17% | +3.16% | -36.7% … +393.3% |
| Median ITL | lower is better | 5 / 9 | -0.03% | -0.03% | -0.3% … +0.1% |
| Mean ITL | lower is better | 7 / 9 | +4.35% | -0.85% | -4.4% … +33.0% |
| Mean TTFT | lower is better | 15 / 18 | +90.86% | -8.99% | -38.3% … +1816.2% |
| Median E2E | lower is better | 17 / 18 | -9.15% | -4.50% | -95.6% … +7.7% |

**Reading the summary**: Median ITL is the steady-state decode metric — it's effectively flat (±0.3%) across all 9 OSL=1024 configs. Throughput and TTFT means are skewed by first-request JIT-compile spikes on the cold-start SP run (see the `1024/1024, conc=8` row which shows TTFT +1816% — that's a single cold-compile request inflating the mean, not a real regression).

### Side-by-side (all 18 configs)

| ISL / OSL | Conc | TotTput main | TotTput sp | Δ Tput | ITL med main | ITL med sp | Δ ITL med | ITL P95 main | ITL P95 sp | ITL P99 main | ITL P99 sp | TTFT main (ms) | TTFT sp | Δ TTFT |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| 1024 / 1 | 8 | 192.31 | 189.22 | -1.6% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 42615 | 43315 | +1.6% |
| 1024 / 1 | 16 | 296.66 | 401.82 | +35.4% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 45105 | 30695 | -31.9% |
| 1024 / 1 | 64 | 2044.94 | 2064.04 | +0.9% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 9554 | 9279 | -2.9% |
| 1024 / 1024 | 8 | 844.86 | 534.69 | -36.7% | 18.68 | 18.69 | +0.1% | 19.69 | 19.55 | 37.91 | 37.73 | 453 | 8684 | +1816.2% |
| 1024 / 1024 | 16 | 1421.08 | 1435.53 | +1.0% | 21.20 | 21.20 | +0.0% | 22.19 | 21.93 | 41.66 | 42.19 | 919 | 792 | -13.8% |
| 1024 / 1024 | 64 | 2112.16 | 1683.62 | -20.3% | 30.45 | 30.44 | -0.0% | 32.00 | 31.73 | 111.22 | 102.72 | 12339 | 12455 | +0.9% |
| 4096 / 1 | 8 | 2441.95 | 12114.96 | +396.1% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 3894 | 2402 | -38.3% |
| 4096 / 1 | 16 | 10821.91 | 11501.83 | +6.3% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 5264 | 4849 | -7.9% |
| 4096 / 1 | 64 | 10718.07 | 11361.49 | +6.0% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 20467 | 19264 | -5.9% |
| 4096 / 1024 | 8 | 1852.57 | 1868.11 | +0.8% | 18.73 | 18.75 | +0.1% | 19.27 | 19.55 | 19.55 | 19.95 | 1866 | 1846 | -1.1% |
| 4096 / 1024 | 16 | 2946.78 | 2995.21 | +1.6% | 21.28 | 21.21 | -0.3% | 21.95 | 22.03 | 22.41 | 22.47 | 3493 | 3173 | -9.2% |
| 4096 / 1024 | 64 | 5875.50 | 6042.51 | +2.8% | 30.69 | 30.66 | -0.1% | 31.90 | 31.62 | 32.85 | 32.84 | 12645 | 11510 | -9.0% |
| 8192 / 1 | 8 | 13206.16 | 14931.27 | +13.1% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 4242 | 3752 | -11.5% |
| 8192 / 1 | 16 | 13241.09 | 14970.32 | +13.1% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 8356 | 7391 | -11.5% |
| 8192 / 1 | 64 | 13255.88 | 14993.04 | +13.1% | 0.00 | 0.00 |  | 0.00 | 0.00 | 0.00 | 0.00 | 33072 | 29240 | -11.6% |
| 8192 / 1024 | 8 | 2939.56 | 3010.09 | +2.4% | 19.63 | 19.62 | -0.1% | 20.20 | 20.27 | 20.47 | 20.64 | 2846 | 2517 | -11.5% |
| 8192 / 1024 | 16 | 4388.20 | 4540.47 | +3.5% | 23.11 | 23.13 | +0.1% | 23.77 | 24.05 | 24.07 | 24.48 | 4936 | 4535 | -8.1% |
| 8192 / 1024 | 64 | 6735.37 | 7116.67 | +5.7% | 36.45 | 36.44 | -0.0% | 37.29 | 37.32 | 38.29 | 38.22 | 24908 | 22666 | -9.0% |

## Checklist

### Test command

```bash
# Unit test for the new helper
python -m pytest python/sgl_jax/test -x -v --tb=short --ignore=python/sgl_jax/test/multimodal 2>&1 | tee /cache/logs/tests/unit-test-sp-0423.txt

# Accuracy (default-off path, as tested here):
python3 -m sgl_jax.launch_server --model xai-org/grok-2 --trust-remote-code \
  --tp-size=8 --device=tpu --dtype=bfloat16 \
  --quantization-config-path=fp8_grok.yaml --page-size=128 \
  --tokenizer-path alvarobartt/grok-2-tokenizer \
  --disable-radix-cache --chunked-prefill-size 16384 --max-running-requests 256

evalscope eval --model xai-org/grok-2 \
  --api-url http://127.0.0.1:30000/v1/chat/completions --api-key EMPTY \
  --eval-type openai_api --datasets gpqa_diamond --dataset-hub huggingface \
  --eval-batch-size 198 \
  --dataset-args '{"gpqa_diamond": {"dataset_id": "Idavidrein/gpqa", "subset_list": ["gpqa_diamond"]}}' \
  --generation-config '{"temperature": 0.5,"top_p":0.8,"top_k":40, "max_tokens": 4096}'

# Benchmark
#!/bin/bash

set -e

if [ -z "$1" ]; then
    echo "Usage: $0 <engine>"
    echo "engine: sgl-jax, sglang-oai or vllm"
    exit 1
fi

backend=${1}
num_prompts_per_concurrency=3
input_seq_lens=(1024 4096 8192)
output_seq_lens=(1 1024) # 1 for TTFT
max_concurrencies=(8 16 64)

for input_seq_len in "${input_seq_lens[@]}"; do
    for output_seq_len in "${output_seq_lens[@]}"; do
        echo "======================================="
        echo "Testing with ISL/OSL: $input_seq_len/$output_seq_len"
        echo "======================================="
        for max_concurrency in "${max_concurrencies[@]}"; do
            echo "benchmark on max_concurrency=$max_concurrency"
            num_prompts=$((num_prompts_per_concurrency * max_concurrency))
            uv run python -m sgl_jax.bench_serving \
              --backend ${backend} \
              --dataset-name random \
              --num-prompts ${num_prompts} \
              --random-input-len ${input_seq_len} \
              --random-output-len ${output_seq_len} \
              --max-concurrency ${max_concurrency} \
              --random-range-ratio 1 \
              --disable-ignore-eos \
              --warmup-requests 0 \
              --tokenizer alvarobartt/grok-2-tokenizer
        done
    done
done

# Accuracy + perf with the feature enabled: add --enable-sequence-parallel
# to the launch-server command above.
```

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -1,5 +1,7 @@
 """GMM-based Expert-Parallel MoE layer and weight mapping utilities."""
 
+from functools import partial
+
 import jax
 from flax import nnx
 from jax import numpy as jnp
@@ -407,6 +409,20 @@ class EPMoE(nnx.Module):
         # Activation quantization is now handled per-GEMM inside _gmm_compute
         # (aligned with sglang-gpu scheme: quantize before each GEMM, dequantize after)
 
+        # Decide once whether SP should reduce-scatter the post-MoE result.
+        # Three things downstream consume this: the shard_map ``out_specs``,
+        # ``_forward``'s ``psum`` vs ``psum_scatter`` choice, and the
+        # post-shard_map reshard back to the original mesh. They MUST agree
+        # — if they ever diverge, ``out_specs`` lies about what the kernel
+        # produced and shard_map concatenates duplicates (see the bug fixed
+        # in ``QuantizedLinear``). Safe today only because input is
+        # pre-replicated to ``P(None)``; hoisting the decision keeps it safe
+        # if that ever changes.
+        do_scatter = (
+            self.enable_sequence_parallel
+            and should_scatter(hidden_states.shape[0], self.tp_size)
+        )
+
         # Run MoE computation on the expert-parallel mesh
         with jax.sharding.use_abstract_mesh(self.updated_mesh):
             hidden_states_reshard = jax.sharding.reshard(hidden_states, P(None))
@@ -430,14 +446,9 @@ class EPMoE(nnx.Module):
                 scale_name="wo_scale",
             )
 
-            if self.enable_sequence_parallel and should_scatter(
-                hidden_states.shape[0], self.tp_size
-            ):
-                out_specs = P("tensor", None)
-            else:
-                out_specs = P(None)
+            out_specs = P("tensor", None) if do_scatter else P(None)
             result = shard_map(
-                self._forward,
+                partial(self._forward, do_scatter=do_scatter),
                 mesh=self.moe_mesh,
                 in_specs=(
                     P(None),
@@ -473,18 +484,12 @@ class EPMoE(nnx.Module):
                 None,
             )
 
-        # Reshard result back to original mesh.
-        # When sequence parallel is on AND the scatter fired inside shard_map,
-        # the token dim is already striped across the moe_mesh "tensor" axis;
-        # preserve that by also striping it across the original mesh's
-        # ("data", "tensor") so the next decoder layer keeps the SP shard.
-        # Otherwise fall back to DP-only sharding on "data".
-        if self.enable_sequence_parallel and should_scatter(
-            hidden_states.shape[0], self.tp_size
-        ):
-            token_axis = ("data", "tensor")
-        else:
-            token_axis = "data"
+        # Reshard result back to original mesh. When the scatter fired inside
+        # shard_map the token dim is already striped across moe_mesh's
+        # "tensor" axis; preserve that by combining ``"data"`` with
+        # ``"tensor"`` on the original mesh so the next decoder layer keeps
+        # the SP shard. Otherwise fall back to DP-only on ``"data"``.
+        token_axis = ("data", "tensor") if do_scatter else "data"
         out_pspec = P(token_axis, *([None] * (result.ndim - 1)))
         return jax.sharding.reshard(result, jax.sharding.NamedSharding(self.mesh, out_pspec))
 
@@ -502,6 +507,8 @@ class EPMoE(nnx.Module):
         w0_kernel_bias=None,
         w1_kernel_bias=None,
         wo_kernel_bias=None,
+        *,
+        do_scatter: bool = False,
     ):
         expert_shard_id = jax.lax.axis_index("expert")
 
@@ -546,8 +553,10 @@ class EPMoE(nnx.Module):
 
         # All-reduce after unpermute: communication volume is (T, hidden_size)
         # instead of (T * top_k, hidden_size), reducing by a factor of top_k.
+        # ``do_scatter`` is decided once in ``__call__`` so this branch and
+        # the outer ``out_specs`` can never disagree.
         if self.tp_size > 1:
-            if self.enable_sequence_parallel and should_scatter(output.shape[0], self.tp_size):
+            if do_scatter:
                 # scatter on sequence/token dimension
                 output = jax.lax.psum_scatter(output, "tensor", scatter_dimension=0, tiled=True)
             else:
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -13,6 +13,7 @@ from sgl_jax.srt.kernels.gmm.megablox_gmm_backend import gmm
 # Re-export for backward compatibility: external code imports from this module.
 from sgl_jax.srt.layers.fused_moe import FusedEPMoE  # noqa: F401
 from sgl_jax.srt.layers.gate import GateLogit, TopK  # noqa: F401
+from sgl_jax.srt.utils.parallel_utils import should_scatter
 from sgl_jax.srt.utils.profiling_utils import named_scope
 from sgl_jax.srt.utils.quantization.quantization_utils import (
     quantize_tensor,
@@ -37,6 +38,7 @@ class EPMoE(nnx.Module):
         quantization_config=None,
         physical_to_logical_map: "jax.Array | None" = None,
         pre_gather_quant_dtype=None,
+        enable_sequence_parallel=False,
     ):
         self.num_experts_per_tok = num_experts_per_tok
         self.physical_to_logical_map = physical_to_logical_map
@@ -57,6 +59,7 @@ class EPMoE(nnx.Module):
         self.mesh = mesh
         self.activation = activation
         self.hidden_size = hidden_size
+        self.enable_sequence_parallel = enable_sequence_parallel
 
         # Get quantization settings from config
         self.quantized_dtype = (
@@ -427,6 +430,12 @@ class EPMoE(nnx.Module):
                 scale_name="wo_scale",
             )
 
+            if self.enable_sequence_parallel and should_scatter(
+                hidden_states.shape[0], self.tp_size
+            ):
+                out_specs = P("tensor", None)
+            else:
+                out_specs = P(None)
             result = shard_map(
                 self._forward,
                 mesh=self.moe_mesh,
@@ -447,7 +456,7 @@ class EPMoE(nnx.Module):
                     P("expert", None, "tensor"),
                     P("expert", None, None),
                 ),
-                out_specs=P(None),
+                out_specs=out_specs,
                 check_vma=False,
             )(
                 hidden_states_reshard,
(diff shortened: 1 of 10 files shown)


sii-xinglong:
### Code review

Found 1 issue:

1. **Outer/inner scatter decision mismatch in `EPMoE`** — The scatter check in `__call__` (line 433) uses `hidden_states.shape[0]` (the **global** sequence dimension), while the check in `_forward` (line 541) uses `output.shape[0]` (the **local** shard dimension inside `shard_map`). When global seq_len is in `[tp_size * 128, tp_size² * 128)` (e.g. TP=8, seq 1024–8191), the outer check passes (`1024 >= 8*128`) setting `out_specs=P("tensor", None)`, but the inner check fails (`128 < 8*128`) using `psum` (all-reduce) instead of `psum_scatter`. This produces a shape mismatch: `out_specs` expects sharded output but `_forward` returns replicated output. For TP=8, any sequence length between 1024 and 8191 tokens will hit this.

https://github.com/sgl-project/sglang-jax/blob/21a6cf8df601a725dd76a5266664bfbcda922062/python/sgl_jax/srt/layers/moe.py#L431-L439

https://github.com/sgl-project/sglang-jax/blob/21a6cf8df601a725dd76a5266664bfbcda922062/python/sgl_jax/srt/layers/moe.py#L539-L546

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

python/sgl_jax/srt/models/grok.py
```diff
@@ -191,12 +192,20 @@ def __init__(
             params_dtype=dtype,
             kernel_axes=("tensor", None),
             mesh=mesh,
+            output_scatter_dimension=0,
```
sii-xinglong: pass output_scatter_dimension when enable_sequence_parallel?

python/sgl_jax/srt/models/grok.py
```diff
@@ -435,6 +452,7 @@ def __init__(
             params_dtype=jnp.bfloat16,
             kernel_axes=("tensor", None),
             mesh=mesh,
+            output_scatter_dimension=0,
```
sii-xinglong: same as  down_proj

python/sgl_jax/srt/layers/moe.py
```diff
@@ -527,7 +538,11 @@ def _forward(
         # All-reduce after unpermute: communication volume is (T, hidden_size)
         # instead of (T * top_k, hidden_size), reducing by a factor of top_k.
         if self.tp_size > 1:
-            output = jax.lax.psum(output, "tensor")
+            if self.enable_sequence_parallel and should_scatter(output.shape[0], self.tp_size):
```
sii-xinglong: is there output.shape is hidden_states / tp_size ?

python/sgl_jax/srt/models/grok.py
hlFu: Yes. Fixed

python/sgl_jax/srt/models/grok.py
hlFu: Fixed

python/sgl_jax/srt/layers/moe.py
hlFu: It happens on seq dim so should be seq_len / tp_size. Not sure if you are asking this.

hlFu:
@sii-xinglong Could you check again?

Bob-Chen222 (approved):

hlFu:
The failed test is not related to this change because the SP flag is not enabled for the test and the test has been failing on main for the recent runs

Run | SHA | Branch | Observed | Pass?
-- | -- | -- | -- | --
25803336220 | ebd04fc1 | main | 9868.40 | ✓ (margin 2.4)
25775778407 | 75125c79 | PR #952 | 9833 / 9862 | ✗
25787533738 | 666a1c9e | main | 9779.24 | ✗
25713412493 | e2c5f910 | main | 9781.89 / 9839.16 | ✗
25668305532 | aebc1f06 | main | 9810.70 | ✗
25542367572 | 0cfbcbc3 | main | (4-tpu(1) fail) | ✗
25538103841 | 801870ea | main | (4-tpu(1) fail) | ✗

<br class="Apple-interchange-newline">

Issue #233, opened 2025-10-11
[Bug] tp_size=1 results in cache_miss in prefill_phase


aolemila:
### Describe the bug

There is a cache_miss when first request comes.

Server: tp_size = 1.

### Reproduction

```bash
# launch server
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache  python3 -u -m sgl_jax.launch_server --model-path Qwen/Qwen3-8B --trust-remote-code  --dist-init-addr=0.0.0.0:10011 --nnodes=1  --tp-size=1 --device=tpu --random-seed=3 --node-rank=0 --mem-fraction-static=0.8 --chunked-prefill-size=2048 --download-dir=/dev/shm --dtype=bfloat16 --max-running-requests 256 --skip-server-warmup --page-size=128 --disable-radix-cache

# curl
curl -X POST http://0.0.0.0:30000/v1/chat/completions  -H "Content-Type: application/json"  -d '{
    "model": "default",
    "messages": [
        {
            "role": "system",
            "content": "You are a helpful AI assistant"
        },
        {
            "role": "user",
            "content": "说出三个水果"
        }
    ],
    "temperature": 0.0,
    "max_tokens": 10
}'

# cache_miss, server log
[2025-10-11 11:18:32] Prefill batch. #new-seq: 1, #new-token: 128, #cached-token: 0, token usage: 0.00, #running-req: 0, #queue-req: 0,
[2025-10-11 11:18:33] Prefill batch. #bid: 1, #cache_miss: 1
[2025-10-11 11:18:33] INFO:     127.0.0.1:54990 - "POST /v1/chat/completions HTTP/1.1" 200 OK
```

### Environment

tpu-v6e

PR #966, opened 2026-04-25
feat: support hybrid attention models with recurrent state pool (Kimi-Linear)

Rodrian7:
## Overview

Adds memory and lifecycle management for mixed-attention models — models that
interleave standard attention layers (writing per-token KV cache) with linear
recurrent attention layers (writing per-request recurrent + conv state). The
target model is Kimi-Linear-48B-A3B-Instruct (27 layers: 7 MLA + 20 KDA), but
the design is generic for future Mamba / GDN / other recurrent variants.

> **Review note**: PR diff is `~1100 LoC business code + ~4400 LoC tests`. Tests
> are lock-in / contract / regression-guard cases. Reviewers can focus on the
> business code first; the test list at the bottom is a navigation map for
> when you want to drill into a specific contract.

### `RecurrentStatePool` (pure buffer pool)

Per-request recurrent + conv state buffer pool for linear recurrent layers.

- Dual list containers (recurrent f32 `[N+1, H, D, D]` + conv bf16 `[N+1, K-1, proj_size]`); slot 0 reserved as dummy (mirrors sglang PyTorch `MambaPool`).
- **Pure buffer pool**: slot allocator state lives in `HybridReqToTokenPool` (mirrors `MHATokenToKVPool` / `TokenToKVPoolAllocator` separation). Pool exposes `clear_slot(idx)`, `replace_buffer(buffers)`, `clear()` primitives.
- Pytree-registered for JIT donate; `tree_unflatten` rebuilds `layers_mapping` (global layer_id → local index).
- Sharded across TP mesh via `NamedSharding`; buffer creation mirrors `MHATokenToKVPool` (`with self.mesh:` + `jax.jit(out_shardings=...)`).
- **GQA-ready API**: `num_k_heads` / `head_k_dim` constructor args (default to V dims; mirrors sglang upstream `KimiLinearStateShape.create()`).
- **Functional write contract**: backend does `cur_layer.at[indices].set(new)` then returns to `replace_buffer`; pool list elements stay referentially unchanged across backend calls (no setter exposed).

### `HybridReqToTokenPool` (subclass with allocator state)

Coordinates KV slot + recurrent state slot allocation.

- `alloc(reqs)` atomic across both pools; pre-checks both capacities before any req-field mutation.
- **`free(req)` polymorphic override**: `free_recurrent_cache(req)` first then `super().free(req)` — recurrent-before-KV release order avoids inconsistent intermediate state. Type-checked guards on both parent and override (raise `TypeError` on legacy int callers with migration hint).
- **`free_chunked(req)` polymorphic override**: no-op so chunked req keeps both KV slot AND recurrent slot for next-round reuse (preserves recurrent state across chunks). Parent `ReqToTokenPool.free_chunked` does free + clear `req.req_pool_idx` for non-hybrid path.
- `recurrent_free_slots` allocator state on the subclass (NOT in pytree donate path) — eliminates a latent corruption hazard where pytree-leaf donate would silently reset `free_slots` after JIT donate / checkpoint-restore.
- `get_linear_recurrent_indices(req_pool_indices)` exposes CPU-side mapping for backend metadata.

### `MemoryPools` container

Pytree container that bundles all pools for a single JIT donate path.
`replace_all(updates)` accepts dict (multi-pool, hybrid) or list (legacy
single-pool, transitional). Per-pool `replace_buffer` carries each pool's
own `tp_size==1` device_put / sharding fix.

### Memory budget

- `--state-to-kv-ratio` (default 0.9, `r = state_budget / kv_budget`, mirroring sglang PyTorch `mamba_full_memory_ratio`) splits available HBM between pool types in `init_memory_pool`'s hybrid branch.
- `max_num_reqs` finalized as `min(kv_side_max_num_reqs, state_side_max_num_reqs)` — matches sglang upstream `_resolve_max_num_reqs`.

### Server args constraints (hybrid only)

- `disable_radix_cache=True` auto-set with `logger.warning` (mirrors sglang upstream `_handle_mamba_radix_cache(support_mamba_cache=False)` for KimiLinear)
- `disable_overlap_schedule` asserted (no double-buffer mechanism in this version)
- `state_to_kv_ratio > 0` validated with actionable error hint

### Lifecycle integration

- `release_req` / `cache_finished_req` go through polymorphic `pool.free(req)` — release-path dispatch lives in pool class hierarchy; no caller-side `hasattr` / helper / branching.
- `chunked_req` cleanup goes through polymorphic `pool.free_chunked(req)` — hybrid no-op preserves slot for cross-chunk reuse; non-hybrid frees + clears `req_pool_idx`.
- `flush_cache` resets recurrent state via `HybridReqToTokenPool.clear()`.

### Entry model migration

13 entry models migrated to return `(logits_output, pool_updates_dict)`;
`_forward` JIT-donates `MemoryPools` and dispatches `replace_all` from
per-model returns. `LlamaForCausalLMEagle3` inherits via parent.

Migrated: `llama, qwen, qwen2, qwen3, qwen2_moe, qwen3_moe, bailing_moe, mimo_mtp, mimo_v2_flash, deepseek_v3, gemma2, umt5, ling`.

## Compatibility

- **Non-hybrid models unchanged**: `MemoryPools.replace_all` accepts both legacy list and dict shapes during the transition window.
- **Existing KV pool paths preserved**: SWA / MLA / MHA pool creation paths in `init_memory_pool` not modified; only wrapped in `MemoryPools`.
- **Compatible with absorbed-MLA** (epic `c3372a9c`): KV pool creation path unchanged.

## Testing (177 cases)

Test files (lock-in / contract / regression-guard):

**Unit / pool**:
- `test_recurrent_state_pool.py` — pool primitives + GQA API + layer-id mapping + pytree roundtrip
- `test_hybrid_req_to_token_pool.py` — atomic alloc + free polymorphism (incl. legacy-int rejection guard) + recurrent slot allocator
- `test_memory_pools.py` — container + replace_all dispatch
- `test_jit_cache_stability.py` — pytree aux equal-by-value across constructions; cache key discrimination on diverging dims/dtypes

**Integration / model_runner**:
- `test_init_memory_pool_hybrid.py` — hybrid pool init + budget split
- `test_state_to_kv_ratio.py` — ratio CLI flag wiring
- `test_phase5_e2e_mock_model.py` — mock model JIT donate + replace_all roundtrip
- `test_recurrent_state_tp1_sharding.py` / `test_recurrent_state_tp4_sharding.py` — TPU real-device TP=1 + TP=4 sharding

**Managers / scheduler**:
- `test_alloc_req_slots_e2e.py` / `test_alloc_req_slots_signature.py` — `alloc_req_slots(reqs)` wiring
- `test_release_req_hybrid.py` / `test_cache_finished_req_hybrid.py` — caller release path uses polymorphic free
- `test_chunked_req_free_path.py` — chunked req `free_chunked` polymorphism
- `test_flush_cache_hybrid.py` — flush cache resets recurrent state

**Models migration**:
- `test_model_signatures.py` — 13 entry models lock-in + transition bridge removed lock-in

## Review focus

1. **Pool design**: list-of-array buffer layout, pool/allocator separation, layer_id-keyed accessor API
2. **Polymorphism**: `pool.free(req)` / `pool.free_chunked(req)` polymorphic dispatch (replaces caller-side `hasattr` branching)
3. **Pytree / JIT donate contract**: `tree_unflatten` aux + per-element sharding fix in `replace_buffer`; cache key stability
4. **Memory budget split**: per-token KV bytes + per-request state bytes formulas; `min()` cap mirroring sglang upstream
5. **Compatibility shims**: `MemoryPools.replace_all` accepting both list and dict during transition

Inline comments welcome.


python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -142,6 +184,113 @@ def clear(self):
...
 
+class HybridReqToTokenPool(ReqToTokenPool):
+    """Hybrid-architecture ReqToTokenPool that coordinates KV slots and recurrent state slots.
+
+    RFC-0015 §HybridReqToTokenPool.
+    - Inherits from ReqToTokenPool (KV slot behavior unchanged).
+    - Holds a reference to a RecurrentStatePool (the same instance also lives inside
+      MemoryPools).
+    - Maintains req_index_to_recurrent_index_mapping (CPU-side np.int32) bridging
+      req_pool_idx -> recurrent_pool_idx.
```
JamesBrianD: Please link this to the RFC issue, or submit the RFC into this pr.

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -121,14 +121,56 @@ def available_size(self) -> int:
...
+
+        Returns: list[int] (the slot per req), or None if capacity is insufficient.
+        """
+        if isinstance(reqs, int):
+            need_size = reqs
+            if need_size > len(self.free_slots):
+                return None
+            select_indices = self.free_slots[:need_size]
+            self.free_slots = self.free_slots[need_size:]
+            return select_indices
```
JamesBrianD: Please avoid hard-coding compatibility. If you change the interface signature to `def alloc(self, reqs: list[Req]) -> Optional[List[int]]:`, please ensure that all calling sites are updated accordingly.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,346 @@
...
+def _resolve_dtype(env_var: str, default):
+    """dtype priority: env var > default.
+
+    Constructor-arg priority is handled in __init__ (constructor arg > env var > default).
+    """
+    name = os.environ.get(env_var)
+    return _DTYPE_MAP[name] if name else default
+
+
+@jax.tree_util.register_pytree_node_class
```
JamesBrianD: First import `from jax.tree_util import register_pytree_node_class`, then use `@register_pytree_node_class`

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,346 @@
...
+        # multi-conv-segment expansion, mirroring PyTorch KimiLinearStateShape.conv: List[tuple]).
+        self.conv_buffers: list = [
+            [
+                jnp.zeros(
+                    (max_num_reqs + 1, conv_kernel_size - 1, self.proj_size),
+                    dtype=self.conv_dtype,
+                )
+            ]
+            for _ in range(self.num_layers)
+        ]
```
JamesBrianD: Please create the JAX array by following the code used for initializing the KV pool in the MHA memory pool. Like these:
```
with self.mesh:
  self.kv_buffer = []
  for _ in range(self.layer_num):
      kv_buf = jax.jit(
          lambda: jnp.zeros(
              shape=fused_buffer_shape,
              dtype=self.dtype,
          ),
          out_shardings=self.kv_sharding,
      )()

      self.kv_buffer.append(kv_buf)
```

python/sgl_jax/srt/models/gemma2.py
```diff
@@ -418,12 +418,13 @@ def _create_layer_mappings(self, layer_idx: int) -> dict:
     def __call__(
         self,
         forward_batch: ForwardBatch,
-        token_to_kv_pool: KVCache,
+        memory_pools,
```
JamesBrianD: Please add a type annotation for memory_pools: MemoryPools to keep the signature consistent with the other parameters.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -51,6 +54,142 @@
 logger = logging.getLogger(__name__)
 
 
+def _has_recurrent_state(hf_config) -> bool:
+    """Return True iff hf_config exposes a non-empty linear_attn_config["kda_layers"].
+
+    NOTE: linear_attn_config is a `dict | None` in real Kimi-Linear HF config
+    (see Decision #12); subscript access is required, not attribute access.
```
JamesBrianD: Please delete the meaningless comments in the docstring. The irrelevant note is unnecessary and should be removed completely.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -212,13 +351,16 @@ def jitted_run_model(
...
             logits_metadata,
         ):
             model_state = jax.tree_util.tree_unflatten(model_state_def, model_state_leaves)
             model = nnx.merge(model_def, model_state)
+            # Phase 2: model __call__ still expects token_to_kv_pool; bridge here.
+            # Phase 3 will migrate models to take memory_pools directly.
+            kv_pool = memory_pools.token_to_kv_pool
             with LoraBatchContext.set_batch(forward_batch):
-                return model(forward_batch, token_to_kv_pool, logits_metadata)
+                return model(forward_batch, kv_pool, logits_metadata)
```
JamesBrianD: The model implementation uses memory_pools, while we adopt kv_pool here for compatibility.

cjx0709: since model signature be mempool.token_to_kv_pool, why this place still do the `kv_pool = memory_pools.token_to_kv_pool`

python/sgl_jax/srt/server_args.py
```diff
@@ -129,6 +129,11 @@ class ServerArgs:
     disable_overlap_schedule: bool = False
     enable_precision_tracer: bool = False
 
+    # Hybrid recurrent state (RFC-0015): ratio of state buffer budget to KV cache
+    # budget for hybrid models. r = state / kv (matches sglang PyTorch
+    # mamba_full_memory_ratio). state_budget = available * r/(1+r).
+    state_to_kv_ratio: float = 0.9
```
cjx0709, line 135: this arg seems not effect for the memory alloc, now still alloc all memory for full attention

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -1092,3 +1241,57 @@ def load_cpu_copy(self, kv_cache_host, indices):
...
+            return self._pools[name]
+        except KeyError:
+            raise AttributeError(f"MemoryPools has no pool '{name}'") from None
+
+    def tree_flatten(self):
+        keys = sorted(self._pools.keys())
+        return [self._pools[k] for k in keys], tuple(keys)
+
+    @classmethod
+    def tree_unflatten(cls, keys, children):
```
JamesBrianD: Please use the standard signature: `def tree_unflatten(cls, aux_data, children):`

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,346 @@
...
+        # K=1 would make conv_buffers[l][i] second dim (K-1) zero; min meaningful value is 2.
+        assert conv_kernel_size >= 2, (
+            f"conv_kernel_size must be >= 2 (got {conv_kernel_size}); "
+            "K=1 produces empty conv buffers."
+        )
+        assert self.proj_size > 0, f"proj_size must be > 0, got {self.proj_size}"
+
+        # Dual list containers; each element has +1 row reserved for dummy slot 0.
+        # recurrent: list[Array] of length L, each [N+1, H, D, D].
+        self.recurrent_buffers: list = [
```
cjx0709: should think about shard which can accord to `MHATokenToKVPool`. Even using a function like `_create_buffers`.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,311 @@
+"""RecurrentStatePool — recurrent + conv state management for linear recurrent layers.
```
JamesBrianD: A Pallas kernel for updating `recurrent_state_pool` is missing.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,311 @@
...
+        aux = (
+            tuple(self.linear_recurrent_layer_ids),
+            self.max_num_reqs,
+            self.num_heads,
+            self.head_dim,
+            self.conv_kernel_size,
+            self.temporal_dtype,
+            self.conv_dtype,
+        )
+        return children, aux
```
JamesBrianD, line 353: Miss free_slots

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,311 @@
...
+
+        indices = self.free_slots[:need_size]
+        self.free_slots = self.free_slots[need_size:]
+
+        # Cross-layer list element mutation + intra-layer vectorized scatter.
+        idx_arr = jnp.asarray(indices, dtype=jnp.int32)
+        for layer in range(self.num_layers):
+            self.recurrent_buffers[layer] = self.recurrent_buffers[layer].at[idx_arr].set(0)
+            for inner in range(len(self.conv_buffers[layer])):
+                self.conv_buffers[layer][inner] = self.conv_buffers[layer][inner].at[idx_arr].set(0)
```
JamesBrianD: Here, the `for` loop iterates `layer_num` times; this should correspond to the number of linear layers.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -148,6 +148,17 @@ def to_json(self):
...
 
+def _maybe_free_recurrent_cache(req_to_token_pool, req) -> None:
+    """Phase 5: release recurrent state slot if pool is hybrid.
+
+    Mirrors Phase 4's _maybe_free_chunked_req_slot pattern (D1): duck-type
+    hybrid pool via hasattr(...,'free_recurrent_cache') and forward the
+    release. No-op for non-hybrid pools (keeps non-hybrid call sites clean).
+    """
+    if hasattr(req_to_token_pool, "free_recurrent_cache"):
+        req_to_token_pool.free_recurrent_cache(req)
```
JamesBrianD: The right fix is to **override `free` on `HybridReqToTokenPool`** so the polymorphism lives in the pool, not in every caller:

```python
# HybridReqToTokenPool
def free(self, req):
    self.free_recurrent_cache(req)        # idempotent when no recurrent slot
    super().free(req.req_pool_idx)
```

Then every call site is just `pool.free(req)` — no `hasattr` check, no `_maybe_free_recurrent_cache` helper, no risk of forgetting one of two calls.

Why this is better than `hasattr(..., "free_recurrent_cache")`:

1. **Polymorphism lives in the pool** — the scheduler shouldn't be probing pool internals to figure out which pool type it has. That's exactly what method override is for.
2. **Impossible to forget** — currently every caller must remember "if hybrid, also call free_recurrent_cache". That's a contract enforced by convention; the override turns it into a contract enforced by the type system.
3. **Removes duplicated `hasattr` branches** — the same check is repeated in `scheduler.py`, `_maybe_free_recurrent_cache`, `scheduler_output_processor_mixin.py`. All gone.
4. **One signature change**: `ReqToTokenPool.free(idx)` → `free(req)`. Mechanical refactor, plus it matches `alloc(reqs)` which already takes `Req` objects.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,152 @@
+"""RecurrentStatePool — KDA 层的 recurrent + conv state 管理。
+
+设计依据：RFC-0015 §RecurrentStatePool 对象设计（v2，list 容器）。
```
hlFu: Can you change to English comments?

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,152 @@
...
+        self.num_layers = num_layers
+        self.max_num_reqs = max_num_reqs
+        self.num_heads = num_heads
+        self.head_dim = head_dim
+        self.conv_kernel_size = conv_kernel_size
+
+        # proj_size 必须按 proj_v + 2·proj_k 写（开发指南 1.1 注意事项 #4 强制）
+        proj_v = num_heads * head_dim
+        proj_k = num_heads * head_dim
+        self.proj_size = proj_v + 2 * proj_k
```
hlFu, line 117: Is this true for every model? How about making it configurable and passed as argument for each model

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # 边界 assert
+        assert num_layers > 0, f"num_layers must be > 0, got {num_layers}"
+        assert max_num_reqs > 0, f"max_num_reqs must be > 0, got {max_num_reqs}"
+        assert num_heads > 0, f"num_heads must be > 0, got {num_heads}"
+        assert head_dim > 0, f"head_dim must be > 0, got {head_dim}"
+        # K=1 会让 conv_buffers[l][i] 第二维（K-1）为 0；最小有意义值是 2
+        assert conv_kernel_size >= 2, (
```
hlFu, line 127: Is this true? I think Ling model has kernel_size = 1?

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -688,6 +728,82 @@ def init_memory_pool(
...
+                )
+            logger.info(
+                "Hybrid budget split: max_num_reqs %s -> %s "
+                "(state_budget=%s bytes, per_req=%s bytes)",
+                max_num_reqs,
+                state_max_reqs,
+                state_budget,
+                per_req_state_bytes,
+            )
+            max_num_reqs = state_max_reqs
```
hlFu: Is it expected that max_num_reqs is overridden by state_max_reqs? instead of min(max_num_reqs, state_max_reqs)

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -51,6 +54,82 @@
...
+    for hybrid recurrent state models.
+
+    - disable_radix_cache: prefix slots being all-zero would corrupt suffix
+      computation in the recurrent path (matches sglang PyTorch
+      _handle_mamba_radix_cache(support_mamba_cache=False)).
+    - disable_overlap_schedule: this version does not implement double-buffer
+      ping-pong (mamba_ping_pong_track_buffer_size); overlap scheduler would
+      race on shared recurrent state.
+    """
+    if not server_args.disable_radix_cache:
```
hlFu: doesn't look like a common practice. I think explicitly throwing error makes the behavior more transparent

hlFu:
I think usually it's recommended to squash the commits into self-contained and stable ones so reviewers won't see intermediate changes.

hlFu:
Could we have some test to check whether JIT cache hit rate is expected? Just want to make sure the pytree generates stable cache key as expected so we can avoid perf issue.

python/sgl_jax/srt/mem_cache/memory_pool.py
Rodrian7: Done. Removed all internal RFC / design-doc references from docstrings across 18 files. The pool's behavior is now fully self-documented in code without external reference dependencies.

python/sgl_jax/srt/mem_cache/memory_pool.py
Rodrian7: Done. ReqToTokenPool.alloc / free both standardize on Req objects (no int compat shim). All call sites updated:

alloc: schedule_batch.alloc_req_slots(reqs) is the single entry
free: 10 call sites across 6 files (schedule_batch, scheduler, scheduler_output_processor_mixin, radix_cache, swa_radix_cache, chunk_cache)
Defensive TypeError guard with migration hint catches any missed legacy caller.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Done. _create_buffers now mirrors MHATokenToKVPool._create_buffers exactly:
```
with self.mesh:
    for _ in range(self.num_linear_recurrent_layers):
        buf = jax.jit(
            lambda: jnp.zeros(shape=recurrent_shape, dtype=temporal_dtype),
            out_shardings=self.recurrent_sharding,
        )()
        recurrent_buffers.append(buf)
```
Each buffer ships with persisted sharding from the start; replace_buffer's tp_size==1 device_put fix uses the persisted sharding (no probing).

python/sgl_jax/srt/models/gemma2.py
Rodrian7: Done. All 13 entry-model __call__ methods now annotate memory_pools: MemoryPools (gemma2 included), with MemoryPools imported at module top.

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done. _build_hybrid_pools docstring now contains only the functional description (constructor coordination + sharding contract), no internal plan numbering.

python/sgl_jax/srt/mem_cache/memory_pool.py
Rodrian7: Done. Both MemoryPools.tree_unflatten and RecurrentStatePool.tree_unflatten now use the standard (cls, aux_data, children) signature (aligned with the 13 other pytree classes in the codebase).

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Resolved by splitting allocator state from the buffer pool: recurrent_free_slots now lives on HybridReqToTokenPool (alongside the parent ReqToTokenPool.free_slots for KV), mirroring the existing MHATokenToKVPool vs TokenToKVPoolAllocator separation in sgl-jax. RecurrentStatePool is now a pure buffer pool exposing only clear_slot(idx) / replace_buffer(buffers) / clear(). This also eliminates a latent corruption hazard where pytree-leaf donate would silently reset free_slots after a JIT donate / checkpoint cycle.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Done. Renamed num_layers → num_linear_recurrent_layers (disambiguates from KV pool layer count). Loops in clear_slot / clear / replace_buffer / _create_buffers all iterate over the linear-recurrent layer count explicitly.

python/sgl_jax/srt/managers/schedule_batch.py
Rodrian7: Done — implemented exactly as suggested:


```
# HybridReqToTokenPool
def free(self, req):
    self.free_recurrent_cache(req)  # idempotent when no recurrent slot
    super().free(req)
```
All 10 call sites across 6 files now do pool.free(req); _maybe_free_recurrent_cache helper removed; hasattr(...,'free_recurrent_cache') checks in scheduler.py / scheduler_output_processor_mixin.py all gone. ReqToTokenPool.free signature changed from free(int | list[int]) to free(req) (matches alloc(reqs)).

Same polymorphism pattern applied to free_chunked(req) for the chunked-prefill release path: HybridReqToTokenPool.free_chunked is no-op (preserves both KV + recurrent slots for next-round reuse), parent does free + clear req_pool_idx. Replaced the _maybe_free_chunked_req_slot module-level helper.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -208,6 +208,11 @@ def __init__(
 
         # Memory pool info
         self.req_pool_idx: int | None = None
+        # HybridReqToTokenPool slot for recurrent + conv state (Kimi-Linear
```
cjx0709: don't need such detailed note(it will be used by KDA, GDN(qwen3), and other like GLA)

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Done. Imported register_pytree_node_class and applied as decorator on RecurrentStatePool.

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done. --state-to-kv-ratio now actually wires through to init_memory_pool's hybrid branch:

_split_state_kv_budget(available_bytes, ratio) splits HBM into state_budget + kv_budget (formula r/(1+r), mirrors sglang upstream mamba_full_memory_ratio)
KV pool sized from kv_budget; recurrent pool's max_num_reqs derived from state_budget // per_req_state_bytes
Final max_num_reqs = min(kv_side, state_side) (matches sglang _resolve_max_num_reqs)

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Done — aligned with MHATokenToKVPool exactly:

Extracted _create_buffers() method (same name as MHA pool)
with self.mesh: jax.jit(..., out_shardings=sharding)() pattern
recurrent_partition_axis / conv_partition_axis ctor args (default "tensor", matches MHA pool's kv_partition_axis)
replace_buffer per-element device_put for tp_size==1 (mirrors MHA pool's issue #233 fix)

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Done. All comments and docstrings are in English now.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Done. Added num_k_heads / head_k_dim constructor args (mirroring sglang upstream KimiLinearStateShape.create() defaults of num_k_heads = num_heads, head_k_dim = head_dim). proj_size formula now decoupled from V dims:


proj_v = num_heads * head_dim
proj_k = num_k_heads * head_k_dim  # ctor-overridable
self.proj_size = proj_v + 2 * proj_k
Default behavior unchanged for current Kimi-Linear callers; ready for GQA-style models.

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done. Changed from direct override to min(max_num_reqs, state_max_reqs) — matches sglang upstream model_runner_kv_cache_mixin._resolve_max_num_reqs:


max_num_reqs = min(max_num_reqs, self.server_args.max_mamba_cache_size // ratio)
Direct override would silently raise the cap when recurrent budget exceeds the KV-side cap; min() preserves the tighter bound.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Out of scope for this PR — current scope is Kimi-Linear only (which uses kernel_size >= 2). Will revisit the assert when adding Ling / other models with different conv kernel constraints. The pool API is already configurable through the conv_kernel_size ctor arg if a future model needs K=1.

python/sgl_jax/srt/server_args.py
Rodrian7: Done. --state-to-kv-ratio is now fully wired through init_memory_pool's hybrid branch — KV pool no longer consumes all available HBM:

_split_state_kv_budget(available_bytes, ratio) splits HBM as state_budget = available * r/(1+r), kv_budget = available - state_budget (mirrors sglang upstream mamba_full_memory_ratio).
KV pool is sized from kv_budget only (was previously sized from available_bytes, which is what your comment correctly flagged).
Recurrent pool's capacity comes from state_budget // per_req_state_bytes (per-request bytes derived from num_layers * num_heads/tp * head_dim^2 * dtype.itemsize for recurrent + num_layers * (K-1) * proj_size/tp * dtype.itemsize for conv).
Final max_num_reqs = min(kv_side, state_side) (mirrors sglang upstream _resolve_max_num_reqs).
Fail-fast _check_state_to_kv_ratio_for_hybrid raises if state_to_kv_ratio <= 0 for hybrid models (prevents the "ratio set but ignored" failure mode).
logger.info prints the actual split + max_num_reqs reduction so users can verify the wiring took effect.

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done. Removed the kv_pool = memory_pools.token_to_kv_pool bridge entirely. The JIT'd _forward now passes memory_pools directly to the model:


with LoraBatchContext.set_batch(forward_batch):
    return model(forward_batch, memory_pools, logits_metadata)
Both sides aligned:

Input side (the bridge you flagged): memory_pools: MemoryPools is the model's 2nd positional arg (was: token_to_kv_pool). All 13 entry models migrated; type-annotated. Inner attention layers receive the unpacked kv_pool from the entry model — keeping the entry model as the only MemoryPools boundary.
Output side: model returns (logits, pool_updates_dict); _forward dispatches MemoryPools.replace_all(pool_updates_dict) after the JIT donate cycle.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -51,6 +54,181 @@
...
+    tp_size: int,
+    token_to_kv_pool,
+    mesh,
+):
+    """Build the hybrid pool stack: RecurrentStatePool + HybridReqToTokenPool + MemoryPools wrapper.
+
+    Caller must pass the runner's linear_recurrent_config (a config whose
+    is_linear_attn is True with non-empty kda_layers); mesh is forwarded so
+    RecurrentStatePool's buffers get the same TP-aware sharding pattern as
+    MHATokenToKVPool.
```
aolemila: Please replace MHATokenToKVPool  with token_to_kv_pool, because we will support other pool like MLATokenToKVPool into the future.

python/sgl_jax/srt/model_executor/model_runner.py
aolemila: agree with you, just raise or assert @Rodrian7

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -550,8 +728,102 @@ def init_memory_pool(
...
+        if has_recurrent_state:
+            _enforce_recurrent_state_server_constraints(self.server_args)
+            _check_state_to_kv_ratio_for_hybrid(self.server_args.state_to_kv_ratio)
+
+            # Budget split must happen BEFORE token_to_kv_pool construction
+            # so the KV pool is sized against kv_budget instead of the full
+            # profile (otherwise state_to_kv_ratio is dead code: KV pool keeps
+            # all HBM and the state pool starves at runtime).
+            if self.is_hybrid or self.use_mla_backend:
+                # SWA-hybrid (full + swa KV pools) and absorbed-MLA (latent KV
```
aolemila: Current @zhengkezhou1 adds patch for kimi_linear which result the use_mla_backend=True, please discuss with @zhengkezhou1 to support MLATokenToKVCache or directly remove or annotate the `config.attention_arch = AttentionArch.MLA` for this version.

aolemila: In order to speedup, we can support MLATokenToKVPool in other PR

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -550,8 +728,102 @@ def init_memory_pool(
...
+                # those out of scope for now keeps the MHA + recurrent path
+                # honest. Removing this guard requires per-path per_token
+                # byte helpers that match each pool's actual storage shape.
+                raise NotImplementedError(
+                    "state_to_kv_ratio budget split is implemented only for "
+                    "the MHA + recurrent_state path; SWA-hybrid and MLA paths "
+                    "are TODO."
+                )
+
+            per_token_kv_bytes = _per_token_mha_kv_bytes(
```
aolemila: this calculation may be not right, please refer to `_compute_cell_size`.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -550,8 +728,102 @@ def init_memory_pool(
...
+                per_req_state_bytes,
+            )
+            # Take the smaller of the previously computed max_num_reqs (KV-side
+            # budget / context-len heuristic) and the recurrent-state budget cap.
+            # Mirrors sglang upstream model_runner_kv_cache_mixin._resolve_max_num_reqs:
+            #     max_num_reqs = min(max_num_reqs, max_mamba_cache_size // ratio)
+            # Direct override would silently raise the cap when the recurrent
+            # budget exceeds the KV-side cap; min() preserves the tighter bound.
+            max_num_reqs = min(max_num_reqs, state_max_reqs)
+
```
aolemila, line 650: please update codes according to sglang

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done. Generalized to "the token_to_kv_pool".

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done. Switched to `assert disable_radix_cache` (same idiom as the adjacent `disable_overlap_schedule`).

python/sgl_jax/srt/managers/schedule_batch.py
Rodrian7: Done. Simplified to one line and removed the Kimi-Linear/KDA reference.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -818,7 +828,7 @@ def prepare_for_extend(self):
 
         # Allocate req slots
         bs = len(self.reqs)
```
cjx0709, line 830: need to delete `bs`, this's a dead code

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -121,27 +121,249 @@ def available_size(self) -> int:
...
+
+        Returns the slot to recurrent_free_slots, sets req.recurrent_pool_idx
+        back to None and clears the corresponding mapping entry to the dummy
+        slot (0). Safe to call when the req has no recurrent slot allocated.
+        """
+        if req.recurrent_pool_idx is None:
+            return
+        self.recurrent_free_slots.append(req.recurrent_pool_idx)
+        self.req_index_to_recurrent_index_mapping[req.req_pool_idx] = 0
+        req.recurrent_pool_idx = None
```
aolemila, line 315: It looks like no one set req.req_pool_idx to None

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -1072,7 +1300,13 @@ def set_kv_buffer(
             "the MLA v2 kernel writes the cache in-place via input_output_aliases."
         )
 
-    def replace_kv_buffer(self, kv_buffer: list[jax.Array]) -> None:
+    def replace_buffer(self, kv_buffer: list[jax.Array]) -> None:
+        # tp_size==1 sharding fix (issue #233): NamedSharding constraint can be
+        # lost on JIT output; explicit device_put with self.kv_sharding restores
+        # it before the slice assign so the next JIT trace sees a stable shape.
+        # Mirrors RecurrentStatePool.replace_buffer  per-element fix.
+        if hasattr(self, "kv_sharding") and len(self.kv_sharding.device_set) == 1:
```
aolemila: Is tp_size==1 equal to len(self.kv_sharding.device_set) == 1? The same to other codes where this line occurs.

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -121,27 +121,249 @@ def available_size(self) -> int:
...
+      allocated). Public name mirrors the parent ReqToTokenPool.free_slots; the
+      separation between buffer pool and slot allocator follows the same shape as
+      MHATokenToKVPool vs TokenToKVPoolAllocator -- it keeps the buffer pool out
+      of JIT donate while leaving allocator state on the host-side scheduler
+      object.
+    - Maintains req_index_to_recurrent_index_mapping (CPU-side np.int32) bridging
+      req_pool_idx -> recurrent_pool_idx; default 0 (dummy) until a real slot is
+      written, so an unallocated req maps to the reserved slot 0.
+
+    **JIT compatibility warning (must NOT be passed to JIT)**:
```
cjx0709: the notes seems too much, maybe less it

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -1092,3 +1326,69 @@ def load_cpu_copy(self, kv_cache_host, indices):
...
+        if name.startswith("_"):
+            raise AttributeError(name)
+        try:
+            return self._pools[name]
+        except KeyError:
+            raise AttributeError(f"MemoryPools has no pool '{name}'") from None
+
+    def tree_flatten(self):
+        keys = sorted(self._pools.keys())
+        return [self._pools[k] for k in keys], tuple(keys)
```
aolemila, line 1337: Have you tested this tree_flatten works?

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -121,27 +121,249 @@ def available_size(self) -> int:
...
+        # Defensive type check: legacy callers passing bare int / list[int]
+        # hit an immediate, clear error instead of silent AttributeError
+        # downstream when self.free_slots.append(req.req_pool_idx) accesses
+        # a missing attribute.
+        if not hasattr(req, "req_pool_idx"):
+            raise TypeError(
+                f"ReqToTokenPool.free expects a Req object with req_pool_idx attr; "
+                f"got {type(req).__name__}. Legacy int/list[int] callers must migrate."
+            )
+        self.free_slots.append(req.req_pool_idx)
```
cjx0709, line 181: why here are not using `chunked_req.req_pool_idx = None`

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -51,6 +54,180 @@
...
+    """
+    if state_to_kv_ratio <= 0:
+        raise ValueError(
+            f"state_to_kv_ratio={state_to_kv_ratio} <= 0 is invalid for "
+            f"has_recurrent_state model: state budget would be 0; "
+            f"set --state-to-kv-ratio > 0 (default 0.9)"
+        )
+
+
+def _enforce_recurrent_state_server_constraints(server_args) -> None:
```
cjx0709: it looks too much for the util file in model_runner, maybe think about add a util file to put the functions. Model runner only with the core logic.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -764,33 +1058,16 @@ def _forward(
...
         # layers_topk_ids required real_bs and original_input_len which could not be stored in ForwardBatch
         return output, cache_miss_count, layers_topk_ids
 
-    def _set_kv_cache_after_forward(self, layers_kv_fused):
-        # Note: For tp_size == 1, we need to put the layers_kv_fused on the device with the target_sharding
-        # because sharding P(None, 'tensor') constraint has lost and this results in cache_miss for first prefill phase.
-        # Issue: https://github.com/sgl-project/sglang-jax/issues/233
-        # Q: Why does not call device_put in every layer?
-        # A: Because it does not work and cache_miss still happens. According to benchmark(https://github.com/sgl-project/sglang-jax/pull/234), the performance is not influenced.
-        if self.tp_size == 1:
```
cjx0709, line 782: the tp_size == 1 logic now be included in replace_all? I don't see it in the `memory_pool.py`

cjx0709 (commented):
the notes are too long to read, if the function is not very complex, maybe we can less it.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,400 @@
...
+          updates from layers 0..N-1 would all be lost.
+        - No-op for an empty index iterable.
+        """
+        indices = [idx_or_indices] if isinstance(idx_or_indices, int) else list(idx_or_indices)
+        if not indices:
+            return
+
+        idx_arr = jnp.asarray(indices, dtype=jnp.int32)
+        for layer in range(self.num_linear_recurrent_layers):
+            self.recurrent_buffers[layer] = self.recurrent_buffers[layer].at[idx_arr].set(0)
```
aolemila, line 231: no mapping for idx_arr?

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
```diff
@@ -0,0 +1,400 @@
...
+            for inner in range(len(self.conv_buffers[layer])):
+                self.conv_buffers[layer][inner] = self.conv_buffers[layer][inner].at[idx_arr].set(0)
```
aolemila, line 233: the same to above

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done in 9439937a — replaced `_per_token_mha_kv_bytes` with `self._compute_cell_size()`. The old helper used TP-total head count + raw `num_hidden_layers`, overestimating `available_bytes` by tp_size× and silently inflating `state_max_reqs` by the same factor. `_compute_cell_size` handles per-device heads, align128, adjust_layer_num, and MLA latent fast path — single source of truth for per-token KV bytes.

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Correct — `replace_all` only does shape normalization + key dispatch. The tp_size==1 sharding fix lives inside each pool's `replace_buffer` (e.g. `MHATokenToKVPool.replace_buffer:1308`, mirrored in MLA and `RecurrentStatePool`). Each pool knows its own sharding; `MemoryPools` only routes.

python/sgl_jax/srt/managers/schedule_batch.py
Rodrian7: `bs` is used at line 956

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done in 9439937a — moved 8 module-level hybrid recurrent helpers (`_build_hybrid_pools`, `_compute_recurrent_per_req_bytes`, `_split_state_kv_budget`, etc.) from `model_runner.py` to a dedicated `python/sgl_jax/srt/model_executor/hybrid_recurrent_utils.py`. Pure relocation, no logic change. `model_runner.py` now stays focused on the core forward / profile / init flow.

python/sgl_jax/srt/mem_cache/memory_pool.py
Rodrian7: Done in 9439937a — `ReqToTokenPool.free` now clears `req.req_pool_idx = None` (paired with simplified `free_chunked`).

python/sgl_jax/srt/mem_cache/memory_pool.py
Rodrian7: Done in 9439937a — trimmed `HybridReqToTokenPool` docstring (22 lines → 9 lines).

python/sgl_jax/srt/mem_cache/memory_pool.py
Rodrian7: Done in 9439937a — `ReqToTokenPool.free` now clears `req.req_pool_idx = None` (paired with simplified `free_chunked`).

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: KDA Pallas backend lives in sgl-project/sglang-jax#970; this PR is the memory/lifecycle infrastructure.

python/sgl_jax/srt/mem_cache/memory_pool.py
Rodrian7: Yes — covered by:
- `test_memory_pools.py::TestMemoryPoolsContainer::test_pytree_roundtrip_single_pool` / `test_pytree_roundtrip_two_pools`
- `test_memory_pools.py::TestMemoryPoolsJitDonate::test_jit_donate_then_replace_updates_state`
- `test_jit_cache_stability.py::TestMemoryPoolsJitCacheKey::test_distinct_memory_pools_same_params_share_cache` (locks aux equal-by-value semantics)

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done — switched to PR #968's MLA support: guard removed, _compute_cell_size() handles MLA latent path, MLATokenToKVPool auto-selected via existing use_mla_backend + fa branch. SWA-hybrid (is_hybrid=True) guard kept — that's a separate path.

python/sgl_jax/srt/model_executor/model_runner.py
Rodrian7: Done in 9439937a — switched to min(max_num_reqs, state_max_reqs), mirroring sglang upstream model_runner_kv_cache_mixin._resolve_max_num_reqs. Same fix as @hlFu's #25.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: idx_arr is a slot index (request-pool slot in [0, max_num_reqs]), not a layer id. layers_mapping only translates model-global layer_id → local layer index, which happens in get_linear_recurrent_layer_cache(layer_id). The .at[idx_arr] axis here is the leading slot dimension of recurrent_buffers[layer] (shape [N+1, H, D, D]), so no mapping is needed.

python/sgl_jax/srt/mem_cache/recurrent_state_pool.py
Rodrian7: Same as above

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -550,8 +559,99 @@ def init_memory_pool(
...
+                )
+            logger.info(
+                "Hybrid budget split (state_to_kv_ratio=%s): "
+                "max_total_num_tokens %s -> %s (kv_budget=%s bytes)",
+                self.server_args.state_to_kv_ratio,
+                self.max_total_num_tokens,
+                new_max_tokens,
+                kv_budget,
+            )
+            self.max_total_num_tokens = new_max_tokens
```
aolemila, line 618: This method is too wasteful and incorrect. Please fix it immediately in next PR. @Rodrian7

aolemila (approved):

PR #989, opened 2026-04-28
fix: support SWA radix cache

### LINKED ISSUE #993: [Feature] support radix cache for swa situation
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

Now swa memory pool only support for chunk cache situation, we need to support radix cache for swa model.

### Related resources

_No response_

cjx0709:
> close #993

## Summary

This PR adds sliding-window attention (SWA) support to the radix cache, enabling KV prefix reuse for hybrid models (e.g., MiMo) that mix full-attention and sliding-window-attention layers.

Previously, SWA was only supported with `ChunkCache` (`--disable-radix-cache`). This PR extends support to the radix tree cache (`SWARadixCache`), allowing prefix sharing across requests while correctly managing the dual-pool (full + SWA) KV memory.

## Core Changes

| Area | Change |
|------|--------|
| `SWARadixCache.evict_req_swa` | Per-request SWA eviction with tree-owned prefix protection. Uses `_node_prefix_len(req.last_node)` to derive the protected boundary, preventing eviction of tree-cached SWA slots. |
| `_evict_swa` safety margin | Eviction formula changed from `pre_len - sliding_window` to `pre_len - sliding_window - page_size`. The extra `- page_size` margin prevents `swa_evicted_seqlen` from reaching the live leaf boundary, ensuring `_insert_helper` can always create a non-tombstone leaf. |
| `_insert_helper` tombstone logic | Two sets of tombstone branch handling: (1) **Existing tombstone healing** (while loop) — 3 branches based on `swa_evicted_seqlen` vs `[node_start, node_end)`: full revive, split+partial revive, keep tombstone. (2) **New node tombstone split** (after while loop) — 3 cases ensuring leaf-must-not-be-tombstone invariant. |
| `_match_prefix_helper` | Stops at tombstone nodes to prevent returning SWA-freed slots as valid prefix. Unconditionally resets tombstone counter on each call. |
| Phase 2 SWA-only eviction | Uses `swa_lru_list` to find LRU nodes. Internal nodes become tombstones (SWA freed, full retained). Leaf nodes are fully deleted (both pools) to preserve the leaf-must-not-be-tombstone invariant, with cascade delete on parent. |
| Extend path gating | `isinstance(self.tree_cache, ChunkCache)` check — SWA eviction during extend only fires for `ChunkCache`, not `SWARadixCache`. For radix cache, SWA reclamation is deferred to tree-level Phase 2 eviction. |
| `release_req` | Use `req.last_matched_prefix_len` instead of `len(req.prefix_indices)` for cache boundary calculation. |
| `set_num_token_hybrid` | Fix memory allocation bug: SWA layer cost was double-discounted when distributing KV cache budget between full and SWA pools. The distribution equation now correctly weights SWA token-layers by the head ratio, recovering ~38.8% previously wasted HBM. |
| `evict_interval` formula | Aligned with upstream sglang ([#22645](https://github.com/sgl-project/sglang/pull/22645)). Lower bound changed from `max(min(sw, ps), 1)` to `max(page_size, int(sliding_window_size * multiplier))` with page alignment. Adds `SGL_JAX_SWA_EVICTION_INTERVAL_MULTIPLIER` env var (default `1.0`) to tune the SWA-waste vs. eviction-overhead tradeoff at runtime. |

## Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `SGL_JAX_SWA_EVICTION_INTERVAL_MULTIPLIER` | `1.0` | Controls the trade-off between SWA memory waste and eviction overhead during decode. The eviction interval is computed as `max(page_size, int(sliding_window_size * multiplier))`, then page-aligned. A larger multiplier reduces eviction frequency (lower overhead) but holds more SWA tokens beyond the sliding window (higher memory usage). |

**Example:**

```bash
# Default: evict every sliding_window_size tokens (page-aligned)
# For sliding_window=128, page_size=256 → evict_interval=256

# Memory-tight large-batch: evict more aggressively (smaller interval)
export SGL_JAX_SWA_EVICTION_INTERVAL_MULTIPLIER=0.5
# → evict_interval = max(256, int(128 * 0.5)) = 256 (page_size dominates)

# Memory-rich low-QPS: evict less often to reduce overhead
export SGL_JAX_SWA_EVICTION_INTERVAL_MULTIPLIER=4.0
# → evict_interval = max(256, int(128 * 4.0)) = 512
```

## Tombstone Healing (3 Branches)

When `_insert_helper` walks through an existing tombstone node beyond `update_kv_after_len`, it checks `swa_evicted_seqlen` against `[node_start, node_end)`:

| Branch | Condition | Action |
|--------|-----------|--------|
| 1 | `swa_evicted <= node_start` | Revive entire node (re-allocate SWA slots) |
| 2 | `node_start < swa_evicted < node_end` | Split node, revive back half only |
| 3 | `swa_evicted >= node_end` | Keep tombstone, free incoming SWA value |

## Test Coverage

| Category | Tests | Description |
|----------|-------|-------------|
| Tombstone lifecycle | 6 | Creation, prefix match, 3-branch healing, cascade delete |
| Size tracking | 3 | Consistency checks, fuzz testing (50-step random ops) |
| Protected prefix | 3 | Tree boundary protection, page alignment, `evict_req_swa` |
| New node invariants | 3 | Tombstone split, leaf-never-tombstone, exact boundary |
| Delete correctness | 2 | `get_child_key_fn` usage for O(1) delete |
| Writeback | 1 | `cache_unfinished_req` writeback range |
| DP rank isolation | 6 | Namespace isolation, shared namespace, multi-rank batching |
| Allocator safety margin | 5 | Updated expectations for `- page_size` formula |
| **Total** | **69** | 42 radix cache + 27 allocator |

### Server Launch Command

```bash
export RANK=0  # 0-3 for each node
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache python -u -m sgl_jax.launch_server \
    --model-path /models/MiMo-V2-Flash \
    --trust-remote-code \
    --tp-size 16 --dp-size 4 --ep-size 16 \
    --moe-backend fused \
    --nnodes 4 --node-rank "$RANK" \
    --dist-init-addr <node0-ip>:30000 \
    --host 0.0.0.0 --port 30271 \
    --page-size 256 \
    --context-length 262144 \
    --chunked-prefill-size 2048 \
    --dtype bfloat16 \
    --mem-fraction-static 0.85 \
    --swa-full-tokens-ratio 0.2 \
    --skip-server-warmup \
    --log-level info \
    --max-running-requests 128
```

### AIME2025 Results

| Subset | Num | Score |
|--------|-----|-------|
| AIME2025-I | 15 | **0.8667** |
| AIME2025-II | 15 | **0.9333** |
| **OVERALL** | **30** | **0.9000** |

### MMLU-Redux Results

| Benchmark | Num | Accuracy |
|-----------|-----|----------|
| MMLU-Redux | 5330 | **91.33%** |

Config: `temperature=0.0, top_p=1, max_tokens=16384, repeat=1`

## Design Document

Added `docs/design/swa_eviction_and_lru_strategy.md` covering:
- Dual-pool architecture (full pool + SWA pool)
- Per-request SWA eviction (ChunkCache vs SWARadixCache paths)
- Extend phase behavior (proactive vs deferred)
- Tree-level eviction (Phase 1 full delete + Phase 2 SWA-only tombstone)
- Tombstone insert logic (existing healing + new node split)
- LRU policy (dual lists: `full_lru_list` + `swa_lru_list`)

## Motivation

SWA layers only need recent tokens within the sliding window, but without radix cache support, every request allocated full-length SWA KV buffers with no prefix reuse. This wastes significant memory for models with many SWA layers (e.g., MiMo-V2-Flash has 39 SWA layers out of 48 total).

With this PR, SWA slots outside the window are freed progressively (per-request during decode, tree-level Phase 2 under memory pressure), while full KV remains in the tree for prefix matching. This reduces SWA memory usage to approximately `sliding_window + page_size` tokens per request for SWA layers.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -19,6 +19,7 @@ ScheduleBatch -> ModelWorkerBatch -> ForwardBatch
 
 import dataclasses
 import logging
+import os
 import threading
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Any
@@ -1333,7 +1334,9 @@ class ScheduleBatch:
         )
 
         if self.forward_mode is not None and self.forward_mode.is_decode():
-            evict_interval = max(min(sliding_window_size, page_size), 1)
+            multiplier = float(os.environ.get("SGL_JAX_SWA_EVICTION_INTERVAL_MULTIPLIER", "1.0"))
+            evict_interval = max(page_size, int(sliding_window_size * multiplier))
+            evict_interval = (evict_interval // page_size) * page_size
             for dp_rank, info in enumerate(self.reqs_info):
                 if not info.reqs:
                     continue
@@ -1361,9 +1364,13 @@ class ScheduleBatch:
                 continue
             for req in info.reqs:
                 pre_len = len(req.prefix_indices)
-                if self.enable_overlap and req.is_chunked > 0:
-                    if chunked_prefill_size is not None and chunked_prefill_size > 0:
-                        pre_len -= chunked_prefill_size
+                if (
+                    self.enable_overlap
+                    and req.is_chunked > 0
+                    and chunked_prefill_size is not None
+                    and chunked_prefill_size > 0
+                ):
+                    pre_len -= chunked_prefill_size
                 self._evict_swa(req, pre_len, sliding_window_size, page_size, dp_rank)
 
     def _evict_swa(
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -1357,10 +1357,10 @@ class ScheduleBatch:
 
         chunked_prefill_size = global_server_args_dict["chunked_prefill_size"]
         for dp_rank, info in enumerate(self.reqs_info):
-            if not info.reqs or info.prefix_lens is None:
+            if not info.reqs:
                 continue
-            for i, req in enumerate(info.reqs):
-                pre_len = info.prefix_lens[i]
+            for req in info.reqs:
+                pre_len = len(req.prefix_indices)
                 if self.enable_overlap and req.is_chunked > 0:
                     if chunked_prefill_size is not None and chunked_prefill_size > 0:
                         pre_len -= chunked_prefill_size
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -1045,8 +1045,7 @@ class ScheduleBatch:
             )
 
         # Evict SWA tokens outside sliding window
-        if self.is_hybrid:
-            self.maybe_evict_swa()
+        self.maybe_evict_swa()
 
     def new_page_count_next_decode(
         self,
@@ -1264,7 +1263,7 @@ class ScheduleBatch:
             self.req_to_token_pool.free(req.req_pool_idx)
         else:
             last_uncached_pos = (
-                len(req.prefix_indices) // server_args.page_size
+                req.last_matched_prefix_len // server_args.page_size
             ) * server_args.page_size
             token_indices = self.req_to_token_pool.req_to_token[
                 req.req_pool_idx, last_uncached_pos : seq_lens_cpu[idx]
@@ -1334,10 +1333,6 @@ class ScheduleBatch:
         )
 
         if self.forward_mode is not None and self.forward_mode.is_decode():
-            # Evict at the smaller of sliding_window_size and page_size to avoid
-            # stale SWA slot accumulation. For large windows (e.g., 4096) this
-            # prevents holding memory far beyond what's needed; for small windows
-            # (e.g., 128) it aligns with the natural eviction boundary.
             evict_interval = max(min(sliding_window_size, page_size), 1)
             for dp_rank, info in enumerate(self.reqs_info):
                 if not info.reqs:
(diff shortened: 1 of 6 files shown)


python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -1334,10 +1333,6 @@ def maybe_evict_swa(self, sliding_window_size=None):
         )
 
         if self.forward_mode is not None and self.forward_mode.is_decode():
-            # Evict at the smaller of sliding_window_size and page_size to avoid
-            # stale SWA slot accumulation. For large windows (e.g., 4096) this
-            # prevents holding memory far beyond what's needed; for small windows
-            # (e.g., 128) it aligns with the natural eviction boundary.
             evict_interval = max(min(sliding_window_size, page_size), 1)
```
JamesBrianD: The current formula min(sliding_window_size, page_size) has two issues worth reconsidering:

  1. Sub-page intervals are wasted work. _evict_swa aligns new_evicted down to a page boundary. If interval < page_size, two consecutive triggers can land on the same page boundary and the second one frees zero slots —   pure overhead. So the effective lower bound for a useful interval is page_size, not 1.
  2. No knob for the SWA-waste vs. eviction-overhead tradeoff. Between two evictions, each request holds up to interval extra SWA tokens beyond the sliding window. With a fixed formula, users can't tune this per  workload (memory-tight large-batch vs. memory-rich low-QPS).

I'd suggest mirroring the upstream sglang approach https://github.com/sgl-project/sglang/pull/22645 — add an envknob and use:
```
  # Lower bound: page_size (anything smaller frees nothing after page alignment).
  # Upper bound scaled by sliding_window so large windows trigger less often.
  # Multiplier exposes the waste-vs-overhead tradeoff at runtime.
  multiplier = envs.SGL_JAX_SWA_EVICTION_INTERVAL_MULTIPLIER.get()  # default 1.0
  evict_interval = max(page_size, int(sliding_window_size * multiplier))
  evict_interval = (evict_interval // page_size) * page_size
```

cjx0709: now add the env SGL_JAX_SWA_EVICTION_INTERVAL_MULTIPLIER for control the different scenarios

JamesBrianD (approved):
LGTM

PR #991, opened 2026-04-28
refactor(model_runner): align KV cache budget split with sglang upstream

Rodrian7:
Hybrid path now dispatches at top of `init_memory_pool`: calls `_profile_available_kv_cache_bytes` then splits raw bytes via `_split_state_kv_budget` (sglang `handle_max_mamba_cache` style — state count is floored, KV reclaims the leftover) instead of overwriting `profile_max_num_token`'s full-available result.

### Design notes

- **Top-level if/else dispatch, not sglang's hidden side-channel.** sglang's `handle_max_mamba_cache` mutates `server_args.max_mamba_cache_size` so downstream `_resolve_max_num_reqs` can read it. We rejected that pattern: sgl-jax has no `mambaish_config` / pool configurator multiplicity to absorb the side-channel, and explicit dispatch is more readable for a smaller codebase. Same algorithm, different delivery.
- **`_split_state_kv_budget` returns `(state_max_reqs, kv_budget_bytes)`** as a heterogeneous tuple, not a `NamedTuple`. Asymmetry is inherent (state pool sizes by count, KV pool sizes by bytes which depends on caller's `cell_size`). Type annotation + docstring document the units.
- **Budget logic inlined in `init_memory_pool` A-section**, not extracted into an orchestrator helper. The change is structural relocation + algorithm tweak; an orchestrator wrapper would hide the move from reviewers. Sub-helpers in `hybrid_recurrent_utils.py` (split / per_req / enforce / check) stay extracted.
- **`max_num_reqs` cap mirrors sglang `_resolve_max_num_reqs`**: when user-unspecified, apply context-len heuristic (`max(int(... / context_len * 512), 2048)` clamped to 4096), then `min(heuristic, state_max_reqs)`. Avoids `state_max_reqs > 4096` blowing up downstream `req_to_token_pool`.
- **`disable_radix_cache + disable_overlap_schedule` asserted, not auto-overridden.** Same constraints as sglang (radix cache poisons recurrent state, overlap scheduler races on shared state), but fail-fast with actionable error rather than silently overriding user config.

Test: `pytest python/sgl_jax/test/model_executor/test_kv_pool_layer_count.py`

### DIFF
--- a/python/sgl_jax/srt/model_executor/model_runner.py
+++ b/python/sgl_jax/srt/model_executor/model_runner.py
@@ -46,7 +46,6 @@ from sgl_jax.srt.model_executor.hybrid_recurrent_utils import (
     _build_hybrid_pools,
     _build_non_hybrid_memory_pools,
     _check_state_to_kv_ratio_for_hybrid,
-    _compute_max_num_reqs_from_state_budget,
     _enforce_recurrent_state_server_constraints,
     _per_req_state_bytes_from_config,
     _split_state_kv_budget,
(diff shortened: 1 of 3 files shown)


python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -507,18 +505,74 @@ def init_memory_pool(
...
+                max_num_reqs = min(max_num_reqs, state_max_reqs)
+            if self.max_total_num_tokens <= 0:
+                raise RuntimeError(
+                    "state_to_kv_ratio split left no KV budget. "
+                    "Lower --state-to-kv-ratio or increase --mem-fraction-static."
+                )
+            if max_num_reqs <= 0:
+                raise RuntimeError(
+                    "state_to_kv_ratio split left no recurrent state budget. "
+                    "Raise --state-to-kv-ratio or increase --mem-fraction-static."
```
aolemila: Increase --state-to-kv-ratio ?

Rodrian7: Done in 4ce34cdc.

aolemila (approved):

PR #997, opened 2026-04-28
docs(mimo): enable DP and radix cache in MiMo cookbooks

JamesBrianD:
## Summary

- DP attention (#939) and SWA radix cache (#989) are now supported, so the MiMo-V2-Flash and MiMo-V2.5-Pro cookbooks can drop `--disable-radix-cache` and recommend a default DP layout.
- `--dp-size` added to every launch command (multi-node v6e-16 / single-node v7x-8 for Flash; v7x-16 / v6e-64 / GKE manifest for Pro), keeping `tp_size` and `ep_size` unchanged so `attn_tp = tp_size / dp_size` lands at 4 (Flash) and 8 (Pro).
- Tip sections updated, `Known Issues` (radix cache) removed in the Pro cookbook, and the v2-Flash TPU configuration table gains a `DP Size` column.

## Test plan

- [ ] Re-run a v6e-16 MiMo-V2-Flash launch with the new flags and confirm radix cache is active.
- [ ] Re-run a v7x-16 MiMo-V2.5-Pro launch with the new flags and confirm DP attention works end-to-end.


cjx0709 (approved):
LGTM

PR #998, opened 2026-04-28
fix(memory_pool): allow alloc reuse on hybrid model's final prefill chunk

aolemila:
## Summary
- Fix `AssertionError` in `ReqToTokenPool.alloc` triggered on the final prefill chunk of any chunked-prefill req in hybrid models (e.g. Kimi-Linear).
- The previous `is_chunked > 0` assertion was incompatible with the chunked-prefill state machine on the final chunk: `HybridReqToTokenPool.free_chunked` retains the slot across chunks, while `is_chunked` is intentionally 0 on the final chunk (the per-chunk `+=1` is skipped because `add_chunked_req` returns `None`, and `is_chunked == 0` is required for the output processor's finalize branch).
- Replace the implicit `is_chunked` signal with an explicit `_chunked_slot_pinned` boolean on `Req`, set by `HybridReqToTokenPool.free_chunked` and cleared by `ReqToTokenPool.free`.

## Test plan
- [ ] Kimi-Linear (or other hybrid model) with a prompt long enough to span 3+ `chunked_prefill_size` chunks; verify prefill completes and decode begins.
- [ ] Non-hybrid model with chunked prefill; verify no regression (non-hybrid `free_chunked` still calls `free()` which leaves the pin cleared).
- [ ] Retraction path: `reset_for_retract` clears the pin.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -210,6 +210,9 @@ class Req:
         self.req_pool_idx: int | None = None
         # Slot in HybridReqToTokenPool's recurrent state pool; None for non-hybrid models.
         self.recurrent_pool_idx: int | None = None
+        # True iff the slot(s) were retained by a free_chunked() call and the
+        # next alloc(reqs) should reuse them instead of fresh-allocating.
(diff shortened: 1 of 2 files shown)


Iamleos (approved):
LGTM

Issue #992, opened 2026-04-28
[Feature] Memory Cache System Refactoring


JamesBrianD:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Support arbitrary feature combinations regarding the KV cache management sub-system.
- Prefix caching
- Request retraction/abortion
- Hybrid and Sparse attention
- Spec Decode

### Roadmap
- Support SWA Radix Cache @cjx0709 
- Fix partial rollout release kv cache
- Unify memory management across (overlap, non-overlap) x (page>=1) x (spec, non-spec, spec v2) x (retract, finished) https://github.com/sgl-project/sglang-jax/pull/994
- Stable req_to_token_pool slot ownership across chunked prefill — refactor alloc(reqs) to reuse slots in place for continuing chunked requests instead of free + realloc, and make free(req) req-aware. Eliminates a potential cross-stream/dispatch race where the scheduler can overwrite a slot still being read by an in-flight forward. https://github.com/sgl-project/sglang-jax/pull/1008 @Rodrian7 

### Related resources

- https://github.com/sgl-project/sglang/issues/12587
- https://github.com/sglang/sglang/issues/17850

PR #1008, opened 2026-04-30
fix(req_to_token_pool): eliminate cross-stream race on chunked prefill

Rodrian7:
## Motivation

Closes the chunked-prefill ownership item in #992 by aligning `ReqToTokenPool` with the upstream sglang fix in https://github.com/sgl-project/sglang/pull/17850 .

Under chunked prefill with the request pool near capacity, the scheduler used to free a chunked request's `req_pool` slot in `get_next_batch_to_run` and rely on the next allocation to hand the slot back. The freed slot could be re-handed to a different request whose matched-prefix indices were then written to the same row of `req_to_token` while the previous batch's forward was still reading that row's host buffer. JAX provides no happen-before between host numpy mutations and an in-flight dispatch, and on the PJRT host-CPU zero-copy path the host write is the device buffer; the result is silent KV gather corruption.

## Modifications

- `ReqToTokenPool.alloc(reqs)` is now req-aware. Reqs whose `req_pool_idx` is already non-None are treated as in-place reuse and do not consume `free_slots`. Atomic on capacity miss: returns `None` and leaves every req field untouched.
- `ReqToTokenPool.free(req)` releases the slot and clears `req.req_pool_idx`, preventing a stale value from leaking to a later caller after the slot has been re-handed.
- The explicit `req_to_token_pool.free(...)` for chunked requests in `Scheduler.get_next_batch_to_run` is removed; the chunked req keeps its `req_pool_idx` and the next `alloc(reqs)` reuses it in place.
- All call sites switch to the req-aware API:
  - `ScheduleBatch.alloc_req_slots(reqs)`
  - `ScheduleBatch.release_req` (retract path, both branches)
  - `ChunkCache.cache_finished_req`
  - `RadixCache.cache_finished_req` (both branches)
  - `SWARadixCache.cache_finished_req` (both branches)
- New unit tests under `python/sgl_jax/test/mem_cache/test_req_to_token_pool.py` lock in: fresh allocation, in-place reuse, atomic capacity miss (with and without partial reuse), free-clears-req-field, alloc/free roundtrip, and end-to-end chunked-req survival across batches. Registered in `test/srt/run_suite.py`.

## Accuracy Tests

N/A — semantics-preserving for non-chunked paths; chunked paths gain slot-ownership stability that the new unit tests cover.

## Benchmarking and Profiling

N/A — perf-neutral by design (one fewer free + alloc pair per chunked batch boundary, slight positive but not measured).

## Checklist

- [x] Format your code (pre-commit run --files passes on all touched files).
- [x] Add unit tests (test_req_to_token_pool.py, registered in run_suite.py).
- [ ] Update documentation - N/A.
- [ ] Provide accuracy and speed benchmark results - N/A (see above).
- [x] Follow the SGLang code style guidance.
- [ ] Work with maintainers to merge your PR.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -750,14 +750,14 @@ class ScheduleBatch:
         """Check if batch is empty (no requests in any DP rank)."""
         return self.batch_size() == 0
 
-    def alloc_req_slots(self, num_reqs: int):
-        req_pool_indices = self.req_to_token_pool.alloc(num_reqs)
+    def alloc_req_slots(self, reqs):
+        req_pool_indices = self.req_to_token_pool.alloc(reqs)
         if req_pool_indices is None:
             raise RuntimeError(
                 "alloc_req_slots runs out of memory. "
                 "Please set a smaller number for `--max-running-requests`. "
                 f"{self.req_to_token_pool.available_size()=}, "
-                f"{num_reqs=}, "
+                f"{len(reqs)=}, "
             )
         return req_pool_indices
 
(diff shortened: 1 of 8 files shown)


JamesBrianD:
LGTM. Should wait for CI tests.

JamesBrianD (approved):

PR #907, opened 2026-04-10
Adding GLM-4.7 

notabee:
## Motivation

This pull request adds support for the GLM-4.7 MoE model in `sglang-jax`. This enables running inference for GLM-4.7 MoE models on TPU clusters.

## Modifications

- Added `python/sgl_jax/srt/models/glm4_moe.py` which implements the GLM-4.7 MoE model architecture, including support for expert parallelism and static quantization.

## Accuracy Tests

Verified that the model generates coherent and correct output for a sample prompt.

**Server Launch Command:**
```bash
PYTHONPATH=<path_to_sglang_jax>/python TPU_HOST_BOUNDS=1,1,1 TPU_TOPOLOGY=2,2,1 JAX_COMPILATION_CACHE_DIR=<path_to_cache> python3 -u -m sgl_jax.launch_server \
  --model-path <path_to_model>/GLM-4.7-FP8 \
  --trust-remote-code \
  --dist-init-addr=0.0.0.0:10011 \
  --nnodes=1 \
  --tp-size=8 \
  --ep-size=1 \
  --device=tpu \
  --random-seed=3 \
  --mem-fraction-static=0.95 \
  --max-running-requests 1 \
  --max-total-tokens 2048 \
  --skip-server-warmup \
  --page-size=32 \
  --disable-radix-cache \
  --chunked-prefill-size 1024 \
  --precompile-token-paddings 1024 
```

**Test Command:**
```bash
curl -s http://127.0.0.1:30000/generate   -H "Content-Type: application/json"   -d '{
    "text": "[gMASK]<sop><|user|>\nWhat is the largest planet in the solar system?<|assistant|>\n",
    "sampling_params": {
      "max_new_tokens": 128,
      "temperature": 0.0,
      "stop": ["<|user|>", "<|observation|>"]
    }
  }'
```

**Output:**
```json
{
  "text": "The largest planet in the solar system is **Jupiter**.",
  "output_ids": [785, 7772, 11575, 304, 279, 12934, 1849, 374, 3070, 41, 19482, 334, 13],
  "meta_info": {
    "id": "7a0a8c0b600740789f3c17ffdbbd8a63",
    "finish_reason": {
      "type": "stop",
      "matched": 151336
    },
    "prompt_tokens": 16,
    "completion_tokens": 14,
    "cached_tokens": 0,
    "routed_experts": null,
    "cache_miss_count": 0,
    "e2e_latency": 1.6975910663604736
  }
}
```

## Benchmarking and Profiling

[...]

PR #1012, opened 2026-04-30
perf(moe): align expert tokens to gmm tile size once upstream

alexnails:
## Motivation

`EPMoE._gmm_compute` calls `gmm()` three times per layer (w0, w1, then wo via the SiLU/multiply'd intermediate). The current upstream pad in `_gmm_compute` only aligns x to the sublane tiling (~8 for bf16), which is enough for the LHS layout but not for the GMM kernel's row-tile constraints:

- `gmm_v2` produces NaNs when m falls below its 32-row tile, because `out_index_map` (`gmm_v2.py:106-117`) uses `floor` for non-last groups and `cdiv` for the last group — intermediate group ends on sub-tile boundaries leak NaN accumulator lanes.
- `gmm_v1` underutilizes the MXU below `tile_m = 128`.

Without coordinating these, any fallback inside `gmm()` that pads on the way in allocates a fresh padded LHS per call — for the three calls in `_gmm_compute`, that is **three** `jnp.pad`s per MoE layer, all on the same x (or its activated form), times ~80 MoE layers per forward.

## Change

Pad x once upstream in `_gmm_compute` to `max(sublane_align, kernel_tile_align)`, where `kernel_tile_align` = 32 for v2 / 128 for v1. The padded m flows through `layer_w0` / `layer_w1` / `intermediate_layer` / wo unchanged; `_unpermute` already trims `intermediate.shape[0] > expected_tokens` (`moe.py:691-693`), so the existing exit path handles the trim.

## Test plan

- [ ] Run GSM8K on a non-quantized MoE model on v5p/v7 (e.g. Qwen3-MoE) — score should match main.
- [ ] Run GSM8K on an FP8 MoE model (e.g. DeepSeek-V3-FP8 or GLM-4.7-FP8) — score should match main.
- [ ] Sanity check `gmm_v2` no longer NaNs at small unaligned m. Quick repro:
  ```python
  for m in (8, 33, 64, 96, 160):
      lhs = jnp.ones((m, k), jnp.bfloat16)
      sizes = jnp.array(...)
      out = gmm(lhs, rhs, sizes)
      assert not jnp.isnan(out).any()
  ```
- [ ] Profile a decode batch=1 forward and confirm the three per-layer `jnp.pad` HLOs in EPMoE collapse to one.

## Notes

- Independent of #907's `gmm()`-internal padding; if both land, this PR's upstream pad makes #907's internal pad a no-op for the EPMoE path.
- Does **not** address NaN/perf for non-MoE direct callers of `gmm()` (none in tree today).

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -9,6 +9,7 @@ from jax.sharding import PartitionSpec as P
 
 from sgl_jax.srt.eplb.expert_location import get_global_expert_location_metadata
 from sgl_jax.srt.kernels.gmm.megablox_gmm_backend import gmm
+from sgl_jax.srt.kernels.gmm.megablox_gmm_kernel.gmm_v2 import is_supported_by_gmm_v2
 
 # Re-export for backward compatibility: external code imports from this module.
 from sgl_jax.srt.layers.fused_moe import FusedEPMoE  # noqa: F401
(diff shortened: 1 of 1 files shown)


Bob-Chen222:
Could you push again to run CI? I think it is still regarded as a draft now

PR #1015, opened 2026-05-02
[Feature]extract precompile logic into CompilationManager

### LINKED ISSUE #609: [RFC][Feature] Refactor JIT Pre-Compilation
### **Checklist**

- [X] If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [X] Please use English, otherwise it will be closed.

### Goal

- Move precompile logic out of workers into a dedicated CompilationManager.
- Let model/spec workers submit compile tasks; manager handles dummy array creation, compilation, logging, and skip logic.
- **Cover aux_data variant combinations** to eliminate runtime recompilation.

### Current Pain Points

1. **Duplicated padding logic**: Bucket computation lives in `tp_worker.normalize_token_paddings()`, while runtime padding lives in `schedule_batch._compute_global_padding_sizes()` + `find_padding_size()`. Two independent implementations that must stay in sync.

2. **Incomplete aux_data variant coverage**: Pre-compilation only covers one aux_data combination per JIT function. When runtime sees a different aux_data value (e.g. `return_logprob=True` for the first time), it triggers recompilation.

3. **Tight coupling**: `tp_worker.generate_model_worker_batch()` (~70 lines) constructs synthetic batches for pre-compilation, duplicating knowledge of `ModelWorkerBatch` internals that also lives in `schedule_batch.get_model_worker_batch()`.

### Current Variant Handling (What's Already Done Well)

The sampler already handles most boolean flags **without recompilation** via two mechanisms:

- **`lax.cond`** for branching (`sampler.py:182-205`): `do_penalties`, `apply_vocab_mask`, `is_all_greedy` all use `lax.cond` — both branches are traced into a single HLO, no recompilation on value change.
- **`data_fields`** (`sampling_batch_info.py:63-77`): `is_all_greedy`, `do_penalties`, `need_min_p_sampling`, `apply_vocab_mask` are all pytree children, not aux_data. `linear_penalty` always passes a zero matrix when no penalties exist, keeping shape constant.

### Actual aux_data That Can Trigger Recompilation

**SamplingMetadata** — only 3 aux_data fields:
- `return_logprob` (bool) — pre-compiled as `False` only
- `top_logprobs_nums` (list|None) — pre-compiled as `None` only
- `token_ids_logprobs` (list|None) — pre-compiled as `None` only

**LogitsMetadata** — key aux_data risks:
- `extend_return_logprob`, `extend_return_top_logprob`, `extend_token_ids_logprob` (bool) — pre-compiled as `False`
- `extend_seq_lens_cpu`, `extend_logprob_start_lens_cpu`, `extend_logprob_pruned_lens_cpu` (list|None) — pre-compiled as `None`
- `top_logprobs_nums`, `token_ids_logprobs` (list|None) — pre-compiled as `None`
- `capture_hidden_mode` — pre-compiled as `NULL` (or `FULL` if multimodal)

### Prior Art

- **tpu-inference** (`vllm-project/tpu-inference`): `CompilationManager` owns bucket lists, iterates all variants, constructs dummy tensors. Sampler has only 2 `meta_fields` (`do_sampling`, `logprobs`), precompiles 4×N variants. Uses `_cache_collision_dummy` trick to prevent persistent cache collisions. **Critically, logprobs computation is a separate JIT function called outside the sampler**, so the sampler itself has no logprob-related aux_data.
- **MaxText**: Uses fully static shapes (training) or per-bucket AOT compilation caches (inference prefill). No runtime dynamic shape management.
- **sglang GPU** (`cuda_graph_runner.py`): Uses `recapture_if_needed()` for lazy re-capture — only re-captures all CUDA graphs when `capture_hidden_mode` escalates (NULL → LAST → FULL). This avoids paying the capture cost upfront for uncommon modes.

### Proposed Architecture

```
CompilationManager (new: srt/model_executor/compilation_manager.py)
├── Bucket management
│   ├── token_buckets, bs_buckets, cache_loc_buckets
│   └── pad_to_bucket(actual_tokens, actual_bs, mode) → (padded_tokens, padded_bs, cache_loc)
│
├── Variant matrix
│   ├── Forward: mode × token_bucket (return_logprob=False only at startup)
│   ├── Sampler: bs_bucket (return_logprob=False only at startup)
│   └── Logprobs: bs_bucket × token_bucket
│
├── Pre-compilation
│   ├── precompile_all()           ← single entry point
│   ├── _precompile_forward()
│   ├── _precompile_sampler()
│   └── _precompile_compute_logprobs()
│
├── Lazy compilation (inspired by cuda_graph_runner.recapture_if_needed)
│   └── _ensure_compiled(variant_flags) ← compile on first encounter, then cache
│
└── Dummy batch construction
    ├── _make_dummy_batch(mode, tokens, bs, **variant_flags)
    └── _make_dummy_sampling_metadata(bs, **variant_flags)
```

### Bucket Management

Bucket computation moves from `tp_worker.normalize_token_paddings()` and `tp_worker.__init__` into `CompilationManager.__init__()`:

- **Token buckets**: power-of-2 from 64 to `max_padded_num_tokens`, filtered by `dp_size` divisibility. User-overridable via `--precompile-token-paddings`.
- **BS buckets**: power-of-2 from 1 to `max_padded_batch_size`. User-overridable via `--precompile-bs-paddings`.
- **Cache-loc buckets**: derived from BS buckets as `bs * ceil(max_req_len / page_size) * page_size`.

`schedule_batch` continues to receive `(token_paddings, bs_paddings, cache_loc_paddings)` as arguments (signature unchanged). `tp_worker.get_precompile_paddings()` reads from `CompilationManager` instead of its own fields.

### Variant Strategy: Eager + Lazy

Inspired by `cuda_graph_runner.recapture_if_needed()`, we use a two-tier approach:

**Eager (at startup)**: Pre-compile the common path only:
- Forward: `{return_logprob=False}` × all token buckets × `{EXTEND, DECODE}`
- Sampler: `{return_logprob=False}` × all BS buckets
- This is what we do today — no additional startup cost.

**Lazy (at runtime)**: When a new aux_data combination is first encountered (e.g. `return_logprob=True`), `_ensure_compiled()` triggers compilation for that variant across all relevant buckets, then caches the results. Subsequent requests with the same aux_data hit the cache.

Benefits:
- **Zero additional startup time** — no logprob variant pre-compilation
- **First logprob request pays a one-time compilation cost** (~30-60s for all buckets), then all subsequent logprob requests are instant
- **Consistent with GPU sglang's pattern** — lazy re-capture on mode change

### Integration with tp_worker

`CompilationManager` is created in `tp_worker.__init__()`, receives a `ModelRunner` reference:

- `tp_worker.run_precompile()` → delegates to `compilation_manager.precompile_all()`
- `tp_worker.get_precompile_paddings()` → reads from `compilation_manager` attributes
- `tp_worker.generate_model_worker_batch()` → moves to `compilation_manager._make_dummy_batch()`
- `tp_worker.normalize_token_paddings()` → moves to `compilation_manager._compute_token_buckets()`

`tp_worker.forward_batch_generation()` and `model_runner.initialize_jit()` are unchanged.

### Out of Scope

- AOT compilation (MaxText-style `jit.lower.compile`)
- Speculative decoding pre-compilation refactor

### Future Work: Eliminate `return_logprob` as aux_data

Currently `return_logprob` remains an aux_data field because the sampler uses a Python `if` branch (`sampler.py:214`):

```python
if sampling_metadata.return_logprob:
    new_logits_output = self._process_logprob_results(logprob_operands)
```

This cannot be trivially converted to `lax.cond` because `_process_logprob_results` returns a `LogitsProcessorOutput` with multiple fields, while the `False` branch returns `None` — `lax.cond` requires both branches to return the same pytree structure. Additionally, `top_logprobs_nums` is a variable-length list that determines `jax.lax.top_k`'s `k` parameter, making it inherently a compile-time value.

**tpu-inference's approach**: Extract logprobs computation into a **separate JIT function** called conditionally from Python outside the sampler boundary. This way the sampler JIT has zero logprob-related aux_data. The logprobs function is precompiled independently.

This refactor would:
- Remove `return_logprob`, `top_logprobs_nums`, `token_ids_logprobs` from `SamplingMetadata` aux_data
- Remove `extend_return_logprob`, `extend_return_top_logprob`, `extend_token_ids_logprob` and associated list fields from `LogitsMetadata` aux_data
- Eliminate the need for lazy compilation of logprob variants entirely
- Reduce to a single compilation per bucket (1×N instead of 2×N)

This is a larger change that touches the sampler return value structure and the tp_worker calling convention, so it is deferred as a follow-up.

### Verification

1. **Unit test**: `test_compilation_manager.py` — bucket computation, `pad_to_bucket` edge cases
2. **Integration**: Run existing e2e tests, verify `cache_miss_count == 0` for normal requests
3. **Logprob regression**: Send logprob requests, verify lazy compilation triggers once then cache hits
4. **Startup time**: Compare pre/post refactor warmup duration; should be same or faster (no logprob pre-compilation)

### Related Resources

- [tpu-inference CompilationManager](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/runner/compilation_manager.py)
- [sglang cuda_graph_runner](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/model_executor/cuda_graph_runner.py)

JamesBrianD:
## Summary

- Extract bucket computation, dummy batch construction, and pre-compilation orchestration from `tp_worker.py` into a new `CompilationManager` class (`model_executor/compilation_manager.py`)
- `tp_worker` delegates to `CompilationManager.precompile_all()` while maintaining backward-compatible attribute aliases (`precompile_token_paddings`, `precompile_bs_paddings`, `precompile_cache_loc_paddings`)
- Add lazy compilation variant tracking (`is_new_variant()`) for future `return_logprob=True` support
- Fix BF16 MoE weight loading: `_view_as_fp8_if_needed` → `_reinterpret_dtype_if_needed`, add V2→bfloat16 view for bulk_read path
- Net: -249 lines from tp_worker, +330 lines in focused CompilationManager module

Closes #609

## Design

See [RFC: Refactor JIT Pre-Compilation](https://github.com/sgl-project/sglang-jax/issues/609) for full design rationale.

**Key decisions:**
- `CompilationManager` receives primitive params (dp_size, tp_size, etc.) instead of a full worker reference — single-direction dependency
- `forward_fn` callback pattern keeps CompilationManager decoupled from ModelRunner
- Lazy imports in CompilationManager avoid circular import issues

## Test results

Tested on TPU v6e pod (`ramezes-retract-chunked1024`):

**Qwen3-8B (dense, tp=4):**
- [x] Precompile: EXTEND 19s, DECODE 26s (with JIT cache)
- [x] Zero cache misses (53 requests)
- [x] Evalscope GSM8K: **100% (50/50)**, 161.56 tok/s

**Qwen3-30B-A3B-Instruct-2507 (MoE 128 experts, ep=4):**
- [x] BF16 bulk_read loading fixed (was: `Dtype |V2 is not a valid JAX array type`)
- [x] Precompile: EXTEND 90s, DECODE 91s
- [x] Zero cache misses
- [x] Evalscope GSM8K: **100% (50/50)**, 179.86 tok/s


JamesBrianD:
Review notes:

1. Blocking: this PR removes `find_padding_size` from `schedule_batch.py`, but `python/sgl_jax/test/test_flashattention_dp.py` still imports it from `sgl_jax.srt.managers.schedule_batch` and uses it in `create_test_data`. That file will fail import/test collection after this PR. Please either update the test to import/use `pad_to_bucket` from `common_utils`, or keep a compatibility wrapper/export in `schedule_batch.py`.

2. Follow-up: `pad_to_bucket` is replacing `find_padding_size`, but the old helper sorted its input defensively while the new helper assumes `buckets` is already sorted because it uses `bisect_left`. The new `CompilationManager` does sort its generated bucket lists, so the production path looks okay, but if this is meant to be the general replacement helper it should either sort internally or assert/validate sorted input to fail loudly.

3. Test coverage gap: the new tests cover bucket computation and `is_new_variant`, but they do not exercise the behavior-preserving part of the refactor (`precompile_all` / `_make_dummy_batch`). Since the PR primarily moves precompile batch construction out of `TpModelWorker`, please add at least a lightweight unit test that builds a dummy extend/decode batch and checks the key shapes/metadata against the previous expectations.

JamesBrianD:
All three review items addressed in commits `ba59e1b` and `6ce1411`:

**1. Blocking — broken `find_padding_size` import**
Fixed: `test_flashattention_dp.py` now imports `pad_to_bucket` from `common_utils` instead of the deleted `find_padding_size` from `schedule_batch`.

**2. Follow-up — sorted input validation**
Fixed: `pad_to_bucket` now includes `assert buckets == sorted(buckets)` — validates in debug mode, zero cost under `python -O`.

**3. Test coverage gap — dummy batch tests**
Fixed: Added `TestDummyBatch` class with 5 test cases covering:
- Extend batch shapes (input_ids, out_cache_loc, positions, cache_loc, extend_seq_lens, logits_indices)
- Decode batch shapes (no extend fields)
- DP metadata (dp_size, per_dp_bs_size, real_bs_per_dp)
- Multimodal CaptureHiddenMode.FULL
- Invalid cache_loc_size raises ValueError

Also added `test_compilation_manager.py` to `unit-test-tpu-v6e-1` suite in `run_suite.py`.

python/sgl_jax/srt/model_executor/compilation_manager.py
```diff
@@ -0,0 +1,312 @@
...
+            lora_ids=lora_ids,
+            dp_size=dp_size,
+            per_dp_bs_size=per_dp_bs_size,
+            real_bs_per_dp=[bs] * dp_size,
+            logits_indices_selector=np.arange(bs, dtype=np.int32),
+        )
+
+    # ---- Lazy compilation tracking ----
+
+    def is_new_variant(self, variant_key: tuple) -> bool:
```
Bob-Chen222: is_new_variant() has check-and-set semantics, but reads as a query — compilation_manager.py:307-312


def is_new_variant(self, variant_key: tuple) -> bool:
    if variant_key in self._compiled_variants:
        return False
    self._compiled_variants.add(variant_key)   # mutation hidden in is_*()
    return True
A predicate that mutates state is not helpful. Rename to register_variant_if_new(), or split into has_variant(key) + mark_compiled(key)

python/sgl_jax/srt/model_executor/compilation_manager.py
```diff
@@ -0,0 +1,312 @@
...
+                    batch, 0, mesh, self.vocab_size
+                )
+                batch.forward_batch = ForwardBatch.init_new(batch, model_runner)
+                if future_token_ids_map is not None:
+                    from sgl_jax.srt.managers.utils import resolve_future_token_ids
+
+                    batch.forward_batch.input_ids = resolve_future_token_ids(
+                        batch.forward_batch.input_ids, future_token_ids_map, mesh
+                    )
+                forward_fn(batch, None, False, sampling_metadata)
```
Bob-Chen222: Magic positional args in the forward callback — [compilation_manager.py:158, 218](vscode-webview://1ler07n9n230dmnsn58h7t90qail84qq6ivqmif6qq0t8edfrvlf/python/sgl_jax/srt/model_executor/compilation_manager.py)


forward_fn(batch, None, False, sampling_metadata)
The None, False are unlabeled here and inside forward_batch_generation. Pass them by name (e.g. launch_done=None, skip_attn_backend_init=False) so the call survives a signature reorder.

python/sgl_jax/srt/utils/common_utils.py
```diff
@@ -37,6 +38,29 @@
...
+        value: The actual size to pad.
+        buckets: Sorted list of bucket sizes.
+
+    Returns:
+        (padded_size, index) into *buckets*.
+
+    Raises:
+        ValueError: If value exceeds all buckets.
+    """
+    assert buckets == sorted(buckets), f"buckets must be sorted, got {buckets}"
```
Bob-Chen222: pad_to_bucket re-validates sortedness on every call — [common_utils.py:55](vscode-webview://1ler07n9n230dmnsn58h7t90qail84qq6ivqmif6qq0t8edfrvlf/python/sgl_jax/srt/utils/common_utils.py#L55)


assert buckets == sorted(buckets), f"buckets must be sorted, got {buckets}"
This is in the hot path for every batch padding decision (schedule_batch.py:1697,1872). Drop the assert — the callers control these lists, and bisect_left already requires sorted input. If you want the safety net, keep it behind __debug__ or move it to a one-shot check in the caller.

python/sgl_jax/srt/model_executor/compilation_manager.py
```diff
@@ -0,0 +1,312 @@
...
+            logits_indices_selector=np.arange(bs, dtype=np.int32),
+        )
+
+    # ---- Lazy compilation tracking ----
+
+    def is_new_variant(self, variant_key: tuple) -> bool:
+        """Returns True if this variant has not been compiled yet."""
+        if variant_key in self._compiled_variants:
+            return False
+        self._compiled_variants.add(variant_key)
```
Bob-Chen222, line 326: The compiled_variants set is updated in _precompile_extend / _precompile_decode, but it does not appear to be read anywhere, and is_new_variant() currently has no callers. In its current form, it seems to duplicate information that JAX already tracks internally, without any consumer.

One useful consumer would be to check this at runtime and log a warning, or bump a metric, when a request hits a variant that was not precompiled. That would let users know they paid a cold-compile cost and may need to extend their precompile_*_paddings.

python/sgl_jax/srt/managers/tp_worker.py
```diff
@@ -207,38 +194,22 @@ def __init__(
...
-        ):
-            self.precompile_bs_paddings.append(self.max_padded_batch_size)
-
-        # padding cache_loc_paddings
-        # note: the length of following two cache_loc_paddings must keep the same to length of separate bs_paddings.
-        self.precompile_cache_loc_paddings = [
-            item * ((self.max_req_len + self.page_size - 1) // self.page_size * self.page_size)
-            for item in self.precompile_bs_paddings
-        ]
+        self.precompile_token_paddings = self.compilation_manager.token_buckets
```
Bob-Chen222: After this PR, the buckets live in two places:

self.compilation_manager.{token,bs,cache_loc}_buckets — the new owner.
self.precompile_{token,bs,cache_loc}_paddings — shadow copies on the worker, kept only so the pre-existing get_precompile_paddings() accessor still works.
Since CompilationManager is now the source of truth, the shadow copies are redundant. Suggest pointing the accessor at the manager directly and deleting the copies:


# delete from __init__:
self.precompile_token_paddings     = self.compilation_manager.token_buckets
self.precompile_bs_paddings        = self.compilation_manager.bs_buckets
self.precompile_cache_loc_paddings = self.compilation_manager.cache_loc_buckets

# update the accessor:
def get_precompile_paddings(self):
    return (
        self.compilation_manager.token_buckets,
        self.compilation_manager.bs_buckets,
        self.compilation_manager.cache_loc_buckets,
    )

Bob-Chen222 (commented):
Claude helped with the review and I changed and selected comments that I think make sense. Thanks🙏

python/sgl_jax/srt/model_executor/compilation_manager.py
JamesBrianD: Fixed in `5058ee0`. Renamed to `register_variant_if_new()` — the check-and-set semantics are now explicit in the name.

python/sgl_jax/srt/model_executor/compilation_manager.py
JamesBrianD: Fixed in `5058ee0`. Now using keyword args: `forward_fn(batch, launch_done=None, skip_sample=False, sampling_metadata=sampling_metadata)`.

python/sgl_jax/srt/utils/common_utils.py
JamesBrianD: Agreed — removed the per-call assert. The callers (CompilationManager bucket builders and schedule_batch) already produce sorted lists, so the hot-path check was unnecessary overhead.

python/sgl_jax/srt/model_executor/compilation_manager.py
JamesBrianD: Good point. Added a TODO comment on `_compiled_variants` noting that a runtime consumer (warn on cold-compile cache miss) should be added in a follow-up refactor. The current set is pre-populated during precompile and will be consumed by upcoming `return_logprob=True` support (ref: issue #609).

python/sgl_jax/srt/managers/tp_worker.py
JamesBrianD: Fixed in `5058ee0`. Deleted the three shadow copies (`self.precompile_{token,bs,cache_loc}_paddings`) and pointed `get_precompile_paddings()` directly at `self.compilation_manager.{token,bs,cache_loc}_buckets`.

Bob-Chen222 (approved):

PR #862, opened 2026-03-05
Feat/support block quant

cjx0709:
## Motivation

Add block-wise quantization support from TPU inference, with coverage for both dense Linear layers and EPMoE.

Closes #879

## Modifications

- Add block-wise quant support for 'QuantizedLinear', basing on adapted quanted linear kernel code from TPU Inference project includeing:
  - per-channel
  - block-channel/ sub-channel
  - 2D block scale formats
- Add block-wise quant support for 'EPMoE', includeing:
  - dynamic/static scale loading
  - dynamic/static quantization
  - scale normalization to the GMM kernel layout 
- Add block-wise supported quantized matmul kernel for TPU inference
  - Integrate the TPU blockwise quantized matmul kernel with tuned fallback logic and safer validation checks.
- Move block-quant scale expansion (`jnp.repeat`) from runtime to init/load time for better inference performance

## Accuracy Tests
This PR will not effects for existing non-quantized model or channel-wise kernel. And add test List into CI about:
- Unit tests:
  - `python/sgl_jax/test/kernels/quantized_linear_test.py`
  - `python/sgl_jax/test/kernels/moe_block_quant_test.py`
  - `python/sgl_jax/test/test_linear_tp.py`
- TPU E2E tests:
  - `test/srt/test_moe_block_quant_e2e.py`
  - `test/srt/quantization/test_w8_block_dynamic_quantization.py`
  - `test/srt/quantization/test_w8_moe_block_linear_channel_quantization.py`

### Qwen3-30B-A3B (`tp=4, ep=4`)

| Eval Set | Variant | Score | Delta vs BF16 |
|---|---:|---:|---:|
| GSM8K | BF16 | 0.939 | - |
| GSM8K | INT8 mixed (`Linear per-channel + MoE block`) | 0.941 | +0.002 |
| GPQA Diamond | BF16 | 0.152 | - |
| GPQA Diamond | INT8 mixed (`Linear per-channel + MoE block`) | 0.141 | -0.011 |

### Qwen3-8B (`tp=4`)

| Eval Set | Variant | Score | Delta vs BF16 |
|---|---:|---:|---:|
| GSM8K | BF16 | 0.861 | - |
| GSM8K | INT8 block dynamic | 0.897 | +0.036 |
| GPQA Diamond | BF16 | 0.066 | - |
| GPQA Diamond | INT8 block dynamic | 0.121 | +0.055 |

## Benchmarking and Profiling

**Hardware:** TPU v6e-4
[...]

PR #1019, opened 2026-05-03
deepseekv3 update

Bob-Chen222:
## Motivation
add the quantization patch to make deepseekV3 runnable

## Modifications

Changed the quantization config and fixed quantization path in deepseek model file

## Benchmarking and Profiling

This is the current throughput of deepseekV3 with dp=2 and tp=16 (total 32 devices). Will do optimization in the following up.

Input lens: (8192,). Output lens: (1024,).
| batch size | latency (s) | input throughput (tok/s)  | output throughput (tok/s) | acc length | ITL (ms) | input cost ($/1M) | output cost ($/1M) |
| ---------- | ----------- | ------------------------- | ------------------------- | ---------- | -------- | ----------------- | ------------------ |
| 1 | 93.83 | 1335.66 | 11.68 | n/a | 85.64 | 0.59 | 47.58 |
| 4 | 117.34 | 2451.43 | 39.39 | n/a | 101.54 | 0.32 | 14.10 |
| 8 | 158.57 | 2448.42 | 62.15 | n/a | 128.72 | 0.32 | 8.94 |
| 16 | 225.31 | 2449.91 | 95.36 | n/a | 167.78 | 0.32 | 5.83 |
| 32 | 344.69 | 2444.34 | 138.00 | n/a | 231.88 | 0.32 | 4.03 |

## Checklist

### DIFF
--- a/python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
+++ b/python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
@@ -8,7 +8,6 @@ from jax import lax
 from sgl_jax.srt.kernels.quantized_matmul.blockwise_utils import (
     get_blockwise_kernel,
     get_safe_blockwise_tuned_value,
-    should_use_blockwise_kernel,
 )
 from sgl_jax.srt.utils.quantization.quantization_utils import quantize_tensor_simple
 
@@ -20,7 +19,6 @@ def xla_quantized_matmul_local(
     quantize_activation: bool = True,
     reduce_axis: str | None = None,
     compute_dtype: jnp.dtype | None = None,
-    weight_block_size: tuple[int, int] | None = None,
     activation_quant_dtype: jnp.dtype | None = None,
 ) -> jax.Array:
     """
@@ -58,26 +56,12 @@ def xla_quantized_matmul_local(
         in_blocks = w_scale.shape[0]
         block_size_in = in_dim // in_blocks
 
-        if weight_block_size is not None:
-            block_size_out = int(weight_block_size[0])
-        else:
-            block_size_out = block_size_in
-
         blockwise_kernel = get_blockwise_kernel()
         if blockwise_kernel is None:
             raise RuntimeError(
                 "Block-wise quantized matmul requires the blockwise kernel, "
                 "but it failed to load. Please check your installation."
             )
-        if not should_use_blockwise_kernel(
-            out_dim=int(out_dim),
-            block_size_out=int(block_size_out),
-        ):
-            raise RuntimeError(
-                f"Block-wise kernel does not support out_dim={out_dim} with "
-                f"block_size_out={block_size_out} (known to cause NaNs)."
-            )
-
         # w_scale is already in kernel-ready layout [in_blocks, 1, n_out].
         x_q_dtype = act_quant_dtype if quantize_activation else x.dtype
         tuned_value = get_safe_blockwise_tuned_value(
--- a/python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
+++ b/python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
@@ -20,6 +20,7 @@ def xla_quantized_matmul_local(
     quantize_activation: bool = True,
     reduce_axis: str | None = None,
     compute_dtype: jnp.dtype | None = None,
+    weight_block_size: tuple[int, int] | None = None,
     activation_quant_dtype: jnp.dtype | None = None,
     allow_narrow_n_blockwise: bool = False,
 ) -> jax.Array:
@@ -58,6 +59,11 @@ def xla_quantized_matmul_local(
         in_blocks = w_scale.shape[0]
         block_size_in = in_dim // in_blocks
 
+        if weight_block_size is not None:
+            block_size_out = int(weight_block_size[0])
+        else:
+            block_size_out = block_size_in
+
         blockwise_kernel = get_blockwise_kernel()
         if blockwise_kernel is None:
             raise RuntimeError(
@@ -72,7 +78,7 @@ def xla_quantized_matmul_local(
         # allow_narrow_n_blockwise=True to bypass this guard. The guard on the quant config is temporary
         if not allow_narrow_n_blockwise and not should_use_blockwise_kernel(
             out_dim=int(out_dim),
-            block_size_out=block_size_in,
+            block_size_out=block_size_out,
         ):
             raise RuntimeError(
                 f"Block-wise kernel does not support out_dim={out_dim} with "
--- a/python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
+++ b/python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
@@ -69,8 +69,7 @@ def xla_quantized_matmul_local(
         # achieve GSM8K 0.489 vs 0.938 for per-channel). Either fix the blockwise
         # kernel for out_dim <= block_size_out, or fall back to per-channel for those
         # layers. Models known to be unaffected (e.g. DeepSeek V3/R1) can set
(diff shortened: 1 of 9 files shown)


python/sgl_jax/srt/kernels/quantized_matmul/quantized_matmul_kernels/blockwise_kernel.py
```diff
@@ -130,12 +134,12 @@ def quantized_matmul_kernel(
...
     # TODO(amandaliang): use pltpu.get_tpu_info().mxu_column_size when JAX version is newer
     compute_tile_n = MXU_SIZE * n_lane_multiplier
-    steps_n = out_block_size // compute_tile_n
+    steps_n = max(1, out_block_size // compute_tile_n)
 
     def kernel(lhs_ref, rhs_ref, w_scales_ref, out_ref, acc_scratch):
         pid_k = pl.program_id(2)
         is_first_step = pid_k == 0
-        is_last_step = pid_k == (n_in - 1)
+        is_last_step = pid_k == (orig_n_in // in_block_size - 1)
```
JamesBrianD: Could you explain the is_last_step change? The grid dimension is n_in = padded_n_in // in_block_size, so pid_k ranges over [0, n_in) — meaning the old pid_k == (n_in - 1) should still trigger on the last iteration. What specific case does using orig_n_in fix here

python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
```diff
@@ -58,26 +56,12 @@ def xla_quantized_matmul_local(
...
             )
-        if not should_use_blockwise_kernel(
-            out_dim=int(out_dim),
-            block_size_out=int(block_size_out),
-        ):
-            raise RuntimeError(
-                f"Block-wise kernel does not support out_dim={out_dim} with "
-                f"block_size_out={block_size_out} (known to cause NaNs)."
-            )
-
```
JamesBrianD, line 80: This nan bugs have been fixed?

JamesBrianD (commented):
Thanks for the PR! Two requests:

1. More context on kernel/MoE changes: Could you explain what specific failure each of these three blockwise kernel fixes addresses? What errors or incorrect outputs did you hit before these changes?

Also, `should_use_blockwise_kernel` in `blockwise_utils.py` is now dead code after this PR — please clean it up.

2. Quantized vs non-quantized benchmark: The throughput table only shows FP8 results. Could you add a comparison against non-quantized (bf16) baseline on the same setup?

python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
Bob-Chen222: Oh actually I only tested on the deepseekV3 on the quantization and I wasn't able to reproduce the error. Can I get a reproduce command for NaNs on Qwen3-MoE k/v projections?

cjx0709: I add the Nan check when I merge the block-wise linear kernel, because I test qwen3-moe with Nan output when I try to block-wise quant the linear layer. 
I will test it today, it maybe not cause by the kernel but the model.

cjx0709: @Bob-Chen222 I add the guard in this PR: https://github.com/sgl-project/sglang-jax/pull/862. For fp8 block-wise quant, I tested with Nan output in qwen3-moe, I will check it agian tody.

Bob-Chen222: Thanks a lot

python/sgl_jax/srt/kernels/quantized_matmul/quantized_matmul_kernels/blockwise_kernel.py
cjx0709: it looks confict with `get_safe_blockwise_tuned_value` function: `    in_block_size = max(int(tuned.in_block_size), int(block_size_in))
    in_block_size = _next_multiple(in_block_size, int(block_size_in))
    in_block_size = min(in_block_size, _floor_multiple(int(n_in), int(block_size_in)))` can change the `in_block_size`. If you will add this, maybe delete the `get_safe_blockwise_tuned_value`

Bob-Chen222:
> Thanks for the PR! Two requests:
> 
> 1. More context on kernel/MoE changes: Could you explain what specific failure each of these three blockwise kernel fixes addresses? What errors or incorrect outputs did you hit before these changes?
> 
> Also, `should_use_blockwise_kernel` in `blockwise_utils.py` is now dead code after this PR — please clean it up.
> 
> 2. Quantized vs non-quantized benchmark: The throughput table only shows FP8 results. Could you add a comparison against non-quantized (bf16) baseline on the same setup?

For bf16 deepseekV3, I think TP=32 is the minimum requirement but for the fp8 deepseekV3, there are some difficulties to combine block quant and TP=32 (because shards of MLP layers will not be divisible by block quant size 128).

python/sgl_jax/srt/kernels/quantized_matmul/quantized_matmul_kernels/blockwise_kernel.py
Bob-Chen222: Thanks for the reminder. These changes reflected what is on tpu-inference so I will double check

JamesBrianD:
> For bf16 deepseekV3, I think TP=32 is the minimum requirement but for the fp8 deepseekV3, there are some difficulties to combine block quant and TP=32 (because shards of MLP layers will not be divisible by block quant size 128).

Understood. Could you please provide a separate comparison of the efficiency differences between bf16 linear and block-wise quantized linear? We want to understand the current state of quantitative performance.

python/sgl_jax/srt/kernels/quantized_matmul/kernel.py
cjx0709: @Bob-Chen222 I test it.                                                                                                                                                                                                                                        On a TPU v6e-4 pod with Qwen3-30B-A3B:                                                                                                                                                                                                     
  ```
  # Baseline (per-channel linear) — works fine                                                                                                                                                                                               
  python -m sgl_jax.launch_server --model-path /models/Qwen3-30B-A3B \                                                                                                                                                                       
    --tp-size 4 --ep-size 4 --moe-backend epmoe \                                                                                                                                                                                            
    --quantization-config int8_moe_block_128_linear_channel_dynamic.yaml                                                                                                                                                                     
                                                                                                                                                                                                                                             
  # Block-wise all layers — accuracy collapse                                                                                                                                                                                                
  python -m sgl_jax.launch_server --model-path /models/Qwen3-30B-A3B \          
    --tp-size 4 --ep-size 4 --moe-backend epmoe \                                                                                                                                                                                            
    --quantization-config int8_block_128_dynamic.yaml 
```
with GSM8K channel-wise quant achived 0.938 while block-wise achived 0.489.(and moe using block quant, linear use channel quant also works fine)
I revise my previous comment description: it doesn't appear to cause NaN, but it will introduce a significant accuracy error.
But I'm not sure this is caused by the quant kernel or only for this special model, if this gurad you think this too strict or in another model, this guard situation can work fine, you can delete this guard but maybe commit with a note for Qwen3-30B-A3B with block-wise quant can collapose the accuracy.

Bob-Chen222:
Hi @cjx0709, could you review this again? I reverted the changes related to the original blockwise implementation and added a temporary bypass flag in the quant config for the DeepSeek-V3 series. Thanks!

cjx0709:
@Bob-Chen222 LGTM, It now appears that the kernel's behavior is related to the specific model, so let's disable it by default and allow it case by case.

cjx0709 (approved):

PR #1026, opened 2026-05-06
fix pallas-kernel

pathfinder-pf:
## Motivation
The write operation does not involve writing the entire NT block at once; instead, it writes block by block.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/kernels/kda/kda.py
+++ b/python/sgl_jax/srt/kernels/kda/kda.py
@@ -592,7 +592,7 @@ def _chunk_gated_delta_rule_fwd_kernel(
 
     @pl.when(idx_nt < real_NT)
     def _():
-        h_ref[0, boh + idx_nt, 0] = scratch_ref[...].astype(h_ref.dtype)
+        h_ref[0, 0, 0] = scratch_ref[...].astype(h_ref.dtype)
 
         b_w = w_ref[0, 0]
         b_v = jnp.dot(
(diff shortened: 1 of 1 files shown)


aolemila:
We have tested this branch for three times. They are 0.6592, 0.6582 and 0.6567. The average of grades is 0.6580. The naive version is 0.66. So we regard this result means using pallas kernel in prefill for KDA has aligned the accuracy.

## More details

We run in tpu-v6e.

### Instructions 
```bash
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache \
uv run python -u -m sgl_jax.launch_server \
--model-path /models/Kimi-Linear-48B-A3B-Instruct/ \
--trust-remote-code  \
--tp-size=4 \
--device=tpu \
--mem-fraction-static=0.95 \
--chunked-prefill-size=512 \
--download-dir=/tmp \
--dtype=bfloat16 \
--max-running-requests 64 \
--skip-server-warmup \
--page-size=256  \
--disable-radix-cache \
--disable-overlap-schedule \
--precompile-bs-paddings 1 16 64 \
--precompile-token-paddings 16 128 512


evalscope eval --model /models/Kimi-Linear-48B-A3B-Instruct/  --api-url http://127.0.0.1:30000/v1   --api-key EMPTY  --eval-type service --datasets mmlu_pro  --eval-batch-size 32 --generation-config '{"temperature": 1.0}' --timeout 10000000
```

### Results

@aolemila 
image

@pathfinder-pf 
image

@[MokusMokun](https://github.com/MokusMokun)
image

PR #1050, opened 2026-05-10
feat(models): MiMoV2FlashMTPForCausalLM + restore EAGLE path broken since #939

### LINKED ISSUE #192: [Feature] Multi-Token Prediction (MTP) support for sgl-jax
### Checklist

### Motivation
Current autoregressive language models generate tokens sequentially, which creates inherent bottlenecks in inference throughput. While speculative decoding techniques like EAGLE improve performance through draft-verify mechanisms, they still rely on single-token predictions from the base model. Multi-Token Prediction addresses this limitation by enabling the model to directly predict multiple tokens, reducing the number of forward passes required for sequence generation.

### Design

 MTP extends the EAGLE speculative decoding framework by leveraging models with built-in multi-token prediction capabilities. Instead of generating single draft tokens, the system generates multiple tokens
  simultaneously from both draft and target models.

`Scheduler -> TpModelWorker/EgaleModelWorker -> ModelRunner`

The EAGLE (Extrapolation Algorithm for Greater Language-model Efficiency) speculative sampling algorithm accelerates large language model (LLM) inference by predicting multiple tokens ahead and verifying them in parallel. Here's a concise breakdown:

`1. Drafting Phase (Small Model) -> 2. Verification Phase (Large Model) -> 3. Fallback (if All Draft Tokens Are Rejected)`

A smaller, faster draft model generates a sequence of K candidate tokens (draft) autoregressively.
The target (large) model processes the draft tokens in parallel, computing their probabilities.
Using a modified rejection sampling method:
Accept each draft token if its probability under the large model is higher than under the draft model.
If rejected, resample a corrected token from the large model and discard the rest of the draft.

If no tokens are accepted, the large model generates a single token normally.

### Related resources

_No response_

zorrofox:
## Summary

Adds NEXTN/EAGLE speculative-decoding draft model for MiMo-V2-Flash / V2.5 / V2.5-Pro
(single-layer SWA attention + dense FFN, loaded from `model.mtp.layers.0` in the same
checkpoint), and fixes the EAGLE path which has been broken on `main` since #939 (DP
refactor) / #940 (sampler RNG fold_in).

Closes #192.

### Verified on v6e-16 (V2-Flash, tp=16 ep=16)
- **accept-len mean 2.77** (topk=1, num_steps=3, num_draft=4) — H100 reference: 2.92
- greedy output **token-identical** to no-MTP on 3/4 test prompts (4th diverges at
  token[1] due to bf16 batch-4 vs batch-1 reduction order, both outputs valid)
- arithmetic / code-completion prompts produce correct results

### Launch
```bash
    --model-path <hf-or-local> --trust-remote-code \
    --tp-size 16 --ep-size 16 --moe-backend epmoe \
    --speculative-algorithm EAGLE \
    --speculative-draft-model-path <same-as-model-path> \
    --speculative-num-steps 3 --speculative-eagle-topk 1 \
    --speculative-num-draft-tokens 4 \
    --disable-overlap-schedule \
    ...
```

### A. EAGLE-path regressions since #939/#940 (independent of MiMo)
Any EAGLE/NEXTN draft model crashes on current `main`:
- `model_runner._sampler_base_rng` only set when `not is_draft_worker`, but
  `EAGLEWorker` calls `initialize_jit()` which reads it (#940)
- `get_spec_model_worker_batch` still references pre-#939 flat attrs
  (`self.extend_lens` / `self.seq_lens` / `self.spec_info` / ...) and misses
  `real_bs_per_dp` / `_merge_sampling_info` / `per_dp_bs_size`
- explicit-sharding mode: `top_k` on vocab-sharded logits fails; host-side
  `jnp.at[].set` / `concatenate` / gather hit `ShardingTypeError` because jit
  outputs carry P("data","tensor") into eager ops
- `mimo_mtp.py` still returns 3-tuple (model_runner unpacks 4)
- `prepare_for_extend_after_verify` doesn't set `logits_indices`

Fix: `__getattr__`/`__setattr__` shim on `ScheduleBatch` routing to `reqs_info[0]`
(spec is dp=1 only for now), replicate jit outputs to `P()` at every host-side
boundary, add missing MWB fields.

### B. `ragged_paged_attention_v3` custom_mask DMA at TARGET_VERIFY
Mosaic requires dynamic slice offset+size to be tiling(8)-aligned, but
`kv_len = seq_len + num_draft_tokens` is not. `_fetch_mask` fails to compile.

Fix: host pads each draft-token row of `custom_mask` to page-aligned `kv_len`
(stable shape across decode steps → no per-step recompile), kernel uses
`mask_kv_len = cu_kv_lens[seq+1]-cu_kv_lens[seq]` for mask stride only, keeping
the unaligned `kv_len` for `_fetch_bkv`'s cache/new-KV split. `pl.multiple_of`
hints prove divisibility.

Also: `get_eagle_forward_metadata` now computes `swa_page_indices` for hybrid
SWA+full targets (otherwise SWA layers index the swa sub-pool with full-pool
page ids during verify).

Ref `q_span/kv_span` hoist (already noted in #192).

### C. `MiMoV2FlashMTPForCausalLM`
- single `MiMoV2FlashMTPLayer` reusing `MiMoV2Attention` + `MiMoV2MLP`
- weight load mirrors target's FP8→bf16 dequant pipeline (`dequant_fp8_layers` +
  `dequant_fused_kv` + `replicate_kv_heads`); K/V via `__KV_*__0` buffers
- `model_config` maps `MiMoV2{Flash,}ForCausalLM` → MTP class when `is_draft_model`,
  forces `num_hidden_layers=1`, adds `eh_proj`/`o_proj` to `ignored_layers`
- `MiMoV2FlashForCausalLM.get_embed_and_head()` for draft head sharing

## Known limitations / follow-ups
- **host-orchestration overhead**: per-decode-round `device_put` resharding +
  eager `update_eagle_lists` make MTP ITL currently *worse* than no-MTP at bs=1
  (~18ms vs ~7ms warm). Needs jit-fusion of draft loop. accept-len is correct.
- spec path is dp=1 only (`__getattr__` shim assumes `reqs_info[0]`)
- `--disable-precompile` required (precompile_spec_* not yet adapted)
- KV-pool free on req finish leaks 1 page/req under spec path (allocate_lens vs output_ids accounting) — long-running will OOM
- only `mtp.layers.0` is loaded (3-layer multi-step EAGLE is a follow-up)

## Test plan
- [ ] `pytest test/test_rpa_custom_mask_verify.py` on v6e-4 (TBD)
- [x] v6e-16 V2-Flash e2e: accept-len 2.77, greedy 3/4 token-identical
- [ ] GSM8K-100 accuracy parity
- [ ] V2.5-Pro v6e-64

### DIFF
--- a/python/sgl_jax/srt/configs/model_config.py
+++ b/python/sgl_jax/srt/configs/model_config.py
@@ -141,6 +141,12 @@ class ModelConfig:
 
         if is_draft_model and self.hf_config.architectures[0] == "MiMoForCausalLM":
             self.hf_config.architectures[0] = "MiMoMTPForCausalLM"
+
+        if is_draft_model and self.hf_config.architectures[0] in (
+            "MiMoV2FlashForCausalLM",
+            "MiMoV2ForCausalLM",
+        ):
+            self.hf_config.architectures[0] = "MiMoV2FlashMTPForCausalLM"
         # Check model type
         self.is_generation = is_generation_model(self.hf_config.architectures, is_embedding)
         self.is_multimodal = False
--- a/python/sgl_jax/srt/configs/model_config.py
+++ b/python/sgl_jax/srt/configs/model_config.py
@@ -147,6 +147,10 @@ class ModelConfig:
             "MiMoV2ForCausalLM",
         ):
             self.hf_config.architectures[0] = "MiMoV2FlashMTPForCausalLM"
+            # Draft uses only mtp.layers.0 (single SWA layer); without this the
+            # KV pool sizes for 48 target layers and OOMs.
+            self.hf_config.num_nextn_predict_layers = 1
+            self.hf_config.num_hidden_layers = 1
             if self.quantization_config is not None:
                 # Draft MTP head: eh_proj and o_proj are bf16 in the checkpoint
                 # (no weight_scale_inv); keep them as plain LinearBase so the
--- a/python/sgl_jax/srt/configs/model_config.py
+++ b/python/sgl_jax/srt/configs/model_config.py
@@ -147,6 +147,14 @@ class ModelConfig:
             "MiMoV2ForCausalLM",
         ):
             self.hf_config.architectures[0] = "MiMoV2FlashMTPForCausalLM"
+            if self.quantization_config is not None:
+                # Draft MTP head: eh_proj and o_proj are bf16 in the checkpoint
+                # (no weight_scale_inv); keep them as plain LinearBase so the
+                # weight_mappings can target `.weight` instead of `.weight_q`.
+                self.quantization_config.ignored_layers = [
+                    "model.eh_proj",
+                    "model.self_attn.o_proj",
+                ]
         # Check model type
         self.is_generation = is_generation_model(self.hf_config.architectures, is_embedding)
         self.is_multimodal = False
--- a/python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention.py
+++ b/python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention.py
@@ -170,9 +170,9 @@ def ref_ragged_paged_attention(
         v = jnp.repeat(v, num_query_per_kv, axis=1)
         attn = jnp.einsum("qhd,khd->hqk", q, k, preferred_element_type=jnp.float32)
         attn *= sm_scale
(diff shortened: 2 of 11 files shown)


zorrofox:
Closing — will reopen once perf optimization (host orchestration overhead, KV-pool leak) is done and numbers are net-positive vs baseline.

PR #1063, opened 2026-05-12
fix(eagle): restore dp=1 dense EAGLE3 path on main + add E2E to CI

zorrofox:
Part of #1053 (P1-0 prerequisite).

## Motivation

`speculative/` was last touched by #939 (4/28). Since then #939/#940/#1015 changed sampler / `schedule_batch` / `compilation_manager` that the EAGLE path depends on, with no EAGLE-side updates; `test_speculative_decoding.py` was not in `run_suite.py` so CI did not catch it. This restores the **dp=1 dense EAGLE3** path (Qwen3-32B + EAGLE3) end-to-end as the baseline for #1053 P1-2's class refactor.

### Regression fixes — #939 (`ScheduleBatch` flat → `reqs_info`)
- `ScheduleBatch.__getattr__/__setattr__` dp=1 compat shim passing through `reqs_info[0]` until #1053 P1-5 lands the DP-aware data contract
- `get_spec_model_worker_batch`: populate `per_dp_bs_size` / `real_bs_per_dp` / `dp_size` / `logits_indices_selector` / merged `sampling_info`
- `ModelWorkerSamplingInfo`: add `__len__` / `filter_batch`
- scheduler spec branch: write `seq_lens`/`spec_info` via `reqs_info[0]`; `device_get` `next_token_ids` before `_extract_dp_output_ids`
- `EagleDraftInput.filter_batch`: coerce list indices to `np.asarray`
- `prepare_for_extend_after_verify`: set `logits_indices`

### Regression fixes — #940 (sampler RNG)
- `model_runner`: init `_sampler_base_rng`/`_sampler_step` for draft worker too (`initialize_jit()` needs it; only the call itself stays guarded by `is_draft_worker`)

### Regression fixes — #1015 (CompilationManager)
- `eagle_worker` precompile: `generate_model_worker_batch` → `compilation_manager._make_dummy_batch` (with `dp_size`/`per_dp_bs_size`)
- `compilation_manager` spec branch: use `ModelWorkerSamplingInfo` (not `SamplingBatchInfo`) for dummy batches

### Explicit-sharding reshard at jit-exit points
- `topk_probs_from_logits`: reshard vocab-sharded logits inside jit
- `_replicate` helper; replicate logits/hidden after target verify, `capture_for_decode`, `draft_forward` step, `draft_extend_after_verify` gather
- `build_tree_kernel_efficient`: replicate inputs at entry
- `update_eagle_lists`: enable `@jit` (inputs now replicated)
[...]

Issue #1053, opened 2026-05-11
[Feature] Refactor Eagle/MTP RFC


pengchengneo:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

# RFC: Eagle/MTP Refactoring

## Background

This RFC aims to support `mimo-v2.5-pro` MTP speculative decoding in `sglang-jax` and, at the same time, refactor the current EAGLE/MTP control flow that is concentrated in `EAGLEWorker`. The existing implementation can support standard EAGLE, but the control plane, draft runner, target verify, and scheduler update logic are tightly coupled. This structure is not a good long-term foundation for MTP, DP attention, or scheduler overlap.

The final goal is an EAGLE/MTP implementation with stable behavior, acceptable performance, clear boundaries, and compatibility with DP attention and future overlap scheduling.

## Overall Plan

This refactor is split into three phases. Each phase focuses on one main problem:

1. **Phase 1: Functional baseline and data-flow refactoring.** Split `BaseSpecWorker` / `BaseDraftWorker`, so standard EAGLE and `mimo-v2.5-pro` MTP share the same verify/data contract. The Phase 1 data contract must also become DP-attention compatible and tighten explicit state boundaries for future scheduler overlap.
2. **Phase 2: EAGLE/MTP pipeline performance optimization.** After the functional path is stable, use profiling to optimize real hotspots. The target is parity with the sglang `mimo-v2.5-pro` MTP path without scheduler overlap. Kernel work is introduced only if profiling proves it is necessary.
3. **Phase 3: Spec decode and scheduler overlap.** After functionality is stable and performance is acceptable, handle compatibility between spec decode and scheduler overlap. A demo has shown that async dispatch is feasible, but whether it should replace the current multithreaded overlap mechanism needs a separate RFC.

### Phase 1: EAGLE Functional Baseline and Data-Flow Refactoring (DDL: 5-15)

Phase 1 refactors the EAGLE/MTP functional path and data boundaries. Standard EAGLE uses the refactored `EAGLEWorker` + `EagleDraftWorker`. `mimo-v2.5-pro` MTP uses `MultiLayerEAGLEWorker` + multi-runner `MultiLayerDraftWorker`. Both share the target verify path and data contract logic.

Main work:

[...]

PR #1064, opened 2026-05-12
feat(spec): SpecInput data contract + host/device boundary annotations

zorrofox:
Part of #1053 (P1-5a). **Draft for early interface review** — P1-2 (`BaseSpecWorker`/`BaseDraftWorker` class refactor) builds on this contract, so opening early per [discussion](https://github.com/sgl-project/sglang-jax/issues/1053#issuecomment-4427601694).

> [!NOTE]
> **Depends on #1063** (P1-0). This branch is stacked on it; the actual P1-5a diff is the last commit only (4 files, +201/−26). Will rebase onto main once #1063 merges.

## What

Defines the spec-decode data contract from the #1053 RFC tables so the P1-2 abstraction layer has a fixed interface to target:

- **`spec_info.SpecInput`** — `@runtime_checkable` Protocol exposing:
  - `is_draft_input()` / `is_verify_input()`
  - `get_logical_token_num()` / `get_allocated_token_num()` / `get_verify_token_num()` — the three token counts the RFC requires to be separated (scheduler advance / KV trimming / verify-metadata+DP-accounting respectively)
  - `get_spec_adjust_token_coefficient()` / `filter_batch()` / `merge_batch()`
  - Docstring fixes the host/device boundary rules and DP-padded layout (Route 1: target+draft both DP) semantics
- **`EagleDraftInput`** — implements `SpecInput`; per-field docstrings annotate device vs host, shape, JIT-cache-key participation, and the `allocate_lens` ≠ `accept_length` ≠ `new_seq_lens` distinction. Multi-layer MTP cross-round hidden_states semantics noted.
- **`EagleVerifyInput`** — implements `SpecInput`; per-field docstrings annotate device vs host and static-metadata fields; `filter_batch`/`merge_batch` raise (verify input is single-round).
- **`test_spec_info.py`** — Protocol conformance + three-token-count distinction.

## Non-changes

- No behavior change to the existing dp=1 path (pure additions: docstrings, Protocol, new methods).
- `GenerationBatchResult` already has all RFC-required fields except optional `new_seq_lens` (derivable from `seq_lens + accept_lens`); not adding it here.
- Type annotations on device fields changed `np.ndarray` → `jax.Array | None` to reflect actual runtime types after #1063's reshard changes — annotations only, no code path change.

## Open questions for P1-2 author

1. Should `SpecInput` be a `Protocol` (current, duck-typed) or an `ABC` base class? Protocol avoids inheritance coupling but `ABC` lets `BaseSpecWorker` type-hint more strictly.
2. `get_logical_token_num` currently returns a scalar `int` (sum across batch). Should it return per-request `np.ndarray` instead for DP per-rank accounting?
3. `EagleDraftInput.merge_batch` currently uses `np.concatenate` (host-side). RFC says these are device arrays — should merge happen on device (`jnp.concatenate`)?

## Test plan

- [x] `test_spec_info.py` 3/3 passed (CPU)
- [x] pre-commit clean
- [ ] dp=1 EAGLE E2E unchanged (will verify after rebasing onto merged #1063)

### DIFF
--- a/python/sgl_jax/srt/speculative/eagle_util.py
+++ b/python/sgl_jax/srt/speculative/eagle_util.py
@@ -379,42 +379,75 @@ def build_tree_kernel_efficient(
 @register_pytree_node_class
 @dataclass
 class EagleDraftInput:
-    # Constant: alloc length per decode step
+    """Next-round draft state — the only persistent cross-round spec state.
+
+    Implements ``SpecInput``. MUST NOT hold worker/runner/pool/future handles.
+    Under DP (Route 1), per-request fields use DP-padded order.
+    """
+
     ALLOC_LEN_PER_DECODE: ClassVar[int] = None
 
-    # The inputs for decode
-    # shape: (b, topk)
-    topk_p: np.ndarray = None
-    topk_index: np.ndarray = None
-    # shape: (b, hidden_size)
-    hidden_states: np.ndarray = None
+    # --- Cross-round draft state (device arrays, consumed by next draft) ---
+    #: device ``(b, topk)`` — top-k probs from previous draft/draft_extend.
+    topk_p: jax.Array | None = None
+    #: device ``(b, topk)`` — top-k token ids.
+    topk_index: jax.Array | None = None
+    #: device ``(b, hidden_size)`` — minimal hidden state for next draft step.
+    #: Multi-layer MTP keeps per-step hidden locally inside one
+    #: ``MultiLayerDraftWorker.draft()``; only this cross-round slice persists.
+    hidden_states: jax.Array | None = None
+    #: static metadata (pytree aux); changing it triggers a new compile shape.
     capture_hidden_mode: CaptureHiddenMode = CaptureHiddenMode.FULL
 
-    # Inputs for extend
-    # shape: (b,)
-    verified_id: np.ndarray = None
-    accept_length: np.ndarray = None
+    # --- Draft-extend inputs (device unless ``_cpu`` suffixed) ---
+    #: device ``(b,)`` — verified token starting the next draft.
+    verified_id: jax.Array | None = None
+    #: device ``(b,)`` — accepted length used to select hidden in draft-extend.
+    accept_length: jax.Array | None = None
+    #: host ``(b,)`` int32 mirror of ``accept_length`` for scheduler bookkeeping.
     accept_length_cpu: np.ndarray | None = None
 
-    # Inputs for the attention backends
-    # shape: (b + 1,)
-    kv_indptr: np.ndarray = None
-    kv_indices: np.ndarray = None
+    # --- Attention-backend metadata (host, participates in metadata build) ---
+    kv_indptr: np.ndarray | None = None
+    kv_indices: np.ndarray | None = None
 
-    # Shape info for padding
+    # --- Padding shape (static; participates in JIT cache key) ---
     num_tokens_per_batch: int = -1
     num_tokens_for_logprob_per_batch: int = -1
 
-    # Inputs for draft extend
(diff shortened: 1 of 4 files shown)


python/sgl_jax/srt/speculative/spec_info.py
```diff
@@ -1,13 +1,57 @@
...
 import jax
+import numpy as np
 
 from sgl_jax.srt.layers.logits_processor import LogitsProcessorOutput
 
 logger = logging.getLogger(__name__)
 
 
+@runtime_checkable
+class SpecInput(Protocol):
```
pengchengneo, line 16: I’d keep `SpecInput` as a Protocol for now. In this repo it is mainly a contract `ModelWorkerBatch.spec_info`, not shared implementation, and the current inputs are dataclass+pytree objects. An ABC would add inheritance coupling before P1-2 proves we need shared base behavior.

We can make the Protocol strict with clear return semantics and boundary checks, then add an ABC/mixin later if real shared logic appears.

python/sgl_jax/srt/speculative/eagle_util.py
```diff
@@ -687,22 +720,65 @@ class EagleVerifyOutput:
...
+    def get_allocated_token_num(self) -> np.ndarray | None:
+        return None
+
+    def get_verify_token_num(self, bs: int) -> int:
+        return bs * self.draft_token_num
+
+    def filter_batch(self, new_indices: np.ndarray, has_been_filtered: bool = True) -> None:
+        raise NotImplementedError("EagleVerifyInput is consumed within one round")
+
+    def merge_batch(self, other) -> None:
```
pengchengneo, line 779: Although `merge_batch` is called from the scheduler host path, these fields are next-round draft state and are consumed on device.

You can replace this with jnp.concatenate

and in phase3 we will put these field on FutureMap, just concatenate future indices will be ok

python/sgl_jax/srt/speculative/spec_info.py
```diff
@@ -1,13 +1,57 @@
...
+    """
+
+    def is_draft_input(self) -> bool: ...
+    def is_verify_input(self) -> bool: ...
+
+    def get_spec_adjust_token_coefficient(self) -> int:
+        """Multiplier for scheduler new-token budgeting (e.g. draft_token_num)."""
+        ...
+
+    def get_logical_token_num(self, bs: int) -> int: ...
```
pengchengneo: Returning only a batch-level int loses DP rank structure, this function can return per-request counts, e.g. `np.ndarray`, and let callers sum when they need a scalar.

python/sgl_jax/srt/speculative/spec_info.py
zorrofox: Kept as Protocol. Will revisit if P1-2 surfaces shared base behavior worth an ABC/mixin.

python/sgl_jax/srt/speculative/eagle_util.py
zorrofox: Switched the four device fields (`hidden_states`/`verified_id`/`topk_p`/`topk_index`) to `jnp.concatenate`; left `allocate_lens` on `np.concatenate` (host array per the contract). 69113de5.

python/sgl_jax/srt/speculative/spec_info.py
zorrofox: Changed to per-request `np.ndarray` (host int32 `(bs,)`); callers sum for batch totals. Updated both implementations + the Protocol docstring + the unit test. 69113de5.

pengchengneo (approved):

pengchengneo:
LGTM

PR #1066, opened 2026-05-12
fix(eagle): align draft KV pool size with hybrid target's full slot range

zorrofox:
Part of #1053 (P1-1).

> [!NOTE]
> **Depends on #1063** (P1-0). Stacked for E2E verification only — the fix itself (`model_runner_kv_cache_mixin.py`) and the assertion are independent of #1063's changes. Actual P1-1 diff is the last commit (2 files, +21).

## Problem

Hybrid target's `set_num_token_hybrid()` (step 7 in `init_memory_pool`) expands `max_total_num_tokens` to the full-pool slot range — e.g. **162304 → 463616** for MiMo-V2-Flash on v6e-16 — **after** `server_args.draft_runner_cache_size` was set in step 5.

The draft worker shares the target's allocator (slot range = post-hybrid full pool) but creates its own KV pool at the **pre-hybrid** size. Any slot ≥ pre-hybrid size reads/writes garbage in the draft KV cache.

**Symptom**: `accept[1:]=1` (only bonus token accepted) for requests allocated at high slots. Low-bs clean-server tests pass only because slots stay in `[0, 512)`.

## Fix

- `model_runner_kv_cache_mixin.py`: after `set_num_token_hybrid()`, if target+spec, overwrite `draft_runner_cache_size` with the post-hybrid `max_total_num_tokens`.
- `eagle_worker.py`: startup-time sanity assert `draft_pool_size >= target_slot_range` so the mismatch surfaces at server start instead of as a silent accept-len degradation.

## Verification

- v6e-4, dense Qwen3-32B + EAGLE3 (non-hybrid path): assertion passes (draft=target=160064), server reaches ready, `health_generate` 200 OK, no regression vs #1063.
- Hybrid-path E2E (MiMo-V2-Flash / V2.5-Pro) deferred to #1053 P1-4 — those models need ≥v6e-16; P1-4's "bs=1 accept-len ≥2.5, output 100%=nospec" acceptance is the direct hybrid+draft-pool correctness check.

## Test plan

- [x] pre-commit clean
- [x] dense EAGLE3 server start + assertion pass (v6e-4)
- [x] dense EAGLE3 MMLU-64 score 0.797 (vs #1063 0.766, within MMLU-64 sampling variance)
- [ ] hybrid target E2E (P1-4)

### DIFF
--- a/python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
+++ b/python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
@@ -555,14 +555,32 @@ class ModelRunnerKVCacheMixin:
                 self.max_total_num_tokens = self.server_args.draft_runner_cache_size
                 self.server_args.max_num_reqs = max_num_reqs
 
-        # 6. Apply constraints (CI, user cap, page align, dp)
+        # 6. Apply constraints (CI, user cap, page align, dp). Draft worker's
+        # pool size is target-derived (must cover target's allocator slot
+        # range), so the user --max-total-tokens cap is not re-applied — it
+        # could otherwise shrink the draft pool below the (post-hybrid) target
+        # range and reintroduce the high-slot garbage this commit fixes.
         self.max_total_num_tokens = self._apply_token_constraints(
-            self.max_total_num_tokens, max_total_tokens, dp_size
+            self.max_total_num_tokens,
+            None if self.is_draft_worker else max_total_tokens,
+            dp_size,
         )
 
         # 7. Hybrid SWA token split (existing logic, not moved)
         if self.is_hybrid:
             self.set_num_token_hybrid()
+            if (
+                not self.is_draft_worker
+                and self.spec_algorithm is not None
+                and not self.spec_algorithm.is_none()
+            ):
+                # Draft shares target's allocator, whose slot range is the
+                # *post-hybrid* full-pool size. The draft_runner_cache_size set
+                # in step 5 was pre-hybrid; without this overwrite, draft's own
+                # KV pool is smaller than the slot range it indexes into, so any
+                # slot >= pre-hybrid size reads/writes garbage (manifests as
+                # accept[1:]=1 for reqs allocated at high slots).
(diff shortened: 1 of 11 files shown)


python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
```diff
@@ -563,6 +563,18 @@ def init_memory_pool(
...
+                and self.spec_algorithm is not None
+                and not self.spec_algorithm.is_none()
+            ):
+                # Draft shares target's allocator, whose slot range is the
+                # *post-hybrid* full-pool size. The draft_runner_cache_size set
+                # in step 5 was pre-hybrid; without this overwrite, draft's own
+                # KV pool is smaller than the slot range it indexes into, so any
+                # slot >= pre-hybrid size reads/writes garbage (manifests as
+                # accept[1:]=1 for reqs allocated at high slots).
+                self.server_args.draft_runner_cache_size = self.max_total_num_tokens
```
pengchengneo, line 583: One edge case: `draft_runner_cache_size` is overwritten with the post-hybrid full slot range on the target side, but the draft worker still runs `_apply_token_constraints(..., max_total_tokens, dp_size)` after reading it.

If `--max-total-tokens` is set, this can cap the draft KV pool back below the target's post-hybrid slot range, defeating this fix or tripping the new startup assert. 

Should the draft-worker path skip the user cap when `draft_runner_cache_size` is already an explicit target-derived size, or should the target write a value that is not re-capped later?

zorrofox: Good catch. Took option A: draft worker passes `max_total_tokens=None` to step 6, so the user cap is not re-applied (page-align/dp still are; `set_num_token_hybrid` already page-aligns the value the target writes). Draft pool size is target-derived and must cover the target slot range, so an independent user cap on the draft side has no useful semantics — the startup assert now holds regardless of `--max-total-tokens`. Updated in 498cd93f.

pengchengneo:
LGTM

pengchengneo (approved):

PR #1080, opened 2026-05-14
refactor(spec): split EAGLEWorker into BaseSpecWorker/BaseDraftWorker…

pengchengneo:
… + EagleDraftWorker (P1-2)

Extract abstract base classes (BaseSpecWorker, BaseDraftWorker) and move draft logic into EagleDraftWorker so MultiLayerEAGLEWorker/MultiLayerDraftWorker (P1-4) can reuse the same verify/data-contract path.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/speculative/base_worker.py
+++ b/python/sgl_jax/srt/speculative/base_worker.py
@@ -3,8 +3,24 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
+import jax
+from jax.sharding import NamedSharding
+from jax.sharding import PartitionSpec as P
+
 if TYPE_CHECKING:
     from sgl_jax.srt.managers.tp_worker import ModelWorker
+
+
+def replicate_to_mesh(
+    mesh: jax.sharding.Mesh, *arrs: jax.Array
+) -> tuple[jax.Array, ...] | jax.Array:
+    """Replicate arrays across a mesh under explicit sharding.
+
+    JIT outputs are typically vocab/data-sharded; spec-decode host orchestration
+    (top_k, gather, build_tree) needs replicated arrays.
+    """
+    out = jax.device_put(arrs, NamedSharding(mesh, P()))
+    return out[0] if len(out) == 1 else out
 
 
 class BaseDraftWorker(ABC):
@@ -19,6 +35,16 @@ class BaseDraftWorker(ABC):
     @abstractmethod
     def draft(self):
         pass
+
+    @property
+    @abstractmethod
+    def sampling_rngs(self):
+        """RNGs used by the spec worker's tree-sampling step.
+
+        Concrete implementations decide which runner's rngs to expose.
+        Single-runner workers usually return ``self.draft_model_runner.rngs``;
+        multi-runner workers (MTP) return the rngs from a designated runner.
+        """
 
 
 class BaseSpecWorker(ABC):
--- a/python/sgl_jax/srt/speculative/base_worker.py
+++ b/python/sgl_jax/srt/speculative/base_worker.py
@@ -1,0 +1,93 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import jax
+
+    from sgl_jax.srt.managers.schedule_batch import ModelWorkerBatch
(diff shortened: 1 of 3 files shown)


zorrofox:
Reviewed against the RFC and our in-progress P1-4 (`MultiLayerDraftWorker`/`MultiLayerEAGLEWorker`) which is stacked on this. P1-0 fixes all survived the split (`_replicate` ×4, `safe_index`, `topk_probs_from_logits` reshard, `device_get` verified_id) — nice.

A few things from the P1-4 consumer side:

**1. `verify()` placement** (`eagle_worker.py:128`)
RFC puts `verify()` on `BaseSpecWorker`; here it's on `EAGLEWorker`. The body only touches `BaseSpecWorker` state (`target_worker`, `mesh`, `speculative_num_*`, `draft_worker.draft_model_runner.rngs`) — nothing EAGLE-specific. If `verify()` + `_replicate()` + `forward_target_extend()` move up, `forward_batch_speculative_generation` can become a concrete default on `BaseSpecWorker`, and `MultiLayerEAGLEWorker` can subclass `BaseSpecWorker` directly. Right now P1-4 has to inherit `EAGLEWorker` and call `BaseSpecWorker.__init__` directly to skip `EagleDraftWorker` instantiation, which is fragile.

**2. P1-1 assertion message regressed** (`eagle_draft_worker.py:73-77`)
The original #1066 message named the root cause ("Hybrid target without the post-`set_num_token_hybrid` `draft_runner_cache_size` overwrite"). The new text says "Check `--mem-fraction-static` or `--kv-cache-dtype`" — those flags don't fix the hybrid slot-range mismatch this assert guards. Suggest restoring the original message.

**3. `BaseDraftWorker` ABC missing precompile attrs** (`base_worker.py:14-58` vs `eagle_worker.py:275,286,312,316,333`)
`run_spec_decode_precompile` reaches into `draft_worker.{compilation_manager, model_config, max_req_len, get_max_padded_size}` — none declared on the ABC. Works because `EagleDraftWorker` happens to expose them via `_worker`. P1-4 sets `self._worker = _workers[0]` so it resolves, but precompile then only warms layer 0 (layers 1..N-1 JIT on first real decode). Could either declare these on `BaseDraftWorker`, or push dummy-batch construction down to a `draft_worker.precompile_one_shape(...)` so multi-layer can override it.

**4. `EagleDraftWorker.__init__` not subclass-friendly** (`eagle_draft_worker.py:42-88`)
Unconditionally creates `self._worker = ModelWorker(...)` mid-init. P1-4's `MultiLayerDraftWorker.__init__` ends up copy-pasting ~20 lines of bookkeeping (43-54, 65-67, 81-88). Extracting an `_init_common()` that does everything except the `ModelWorker` creation + `_share_embed_head` + `initialize_jit` would let subclasses reuse it.

Minor:
- `draft_model_runner` (singular) on the ABC is a slight lie for multi-runner. Only external consumer is `verify()` for `.rngs`; could narrow to a `sampling_rngs` property, or have `verify()` use `target_worker.model_runner.rngs` instead.
- `_replicate` is duplicated in both files; goes away if (1) lands.

Happy to fold (1)/(3)/(4) into P1-4 if you'd rather keep this PR's diff small — let me know which.

pengchengneo:
I think it would be better to fold (1)/(3)/(4) into P1-4, cause this will affect your code' arch, and I will fix (2) and minor problems 
@zorrofox

zorrofox:
Sounds good — will fold (1)/(3)/(4) into P1-4. We'll rebase once #1080 lands with your (2) fix.

zorrofox (approved):
Reviewed `77ea2a6f` — assert message restored, `sampling_rngs` removal in `9aed7249` makes sense (`draft_model_runner` override covers the multi-runner case). (1)/(3)/(4) tracked on our side for P1-4. LGTM from the consumer side.

jimoosciuc (approved):

JamesBrianD:
Naming nit: draft_model_runner → draft_runner. sglang's EagleDraftWorker exposes self.draft_runner = self.draft_worker.model_runner as the short alias. Matching that name would shorten call chains here and keep cross-repo greps consistent.

pengchengneo:
> Naming nit: draft_model_runner → draft_runner. sglang's EagleDraftWorker exposes self.draft_runner = self.draft_worker.model_runner as the short alias. Matching that name would shorten call chains here and keep cross-repo greps consistent.

OK, it will be modify in next PR's, we will change abstract relations this PR,   cc @zorrofox

PR #1089, opened 2026-05-14
feat(spec): MultiLayerEAGLEWorker/MultiLayerDraftWorker (#1053 P1-4)

zorrofox:
Part of #1053 Phase 1 (P1-4). Stacked on `epic/mtp-refactor-phase1` (#1080 merged).

## What

3-layer MTP for MiMo-V2.5-Pro via `MultiLayerEAGLEWorker` + `MultiLayerDraftWorker`:
- N independent draft `ModelWorker`s (one per `mtp_layer_idx`), hidden states chained layer→layer in `draft_extend_*`
- `draft_forward` step *i* uses runner *i* (per-layer KV pool, per-layer `forward_metadata`)
- Verify/orchestration reused from `BaseSpecWorker` via `draft_worker` injection

Two commits:
- `5db87f1f` — functional: `MultiLayer*` + NEXTN wire-up + model/config/kernel adaptations
- `b731d0c6` — refactor: lift `verify`/`forward_target_extend`/`forward_batch_speculative_generation` to `BaseSpecWorker` (fold #1080 review item (1)/(4); `EAGLEWorker` reduced to `draft_worker` injection + logprob/precompile)

## Includes

- `SpeculativeAlgorithm.NEXTN` + scheduler dispatch
- `model_config`: draft arch→`MiMoV2MTPForCausalLM`, `num_hidden_layers=1`, quant ignore `eh_proj`/`o_proj` (BF16 in checkpoint)
- `mimo_v2_nextn`: 4-tuple `__call__` + concat reshard; `mimo_v2_flash`: `get_embed_and_head`
- `flashattention_backend`: TARGET_VERIFY `custom_mask` page-align pad + `swa_page_indices` for hybrid
- `rpa_v3`: `_fetch_mask` aligned mask stride + `pl.multiple_of` hints (kernel side of mask pad)
- spec prefill via padded `get_model_worker_batch` (epmoe padding-sensitivity, see below)

## E2E (v6e-64, V2.5-Pro, `--ep-size 64 --moe-backend epmoe`, bs=1 topk=1 greedy)

| | |
|---|---|
| crash-free | ✓ prefill / 3×draft_extend / draft_forward / verify / decode loop |
| accept-len | warm 8-round mean **2.53** (5-prompt mix: natural / code / periodic) |
| output vs nospec | first token matches; first ~13–30 tokens match per prompt |

### Known limitation: spec ≠ nospec on MoE+EP after ~13–30 tokens

Verify-stage divergence is **not** a verify-logic bug — it's upstream `EPMoE` bf16 padding-sensitivity: `get_spec_model_worker_batch` doesn't pad tokens/bs (vs nospec's `get_model_worker_batch`), so the same real tokens land at different row indices inside per-expert GMM groups after `_permute`'s `argsort(topk_ids)` → bf16 accumulation order differs → occasional argmax flip. Padded prefill (this PR) makes the first token match; verify (4-token) vs nospec decode (1-token) are inherently different shapes. Dense targets (P1-0 Llama) unaffected. Tracked in #1090 — likely needs a router-side padding mask.

## TODO (follow-up, not blocking)

- [ ] #1080 review (3): `run_spec_decode_precompile` only warms layer 0 via `_worker` delegation (currently `--disable-precompile` in E2E)
- [ ] `draft_forward` per-layer semantics: tested `runner(i+1)` @ step-0 position — accept-len dropped (3.5→2.0), reverted; current `runner(i)` @ step-i is empirically best but the layer/token-shift contract vs MiMo training needs confirming
- [ ] `EagleDraftWorker.__init__` extract `_init_common()` (review (4) full version) — `MultiLayerDraftWorker.__init__` still copies ~20 lines

## Test plan

- [x] v6e-64 E2E: 3-layer MTP load + serve + 6-prompt generate, no crash
- [x] accept-len ≥2.5 (warm)
- [x] spec output = nospec for first N≥13 tokens
- [ ] CI `test_speculative_decoding.py` (P1-8 will add NEXTN case)

### DIFF
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -299,9 +299,14 @@ class Scheduler(
 
         # launch draft worker
         if self.spec_algorithm is not None and self.spec_algorithm.is_eagle():
-            from sgl_jax.srt.speculative.eagle_worker import EAGLEWorker
+            if self.spec_algorithm.is_nextn():
+                from sgl_jax.srt.speculative.multi_layer_eagle_worker import (
+                    MultiLayerEAGLEWorker as _SpecWorkerCls,
+                )
+            else:
+                from sgl_jax.srt.speculative.eagle_worker import EAGLEWorker as _SpecWorkerCls
 
-            self.draft_worker = EAGLEWorker(
+            self.draft_worker = _SpecWorkerCls(
                 server_args=server_args,
                 target_worker=self.tp_worker,
             )
@@ -1802,7 +1807,15 @@ class Scheduler(
                 next_token_ids = np.array(jax.device_get(next_token_ids_device))
                 self._extract_dp_output_ids(next_token_ids, model_worker_batch, batch)
         else:
-            model_worker_batch = batch.get_spec_model_worker_batch(
+            # Spec prefill must use the same padded mwb as nospec so target
+            # forward sees identical input shapes; get_spec_model_worker_batch
+            # skips token/bs padding which can shift logits via MoE EP dispatch.
+            _get = (
+                batch.get_model_worker_batch
+                if batch.forward_mode.is_extend()
+                else batch.get_spec_model_worker_batch
+            )
+            model_worker_batch = _get(
                 precompile_token_paddings,
                 precompile_bs_paddings,
                 precompile_cache_loc_paddings,
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -303,7 +303,12 @@ class Scheduler(
             # Multi-layer vs single-layer is a model property (how many MTP heads
             # the target ships), not a CLI-algorithm property. NEXTN with a single
             # MTP head behaves exactly like EAGLE (same head run N times).
+            # DeepSeek-style configs expose num_nextn_predict_layers; MiMo-style
+            # configs don't, so fall back to --speculative-num-steps under NEXTN
+            # (one MTP weight set per step).
             n_mtp = getattr(self.tp_worker.model_config.hf_config, "num_nextn_predict_layers", None)
+            if n_mtp is None and self.spec_algorithm.is_nextn():
+                n_mtp = server_args.speculative_num_steps
             self._spec_multi_layer = n_mtp is not None and n_mtp > 1
             if self._spec_multi_layer:
                 from sgl_jax.srt.speculative.multi_layer_eagle_worker import (
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -304,7 +304,9 @@ class Scheduler(
                     MultiLayerEAGLEWorker as _SpecWorkerCls,
                 )
             else:
-                from sgl_jax.srt.speculative.eagle_worker import EAGLEWorker as _SpecWorkerCls
+                from sgl_jax.srt.speculative.eagle_worker import (
+                    EAGLEWorker as _SpecWorkerCls,
+                )
 
             self.draft_worker = _SpecWorkerCls(
                 server_args=server_args,
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -1809,22 +1809,26 @@ class Scheduler(
                 next_token_ids = np.array(jax.device_get(next_token_ids_device))
                 self._extract_dp_output_ids(next_token_ids, model_worker_batch, batch)
         else:
-            # NEXTN (MoE+EP target) prefill must use the same padded mwb as
-            # nospec so target forward sees identical input shapes (#1090).
-            # Dense EAGLE/EAGLE3 targets don't need this and EagleDraftWorker
-            # doesn't yet handle padded prefill mwb, so keep the unpadded path.
-            _get = (
-                batch.get_model_worker_batch
-                if batch.forward_mode.is_extend() and self.spec_algorithm.is_nextn()
-                else batch.get_spec_model_worker_batch
-            )
-            model_worker_batch = _get(
-                precompile_token_paddings,
-                precompile_bs_paddings,
-                precompile_cache_loc_paddings,
-                self.page_size,
-                self.server_args.enable_static_lora,
-            )
+            if batch.forward_mode.is_extend() and self.spec_algorithm.is_nextn():
+                # NEXTN (MoE+EP target) prefill must use the same padded mwb as
+                # nospec so target forward sees identical input shapes (#1090).
+                # Dense EAGLE/EAGLE3 targets don't need this and EagleDraftWorker
+                # doesn't yet handle padded prefill mwb.
+                model_worker_batch = batch.get_model_worker_batch(
+                    precompile_token_paddings,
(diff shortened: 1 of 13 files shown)


python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -299,9 +299,16 @@ def __init__(
 
         # launch draft worker
         if self.spec_algorithm is not None and self.spec_algorithm.is_eagle():
-            from sgl_jax.srt.speculative.eagle_worker import EAGLEWorker
+            if self.spec_algorithm.is_nextn():
```
pengchengneo: nextn's logic == eagle logic

the only difference is nextn's draft model params is from target model layers, here is not correct

we should adjust this judge code to use is there are multi layer draft model params for different steps

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -1802,7 +1809,16 @@ def run_batch(self, batch: ScheduleBatch) -> GenerationBatchResult:
                 next_token_ids = np.array(jax.device_get(next_token_ids_device))
                 self._extract_dp_output_ids(next_token_ids, model_worker_batch, batch)
         else:
-            model_worker_batch = batch.get_spec_model_worker_batch(
+            # NEXTN (MoE+EP target) prefill must use the same padded mwb as
+            # nospec so target forward sees identical input shapes (#1090).
+            # Dense EAGLE/EAGLE3 targets don't need this and EagleDraftWorker
+            # doesn't yet handle padded prefill mwb, so keep the unpadded path.
+            _get = (
```
pengchengneo: this function name alias is not easy for understanding

python/sgl_jax/srt/models/mimo_v2_nextn.py
```diff
@@ -146,6 +146,9 @@ def __call__(
     ) -> tuple[jax.Array, list[jax.Array]]:
         embed = self.embed_tokens(forward_batch.input_ids)
         hidden_in = forward_batch.spec_info.hidden_states
+        emb_sh = jax.typeof(embed).sharding
```
pengchengneo, line 149: can we reuse "replicate_to_mesh" ?

python/sgl_jax/srt/speculative/multi_layer_draft_worker.py
```diff
@@ -0,0 +1,312 @@
...
+        self.speculative_num_draft_tokens = server_args.speculative_num_draft_tokens
+        self.page_size = server_args.page_size
+        from sgl_jax.srt.speculative.spec_info import SpeculativeAlgorithm
+
+        self.speculative_algorithm = SpeculativeAlgorithm.from_string(
+            server_args.speculative_algorithm
+        )
+        self.hot_token_ids = None
+
+        self.num_mtp_layers = self.speculative_num_steps
```
pengchengneo, line 77: Shall we add an assert here ?

python/sgl_jax/srt/managers/scheduler.py
zorrofox: `is_nextn()` is the user's explicit CLI choice (`--speculative-algorithm NEXTN`). Agree the model-side check belongs at construction — added an assert in `MultiLayerDraftWorker.__init__` against `hf_config.num_nextn_predict_layers` (93f881e5). Keeping the dispatch on the algorithm enum so a future single-layer NEXTN can route to `EagleDraftWorker` by checking `num_nextn_predict_layers==1` there if needed.

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Expanded to explicit if/else in 93f881e5.

python/sgl_jax/srt/models/mimo_v2_nextn.py
zorrofox: This needs reshard to the *embed's* sharding (e.g. `P("data", "tensor")`) so the subsequent concat is consistent — `replicate_to_mesh` would force `P()` and add an unnecessary all-gather. Different intent.

python/sgl_jax/srt/speculative/multi_layer_draft_worker.py
zorrofox: Added in 93f881e5 (checks `hf_config.num_nextn_predict_layers == speculative_num_steps`).

python/sgl_jax/srt/managers/scheduler.py
pengchengneo: i think nextn is exacty equals to eagle now. whether use multilayerdraftworker is according to hf_config.num_nextn_predict_layers, not according to nextn or egale

zorrofox: You're right — re-checked and the single-MTP-head case (DeepSeek-style) would have been mis-routed. Switched to `getattr(hf_config, "num_nextn_predict_layers", None) > 1` in 1422a0e3; also moved the #1090 padded-prefill gate to the same flag (only `MultiLayerDraftWorker` handles a padded prefill mwb today). `is_nextn()` now has no callers in `srt/`. No behavior change for V2.5-Pro (n=3) or EAGLE3 (attr absent).

pengchengneo:
@zorrofox  is this ready for merge into epic branch, and can you paste your launch server cmd and test cmd here ?

zorrofox:
Not quite — caught a regression in 1422a0e3 while bringing up dp=2 on v6e-64: MiMo V2.5-Pro's `config.json` doesn't carry `num_nextn_predict_layers` (only DeepSeek-style configs do), so `_spec_multi_layer` resolved to `False` and routed to single-layer `EAGLEWorker`. CI didn't catch it (only tests Qwen3/EAGLE3). Pushed a fallback in f24183fc: when the hf_config field is absent under `--speculative-algorithm NEXTN`, use `--speculative-num-steps` as `n_mtp` (one MTP weight set per step). Verified on v6e-64 that `MultiLayerDraftWorker` is now selected (3 draft `ModelWorker` instances constructed).

Launch (v6e-64, 16 nodes):
```bash
python3 -m sgl_jax.launch_server --model-path <V2.5-Pro> --trust-remote-code \
  --tp-size 64 --dp-size 1 --ep-size 64 --moe-backend epmoe \
  --page-size 256 --context-length 32768 --chunked-prefill-size 4096 --dtype bfloat16 \
  --mem-fraction-static 0.80 --swa-full-tokens-ratio 0.15 --max-running-requests 16 \
  --skip-server-warmup --disable-radix-cache --attention-backend fa \
  --speculative-algorithm NEXTN --speculative-draft-model-path <V2.5-Pro> \
  --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4 \
  --disable-overlap-schedule --disable-precompile \
  --nnodes 16 --node-rank ${RANK} --dist-init-addr <rank0>:5000
```
Test:
```bash
curl -sS http://localhost:30271/generate -H 'Content-Type: application/json' \
  -d '{"text":"The capital of France is","sampling_params":{"max_new_tokens":32,"temperature":0}}'
# greedy first token matches nospec (264), accept-len ~2.5 warm
```
The env-gated `TestNextNV25Pro` in #1097 wraps the same checks. Ready to merge after f24183fc CI.

zorrofox:
f24183fc CI is green (16/16) — ready to merge.

PR #1087, opened 2026-05-14
feat: support Ling-2.6-1T compressed-tensors per-channel FP8

JamesBrianD:
## Summary

Adds support for **Ling-2.6-1T** compressed-tensors W8A16 inference (base 80 layers).

The Ling-2.6-1T checkpoint ships in the compressed-tensors format with **per-channel FP8 weight scales** plus a dynamic per-token activation hint that we deliberately ignore (matching sglang's own `CompressedTensorsW8A16Fp8` scheme). Layer 80 is the NextN/MTP draft layer kept fully BF16 in the checkpoint; this PR covers base inference over layers 0..79 only — MTP support will land in a follow-up.

## Changes

- **`model_config.py`** — extend the existing compressed-tensors auto-detect path to (1) read `weights.strategy` and treat `"channel"` as no block size, (2) load the HF `ignore` list into `QuantizationConfig.ignored_layers`, and (3) log the dynamic activation entry but force `activation_dtype=None` so we land on weight-only FP8.
- **`bailing_hybrid.py`** — thread `num_nextn_predict_layers`, `mtp_loss_scaling_factor`, and `quantization_config` through `PretrainedConfig.__init__` so HF `from_dict` on the 1T config no longer raises.
- **`bailing_moe_linear.py`** — introduce a single `_add_linear` helper covering both the unquantized BF16 path and the static-FP8 `weight_q` + per-channel `weight_scale` sidecar. Wire it through MLA / GQA / linear-attention, dense MLP, MoE expert groups, and shared experts (EPMoE direct, fused via `mlp.{w1,w2,w3}_shared_scale`).
- **`fused_moe.py`** — when `quant_block_k/n` are both `None`, allocate per-channel static placeholders `[E, 1, 1, out_dim]` for routed experts and `[out_dim]` for shared experts so the existing scale loader can populate them.
- **`quantization_utils.py`** — have `apply_moe_quantization` honor `ignored_layers` as a defensive net (no-op for Ling-1T base since layer 80 is not constructed).

The Ling-2.6-flash BF16 path is unchanged: when `is_static_quant=False`, `_add_linear` reduces to the original inline `WeightMapping(target=...weight, transpose=True)` it replaced.

## Test plan

- [ ] Bring up an 8-layer mini-config of Ling-2.6-1T on TPU v6e-16 (`ramezes-ling1t-v6e16` job) and verify weight loader path: no missing-key errors, FP8 sidecars resolved, MoE scale layout matches `[E, 1, 1, out_dim]`.
[...]

PR #1093, opened 2026-05-15
perf(bailing-moe-linear): split GLA QKV at load time to skip post-projection all-to-all

lianfang0320:
…k layout

Replace the per-projection Q/K/V LinearBase trio in the GLA block with a single MergedColumnParallelLinear and rearrange the loaded HF [Q_full|K_full|V_full] columns into [Q_r0|K_r0|V_r0|Q_r1|...] at load time via post_load_weights. The (None,"tensor") shard then places [Q_my|K_my|V_my] contiguously on each device, so the post-projection 3-way split + reshape is a pure local op (no all-to-all collective).

Restripe helpers are hoisted to module scope as jit-cached functions (static_argnames on mesh/tp_size/num_heads_local/head_dim) so all 28 decoder layers reuse one compiled program. tp_size is cached on the attention module to keep __call__ and post_load_weights consistent.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/models/bailing_moe_linear.py
+++ b/python/sgl_jax/srt/models/bailing_moe_linear.py
@@ -68,14 +68,36 @@ class BailingMoELinearAttention(nnx.Module):
 
         inner_size = self.num_heads * self.head_dim
         qkv_bias = getattr(config, "use_bias", False) or getattr(config, "use_qkv_bias", False)
-        self.qkv_proj = LinearBase(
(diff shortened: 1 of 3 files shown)


lianfang0320:
Superseded by #1146 — rebased onto current main, dropped three pairs of mutually-cancelling revert commits from this branch's history, and added the static-quant adaptation that #1087's `_add_linear` helper introduced as a required adaptation point. New branch is `primatrix:feat/gla-split-qkv`.

PR #1094, opened 2026-05-15
perf(lightning): shard per-layer alibi slope across tensor axis

JamesBrianD:
## Summary

Fixes a hidden ``shard_device_array`` reshard fired on every forward step in models that use ``LightningAttnBackend`` (Ling-2.6-flash and friends).

## Root cause

``LightningAttnBackend`` pre-computes one ``(num_heads,)`` float32 ALiBi slope per linear-attention layer at ``__init__`` and stashes them in ``nnx.data({lid: slope})``. The slopes were constructed via ``jnp.asarray`` on the host with no sharding, producing single-device ``jax.Array`` values that landed on device 0.

Because the jitted forward consumes each slope with ``in_sharding`` ``PartitionSpec(\"tensor\")``, JAX reshards the single-device array to all TP devices via ``shard_device_array`` on every forward step — once per linear-attention layer.

On Ling-2.6-flash (28 linear-attention layers, tp=16) this was **28 reshards per decode step**.

## Diagnostic evidence

Hooked ``jax._src.array.shard_device_array`` for the first 3 forward steps:

**Before** (commit ``9470d8a4``, main head):
```
[shard_device_array DIAG] step 0 done, total shard_device_array calls so far: 28
[shard_device_array DIAG] step 1 done, total shard_device_array calls so far: 56
[shard_device_array DIAG] step 2 done, total shard_device_array calls so far: 84

unique pattern: shape=(32,) dtype=float32
                src=(num_devices=1, spec=None)
                -> dst=(num_devices=16, spec=PartitionSpec('tensor',))
```

**After** (this PR):
```
[shard_device_array DIAG] step 0 done, total shard_device_array calls so far: 0
[shard_device_array DIAG] step 1 done, total shard_device_array calls so far: 0
[shard_device_array DIAG] step 2 done, total shard_device_array calls so far: 0
```

## Fix

Construct the slope directly with ``NamedSharding(mesh, P(\"tensor\"))`` via ``make_array_from_callback`` so each device only ever sees its own shard and the ``shard_device_array`` reshard never triggers. ``mesh=None`` keeps backward-compat (test path).

Also verified KDA backend and other linear-attention backends (gdn, qwen3.5 gated_delta_net) — they wrap their per-layer params in ``nnx.Param`` which goes through the weight loader and gets correctly sharded, so they don't have this bug. Only ``LightningAttnBackend`` constructs the slope outside the param loader path.

## Verification

Setup: TPU v6e-16 (4 host), Ling-2.6-flash, epmoe, dp=4, ep=16, conc=256, 1k input / 1k output.

| | before | after |
|--|--|--|
| ``shard_device_array`` calls per forward | 28 | **0** |
| GSM8K accuracy (200 examples) | n/a | **95.5%** (no regression) |

Numerical equivalence: the formula is unchanged — only the destination sharding is fixed up front instead of being inferred and resharded later.

## Test plan
- [x] GSM8K @ 50 examples: 100%
- [x] GSM8K @ 200 examples: 95.5% (matches Ling-2.6-flash baseline)
- [x] DIAG hook confirms ``shard_device_array`` reshards drop from 28/step to 0/step
- [x] ``mesh=None`` backward-compat path returns identical numerical values

### DIFF
--- a/python/sgl_jax/srt/layers/attention/linear/lightning_backend.py
+++ b/python/sgl_jax/srt/layers/attention/linear/lightning_backend.py
@@ -20,7 +20,9 @@ from typing import TYPE_CHECKING
 
 import jax
 import jax.numpy as jnp
+import numpy as np
 from flax import nnx
+from jax.sharding import NamedSharding
 from jax.sharding import PartitionSpec as P
 
 from sgl_jax.srt.layers.attention.hybrid_linear_attn_backend import (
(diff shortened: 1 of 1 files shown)


Rodrian7 (approved):
LGTM

PR #1104, opened 2026-05-18
fix(gmm): ensure default block_m divides m (#1099)

### LINKED ISSUE #1099: EPMoE GMM tile_m=384 fails for dp_size=2 + disable-precompile (nospec, V2.5-Pro)
### Symptom

v6e-64, MiMo-V2.5-Pro, `--moe-backend epmoe --ep-size 64 --dp-size 2 --disable-precompile`, **nospec** (no speculative), single short prompt:

```
[GMM kernel] using default block sizes for key: (1024, 6144, 2048, 384, 6, 'bfloat16', 'float8_e4m3fn', 6144): (384, 2048, 2048)
...
File ".../gmm/megablox_gmm_kernel/gmm.py", line 80, in _calculate_num_tiles
    raise ValueError(f"{x} must be divisible by x-dimension tile size ({tx}).")
ValueError: 1024 must be divisible by x-dimension tile size (384).
```

`dp_size=1` with the same flags works (m=512 picks a different block_m).

### Shape

- per_dp_token_padding = 64 (disable-precompile default first bucket)
- total_tokens = 64 × dp_size(2) = 128
- m = 128 × top_k(8) = 1024 → 128-aligned already
- default block_m chosen for m=1024 is 384 (= num_experts) → 1024 % 384 ≠ 0

`_gmm_compute` already pads `x` to `align=max(sublane, 128)` and adds `pad_size` to `group_sizes[-1]`, but `_calculate_num_tiles` is checked against the *padded* m vs the chosen tile_m, and 384 doesn't divide any 128-multiple between 768 and 1152.

### Not spec-related

Found while bringing up spec+DP (#1053 P1-5b), but reproduces on the plain nospec path (`scheduler.py:1799 → tp_worker.forward_batch_generation`). dp=1 and the cookbook dp=8 (with precompile token buckets) don't hit it because they land on different m / block_m combinations.

### Possible fixes

- `_gmm_compute` pad target should be `lcm(128, block_m)` instead of 128 (or pad to `cdiv(m, block_m) * block_m`)
- or block_m selection should pick a divisor of the post-128-pad m
- or default token bucket under disable-precompile should be `lcm(ep_size, num_experts // top_k)` aware

### Repro

```bash
# v6e-64
python3 -m sgl_jax.launch_server --model-path <V2.5-Pro> \
  --tp-size 64 --dp-size 2 --ep-size 64 --moe-backend epmoe \
  --page-size 256 --disable-precompile --disable-overlap-schedule \
  --attention-backend fa --dtype bfloat16 --nnodes 16 ...
# then any short generate request
```

zorrofox:
## Summary
`get_default_gmm_block_sizes` computes `tm = min(round_up_128(2m/g), m)`; the `min()` only handles `tm > m`, so when `tm < m` and doesn't divide `m` (e.g. `m=1024, g=6 → tm=384`), `_calculate_num_tiles` raises. Hit on V2.5-Pro `--dp-size 2 --disable-precompile` (per_dp 64 × 2 × top_k 8 = 1024).

Fix: after `min()`, step `tm` down to the largest 128-multiple that divides `m`.

Closes #1099. Orthogonal to #1079 (which removed the outer `moe.py` pad — a no-op for `m=1024` which is already 128-aligned).

## Test plan
- [x] `pytest sgl_jax/test/kernels/gmm_test.py -k test_default_block_m_divides_m` (18 cases, m∈{128,512,896,1024,1152,2048} × g∈{6,8,384})
- [x] v6e-64 V2.5-Pro `--tp 64 --dp 2 --ep 64 --moe-backend epmoe --disable-precompile` nospec: GMM log shows `(256, 2048, 2048)` for key `(1024, 6144, 2048, 384, 6, ...)`, `/generate` succeeds

### DIFF
--- a/python/sgl_jax/srt/kernels/gmm/megablox_gmm_kernel/tuned_block_sizes.py
+++ b/python/sgl_jax/srt/kernels/gmm/megablox_gmm_kernel/tuned_block_sizes.py
@@ -559,7 +559,11 @@ def get_default_gmm_block_sizes(m: int, k: int, n: int, g: int) -> tuple[int, in
     # topk=2, m=topk * num_tokens=64, in this case, 2*m//g will be less than
     # 512.
     tm = round_up_to_multiple_of_128_within_limit(2 * m // g, 512)
-    tm = min(tm, m)  # there's a requirement that m % tm == 0
+    tm = min(tm, m)
+    # _calculate_num_tiles requires m % tm == 0; the min() above only handles
+    # tm > m. Step down to the largest 128-multiple that divides m (#1099).
+    while tm > 128 and m % tm != 0:
+        tm -= 128
     # k/n correspond to n_input_features/n_output_features in the matmul so they
     # are normally greater than 2048, unless the num shards is large.
     tk = round_up_to_multiple_of_128_within_limit(k, 2048)
--- a/python/sgl_jax/test/kernels/gmm_test.py
+++ b/python/sgl_jax/test/kernels/gmm_test.py
@@ -152,6 +152,17 @@ def reference_gmm(
 @jtu.with_config(jax_numpy_dtype_promotion="standard")
 class GmmTest(jtu.JaxTestCase):
 
+    @parameterized.product(m=[128, 512, 896, 1024, 1152, 2048], g=[6, 8, 384])
+    def test_default_block_m_divides_m(self, m, g):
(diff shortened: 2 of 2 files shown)


cjx0709 (approved):
LGTM. By the way, if we want to achive the best performance, we can tuner ep moe kernel config before e2e model benchmark, now use the `get_defaut_gmm_block_sizes` is only for functional verification.

pengchengneo:
it merged and rebase onto epic @zorrofox

PR #1037, opened 2026-05-07
Add GLM-5.1 BF16

notabee:
## Motivation

This pull request adds support for the **GLM-5.1** (or GLM-5 MoE) model in `sglang-jax`. This enables running inference for GLM-5.1 models on TPU clusters.

## Modifications

- Added `python/sgl_jax/srt/models/glm5_moe.py` which implements the GLM-5.1 model architecture with absorbed MLA projections and the `GlmDsaIndexer`.
- Updated `python/sgl_jax/srt/configs/model_config.py` to support `glm_moe_dsa` model type and MLA attention.
- Updated `python/sgl_jax/srt/layers/embeddings.py` to support mesh sharding in `RotaryEmbedding`.

## Accuracy Tests

Verified that the model can be served and responds to prompts, although it currently suffers from repetition issues in long generations (see RFC for details).

**Server Launch Command:**
```bash
python -u -m sgl_jax.launch_server \
  --model-path /local/GLM-5.1 \
  --tokenizer-path /local/GLM-5.1 \
  --trust-remote-code \
  --dist-init-addr=10.128.0.249:10011 \
  --nnodes=4 \
  --node-rank=0 \
  --tp-size=32 \
  --ep-size=16 \
  --device=tpu \
  --random-seed=3 \
  --mem-fraction-static=0.65 \
  --max-running-requests 32 \
  --chunked-prefill-size 2048 \
  --port 30000 \
  --disable-precompile \
  --context-length 8192 \
  --attention-backend fa \
  --disable-radix-cache
```

**Test Command 1:**
```bash
curl -s http://127.0.0.1:30000/generate   -H "Content-Type: application/json"   -d '{
    "text": "[gMASK]<sop><|system|>\nYou are a helpful assistant.\n<|user|>\nWhat is the largest planet in the solar system?\n<|assistant|>\n",
    "sampling_params": {
      "max_new_tokens": 128,
      "temperature": 0.0,
      "stop": ["<|user|>", "<|observation|>", "<|system|>", "<|assistant|>"]
    }
  }'
```

**Output 1:**
```json
{
  "text": "The largest planet in the Solar System is Jupiter, the fifth planet from the Sun, and the largest. Jupiter has a diameter of 142,000 km, making it the largest of the planets, with a mass of 1.898 kg, it is 1,898, and the distance is 1,898, and the distance is 1,898, and the distance is 1,898, and the distance is 1,898, and the distance is 1,898, and the distance is 1,898, and the distance is 1,898",
[...]

Issue #1036, opened 2026-05-07
[Feature] Add GLM-5.1


notabee:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

***

#### Summary

This RFC proposes adding inference support for **GLM-5.1** (or GLM-5 MoE) in `sglang-jax`. GLM-5.1 is a large-scale mixture-of-experts (MoE) language model that utilizes Multi-head Latent Attention (MLA). It is a reasoning model designed to generate a chain-of-thought before producing final answers.

*   GLM-5.1 is built on a MoE architecture with 78 layers, where the first 3 layers are dense and the rest are sparse MoE.
*   It uses **Multi-head Latent Attention (MLA)** to compress the KV cache, similar to DeepSeek-V3.
*   It is a reasoning model that expects prompts in a specific format and generates a thinking process within `<think>` tags.
*   The implementation in `sglang-jax` utilizes the Pallas-based MLA kernel and `EPMoE` for expert parallelism.

##### 1.1 LLM Backbone

The LLM is a MoE transformer (model\_type: "glm\_moe\_dsa") characterized by sparsity and memory-efficient attention:

*   **Attention:** Uses **Multi-head Latent Attention (MLA)** to compress the KV cache.
    *   **Latent Ranks:** `kv_lora_rank=512`.
    *   **Head Dimensions:** `qk_nope_head_dim=192`, `qk_rope_head_dim=64`, `v_head_dim=256`.
    *   **Cache Size:** Aligned to 128 bytes, resulting in a 640-dimension fused KV buffer per token.
*   **Mixture-of-Experts (MoE):** 78 layers total.
    *   **Layers 0-2:** Dense (`first_k_dense_replace=3`).
    *   **Layers 3–77:** Sparse MoE with **256 routed experts**.
    *   **Routing:** Selects **top-8** experts per token using `sigmoid` scoring with `noaux_tc` (No Auxiliary Loss, Token-Critical) routing.
    *   **Scaling:** `routed_scaling_factor=2.5`.
*   **Context Window:** 202,752 tokens.
*   **Vocabulary Size:** 154,880.
*   **EOS Tokens:** `[154820, 154827, 154829]`.

##### 2.1 Component Reuse

[...]

PR #1111, opened 2026-05-18
feat(glm5): support FP8 block-wise checkpoint

zorrofox:
## Summary

Enables `zai-org/GLM-5.1-FP8` (block-wise 128×128 e4m3, 750GB) on sglang-jax. PR #1037 shipped the FP8 weight mapping skeleton but it had several bugs that prevented loading; this PR fixes them and gets FP8 serving end-to-end.

Tested on TPU v6e-64 (TP=64, EP=32): sanity + GSM8K correct, **BF16 path verified unchanged** (±2%).

## Changes

**`glm5_moe.py`** (+82/-176)
- `post_load_weights`: dequant FP8 `kv_b_proj` (weight_q × scale) before splitting `w_uk`/`w_uv` for absorbed MLA. Mirrors `deepseek_v3.py`.
- Rewrite linear weight mappings via an `add_linear()` helper (mirrors deepseek_v3). The original `w_name + transpose=True` pattern from #1037 mismatched the `weight_q [out, in]` layout.
- MoE scale mapping `sharding=None → ("expert", None, None)`: replicated scales accumulated ~22 GB/chip and OOM'd at layer 69.
- `patch_model_config`: translate HF `modules_to_not_convert` (`self_attn.indexers_proj`) to the sglang module path `indexer.weights_proj`; set `allow_narrow_n_blockwise=True` (only `indexer.wk` with out_dim=128 hits the narrow-N guard, and its output is discarded).

**`layers/linear.py`** (+30/-7)
- `QuantizedLinear.from_linear`: when a row-parallel block-wise layer's `in_blocks` does not divide TP, fall back to a replicated reduce axis **for that QuantizedLinear only**. GLM-5.1 dense `down_proj` (intermediate=12288 → 96 blocks) and shared `down_proj` (2048 → 16 blocks) don't divide TP=64. The original `LinearBase` keeps row-parallel, so **BF16 is unaffected**.

**`configs/model_config.py`** (+3/-1)
- Read HF `modules_to_not_convert` as `ignored_layers` fallback.

### Correctness (v6e-64, TP=64/EP=32)
| | BF16 (this PR) | BF16 (main) | FP8 W8A16 (this PR) |
|---|---|---|---|
| Sanity / GSM8K | ✅ | ✅ | ✅ |
| KV cache tokens | 25.6K | 25.6K | **127K (5.0×)** |
| BF16 regression | ±2% | — | — |

### Performance (initial → with follow-up PRs)
| config | 1K cc=8 TPOT | notes |
|---|---|---|
| FP8 W8A16, this PR alone | 78.4 ms | **gmm_v2 rejects block-scale** (`shape[1]=48`), falls back to gmm v1 + unfused permute/mask |
| + #1191 (merged) quantized_matmul tuning | 61.0 ms | dense linear tuned, MoE still v1 |
| + **#1259** gmm_v2 block-scale | ~37 ms (proj.) | unblocks gmm_v2 for block-wise MoE; standalone kernel W8A16-blk **268 μs vs W8A8 319 μs** (16% faster) |

### Production-scale reference (W8A8 on-the-fly, same gmm_v2 path that #1259 unlocks for W8A16)
| platform | config | tok/s | TPOT (ms) | tok/s/chip |
|---|---|---|---|---|
| **Trillium (v6e-64)** | W8A8 tp64 dp8 ep64, 1K/1K cc=460 | **2804** | 131 | 43.8 |
| Ironwood (v7x-16) | W8A8 tp32 dp4 ep32, 1K/1K cc=500 | 2122 | 193 | **132.6** |
| H200×8 ([sgl-cookbook §5.1.2](https://github.com/sgl-project/sgl-cookbook/blob/main/docs/autoregressive/GLM/GLM-5.1.md#512-throughput-benchmark)) | FP8 tp8+MTP+DP, 1K/1K cc=100 | 1215 | 38.7 | 151.9 |

(Different cc reflects each platform's KV-cache capacity; all are at/near their respective throughput peaks.)

**Note**: the `GLM-5.1-FP8` checkpoint specifies `"activation_scheme": "dynamic"` (i.e. W8A8), but this PR only wires up **weight loading** — activations stay bf16, so it currently runs as **W8A16** (bf16×fp8 → no FP8 2× compute). With #1259 it should land between BF16 and the W8A8 reference above (decode/memory-bound close to W8A8; prefill/compute-bound close to BF16). Full W8A8 (block-wise weight + dynamic per-token act) is a follow-up.

## Follow-ups
- [x] #1191 (merged): GLM shapes in quantized_matmul tuning table
- [ ] **#1259**: gmm_v2 block-scale support — required for block-wise FP8 MoE to use gmm_v2
- [ ] **W8A8**: implement dynamic act quant for block-wise weight (honor `activation_scheme: dynamic`) → fp8×fp8 matmul, should reach W8A8 reference perf
- [ ] (optional) dequant `down_proj` on load to BF16 row-parallel (+~37 MB/chip, removes replicated fallback)

Refs #1036.

### DIFF
--- a/python/sgl_jax/srt/layers/linear.py
+++ b/python/sgl_jax/srt/layers/linear.py
@@ -305,9 +305,11 @@ class QuantizedLinear(nnx.Module):
                     tp = linear.mesh.shape.get(input_axis, 1) if input_axis else 1
                     if input_axis is not None and in_blocks % tp != 0:
                         logger.warning(
-                            "QuantizedLinear %s: in_blocks=%d not divisible by "
-                            "TP=%d on axis %r; replicating reduce axis",
+                            "QuantizedLinear %s [in=%d, out=%d]: in_blocks=%d not "
+                            "divisible by TP=%d on axis %r; replicating reduce axis",
                             linear.name,
+                            in_features,
+                            out_features,
                             in_blocks,
                             tp,
                             input_axis,
--- a/python/sgl_jax/srt/layers/linear.py
+++ b/python/sgl_jax/srt/layers/linear.py
@@ -1,3 +1,4 @@
+import logging
 from collections.abc import Sequence
 from functools import partial
 
@@ -14,6 +15,8 @@ from sgl_jax.srt.kernels.quantized_matmul.kernel import xla_quantized_matmul_loc
 from sgl_jax.srt.utils.parallel_utils import prepare_scattered_spec_if_needed
 from sgl_jax.srt.utils.profiling_utils import named_scope
 from sgl_jax.srt.utils.quantization.quantization_utils import quantize_tensor
+
+logger = logging.getLogger(__name__)
 
 
 class LinearBase(nnx.Module):
@@ -276,19 +279,13 @@ class QuantizedLinear(nnx.Module):
             tuple(weight_block_size) if weight_block_size is not None else None
         )
 
+        kernel_axes = linear.kernel_axes or (None, None)
         if is_static_input:
             # Static checkpoint already stores pre-quantized weights and scales.
             weight = linear.weight.value
 
             if isinstance(weight, jax.ShapeDtypeStruct):
                 in_features, out_features = map(int, weight.shape)
-                kernel_axes = linear.kernel_axes or (None, None)
-                wq_sharding = NamedSharding(linear.mesh, P(kernel_axes[1], kernel_axes[0]))
-                weight_q = jax.ShapeDtypeStruct(
-                    shape=(out_features, in_features),
-                    dtype=weight_dtype,
-                    sharding=wq_sharding,
-                )
 
                 if (
                     effective_weight_block_size is not None
@@ -298,6 +295,24 @@ class QuantizedLinear(nnx.Module):
                         effective_weight_block_size[1]
                     )
                     in_blocks = (in_features + block_k - 1) // block_k
+                    # Row-parallel block-wise layers shard the reduce axis, so
+                    # the in_blocks dimension of the [in_blocks, 1, n_out]
+                    # scale must also tile across TP. When it doesn't (e.g.
+                    # GLM-5.1 dense/shared down_proj: 96/16 blocks vs TP=64),
+                    # fall back to a replicated reduce axis for this layer
+                    # only. The original LinearBase keeps row-parallel.
+                    input_axis = kernel_axes[0]
+                    tp = linear.mesh.shape.get(input_axis, 1) if input_axis else 1
+                    if input_axis is not None and in_blocks % tp != 0:
+                        logger.warning(
+                            "QuantizedLinear %s: in_blocks=%d not divisible by "
+                            "TP=%d on axis %r; replicating reduce axis",
+                            linear.name,
+                            in_blocks,
+                            tp,
+                            input_axis,
+                        )
+                        kernel_axes = (None, kernel_axes[1])
                     # Pre-expanded kernel-ready layout: [in_blocks, 1, n_out].
                     scale_sharding = NamedSharding(
                         linear.mesh, P(kernel_axes[0], None, kernel_axes[1])
@@ -312,6 +327,12 @@ class QuantizedLinear(nnx.Module):
                     weight_scale = jax.ShapeDtypeStruct(
                         shape=(out_features,), dtype=jnp.float32, sharding=scale_sharding
                     )
+                wq_sharding = NamedSharding(linear.mesh, P(kernel_axes[1], kernel_axes[0]))
+                weight_q = jax.ShapeDtypeStruct(
+                    shape=(out_features, in_features),
+                    dtype=weight_dtype,
+                    sharding=wq_sharding,
+                )
                 bias = linear.bias.value if linear.bias is not None else None
             else:
                 if weight.dtype != weight_dtype:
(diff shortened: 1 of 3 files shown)


zorrofox:
@JamesBrianD do you have any comments on this PR?

python/sgl_jax/srt/models/glm5_moe.py
```diff
@@ -1067,6 +973,15 @@ def patch_model_config(cls, mc: ModelConfig) -> None:
         mc.v_head_dim = getattr(mc.hf_text_config, "v_head_dim", 256)
         # GLM-5 uses MLA architecture
         mc.attention_arch = AttentionArch.MLA
+        # GLM-5.1-FP8 ships modules_to_not_convert with HF naming (e.g.
+        # `self_attn.indexers_proj`); translate to sglang-jax module paths so
+        # quantize_model leaves the unquantized indexer head-gate as LinearBase.
+        if mc.quantization_config is not None and mc.quantization_config.is_static_checkpoint:
+            mc.quantization_config.ignored_layers = ["indexer.weights_proj"]
```
JamesBrianD: Consider merging into ignored_layers instead of overwriting it

zorrofox: Done in 07718874. Note the original HF `modules_to_not_convert` entries (631 full paths like `model.layers.N.self_attn.indexers_proj`, layernorms, `mlp.gate`) don't match any sglang `LinearBase` path even after the dot→bracket normalize, so they're no-ops — but appending is safer for user-supplied configs.

python/sgl_jax/srt/layers/linear.py
```diff
@@ -298,6 +295,24 @@ def from_linear(
...
+                    # the in_blocks dimension of the [in_blocks, 1, n_out]
+                    # scale must also tile across TP. When it doesn't (e.g.
+                    # GLM-5.1 dense/shared down_proj: 96/16 blocks vs TP=64),
+                    # fall back to a replicated reduce axis for this layer
+                    # only. The original LinearBase keeps row-parallel.
+                    input_axis = kernel_axes[0]
+                    tp = linear.mesh.shape.get(input_axis, 1) if input_axis else 1
+                    if input_axis is not None and in_blocks % tp != 0:
+                        logger.warning(
+                            "QuantizedLinear %s: in_blocks=%d not divisible by "
```
cjx0709: the slient sharding will be easy to ignore, maybe also add the out_features/in_features shape info in warning print.

zorrofox: Added in f8567a51 — now prints `QuantizedLinear down_proj [in=12288, out=6144]: in_blocks=96 not divisible by TP=64 ...`.

zorrofox:
`accuracy-test-4-tpu (0)` was cancelled by the runner, not a test failure:

```
##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
```

It died 2 min in while `test_moe_eval_accuracy_large.py` was still in `popen_launch_server` / `time.sleep(10)` — never reached an assert. The same test passed (32 min) on the previous run before rebase. Could a maintainer re-run the failed jobs? I don't have permission from a fork.

(The `review` job failure is the known fork-PR OIDC issue from #1174.)

zorrofox:
Rebased onto main — CI now all green (the previous `accuracy-test-4-tpu` red was a runner shutdown). Both review comments addressed. Ready for another look. cc @JamesBrianD

zorrofox:
## Performance follow-up: #1111 + #1259 = production-ready FP8

This PR alone (block-wise FP8 ckpt loading) currently **falls back to gmm v1** for MoE because `gmm_v2.is_supported_by_gmm_v2()` rejects `rhs_scale.shape[1]>1` (GLM-5.1 has 48 K-blocks). Result: W8A16 cc=8 TPOT 78ms vs BF16 38ms (slower, see profile in #1191 description — gmm v1 + unfused permute/mask dominates).

**#1259** adds block-scale support to gmm_v2. Standalone v6e kernel bench (E=8 M=512 K=7168 N=1536):
| | μs | vs BF16 |
|---|---|---|
| BF16 v2 | 322 | — |
| W8A16 per-channel v2 | 249 | -23% |
| **W8A16 block-scale v2 (#1259)** | **268** | **-17%** |
| W8A8 per-channel v2 | 319 | -1% |

With #1111+#1191+#1259, GLM-5.1 static W8A16 ckpt should match or exceed the W8A8 on-the-fly path, which on v6e-64 dp8 reaches **2804 tok/s @ cc=460** (1K/1K) — 230% of H200×8 cookbook throughput, 116% cost-efficiency at GCP on-demand.

Will post W8A16+#1259 e2e numbers when I get v6e-64 time. Suggest reviewing #1111 and #1259 together.

zorrofox:
@JamesBrianD @cjx0709 — updated the PR body with current performance data. Since the original submission we landed #1191 (kernel tuning) and opened #1259 (gmm_v2 block-scale, the actual perf unblocker for this PR). With the W8A8 runtime config (`dp8 ep64` + precompile buckets) the same gmm_v2 path now hits 2804 tok/s on Trillium-64 / 2122 on Ironwood-16, so #1111+#1259 together make the static FP8 ckpt production-viable rather than just functional. Would appreciate a look at both PRs together.

cjx0709 (approved):
I think we can merge it in the first step. @zorrofox

zorrofox:
Closing out follow-ups:
- **W8A8 block-wise + dynamic act**: not pursuing. e2e on v6e-64 dp8 (see https://github.com/sgl-project/sglang-jax/pull/1259#issuecomment-4591196090) shows W8A16 block-scale is already **+2% per-step** vs W8A8 per-channel — adding act quant would only add overhead (gmm K=7168: W8A16-blk 268μs vs W8A8-ch 319μs). The "should reach W8A8 reference perf" assumption doesn't hold; W8A16 already does.
- **down_proj dequant-on-load**: not pursuing. xprof shows the replicated all-gather is only ~1% of step time.

PR #1154, opened 2026-05-19
fix(weight-loader): split 1-D per-channel weight_scale in fused QKV path

lianfang0320:
## Summary

`_split_qkv_weight` had three dispatch branches: ``.bias`` (1-D), ``"scale"`` with ``ndim == 2`` (block-quant), and an else for 2-D weights. A **1-D per-channel** ``weight_scale`` — as produced by compressed-tensors W8A16 (see ``python/sgl_jax/srt/layers/linear.py:312-314``, ``shape=(out_features,)``) — matched none of them and fell into the else branch, where ``weight[:q_dim, :]`` raises ``IndexError`` on the 1-D input.

Add a dedicated ``"scale" + ndim == 1`` branch that mirrors the bias branch's split-and-pad logic.

## Motivation

Two existing callers register this exact mapping shape on the fused HF ``attention.query_key_value.weight_scale`` key:

- ``python/sgl_jax/srt/models/bailing_moe.py:602`` (BailingMoE V1 GQA static-quant, landed with #1087)
- ``python/sgl_jax/srt/models/bailing_moe_linear.py:885`` (BailingMoE V2.5 GLA static-quant, added in #1146 — merge order independent of this fix)

Both load Ling-style compressed-tensors FP8 checkpoints and would ``IndexError`` today the moment they hit a per-channel-scale checkpoint on a fused QKV layer.

## Modifications

- ``python/sgl_jax/srt/utils/weight_utils.py`` — insert ``elif "scale" in hf_key and weight.ndim == 1:`` between the bias and 2-D-scale branches in ``_split_qkv_weight``. Splits the 1-D tensor at the same Q/K/V offsets the bias branch uses; applies the same reshape→pad→flatten when ``mapping.head_dim_padding=True``. **No other branches modified.**
- ``python/sgl_jax/test/test_weight_loader_qkv_split.py`` (new) — 4 unit tests:
  1. ``test_1d_per_channel_scale_splits_at_correct_offsets`` — GLA/MHA, no padding (the canonical case)
  2. ``test_1d_per_channel_scale_gqa_offsets`` — GQA with ``num_kv_heads < num_heads`` (asymmetric)
  3. ``test_1d_per_channel_scale_head_dim_padding`` — Qwen-style ``head_dim=96`` padded to 128
  4. ``test_pre_fix_else_branch_still_handles_2d_weight`` — regression guard for the 2-D weight else branch

## Test plan

- [x] All 4 new unit tests pass locally
- [x] Existing ``test_bailing_moe_linear.py`` tests pass (no regression)
- [ ] CI

## Blast radius

The new branch is gated on ``"scale" in hf_key and weight.ndim == 1`` — strictly disjoint from the existing three branches. Every pre-existing caller takes the same path it did before:

- ``.bias`` 1-D path: ends at ``hf_key.endswith(".bias")`` gate, never reaches the new branch
- 2-D block-quant scale: gate ``"scale" in hf_key and weight.ndim == 2`` is preserved verbatim
- 2-D weights: else branch is preserved verbatim

### DIFF
--- a/python/sgl_jax/srt/utils/weight_utils.py
+++ b/python/sgl_jax/srt/utils/weight_utils.py
@@ -2557,6 +2557,35 @@ class WeightLoader:
                 v_bias = jnp.reshape(v_bias, (self.num_kv_heads * v_head_dim_padded,))
 
             splits = [q_bias, k_bias, v_bias]
+        elif "scale" in hf_key and weight.ndim == 1:
+            # Per-channel weight_scale (e.g. compressed-tensors W8A16): 1-D
+            # shape [Q_out + K_out + V_out,]. Splits along the single axis with
+            # the same Q/K/V offsets the bias branch uses; applies per-head pad
+            # the same way if mapping.head_dim_padding is set. Falls through to
+            # the 2-D scale branch below for block-quant scales (ndim == 2).
+            q_dim = self.num_heads * self.head_dim_original
+            k_dim = self.num_kv_heads * self.head_dim_original
+            v_dim = self.num_kv_heads * v_head_dim
+
+            q_scale = weight[:q_dim]
+            k_scale = weight[q_dim : q_dim + k_dim]
+            v_scale = weight[q_dim + k_dim : q_dim + k_dim + v_dim]
+
+            if mapping.head_dim_padding and self.head_dim_pad > 0:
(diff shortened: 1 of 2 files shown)


JamesBrianD (approved):
LGTM

PR #1160, opened 2026-05-19
ci: reusable venv setup and flaky test retry (P1-5, P1-14)

### LINKED ISSUE #1133: [CI P1-5] Reusable setup workflow
## Objective

Extract the repeated venv + install setup block into a reusable composite action to reduce YAML duplication.

## Problem

Every test job repeats the same ~4 lines of venv creation + package installation. Changes to the setup process require updating all jobs individually.

## Fix

Create `.github/workflows/_setup.yml` (reusable composite action). Each test job uses `uses: ./.github/workflows/_setup.yml` instead of inline setup steps.

## Acceptance Criteria

1. `_setup.yml` composite action exists and is used by all test jobs
2. Setup behavior is identical to current inline steps (same venv, same packages)
3. Adding a new dependency only requires changing `_setup.yml`, not every job

Rodrian7:
## Summary
- Create `.github/actions/setup-venv/action.yml` composite action that encapsulates the repeated 4-line venv+install pattern
- Replace 21 inline setup blocks across `pr-test.yml`, `nightly-test.yml`, and `nightly-test-daily.yml` with single-line `uses: ./.github/actions/setup-venv`
- Support optional `extra-packages` input for jobs needing additional deps (e.g. `evalscope==1.0.0`)
- Add `.venv/bin` to `GITHUB_PATH` so subsequent steps use the venv automatically without `source activate`
- Net reduction: -100 lines, +73 lines (including the new action file)

## Test plan
- [ ] Verify PR test workflows create venv and install dependencies correctly
- [ ] Verify nightly accuracy jobs install evalscope via `extra-packages` input
- [ ] Verify subsequent run steps can access Python and installed packages from the venv

Closes #1133

### DIFF
--- a/.github/actions/setup-venv/action.yml
+++ b/.github/actions/setup-venv/action.yml
@@ -18,8 +18,11 @@ runs:
         pip install uv
         uv pip install -e "python[all]"
         if [ -n "$EXTRA_PACKAGES" ]; then
+          set -f
+          # shellcheck disable=SC2086
           uv pip install $EXTRA_PACKAGES
+          set +f
         fi
     - name: Add venv to PATH
       shell: bash
-      run: echo "${{ github.workspace }}/.venv/bin" >> $GITHUB_PATH
+      run: echo "$GITHUB_WORKSPACE/.venv/bin" >> $GITHUB_PATH
--- a/.github/actions/setup-venv/action.yml
+++ b/.github/actions/setup-venv/action.yml
@@ -18,9 +18,7 @@ runs:
         pip install uv
         uv pip install -e "python[all]"
         if [ -n "$EXTRA_PACKAGES" ]; then
-          for pkg in $EXTRA_PACKAGES; do
-            uv pip install "$pkg"
-          done
+          uv pip install $EXTRA_PACKAGES
         fi
     - name: Add venv to PATH
       shell: bash
--- a/.github/actions/setup-venv/action.yml
+++ b/.github/actions/setup-venv/action.yml
@@ -1,0 +1,25 @@
+name: 'Setup Python venv'
+description: 'Create Python 3.12 venv, install uv and project dependencies'
+inputs:
+  extra-packages:
+    description: 'Space-separated list of extra pip packages to install (e.g. "evalscope==1.0.0")'
+    required: false
+    default: ''
+runs:
+  using: "composite"
+  steps:
+    - name: Create venv and install dependencies
+      shell: bash
+      run: |
+        python3.12 -m venv .venv
+        source .venv/bin/activate
+        pip install uv
+        uv pip install -e "python[all]"
+        if [ -n "${{ inputs.extra-packages }}" ]; then
+          for pkg in ${{ inputs.extra-packages }}; do
+            uv pip install "$pkg"
+          done
+        fi
+    - name: Add venv to PATH
+      shell: bash
+      run: echo "${{ github.workspace }}/.venv/bin" >> $GITHUB_PATH
--- a/.github/actions/setup-venv/action.yml
+++ b/.github/actions/setup-venv/action.yml
@@ -10,13 +10,15 @@ runs:
   steps:
     - name: Create venv and install dependencies
       shell: bash
+      env:
+        EXTRA_PACKAGES: ${{ inputs.extra-packages }}
       run: |
         python3.12 -m venv .venv
         source .venv/bin/activate
         pip install uv
         uv pip install -e "python[all]"
-        if [ -n "${{ inputs.extra-packages }}" ]; then
-          for pkg in ${{ inputs.extra-packages }}; do
+        if [ -n "$EXTRA_PACKAGES" ]; then
+          for pkg in $EXTRA_PACKAGES; do
             uv pip install "$pkg"
           done
         fi
--- a/.github/workflows/nightly-test.yml
+++ b/.github/workflows/nightly-test.yml
@@ -27,18 +27,16 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+      - name: Setup Python environment
+        uses: ./.github/actions/setup-venv
+        with:
+          extra-packages: 'evalscope==1.0.0'
       - name: Run eval test for text models
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
           HF_HUB_DOWNLOAD_TIMEOUT: ${{ vars.HF_HUB_DOWNLOAD_TIMEOUT }}
         timeout-minutes: 1000
         run: |
-          python3.12 -m venv .venv
-          source .venv/bin/activate
-          pip install uv
-          uv pip install -e "python[all]"
-          uv pip install evalscope==1.0.0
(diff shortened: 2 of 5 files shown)


test/srt/run_suite.py
```diff
@@ -651,6 +672,18 @@ def auto_partition(files, rank, size):
         type=int,
         help="Use auto load balancing. The number of parts.",
     )
+    arg_parser.add_argument(
+        "--reruns",
```
zhengkezhou1, line 688: Why not just name it `--retries`

test/srt/run_suite.py
```diff
@@ -651,6 +672,18 @@ def auto_partition(files, rank, size):
...
+        type=int,
+        default=0,
+        help="Number of times to retry a failed test file (0 = no retry). For pytest runner, also enables per-test retry via pytest-rerunfailures.",
+    )
+    arg_parser.add_argument(
```
zhengkezhou1, line 693: it looks like a timeout param, maybe call it `retry-timeout`?

test/srt/run_suite.py
```diff
@@ -166,6 +172,21 @@ def run_one_file(
 
         try:
             ret_code = run_with_timeout(run_one_file, args=(filename,), timeout=timeout_per_file)
+            if ret_code != 0 and reruns > 0 and file_entry.runner != "pytest":
```
zhengkezhou1: Aren't we using unittest for our tests? See:https://github.com/sgl-project/sglang-jax/tree/main/test#:~:text=SGLang%2DJax%20uses%20the%20built%2Din%20library%20unittest%20as%20the%20testing%20framework.
Is pytest being introduced in this PR?

test/srt/run_suite.py
Rodrian7: --reruns: Kept the same name as pytest-rerunfailures' CLI flag to stay consistent. No extra name mapping
  needed.

test/srt/run_suite.py
Rodrian7: --reruns-delay: It's the wait time between retries, not a timeout. Also matches pytest-rerunfailures' naming

test/srt/run_suite.py
Rodrian7: pytest: Not introducing pytest as a testing framework — tests are still written in unittest. Some files in run_suite.py already use runner: "pytest" to execute unittest files via pytest, that's pre-existing. pytest-rerunfailures only kicks in for those files. For unittest runner files, we do file-level retry instead.

zhengkezhou1 (dismissed):

.github/workflows/nightly-test.yml
```diff
@@ -155,7 +147,7 @@ jobs:
 
           bash scripts/killall_sglang.sh
           cd test/srt
-          python3 run_suite.py --suite nightly-test-perf-text-models-tpu-v6e-1 --auto-partition-id ${{ matrix.part }} --auto-partition-size 8
+          python3 run_suite.py --suite nightly-test-perf-text-models-tpu-v6e-1 --auto-partition-id ${{ matrix.part }} --auto-partition-size 8 --reruns 2
```
cjx0709: Worth discussing: `--reruns 2` on perf suites is risky — a perf test failing once and passing on retry is itself a degradation signal, and retrying to green silences regressions. Same concern for accuracy (sampling/numerics flakes are often real). Retries fit infra-class failures (TPU init flakes, HF timeouts). Suggest skipping retry for perf, or warn loudly when retry passes.

test/srt/run_suite.py
cjx0709: The `runner != "pytest"` condition makes file-level retry mutually exclusive with `pytest-rerunfailures`. But the latter only retries test cases — collection errors / fixture setup failures aren't covered, so the whole file fails without retry. The two should stack; suggest dropping the `runner != "pytest"` check.

test/srt/run_suite.py
```diff
@@ -127,6 +129,8 @@ def run_one_file(
                         "--with",
                         "pytest",
                     ]
+                    if reruns > 0:
+                        cmd.extend(["--with", "pytest-rerunfailures"])
```
cjx0709: `pytest-rerunfailures` isn't pinned — uv pulls latest. For CI, suggest pinning (e.g. `pytest-rerunfailures==14.0`) to avoid upstream breakage.

.github/actions/setup-venv/action.yml
```diff
@@ -0,0 +1,27 @@
...
+        EXTRA_PACKAGES: ${{ inputs.extra-packages }}
+      run: |
+        python3.12 -m venv .venv
+        source .venv/bin/activate
+        pip install uv
+        uv pip install -e "python[all]"
+        if [ -n "$EXTRA_PACKAGES" ]; then
+          for pkg in $EXTRA_PACKAGES; do
+            uv pip install "$pkg"
+          done
```
cjx0709: Whitespace split — fine for `evalscope==1.0.0`, but breaks on inputs like `'pkgA pkgB[extra1,extra2]'`. Could simplify to:

```bash
uv pip install $EXTRA_PACKAGES   # shell split + single uv install
```

Nice-to-have, not blocking.

test/srt/run_suite.py
```diff
@@ -62,7 +62,9 @@ def cleanup_model_cache():
                         print(f"Failed to clean model cache: {e}\n", flush=True)
 
 
-def run_unittest_files(files: list[TestFile], timeout_per_file: float):
+def run_unittest_files(
+    files: list[TestFile], timeout_per_file: float, reruns: int = 0, reruns_delay: int = 0
```
cjx0709: Function default `reruns_delay=0` but CLI `--reruns-delay default=10` (L684). Importing and calling directly behaves differently from CLI; suggest aligning the function default to 10.

.github/workflows/nightly-test.yml
Rodrian7: removed --reruns 2 from all perf and perf-trace suites. Only accuracy suites keep it now.

test/srt/run_suite.py
Rodrian7: Done — pinned to ==14.0.

test/srt/run_suite.py
Rodrian7: Agreed that collection/fixture errors need coverage. Removed the runner != "pytest" guard, but added exit-code coordination to avoid multiplicative retries: pytest exit 1 (test failures already handled by pytest-rerunfailures) skips file-level retry; exit 2/3/5 (infra errors) triggers it.

.github/actions/setup-venv/action.yml
Rodrian7: Simplified to single uv pip install $EXTRA_PACKAGES, also added set -f to disable glob expansion.

test/srt/run_suite.py
Rodrian7: Aligned to 10. Also changed type from int to float to match pytest-rerunfailures upstream.

JamesBrianD (approved):

PR #1164, opened 2026-05-20
feat(weight): support MiMo-V2.5 fused QKV dequant for hybrid attention

cjx0709:
## Summary
- **dequant_fused_qkv**: support per-layer num_kv_heads in hybrid SWA/full-attn models (e.g. MiMo-V2.5: SWA=8, full=4) by selecting the correct head count from `hybrid_layer_pattern` when splitting Q/K/V from the fused tensor
- **_infer_qkv_shards**: search TP candidates in descending order to prefer the largest match. Fixes a bug where SWA layers with per_shard_total exactly divisible by block_size would incorrectly match TP=1, producing a wrong Q/K/V split and garbled output
- **get_rope**: handle `rope_type="default"` (HF transformers convention for "no scaling"), used by MiMo-V2.5 config

## Background
MiMo-V2.5 uses fused qkv_proj with per-TP-shard block-wise FP8 quantization (TP=4). The checkpoint layout is `[shard0_QKV | shard1_QKV | ...]`, where each shard's Q/K/V are concatenated before quantization (scales can span K/V boundaries).

Previously, `_infer_qkv_shards` returned n_shards=1 for SWA layers because per_shard_total=3712 is a multiple of block_size=128, making TP=1 and TP=4 both pass the dimension check. With TP=1, the split treated the first 12288 rows as Q, but they actually contained interleaved Q/K/V data from all 4 shards — corrupting all three projections and producing nonsensical output.

## Test plan
- [x] Deployed MiMo-V2.5 on TPU v6e-16 (4 hosts, tp=16, dp=4, ep=16) with correct text generation
- [x] Verified no impact on MiMo-V2-Flash (uses split QKV, does not enter this path) or MiMo-V2-Pro (TP=8, only one TP value matches due to non-aligned per_shard_total)

### DIFF
--- a/python/sgl_jax/srt/layers/embeddings.py
+++ b/python/sgl_jax/srt/layers/embeddings.py
@@ -620,7 +620,13 @@ def get_rope(
         else:
             raise ValueError("Unknown RoPE scaling type")
 
-        if scaling_type == "llama3":
+        if scaling_type == "default":
+            # HF transformers uses rope_type="default" to mean "no scaling",
+            # equivalent to rope_scaling=None.  Fall back to plain RotaryEmbedding.
+            rotary_emb = RotaryEmbedding(
+                head_size, rotary_dim, max_position, base, is_neox_style, dtype
+            )
+        elif scaling_type == "llama3":
             scaling_factor = rope_scaling["factor"]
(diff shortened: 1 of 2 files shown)


JamesBrianD (approved):
LGTM

PR #1108, opened 2026-05-18
feat(spec): DP-aware spec decode (#1053 P1-5b)

zorrofox:
Stacked on #1089 (merged). Implements attention-backend per-DP metadata + scheduler plumbing for spec decode under `dp_size > 1`.

## Design (resolved — option A)
`spec_info` (`EagleDraftInput`) stays **global-flat** `(real_bs,)` on `reqs_info[0]` as the cross-round state. `_get_spec_decode_mwb_dp` scatters it into DP-padded `(total_bs,)` slots via `logits_indices_selector` when building the mwb; producers (`verify`/`draft_extend_*`) gather back via the same selector. `filter_batch` filters spec_info once globally with cross-rank flat indices. All `[:real_bs]` contiguous-slice callers replaced with `[selector]`.

All `P("data")`-sharded spec metadata (`cache_loc`, `page_indices`, `swa_page_indices`, `custom_mask`, `logits_indices`) is laid out **DP-segmented** (rank r's data at section `[r*per_dp:(r+1)*per_dp)`) so the shard gives each rank its own data. `prepare_for_decode` allocates per-rank (`dp_rank=r`) so SWA mapping / allocator state stay consistent.

## Verified (v6e-64, V2.5-Pro NEXTN, `9610a91f`)
- ✅ **dp=4 n=1/2/4/8**: all reqs coherent; rank>0 32-token output **byte-identical to nospec dp=4**; rank 0 diverges from nospec at token 10 (= #1090 verify-batch-size epmoe variance, not a P1-5b issue)
- ✅ **dp=4 accept-len 2.18-3.00** (vs dp=1 2.31-3.50; was 1.0-1.6 before the 8 fixes in 3ce2fd31/573a0576/9610a91f)
- ✅ **dp=2 n=1/2/4**: complete incl. partial-finish; ids[0:21] match nospec dp=2
- ✅ **nospec dp=4 baseline** captured (`acceptance-nospec-dp4.log`): all ranks coherent, token[0:10] identical, 10+ = #1090 — confirms remaining spec/nospec delta is #1090 not P1-5b
- ✅ CPU `test_spec_dp_shapes.py` (28 cases, 2.3s): mwb shapes / spec_info DP-slot alignment / filter_batch global / `_resolve_spec_decode_token_ids` / DP-segmented page_indices invariant for dp∈{1,2,4} × balanced/unbalanced

## 8 DP-layout fixes since `4613b9c9` (3 commits)
| # | location | bug | introduced |
|---|---|---|---|
[...]

Issue #1090, opened 2026-05-14
EPMoE output is padding-sensitive under bf16 (breaks spec-greedy == nospec-greedy on MoE targets)


zorrofox:
### Symptom

On MiMo-V2.5-Pro (384-expert MoE, `--moe-backend epmoe --ep-size 64`, v6e-64), greedy spec-decode output diverges from nospec greedy:

| prompt | nospec (padded `ids.shape=(64,)`) | spec (unpadded `ids.shape=(5,)`) |
|---|---|---|
| `"The capital of France is"` | `[264] " a"` | `[825] " one"` |

After forcing spec prefill through `get_model_worker_batch` (padded, same as nospec):

| | first token | divergence point |
|---|---|---|
| spec (padded prefill) | `[264] " a"` ✓ | token ~30 (verify still unpadded) |

i.e. **same 5 real tokens, different logits depending on whether 59 padding tokens are present**.

### Root cause

`EPMoE._permute` (`layers/moe.py:705-712`):

```python
flatten_selected_experts = jnp.ravel(top_k_indices)          # includes padding tokens' expert ids
sorted_selected_experts = jnp.argsort(flatten_selected_experts, stable=True)
token_indices = sorted_selected_experts // self.num_experts_per_tok
group_sizes = jnp.bincount(flatten_selected_experts, length=self.num_experts)
```

Padding tokens (`embed_tokens[0]`) go through the router and get real expert assignments. After the stable argsort, a real token's row index inside its expert group depends on how many padding tokens picked a `<=` expert id — so the same real token lands in a different GMM tile under different padding. Under bf16 accumulation that's enough to flip argmax when top-1/top-2 logits are close.

`_unpermute` correctly scatters padding outputs back to padding positions (so this is **not** a logical contamination), and the internal 128-align pad at `_gmm_compute:610-613` is trimmed at `_unpermute:728`. The remaining effect is purely **numeric ordering** inside GMM.

### Why this matters for spec decode

- nospec path: `get_model_worker_batch` pads tokens/bs to precompile buckets
- spec path: `get_spec_model_worker_batch` does **not** pad (`eagle_worker.py:60` `# FIXME(pc) add padding logic here`)
- spec verify forwards `bs × num_draft_tokens` real tokens; nospec decode forwards `padded_bs × 1`

[...]

PR #1171, opened 2026-05-20
refactor(spec): per-rank ScheduleBatch.spec_info

pengchengneo:
Refactor scheduler-persisted \`spec_info\` from PR #1108's "cross-rank flat on \`reqs_info[0]\`" single-point layout to **per-rank** (symmetric with \`reqs\` / \`seq_lens\` / \`req_pool_indices\`). Forward path stays cross-rank-flat / DP-scatter-padded; boundary code does split/concat conversion.

Five commits map to five steps, each independently runnable:

| Commit | Step | Summary |
|---|---|---|
| \`ee14210e\` | A | Add \`_split/_concat_spec_info_per_rank\` pure-function helpers + CPU round-trip tests |
| \`4c34ccd2\` | B | Boundary swap: input side concat→scatter in \`_get_spec_decode_mwb_dp\`, output side split→write back to \`reqs_info[r].spec_info\`; introduce \`MWB.spec_info_padded\` field; drop \`_global_spec\` pull-out/putback from \`filter_batch\`; route all spec extend through \`get_model_worker_batch\` (\`get_spec_model_worker_batch\` shrinks to decode-only) |
| \`464de2ff\` | C | Rename 30+ forward-internal \`mwb.spec_info\` to \`mwb.spec_info_padded\` across single-layer (\`eagle_draft_worker\`) and multi-layer (\`multi_layer_draft_worker\`); drop \`MWB.spec_info\` dataclass field |
| \`3ea953f4\` | D | Simplify \`filter_batch\` (drop manual 5-field filter; \`EagleDraftInput.filter_batch\` covers all 5 and dispatches by length match instead of trusting \`has_been_filtered\`); make \`padding_for_decode\` **DP-segmented per-field pad** (end-padding lets shard_map(P(\"data\")) hand a following rank's data to a prior rank under dp>1) |
| \`88d6fb28\` | E | Drop \`ScheduleBatch.spec_info\` (mixin's dynamic attribute write removed; all \`self.spec_info\` readers switched to \`reqs_info[0].spec_info\`); cherry-pick PR #1108 follow-up's dp>1 precompile fix (\`compilation_manager.real_bs_per_dp\` + EAGLE worker \`precompile_spec_*\`) |

### Testing done

- **CPU**: \`test_spec_dp_shapes.py\` 37/37 (28 baseline + 9 new round-trip + assertions updated for per-rank layout)
- **Pod single-layer EAGLE3** (\`niu-v6e4-sleep\` 4-chip v6e, Llama-3.1-8B-Instruct + EAGLE3-LLaMA3.1-Instruct-8B):
  - dp=1 single req → reference \`[264,3363,315,30363,11,1989,11,11401,11,323]\` ✓
  - dp=1 concurrent x4 → all reqs match reference ✓
  - dp=2 single req → matches reference ✓

### Known issues (not in scope)

- **dp>1 multi-req concurrent token mismatch**: dp=2 with 2+ concurrent reqs produces wrong (repeating) tokens for half the reqs. Diagnostic logging proves split/concat/scatter/merge_batch are all correct (layouts align with selector); the bug is forward-internal (KV cache or attention-metadata vs dp-shard synchronization). This is an **inherent issue in PR #1108's dp>1 spec extend path**: baseline asserts dp>1 spec extend off, so it was never exercised. This PR finally routes dp>1 spec extend through, surfacing the bug. **Tracked separately; follow-up PR.**
- **Multi-layer EAGLE (MiMo-V2-Pro) end-to-end validation**: code rename covered (\`multi_layer_draft_worker.py\` 11 sites); E2E left to the next stage per plan (single-layer first, then multi-layer).

### DIFF
--- a/python/sgl_jax/srt/layers/attention/flashattention_backend.py
+++ b/python/sgl_jax/srt/layers/attention/flashattention_backend.py
@@ -213,15 +213,15 @@ class FlashAttention(AttentionBackend):
 
         if batch.forward_mode == ForwardMode.TARGET_VERIFY:
             # convert custom_mask from bool to int32, because dma not support bool type
-            if batch.spec_info.custom_mask.dtype == jnp.bool:
+            if batch.spec_info_padded.custom_mask.dtype == jnp.bool:
                 # FIXME(pc) rm this dtype convert
                 logger.warning(
-                    "batch.spec_info.custom_mask type is  %s, it may make performance very low",
-                    batch.spec_info.custom_mask.dtype,
+                    "batch.spec_info_padded.custom_mask type is  %s, it may make performance very low",
+                    batch.spec_info_padded.custom_mask.dtype,
                 )
-                metadata.custom_mask = batch.spec_info.custom_mask.astype(jnp.int32)
+                metadata.custom_mask = batch.spec_info_padded.custom_mask.astype(jnp.int32)
             else:
-                metadata.custom_mask = batch.spec_info.custom_mask
+                metadata.custom_mask = batch.spec_info_padded.custom_mask
         else:
             metadata.custom_mask = None
 
@@ -230,7 +230,7 @@ class FlashAttention(AttentionBackend):
         if batch.forward_mode.is_target_verify():
             padded_batch_size = len(batch.seq_lens)
             extend_seq_lens = np.zeros(padded_batch_size, dtype=np.int32)
-            extend_seq_lens[batch.logits_indices_selector] = batch.spec_info.draft_token_num
+            extend_seq_lens[batch.logits_indices_selector] = batch.spec_info_padded.draft_token_num
         else:
             extend_seq_lens = batch.extend_seq_lens
         cu_q_lens = _per_dp_cumsum(extend_seq_lens, dp_size, per_dp_bs)
@@ -246,7 +246,7 @@ class FlashAttention(AttentionBackend):
             # stride. Dense targets (EAGLE3) keep the unpadded path so accept-
             # rate / cache-hit behavior is unchanged.
             if metadata.custom_mask is not None and self.page_size >= 256:
-                q = batch.spec_info.draft_token_num
+                q = batch.spec_info_padded.draft_token_num
                 cm = np.asarray(jax.device_get(metadata.custom_mask))
                 # cm is DP-slot-ordered (build_tree got verified_seq_len = mwb.seq_lens-1
                 # over total_bs). Per-slot cm length = q*(verified_seq_len[s]+q); for pad
@@ -292,7 +292,7 @@ class FlashAttention(AttentionBackend):
             # the DP-segmented layout from padding_for_decode (rank r's pages
             # at [r*per_dp_pg : ...]). page_indices (line 212) is already
             # cache_loc[::page_size]//page_size, so re-gather from it per-rank.
-            allocate_lens = batch.spec_info.allocate_lens
+            allocate_lens = batch.spec_info_padded.allocate_lens
             if hasattr(allocate_lens, "device"):
                 allocate_lens = jax.device_get(allocate_lens)
             allocate_lens = np.asarray(allocate_lens)
@@ -370,7 +370,7 @@ class FlashAttention(AttentionBackend):
         # (== selector order), so the per-req gather below stays aligned.
         sel = np.asarray(batch.logits_indices_selector)
         current_seq_lens = np.asarray(batch.seq_lens)[sel]
-        allocate_lens = np.asarray(batch.spec_info.allocate_lens)[sel]
+        allocate_lens = np.asarray(batch.spec_info_padded.allocate_lens)[sel]
 
         draft_allocs = allocate_lens - current_seq_lens
 
@@ -432,7 +432,7 @@ class FlashAttention(AttentionBackend):
 
         if batch.spec_algorithm.is_none():
             raise RuntimeError("should not reach here")
-        assert isinstance(batch.spec_info, EagleDraftInput)
+        assert isinstance(batch.spec_info_padded, EagleDraftInput)
         topk = batch.speculative_eagle_topk
         cu_q_lens = np.tile(np.arange(0, per_dp_bs * topk + 1, topk, dtype=np.int32), dp_size)
         seq_2d = np.asarray(batch.seq_lens).reshape(dp_size, per_dp_bs)
--- a/python/sgl_jax/srt/layers/attention/flashattention_backend.py
+++ b/python/sgl_jax/srt/layers/attention/flashattention_backend.py
@@ -240,29 +240,31 @@ class FlashAttention(AttentionBackend):
         if batch.forward_mode.is_target_verify():
             seq_lens += extend_seq_lens
             aligned_seq_lens = ((seq_lens + self.page_size - 1) // self.page_size) * self.page_size
-            # Hybrid-SWA targets at page_size>=256 hit Mosaic's tiling(8) proof
-            # in rpa_v3._fetch_mask. Pad each mask row to page-aligned kv_len so
-            # the kernel can use cu_kv_lens delta as a statically-8-divisible
-            # stride. Dense targets (EAGLE3) keep the unpadded path so accept-
-            # rate / cache-hit behavior is unchanged.
-            if metadata.custom_mask is not None and self.page_size >= 256:
+            # At dp>1 the verify mask must be DP-segmented per rank so each rank's
+            # P("data") shard sees its own slots' mask (the kernel computes
+            # cu_seq_mask_lens from per-rank cu_kv/q starting at 0). Without
+            # repack, the build_tree output sits in cross-rank-flat order, and
+            # rank>0 reads the mask tail (padding zeros) → all-masked verify
+            # → garbage. (#1108 P1-7)
+            #
+            # mask col width depends on rpa_v3's mask_aligned_to_cu_kv switch
+            # (page_size>=256 uses aligned cu_kv_lens; else actual seq_lens),
+            # so pad each row to match — otherwise kernel reads wrong stride.
+            if metadata.custom_mask is not None and dp_size > 1:
                 q = batch.spec_info_padded.draft_token_num
                 cm = np.asarray(jax.device_get(metadata.custom_mask))
                 # cm is DP-slot-ordered (build_tree got verified_seq_len = mwb.seq_lens-1
                 # over total_bs). Per-slot cm length = q*(verified_seq_len[s]+q); for pad
-                # slots verified_seq_len=-1 → q*(q-1). Repack per DP rank (real slots'
-                # rows padded to aligned kv, pad slots = zeros), then pad each rank's
-                # section to a common length so the result shards P("data") and each
-                # rank's locally-computed cu_seq_mask_lens (starting at 0) indexes its
-                # own shard. (#1108 P1-7 — n>1 dp>1 verify mask cross-rank read.)
+                # slots verified_seq_len=-1 → q*(q-1).
                 cm_kl = np.where(seq_lens > 0, seq_lens, q - 1).astype(np.int64)
                 cm_off = np.concatenate([[0], np.cumsum(q * cm_kl)])
+                row_width = aligned_seq_lens if self.page_size >= 256 else seq_lens
                 rank_chunks: list[np.ndarray] = []
                 for r in range(dp_size):
                     parts = []
                     for j in range(per_dp_bs):
                         s = r * per_dp_bs + j
-                        kla = int(aligned_seq_lens[s])
+                        kla = int(row_width[s])
                         if seq_lens[s] > 0:
                             kl = int(seq_lens[s])
                             row = cm[cm_off[s] : cm_off[s] + q * kl].reshape(q, kl)
@@ -279,7 +281,7 @@ class FlashAttention(AttentionBackend):
                 ).astype(np.int32)
                 metadata.custom_mask = device_array(
                     packed,
-                    sharding=NamedSharding(self.mesh, P("data") if dp_size > 1 else P()),
+                    sharding=NamedSharding(self.mesh, P("data")),
                 )
         else:
             aligned_seq_lens = (
--- a/python/sgl_jax/srt/layers/logits_processor.py
+++ b/python/sgl_jax/srt/layers/logits_processor.py
@@ -201,10 +201,10 @@ class LogitsMetadata:
             extend_seq_lens=device_array(batch.extend_seq_lens, sharding=sharding),
             logits_indices=device_array(batch.logits_indices, sharding=sharding),
             accept_lens=(
-                device_array(batch.spec_info.accept_length, sharding=sharding)
-                if batch.spec_info is not None
-                and hasattr(batch.spec_info, "accept_length")
-                and batch.spec_info.accept_length is not None
+                device_array(batch.spec_info_padded.accept_length, sharding=sharding)
+                if batch.spec_info_padded is not None
+                and hasattr(batch.spec_info_padded, "accept_length")
+                and batch.spec_info_padded.accept_length is not None
                 else None
             ),
             extend_seq_lens_cpu=extend_seq_lens_cpu,
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -188,6 +188,13 @@ class EPMoE(nnx.Module):
                         f"Unsupported {scale_name} shape {scale.shape} for weight shape {weight.shape}. "
                         f"Expected k_blocks dimension to be 1 or {expected_k_blocks}."
                     )
+            # FIXME(jax-upgrade): incoming sharding is typically P("expert", None, None, None) —
+            # the EPMoE weight loader passes a 3-tuple ("expert", None, None) that JAX pads
+            # with None on the 4D array — which doesn't textually match the downstream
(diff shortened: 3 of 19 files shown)


pengchengneo:
## Multi-host validation (v6e 16 × 4 = 64 chip, MiMo-V2.5-Pro)

Verified the option-C refactor end-to-end on a 64-chip v6e cluster (16 × v6e-8x8 pods, `tp=ep=64`) against `MiMo-V2.5-Pro` with NEXTN speculative decoding, project-pinned venv (jax 0.8.1 + libtpu 0.0.30).

**Launch flags** (single line, abbreviated):
```
--moe-backend epmoe --tp-size 64 --ep-size 64 --dp-size {1,2,4}
--speculative-algorithm NEXTN --speculative-num-steps 3
--speculative-eagle-topk 1 --speculative-num-draft-tokens 4
--max-running-requests 16 --chunked-prefill-size 4096 --disable-precompile
--precompile-token-paddings 128 256 512 1024 2048 4096   # see "pre-existing" note 1 below
```

**Determinism + correctness check** (`temperature=0.0`, `max_new_tokens=16`):
- `A = "The quick brown fox"` — run twice, compare `output_ids`
- `B = "In a hole in the ground there lived a hobbit."` — different prompt sanity

| dp | tp | ep | A1 == A2 (deterministic) | B ≠ A | warm-call latency | notes |
|----|----|----|--------------------------|-------|-------------------|-------|
| 1  | 64 | 64 | n/a (single request)     | n/a   | —                 | 8-token greedy decode, 200 OK |
| 2  | 64 | 64 | ✓                        | ✓     | 0.77 s            | A2 reused warm JIT cache |
| 4  | 64 | 64 | ✓                        | ✓     | 0.67 s            | A2 reused warm JIT cache |

Observed `output_ids` for prompt `A` were byte-identical between dp=2 and dp=4 (`[34208, 916, 279, 15678, 5562, 13, 1096, ...]` decoded to `" jumps over the lazy dog. ..."`), and distinct from prompt `B`'s output (`" Not a nasty, dirty, wet hole, ..."`) — i.e. the per-rank `spec_info` boundary split/concat preserves greedy determinism across DP layouts.

## Pre-existing issues (not introduced by this PR — reproduce on `epic/mtp-refactor-phase1` and `main`)

These are worked around at the launcher level for the validation run; they will be filed separately.

1. **`compilation_manager._compute_token_buckets` lacks the fused-MoE guard** that `_compute_bs_buckets` already enforces. The default token padding list `[64, 128, ...]` lets a 1-token request pad to `num_tokens=64`, which fails `validate_fused_moe_block_config` with `local_num_tokens=1 / t_packing=2` when `ep_size==64` and `dtype=bfloat16`. Worked around with `--precompile-token-paddings 128 256 512 1024 2048 4096`.
2. **Cluster image `/workdir/extras` overlay** shadows the venv with `jax 0.10.1` while `libtpu==0.0.30` (project pin) stays in the venv, producing `Pallas TPU requires a libtpu version that's at most a month old`. Worked around per-pod with `uv pip install -e ".[tpu]"` against `/opt/venv` and dropping `/workdir/extras` from `PYTHONPATH`.

The drive-by `moe.py` FIXME in this PR (`9b4f3fb4`) documents a third sharding-spec asymmetry that does not fire on jax 0.8.1 (degenerate-axis lenience) but will raise on a jax 0.10.x upgrade.

Single-host coverage is in the parent epic PR comment.

pengchengneo:
## Reproducible test recipe
```
  python -u -m sgl_jax.launch_server \
  --model-path /models/MiMo-V2.5-Pro --trust-remote-code \
  --tp-size 64 --dp-size "$DP_SIZE" --ep-size 64 --moe-backend epmoe \
  --page-size 256 --context-length 32768 --chunked-prefill-size 4096 --dtype bfloat16 \
  --mem-fraction-static 0.80 --swa-full-tokens-ratio 0.15 --max-running-requests 16 \
  --skip-server-warmup --disable-radix-cache --attention-backend fa \
  --speculative-algorithm NEXTN --speculative-draft-model-path /models/MiMo-V2.5-Pro \
  --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4 \
  --disable-overlap-schedule --disable-precompile \
  --nnodes 16 --node-rank "$TPU_WORKER_ID" --dist-init-addr "${COORD_IP}:${COORD_PORT}" \
  --decode-log-interval 1 --port 8000 \
  --precompile-token-paddings 128 256 512 1024 2048 4096 
```



### dp=1 sanity (8 tokens, greedy)

```bash
kubectl exec mimo-pro-eagle-0-qkgbx -c jax-tpu -- \
  curl -sS -X POST -H 'Content-Type: application/json' \
    -d '{"text":"x","sampling_params":{"max_new_tokens":8}}' \
    http://127.0.0.1:8000/generate
```

Returned `output_ids=[198,2,5308,14,5543,271,14374,5308]` (`"\n# PRO/II\n\n### PRO"`), 200 OK, e2e 167.8 s (includes JIT compile).

### dp=2 and dp=4 determinism + different prompt

The same 3-call block was run for both dp values:

```bash
GEN='{"sampling_params":{"max_new_tokens":16,"temperature":0.0}}'
call() {
  local txt="$1"
  local req=$(printf "%s" "$GEN" | python3 -c \
    "import sys,json; d=json.load(sys.stdin); d['text']=sys.argv[1]; print(json.dumps(d))" "$txt")
  kubectl exec mimo-pro-eagle-0-qkgbx -c jax-tpu -- \
    curl -sS -X POST -H 'Content-Type: application/json' -d "$req" \
      http://127.0.0.1:8000/generate
}

R1=$(call 'The quick brown fox')                              # A — first call
R2=$(call 'The quick brown fox')                              # A — repeat (must equal R1)
R3=$(call 'In a hole in the ground there lived a hobbit.')    # B — different prompt
```

Pass criteria: `R1.output_ids == R2.output_ids` and `R3.output_ids != R1.output_ids`. See the table in the previous comment for the resulting tokens.

zorrofox (commented):
Went through all 10 commits + ran `test_spec_dp_shapes.py` locally on CPU (37/37 pass). Design looks right — per-rank storage makes `merge_batch`/`filter_batch` symmetric with everything else in `reqs_info[r]`, and the boundary concat→scatter / gather→split keeps the forward path unchanged. A few findings:

### Bugs (minor)

1. **`schedule_batch.py:2415` + `:2516` double `acc_global_bid()`** in the dp=1 `get_spec_decode_model_worker_batch` path. L2415 is the pre-existing throw-away call (epic-base had a latent `NameError` here that never fired because the path was dead post-#1108); L2516 is the new `bid = acc_global_bid()`. Net effect: bid counter increments twice per dp=1 spec decode. Just delete L2415.

2. **`schedule_batch.py:2457` `self.spec_info.draft_token_num`** — `ScheduleBatch.spec_info` no longer exists (step E dropped it), so this is an `AttributeError` waiting to happen. It's currently dead code (L2407 asserts `decode_or_idle()` so the `TARGET_VERIFY` branch at 2455-2457 can't execute), but should be deleted along with the whole `if TARGET_VERIFY` arm.

### Test gap

3. **No `merge_batch` CPU test.** The new tests cover split/concat round-trip but not `ScheduleBatch.merge_batch` itself — which is the path the refactor was motivated by. Suggest adding `test_merge_batch_per_rank_spec_info` parametrized over the cases from the #1053 thread:
   - `self=[r0:2, r1:1]` + `other=[r0:1, r1:2]` (both ranks non-empty both sides → per-rank concat)
   - `self=[r0:2, r1:1]` + `other=[r0:0, r1:2]` (other rank0 empty → old L1685 `continue` case)
   - `self=[r0:0, r1:2]` + `other=[r0:1, r1:0]` (self rank0 empty → old L1700 overwrite case)

   Assert each `reqs_info[r].spec_info.verified_id` matches `[self_r ++ other_r]` after merge. Happy to push this as a commit on top if you want.

4. **Multi-host validation didn't exercise `merge_batch`.** The dp=2/4 determinism check fires one prompt at a time, so `running_batch` is always empty when the next prefill arrives → scheduler.py:1488 direct-assign, never L1492 `merge_batch`. Same blind spot as our #1108 acceptance. If you still have the v6e-64 up, a "fire 2 reqs → sleep 1s (let them enter decode) → fire 2 more" run would close that. (We released our reservation 5/21 so I can't run it.)

### Nits

5. `schedule_batch.py:2153` docstring still says "the cross-round flat state on `reqs_info[0].spec_info` is unchanged" — should be `reqs_info[r].spec_info` (per-rank).
6. `_concat_spec_info_per_rank`: if one rank's `accept_length` is `None` and another's isn't, `nonnull` silently drops the None and the concat length no longer matches `sum(real_bs_per_dp)`. Shouldn't happen in practice (all ranks come from the same verify split), but an `assert len(nonnull) in (0, len(nonempty))` per field would catch it loudly.

### Bonus (nice)

- `prepare_for_extend_after_target_prefill` DP-padded walk (commit `7094341e`) fixes the "prefill draft_extend rank>0 input unshifted" leftover I'd flagged in the #1053 closeout — thanks.
- `EagleDraftInput.filter_batch` now also filters `allocate_lens` (was missing before, the global pull-out path had been masking that).
- `compilation_manager` `real_bs_per_dp=[per_dp_bs_size]*dp_size` and the precompile `dp_size` wiring fix dp>1 precompile shapes.

CI: run 26230440884 still pending; previous run had `unit-test-{1,4}-tpu` failures but logs aren't available yet (run still in progress). Will check back once it completes.

pengchengneo:
## Review feedback addressed (commit `35fdbe67`) + merge_batch exercised on v6e-64

### Code changes (per @zorrofox's review)

| # | Bug / Nit | Resolution |
|---|---|---|
| 1 | Double `acc_global_bid()` at `schedule_batch.py:2415` + `:2516` in dp=1 spec decode | Deleted the L2415 throw-away call; bid counter now increments once per batch as intended. |
| 2 | Dead `if forward_mode == TARGET_VERIFY` arm at `schedule_batch.py:~2455` referencing defunct `self.spec_info.draft_token_num` | Arm removed; surrounding L2407 assert already excludes TARGET_VERIFY, so this was just an `AttributeError` trap. |
| 5 | `_scatter_spec_info_to_dp_slots` docstring still said `reqs_info[0].spec_info` | Fixed to `reqs_info[r].spec_info`. |
| 6 | `_concat_spec_info_per_rank` silently dropping `None` per-field would let concat length drift away from `sum(real_bs_per_dp)` | Added all-or-nothing assert per field; partial-None mixes now fail loudly. |
| 3 | No CPU test for `ScheduleBatch.merge_batch` per-rank spec_info | Added `test_merge_batch_per_rank_spec_info` to `test_spec_dp_shapes.py`, parametrized over the three cases from #1053: (a) both ranks non-empty both sides → per-rank concat via `EagleDraftInput.merge_batch`; (b) other-rank empty → `merge_batch` early-continue keeps self; (c) self-rank empty → overwrite branch pulls other wholesale. Each rank's `verified_id` is tagged with a stable `(rank, side, idx)` offset; the test asserts the post-merge per-rank `verified_id` equals the expected `[self_r ++ other_r]` sequence. |

### #4 Multi-host: `merge_batch` actually exercised on v6e 16×4 dp=2 (mimo V2.5-Pro, NEXTN num_steps=3)

Followed the "2 reqs → sleep → 2 reqs" recipe to force a non-empty `running_batch` at the second prefill:

```bash
# pair A — fire 2 concurrently
for i in 1 2; do
  curl -X POST :8000/generate \
    -d '{"text":"The quick brown fox","sampling_params":{"max_new_tokens":32,"temperature":0.0}}' &
done
sleep 10
# pair B — fire 2 more concurrently while pair A is in decode
for i in 1 2; do
  curl -X POST :8000/generate \
    -d '{"text":"In a hole in the ground there lived a hobbit.","sampling_params":{"max_new_tokens":32,"temperature":0.0}}' &
done
wait
```

Server log shows `scheduler.py:1492` `running_batch.merge_batch(last_batch)` did fire:

```
15:19:24  Prefill batch. #new-seq: 2, #running-req: 0, #prefill per DP: [1, 1], #running per DP: [0, 0]    # pair A — no running yet, L1488 direct-assign
15:20:10  Prefill batch. #new-seq: 2, #running-req: 2, #prefill per DP: [1, 1], #running per DP: [1, 1]    # pair B — 2 already running → L1492 merge_batch fires
15:20:50  Decode batch.  #running-req: 4, #running-req per DP: [2, 2], accept-len 2.50, accept-ratio 0.62  # 4-way decode after merge
```

All four requests returned 200 OK (32-token greedy decode each). `B1.output_ids == B2.output_ids` byte-identical (DP-rank symmetry preserved through merge). `A1.output_ids != A2.output_ids` — this is the expected non-determinism when concurrent reqs share a JIT batch (different padding / draft-accept context per concurrent slot); it's not regressed by this PR (reproduces on `epic/mtp-refactor-phase1`) and the sequential-determinism check in the previous comment still holds.

CI: HEAD `35fdbe67`, run [26233703131](https://github.com/sgl-project/sglang-jax/actions/runs/26233703131) — most jobs green; one `e2e-test-4-tpu (2)` flaked on a GitHub Actions `##[error]The runner has received a shutdown signal` mid-MMLU eval (73/128 examples in, all `200 OK`). Will rerun that single job once the workflow settles.

zorrofox (commented):
Re-ran `test_spec_dp_shapes.py` at `35fdbe67` — 40/40 pass locally including the 3 new `merge_batch` cases. Diff for items 1/2/5/6 all look right. The merge_batch test is exactly what was needed (tag-based, covers all three branch shapes). And the v6e-64 staggered run with `#running-req: 2` → `merge_batch` firing → 4-way decode at `accept-len 2.50` closes the loop on #4.

LGTM from my side.

One non-blocking observation on the `A1 != A2` you noted: same prompt + greedy + concurrent → different output is the #1090 verify-batch variance signature (batch composition shifts MoE-EP routing → target logits drift → spec accept path diverges). Agree it's not introduced here since it reproduces on epic base, but worth linking from #1090 as another repro recipe — it's a cleaner/smaller one than the padded-prefill case we had.

(CI: the `e2e-test-4-tpu(2)` runner-shutdown is infra flake, not this PR.)

zorrofox:
Ran an independent dp=4 verification at `35fdbe67` on v6e-64. Incremental coverage over the runs above: **dp=4** + **7-prompt hetero** (unbalanced per-rank) + **spec vs nospec diff**.

### dp=4 spec — hetero7 (7 distinct prompts, t=0 concurrent)

DP allocation `[2,2,1,2]` (rank2 singleton). All 7 returned 200 OK, accept-len 2.14–3.50 over the run.

### dp=4 spec — staggered hetero (4 prompts → sleep 3s → 3 prompts, max_tok=512)

`merge_batch` fired with **unbalanced** per-rank shapes on both sides:
```
00:29:22  Prefill #new-seq: 4  #running-req: 0  prefill/DP=[1,1,1,1]  running/DP=[0,0,0,0]   ← wave-A
00:29:29  Prefill #new-seq: 3  #running-req: 4  prefill/DP=[1,1,1,0]  running/DP=[1,1,1,1]   ← wave-B → L1492 merge_batch
00:29:36  Decode  #running-req: 7               running/DP=[2,2,2,1]  accept-len 2.00→3.14   ← post-merge
```
Post-merge 7-way decode ran clean (throughput 137–225 tok/s warm, no crash). Partial-finish then drained 7→6→5→4 with per-DP going `[2,2,2,1]→…→[1,1,2,0]`. (Client side hit a 900s timeout because each new bs bucket triggered a fresh JIT compile under `--disable-precompile`; server side completed — harness issue, not the PR.)

### dp=4 nospec — same two scenarios

Both completed; nospec staggered also shows `Prefill #new-seq: 3 #running-req: 4 prefill/DP=[1,1,0,1] running/DP=[1,1,1,1]` → merge.

### spec vs nospec diff (hetero7, dp=4, ids[:32])

| prompt | result |
|---|---|
| [2] `1+1=` / [3] `quick brown fox…` / [4] `quantum entanglement` / [5] `def fibonacci` / [6] `first 10 primes` | **byte-identical** |
| [0] `capital of France` | diverges at token 26 |
| [1] `factorial of n` | diverges at token 11 |

**5/7 identical, 2/7 diverge** — same ratio as our #1108 dp=4 baseline (pre-refactor), so not a #1171 regression. The 2 diverging prompts are both on rank0 (the `[2,_,_,_]` slot). Also worth noting: **nospec-hetero7[0] ≠ nospec-staggered[0]** (token 11: `576` vs `12095`) — i.e. the divergence reproduces between two *nospec* runs with different batch composition, which is squarely #1090 (MoE-EP target logits shift with batch shape), not spec-specific.

Net: per-rank refactor holds under dp=4 hetero + merge_batch + partial-finish; no regression vs #1108 acceptance.

pengchengneo:
Thanks @zorrofox for the LGTM and the extra dp=4 hetero / staggered / partial-finish run — that's exactly the coverage gap I was worried about closing. Net summary across both rounds:

- Per-rank refactor holds on v6e-64 under: dp 1/2/4, balanced + unbalanced (`prefill/DP=[1,1,1,0]`), `merge_batch` fired with non-empty `running_batch` on both sides, partial-finish drain.
- 5/7 prompts byte-identical between spec and nospec at dp=4 hetero; the 2/7 diverging on rank0 reproduce on `epic/mtp-refactor-phase1` at the same ratio as the #1108 baseline — same #1090 (MoE-EP batch-shape sensitivity) signature, not a #1171 regression. I'll cross-link this as a small repro recipe on #1090.
- CI run [26233703131](https://github.com/sgl-project/sglang-jax/actions/runs/26233703131) is fully green after re-running the one runner-shutdown e2e flake.

No code changes from the last round.

Issue #4, opened 2025-08-19
[Feature] Support Qwen3 Moe with Fuse Moe


Prayer3th:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

* Support Qwen3 Moe Model
* Support Fuse Moe
* Support GQA with flash attention

PR #1175, opened 2026-05-21
docs(cookbook): add structured cookbook for autoregressive recipes

lianfang0320:
## Summary

Introduces `docs/cookbook/` as the single home for TPU serving recipes, organized by capability with shared base references and deployment templates:

- `autoregressive/` — text LLMs + vision-language decoders, organized by vendor (`DeepSeek/` · `GLM/` · `Google/` · `Grok/` · `InclusionAI/` · `Llama/` · `Moonshotai/` · `Qwen/` · `Xiaomi/`)
- `diffusion/` — diffusion-style T2V models (`Wan/`)
- `base/` — `tpu-topology-reference.md`, `launch-flags-reference.md`, `basic-api-usage.md`
- `deployment/` — `single-host-docker.md`, `gke-indexed-job.md` (primary multi-host), `skypilot.md` (advanced v6e)
- `troubleshooting.md` — cross-recipe generic failure modes
- `index.md` — landing page

Every recipe follows the same 5-section schema:

1. **Model Introduction** — variants, recommended sampling params, license
2. **Deployment** — 2.1 Hardware Matrix / 2.2 Environment / 2.3 Launch (single-host + multi-host) / 2.4 Configuration Tips
3. **Invocation** — chat completion + reasoning / tool-call patterns where applicable
4. **Benchmark** — Accuracy (§4.1) and Speed (§4.2), each with Test Environment / Deployment Command / Benchmark Command / Test Results
5. **Troubleshooting** — model-specific symptom → cause → fix table

A four-tier validation legend is used across the index:
`✅` Validated · `🧪` Partially validated · `🚧` Starter · `📝` Planned.

Each recipe ships **only the variants that were actually run on `sglang-jax 0.1.0`**; unmeasured siblings and unmeasured second hardware paths are explicitly out of scope for this PR (filed as follow-up work). Banner / §4 Test Environment / footnote tables all pin to `sglang-jax 0.1.0`.

**Base-model writing rules** (codified across recipes): base / pretrain-only models (currently just Grok-2 in scope) must declare "base model — not instruction-tuned" in §1, use `/v1/completions` in §3.1 (not `/v1/chat/completions`), skip §4.1 Accuracy entirely because chat-format datasets via `/v1/chat/completions` produce evalscope extraction artifacts on base models, and carry two specific base-model rows in §5. Lesson came from Grok-2 measuring 17%/20% on GSM8K — root cause was an evalscope extraction artifact from base-model self-continuation, not a model defect.

### ✅ Validated (real `bench_serving` + accuracy on `sglang-jax 0.1.0`)

| Recipe | Validated variant | TPU | Highlight |
|---|---|---|---|
| `Qwen/Qwen.md` | Qwen-7B-Chat | v6e-4 | reference single-host dense recipe |
| `Qwen/Qwen3.md` | Qwen3-8B + Qwen3-32B | v6e-4 | GSM8K 0.975; reference dense recipe with benchmark comparison |
| `Qwen/Qwen2.5-VL.md` | 3B / 7B / 32B / 72B | v6e-4 (3B/7B/32B) | request path verified end-to-end; §4 Benchmark intentionally omitted pending real PR-back MMMU / DocVQA / ChartQA numbers |
| `Qwen/Qwen3-MoE.md` | Qwen3-30B-A3B | v6e-16 | reference single-host-class MoE recipe; `--moe-backend epmoe` (fused requires `intermediate_size % 512 == 0`; 30B-A3B's 768 fails) |
| `Llama/Llama3.1.md` | Llama 3.1 8B-Instruct | v6e-4 | GSM8K 0.825 (200); `--page-size 128` mandatory (default `page-size=1` caps `max_running_requests` → 1, ~9× slower) |
| `Llama/Llama3.3-70B.md` | Llama 3.3 70B | v6e-32 | reference multi-node dense recipe |
| `Google/Gemma2.md` | Gemma 2 27B-it | v6e-4 | 974 tok/s output, GSM8K 0.865 |
| `Grok/Grok2.md` | Grok-2 base | v6e-32 | reference base-model recipe (no chat template, `/v1/completions` flow). Architecture corrected to MoE (8 experts, 2 active). §4.1 Accuracy intentionally omitted per the base-model writing rules. Recipe uses `--moe-backend epmoe` (issue #7 — fused MoE backend init crashes on the (data=1, tensor=64) mesh with 8 experts) |
| `GLM/GLM-4.5.md` | GLM-4.5-Air (106B) | v6e-32 | reference 100B-class MoE recipe with both reasoning + tool-call parsers (`glm45`); GSM8K 0.955 / TPOT 13.35 ms / 2153 tok/s. The transpose-mapping fix from issue #6 shipped in `sglang-jax 0.1.0`. |
| `DeepSeek/DeepSeek-V2.md` | V2-Lite / V2-Lite-Chat | v6e-4 | GSM8K + speed on v6e-4; use `-Chat` checkpoint for accuracy eval (base lacks chat template, evalscope loops) |
| `DeepSeek/DeepSeek-V3.md` | V3 / V3-Base | v6e-64 | GSM8K 0.975 (200), bench_serving 249 tok/s output. `--dp-size 8` mandatory for FP8 shared-expert block-quant compatibility (resolves issue #3); `--moe-backend epmoe` is the validated default |
| `DeepSeek/DeepSeek-R1.md` | R1 | v6e-64 | GSM8K thinking-on 0.98 (50). Inherits V3 mesh fix; adds `--reasoning-parser deepseek-r1` |
| `Moonshotai/Kimi-Linear.md` | 48B-A3B-Instruct | v6e-16 + v6e-32 | reference linear-attention recipe; both hardware paths measured |
| `InclusionAI/Ling-2.6.md` | Ling-2.6-1T | v6e-64 | GSM8K 0.985 (200), bench_serving in §4.2. Requires the channel-wise FP8 QKV split fix that landed in this PR (resolves issue #4); earlier builds crash at weight load |
| `Xiaomi/MiMo-V2-Flash.md` | MiMo-V2-Flash | v7x-8 + v6e-16 | v7x-8 config sweep (winning 480 tok/s), v6e-16 multi-host 1034 tok/s, GSM8K 0.9401 |
| `Xiaomi/MiMo-V2.5-Pro.md` | MiMo-V2.5-Pro | v6e-64 | GSM8K 0.975 (200), bench_serving 469 tok/s output (1.89× DeepSeek-V3 total throughput). v7x-16 is supported alternative hardware (same launch shape) — no current-build rerun required |
| `Xiaomi/MiMo-7B.md` | MiMo-7B-RL | v6e-4 | GSM8K 0.920; reference compact reasoning recipe |
| `diffusion/Wan/Wan2.1.md` | Wan 2.1 T2V-14B | v6e-4 | §3.1 basic-usage smoke run (480*832 / 41 frames / 50 steps, ~4 min wall-clock); Wan has no numeric `evalscope` / `bench_serving` driver — §4 is intentionally narrative |
| `diffusion/Wan/Wan2.2.md` | Wan 2.2 T2V A14B | v6e-4 | §3.1 basic-usage smoke run (480*832 / 41 frames / 50 steps, ~2 min 43 s wall-clock); narrative §4 |

### 🧪 Partially validated

None remaining in this PR — the patch dependencies that previously held GLM-4.5-Air and Qwen3-MoE at 🧪 have both landed in `sglang-jax 0.1.0` (see Resolved issues #4 and #6).

### 🚧 Starter

None remaining in this PR.

#### 1. Multimodal `bench_serving` fails at the final metadata step

After all generation requests succeed, `bench_serving.py` calls `GET /get_server_info`, which goes through `tokenizer_manager.get_internal_state()` → `GetInternalStateReq`. The multimodal path does not wire that request type through:

- `MultimodalTokenizer._result_dispatcher` forwards only batch outputs, `AbortReq`, and `ProfileReqOutput` — `GetInternalStateReqOutput` is dropped.
- `GlobalScheduler` dispatches `ProfileReq` but not `GetInternalStateReq`.

Result: VL benchmarks complete the request path but hang / time out / fail at the metadata collection step. Qwen2.5-VL ships ✅ on request-path verification but the §4 Benchmark block is omitted pending this fix.

#### 2. Reasoning / tool-call parser gaps vs upstream sglang

Several recipes have to point users at other models for `<think>` / tool-call workloads because the corresponding parser isn't wired on the sglang-jax side yet, even though upstream sglang supports them:

- **DeepSeek-V2-Lite / V3** — non-reasoning, no native tool-call format on sglang-jax. Recipes redirect to DeepSeek-R1 for `<think>` and to Qwen3 / MiMo-V2.5-Pro for tool-calling.
- **DeepSeek-R1** — `--reasoning-parser deepseek-r1` works; no tool-call parser, so tool-call workloads still redirect.

Other models (Qwen3, Qwen3-MoE, MiMo, GLM-4.5, Kimi reserved) already have the parsers registered — see the parser-key reference in `autoregressive/index.md`.

#### 7. Grok-2 fused MoE backend crashes on small-expert × (data=1, tensor=64) mesh

`--moe-backend fused` crashes the FusedEPMoE init on the v6e-64 (data=1, tensor=64) mesh shape when `num_local_experts=8` (Grok-2's MoE config). Recipe falls back to `--moe-backend epmoe`, which runs correctly. Same root area as the `t_packing` alignment failure documented for DeepSeek-V3 (resolved issue #3) — the fused MoE kernel assumes per-rank token counts above a threshold that small-expert MoE on a wide mesh doesn't meet.

#### 6. GLM-4.5 (Air) `WeightMapping` `transpose=False` mismatch — **RESOLVED in `sglang-jax 0.1.0`**

`python/sgl_jax/srt/models/glm4_moe.py` previously declared `transpose=False` on every q/k/v/o, dense-MLP, and `shared_experts` `WeightMapping` block, but HF stores those weights as `[out, in]` and `LinearBase` expects `[in, out]`. Without the flip, GLM-4.5-Air on v6e-32 produced garbage output. The 10-mapping `transpose=True` patch landed in `sglang-jax 0.1.0`, unblocking GLM-4.5-Air's promotion to ✅. The recipe banner still carries a one-line "if pre-0.1.0 see §5" pointer for users on older builds.

#### 3. DeepSeek-V3 repetition under bring-up config — **RESOLVED**

Long-output repetition loops on v6e-64 were root-caused as an FP8 block-quant mesh constraint: when the per-rank shared-expert `out_dim = moe_intermediate_size(2048) / tensor_axis` equals `block_size_out=128` (i.e., tensor=16 / `--dp-size 4`), the block-wise quantized matmul kernel enters its documented accuracy-collapse regime — the `epmoe` path asserts explicitly, the `fused` path silently emits garbage tokens.

Fix: `--dp-size 8` so `out_dim = 2048 / 8 = 256 > 128`. The dense MLP block-quant scale `(144, 56)` further requires `144 % tensor == 0`, so on v6e-64 `tensor=8` is the only working option. Validated launch flags are now the DeepSeek-V3 recipe defaults; R1 inherits the same mesh fix.

#### 4. Ling-2.6 channel-wise FP8 QKV scale loading crash — **RESOLVED in commit `d9c98c80`**

`weight_utils.py::_split_qkv_weight()`'s 2D scale branch unconditionally dereferenced `quant_cfg.weight_block_size[0]`, which is `None` for `compressed-tensors + strategy="channel"` checkpoints. Ling-2.6-1T ships per-channel weight scales as `[Q+K+V, 1]`, so it crashed at startup with `TypeError: 'NoneType' object is not subscriptable`.

Fix landed in this PR: branch on quantization semantics (`weight_block_size`) instead of tensor rank alone — when `weight_block_size is None` and the scale's trailing dim is 1, split by output channel and keep the `[out, 1]` shape so the existing `("tensor", None)` sharding spec stays applicable; the post-shard `_maybe_expand_linear_block_scale` call squeezes back to `[out]`. Block-wise path (DeepSeek-V3 / R1) unchanged. Ling-2.6-1T promoted from 🚧 to ✅ on the back of this fix.

## Out of scope for this PR

Variants and second hardware paths that the cookbook recipes do **not** yet cover. Each will be added in a follow-up PR once measured on `sglang-jax 0.1.0` or a successor pin.

**Unmeasured variants of in-scope recipes:**

- **DeepSeek-V2 full 236B** on v6e-32 — needs the `gate.py` one-line patch (`routed_scaling_factor` nested inside the `renormalize` branch; V2 is the only family member with `norm_topk_prob=False` + `routed_scaling_factor=16.0`). Recipe currently ships V2-Lite only.
- **GLM-4.5 full 355B** on v6e-64 — same launch shape as Air; v6e-64 nodepool released this pass before the run could land.
- **Qwen3-MoE 235B-A22B** on v6e-64 — same launch path as 30B-A3B with larger TP / EP.
- **MiMo-7B Base / SFT** — same launch path as MiMo-7B-RL.
- **Llama 3.1 Phi-3 / InternLM3 aliases** — same SGL-JAX runtime path as Llama 3.1 8B-Instruct.
- **Ling-2.6 v7x-16** path — same launch shape as v6e-64 with `--tp-size 32 --dp-size 4`.
- **Wan 2.1 T2V-1.3B** — same launch path as T2V-14B, smaller checkpoint.

**Entirely out-of-scope models (no recipe in this PR):**

- **Grok-1** — upstream `xai-org/grok-1` ships native JAX `ckpt-0/*` shards rather than safetensors; sgl-jax's loader needs `ckpt-0/*` support or a safetensors mirror with `Grok1ForCausalLM` key naming.
- **Ling / Ring 2.5 (1T)** — `--quantization fp8 --quantization-config-path fp8_bailing.yaml` OOMs on v6e-64 because the quant config excludes MoE expert weight paths (`mlp.experts.wi_0|wi_1|wo`), leaving ~1.5–1.7 TB effective on-chip footprint.
- **GLM-5** — runtime supports `Glm5ForCausalLM` / `GlmMoeDsaForCausalLM`, but no public release pin yet.

## Design notes

- **Recipes are self-contained** — each recipe stands on its own for a reader landing cold. Sibling-pointer cross-references in §1 / Additional Resources were dropped; "choose a model with parser support" callouts route through the canonical [Parser key reference](docs/cookbook/autoregressive/index.md#parser-key-reference) table in the autoregressive index; `bench_serving` driver commands are inlined where previously they said "adapt from [Other recipe]"; cross-model Speed comparison footnotes (e.g., "1.89× faster than DeepSeek-V3") are removed because absolute numbers go stale across recipes.
- **Basic-API duplication is the one exception** — short Python OpenAI client snippets are inlined in §3.1 alongside the link to `base/basic-api-usage.md`, because the snippet plus the recipe-specific sampling params is what readers copy-paste.
- **GKE Indexed Job as primary multi-host path** — SkyPilot demoted to v6e advanced/experimental in deployment templates and recipe §2.3.
- **User-perspective constraint** — recipes do not reference internal source paths (e.g., `scheduler.py:273`, `fused_moe/v1/kernel.py`), runtime class names (`FusedEPMoE`, `MultimodalServerArgs`), PR numbers, design RFCs, commit hashes, or dated audit traces ("validated 2026-05-25"). All builds are pinned to `sglang-jax 0.1.0`; only public HF cards and other cookbook pages are cited.
- **Variant scope-narrow per recipe** — each recipe ships only the variants actually validated on `sglang-jax 0.1.0`; HF-card-only siblings are listed under "Out of scope" rather than carried in-recipe as `_Pending_` placeholders.
- **Shipped tree scope** — the cookbook tree only ships user-facing content: `index.md` / `troubleshooting.md` / `docs.json` (landing, cross-recipe, navigation), `autoregressive/` and `diffusion/` (recipes organized by vendor), `base/` (TPU topology, launch flags, basic API), and `deployment/` (launcher templates). Author-tooling scaffolding (`2026-05-18-cookbook-research.md`, `cookbook-recipe-design.md`, `2026-05-21-recipe-command-audit.md`, `mintlify-migration.md`, dated audit / integration directories) was dropped from the shipped tree; the underlying knowledge lives in commit history, design docs, and the per-recipe banners / §5 troubleshooting rows that already absorbed the load-bearing findings.

## Test plan

- [x] Render `docs/cookbook/index.md` and walk all internal links — every `[...](relative/path.md)` resolves
- [x] Validate the ✅ recipes end-to-end on the listed TPU topology and `sglang-jax 0.1.0` (one variant per family per the validation rule above)
- [x] Verify `--reasoning-parser` / `--tool-call-parser` key names against `python -m sgl_jax.launch_server --help`
- [x] Land the Ling-2.6 channel-wise FP8 QKV scale fix (commit `d9c98c80`, resolves issue #4)
- [ ] File follow-up issues for the open runtime gaps: #1 multimodal `get_server_info` (blocks Qwen2.5-VL §4 numbers) · #2 DeepSeek tool-call parser · #7 Grok-2 fused MoE on small-expert mesh


docs/cookbook/autoregressive/DeepSeek/DeepSeek-V2.md
```diff
@@ -0,0 +1,179 @@
...
+|---|---|---|
+| MoE throughput plateau | Wrong `--moe-backend` for EP size | V2-Lite uses `epmoe` at EP=4. |
+| OOM at startup (V2-Lite) | `--mem-fraction-static` too high | Lower from default 0.88 to 0.85. Verify `--tp-size 4` matches v6e-4 chip count. |
+| First request takes ~4 min | JIT cache empty | Persist `JAX_COMPILATION_CACHE_DIR` across restarts. |
+
+## Additional Resources
+
+- [DeepSeek-V2-Lite model card](https://huggingface.co/deepseek-ai/DeepSeek-V2-Lite)
+- [`../../base/launch-flags-reference.md`](../../base/launch-flags-reference.md)
+- [`../../troubleshooting.md`](../../troubleshooting.md) — cross-recipe generic issues.
```
JamesBrianD: Do not include any extra paths in the link name.

docs/cookbook/autoregressive/GLM/GLM-4.5.md
```diff
@@ -0,0 +1,312 @@
+---
+title: "GLM-4.5"
+---
+
+# GLM-4.5 MoE on SGL-JAX
+
+> **Validated recipe** — GLM-4.5-Air (106B) validated on TPU v6e-32 with sglang-jax 0.1.0; sanity + GSM8K + bench all pass, **but requires a 10-line patch to `python/sgl_jax/srt/models/glm4_moe.py`** as of `d9c98c80`: every `transpose=False` on the q/k/v/o, dense-MLP, and shared_experts `WeightMapping` blocks must flip to `transpose=True` (HF stores `[out, in]`, `LinearBase` expects `[in, out]` — compare every other validated recipe). Apply via the in-manifest patch heredoc shown in [`../../2026-05-21-recipe-command-audit/glm45-air-32-patched/manifest.yaml`](../../2026-05-21-recipe-command-audit/glm45-air-32-patched/manifest.yaml).
```
JamesBrianD: The debug content can be deleted or hidden.

docs/cookbook/autoregressive/DeepSeek/DeepSeek-R1.md
```diff
@@ -0,0 +1,320 @@
...
+V6e-64 is the minimum slice that fits the official FP8 checkpoint plus runtime overhead. See [`../../base/tpu-topology-reference.md`](../../base/tpu-topology-reference.md) for the TPU generation reference.
+
+### 2.2 Environment
+
+Install per [`../../../get_started/install.md`](../../../get_started/install.md). Multi-host required — use [`../../deployment/gke-indexed-job.md`](../../deployment/gke-indexed-job.md) as the primary user-facing path. Advanced users running temporary v6e experiments can adapt [`../../deployment/skypilot.md`](../../deployment/skypilot.md). The required JAX TPU container image:
+
+| Hardware Platform               | Docker Image                                                       |
+|---|---|
+| TPU v5e / v5p / v6e (Trillium)  | `us-docker.pkg.dev/cloud-tpu-images/jax-ai-image/tpu:jax0.8.1-rev1` |
+| TPU v7x (Ironwood)              | `us-docker.pkg.dev/cloud-tpu-images/jax-ai-image/tpu:jax0.8.1-rev1` |
```
zhengkezhou1: Users do not need this information.

docs/cookbook/autoregressive/DeepSeek/DeepSeek-R1.md
```diff
@@ -0,0 +1,320 @@
...
+
+For evaluation, additionally install `evalscope` in the client environment:
+
+```bash
+pip install evalscope==0.17.1
+```
+
+### 2.3 Launch
+
+#### Multi-host (GKE Indexed Job) — TPU v6e-64
```
zhengkezhou1: I think this should be differentiated by use case. For one-off workloads such as benchmarking or performance testing, using a Kubernetes Job is a reasonable choice. However, if the goal is to deploy a long-running service, a Deployment would be more appropriate: [https://kubernetes.io/docs/concepts/workloads/controllers/deployment/](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)

docs/cookbook/autoregressive/DeepSeek/DeepSeek-V2.md
```diff
@@ -0,0 +1,179 @@
...
+Install per [`../../../get_started/install.md`](../../../get_started/install.md). For V2-Lite single-host use [`../../deployment/single-host-docker.md`](../../deployment/single-host-docker.md). The required JAX TPU container image:
+
+| Hardware Platform               | Docker Image                                                       |
+|---|---|
+| TPU v5e / v5p / v6e (Trillium)  | `us-docker.pkg.dev/cloud-tpu-images/jax-ai-image/tpu:jax0.8.1-rev1` |
+| TPU v7x (Ironwood)              | `us-docker.pkg.dev/cloud-tpu-images/jax-ai-image/tpu:jax0.8.1-rev1` |
+
+### 2.3 Launch
+
+#### Single-host (Docker) — TPU v6e-4 (DeepSeek-V2-Lite)
```
zhengkezhou1:
```suggestion
#### Single-host — TPU v6e-4 (DeepSeek-V2-Lite)
```

docs/cookbook/autoregressive/Google/Gemma2.md
```diff
@@ -0,0 +1,212 @@
...
+  --skip-server-warmup \
+  --host 0.0.0.0 --port 30000
+```
+
+The 27B-it `--mem-fraction-static 0.85` leaves room for the dual (global + sliding) KV pools.
+
+#### Multi-host
+
+Not needed at the 27B scale — fits single-host on v6e-4.
+
```
zhengkezhou1: This section is unnecessary.

docs/cookbook/autoregressive/Grok/Grok2.md
```diff
@@ -0,0 +1,262 @@
...
+  --chunked-prefill-size 2048 \
+  --page-size 128 \
+  --max-running-requests 128 \
+  --random-seed 3 \
+  --skip-server-warmup
+```
+
+Mount a shared `JAX_COMPILATION_CACHE_DIR` on the same PVC as the model weights — first cold compile is ~5-10 min, much faster than 1T-class models since the MoE kernel shape sweep is smaller.
+
+#### Multi-host (GKE Indexed Job) — TPU v6e-32 (8 nodes, starter)
```
zhengkezhou1: It would be helpful to explain the intended use cases for each configuration. It's not clear why both deployment options are needed or when users should choose one over the other.

docs/cookbook/deployment/gke-indexed-job.md
```diff
@@ -0,0 +1,150 @@
...
+# GKE Indexed Job Launcher
+
+Generic GKE template for launching a multi-host TPU slice serving SGL-JAX. This page is the **single source of truth** for the Indexed Job + headless Service pattern — recipes reference it instead of pasting a full manifest, and fill in only their model-specific fields (`<JOB>`, `<ACCELERATOR>`, `<TOPOLOGY>`, `<LAUNCH_FLAGS>`).
+
+## Why Indexed Job + headless Service
+
+Multi-host JAX needs:
+1. **A stable rank-0 hostname** for `--dist-init-addr`. Indexed Job gives every pod the index `${JOB_COMPLETION_INDEX}`; combined with a headless Service the pod `0` becomes `<job>-0.<svc>` permanently.
+2. **TPU process address fanout**. The JAX TPU runtime expects `TPU_PROCESS_ADDRESSES` listing every node's `host:8471`. Headless DNS makes those names enumerable.
+3. **No restarts on crash**. `restartPolicy: Never` + `backoffLimit: 0` — if one pod dies, the whole job fails fast rather than partial restart that would desync ranks.
```
zhengkezhou1:
```suggestion
```

docs/cookbook/deployment/single-host-docker.md
```diff
@@ -0,0 +1,95 @@
+---
```
zhengkezhou1, line 1: An example showing how to run it with Docker:

> Docker images are published at <https://hub.docker.com/r/lmsysorg/sglang-jax>.
> 
> Pull image
> ```bash
> docker pull lmsysorg/sglang-jax:<TAG>
> ```
> 
> Run service on a TPU VM
> 
> ```bash
> docker run --rm -it \
>   --name sglang-jax \
>   --privileged \
>   --network=host \
>   -e JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache \
>   -e HF_HOME=/tmp/models \
>   -v /tmp/jit_cache:/tmp/jit_cache \
>   lmsysorg/sglang-jax:<TAG> \
>   python3 -u -m sgl_jax.launch_server \
>   --model-path Qwen/Qwen-7B-Chat \
>   --trust-remote-code \
>   --tp-size=4 \
>   --device=tpu \
>   --dtype=bfloat16 \
>   --mem-fraction-static=0.8 \
>   --download-dir=/tmp/models \
>   --host=0.0.0.0 \
>   --port=30000
> ```
> 
> ## Verify the service
> 
> ```bash
> curl -X POST http://127.0.0.1:30000/v1/chat/completions \
>   -H "Content-Type: application/json" \
>   -d '{
>     "model": "Qwen/Qwen-7B-Chat",
>     "messages": [{"role": "user", "content": "What is the capital of France?"}],
>     "max_tokens": 64,
>     "temperature": 0
>   }'
> ```

zhengkezhou1 (changes requested):

docs/cookbook/autoregressive/Moonshotai/Kimi-Linear.md
```diff
@@ -0,0 +1,236 @@
...
+  --max-running-requests 256 \
+  --skip-server-warmup
+```
+
+For temporary v6e experiments, advanced users can adapt [`../../deployment/skypilot.md`](../../deployment/skypilot.md) with the same launch flags. The model recipe does not require users to run repository-local SkyPilot helper scripts.
+
+### 2.4 Configuration Tips
+
+**Mandatory: `--disable-radix-cache` for hybrid recurrent state:**
+- Kimi-Linear is a hybrid recurrent state model — radix prefix sharing is unsafe with the recurrent state pool. Without `--disable-radix-cache` startup asserts: `Hybrid recurrent state models require --disable-radix-cache (prefix sharing is unsafe with recurrent state)`.
```
Rodrian7: Quick note — "prefix sharing is unsafe with recurrent state" isn't quite right. It's not unsafe, we just haven't built prefix/radix caching for hybrid recurrent models yet (planned, just got pushed).

JamesBrianD (approved):

PR #1177, opened 2026-05-21
fix(logprob): scalar input_token_logprobs + bucketed logprob path (host RAM + JIT churn)

### LINKED ISSUE #1176: [Bug] Engine.generate(return_logprob=True) — host RAM blowup and throughput collapse from unbucketed logprob path
## Symptom

`Engine.generate(..., return_logprob=True, logprob_start_len=0)` causes the scheduler subprocess (`sglang::scheduler_NP0`) host RSS to grow until the pod is OOM-killed, and throughput collapses to a small fraction of the no-logprob baseline. Reproduced on TPU v6e 4×4 with Qwen3-30B-A3B (vocab=151,936), driving a long-running, heterogeneous-shape eval workload (MMLU-Pro PPL-style: ~13.5k prompts × 14 categories).

Observed end-to-end with a local fix in place: throughput recovers ~19× (from ~0.025 q/s pre-fix to ~0.47 q/s post-fix), because the scheduler main thread was previously blocked inside `jax/_src/compilation_cache.py::compress_executable` / `_cache_write` between most batches.

## Root cause — unbucketed logprob path with two compounded symptoms

The `return_logprob=True` extend path in `LogitsProcessor.__call__` bypasses the bucketed-shape design used by the main prefill path. The main path pre-pads `input_ids` / `positions` / `out_cache_loc` to `total_token_size` so the JIT cache key stays stable across batches; the logprob path does not, and additionally returns full vocab rows to host. Two symptoms result.

### Symptom 1 — full-vocab `input_token_logprobs` returned to host (true leak)

`meta_info["input_token_logprobs"]` is documented as a list of `(scalar_logprob, token_id, text)` per input token. In practice each entry carries `(vocab_logprobs[1, vocab_size], token_id, text)` — the entire vocab row. For vocab=151,936 × bf16 that is **~297 KB per token**, ~178 MB per ~600-tok prompt, ~36 GB per 200-prompt `generate()` call. The host transfer at `scheduler_output_processor_mixin.py:86-88` does `jax.device_get(logits_output.input_token_logprobs).astype(float)` on the full-vocab tensor; nothing on the path narrows it to the realized token before the host transfer.

### Symptom 2 — JIT cache key instability causes recompile churn (not a leak, but unbounded host growth)

This is **not** a true leak — JAX's live-executable cache has no LRU eviction, so every distinct cache key retains its compiled artifact (~800 MB of Pallas IR + HLO + jaxpr) on host indefinitely. With a churning key, host RSS grows linearly (~1.3 GB/min in our workload) and only plateaus once the key space is exhausted, which never happens on heterogeneous workloads. Two independent shape-instability sources stack:

1. **Aux-data instability.** `LogitsMetadata.tree_flatten` places `extend_seq_lens_cpu` / `extend_logprob_start_lens_cpu` / `extend_logprob_pruned_lens_cpu` (Python `list[int]`) in the pytree aux dict. JAX compares aux by value, so any per-batch variation in these lists is a new cache key.
2. **Traced-shape instability.** Even with aux fixed, `LogitsProcessor.__call__` builds `pruned_states = jnp.concat([hidden_states[pt+start:pt+end] for ...])`, making the traced intermediate length equal to `sum(extend_len - start_len)` — which varies per batch. So the abstract shape entering `_get_logits` is also a per-batch cache key.

Together these drive a sustained ~4% recompile rate on prefill batches. `py-spy dump` caught the scheduler main thread inside `compress_executable` mid-run — the cost shows up in throughput as well as RSS, because the main loop is blocked compiling rather than scheduling the next batch.

## Reproduction

check https://github.com/sgl-project/sglang-jax/issues/1176#issuecomment-4507231466

## Fix

PR: https://github.com/sgl-project/sglang-jax/pull/1177

Results: https://github.com/sgl-project/sglang-jax/issues/1176#issuecomment-4508284599

MokusMokun:
## Summary

Fixes #1176.

- **Symptom 1 (true leak)** — `LogitsProcessor.__call__` now gathers the realized-token logprob on-device before host transfer, so `meta_info["input_token_logprobs"][i][0]` is a scalar (or `None`) instead of a full `[1, vocab_size]` row. `scheduler_output_processor_mixin._materialize_input_token_logprobs` adds a 1-D shape guard so a regression here is caught at the boundary instead of silently OOM-ing the scheduler.
- **Symptom 2 (recompile churn)** — JIT cache key stabilized in the logprob path: when the padded path is active, `LogitsMetadata.tree_flatten` drops the per-batch `extend_seq_lens_cpu` / `extend_logprob_start_lens_cpu` / `extend_logprob_pruned_lens_cpu` aux entries, and `ScheduleBatch.get_model_worker_batch` builds bucketed `input_logprob_indices` / `merged_extend_input_logprob_token_ids` padded to `total_token_size`. `LogitsProcessor.__call__` consumes those instead of rebuilding a variable-length `jnp.concat` per batch.
- **DP>1 chunked-prefill follow-up** — `process_batch_result_prefill` used a single-slot `skip_stream_req` that silently dropped all-but-one chunked req on dp>1, surfacing as `NoneType` crashes / truncated logprobs in `TokenizerManager`. Replaced with a `set` of `id(req)`; `stream_output` accepts both shapes for back-compat.

## Test plan

- [x] `pre-commit run --files <changed>` — all hooks Passed (ruff, black, isort, codespell, mypy)
- [x] New unit test `test/srt/test_logprobs.py::TestLogprobsDpChunkedPrefill::test_dp2_multi_req_chunked_prefill_logprobs` — exercises 4 varying-length prompts (160–280 tokens) with `dp_size=2`, `chunked_prefill_size=128` so multiple reqs are mid-prefill in the same batch; asserts every position is a finite scalar (or `None` only at index 0)
- [x] Existing `TestLogprobsDense::test_logprobs` extended with `assert_scalar_logprobs` to lock in the Symptom 1 contract
- [ ] CI on TPU v6e (gated by `run-ci` label)

### DIFF
--- a/python/sgl_jax/srt/layers/logits_processor.py
+++ b/python/sgl_jax/srt/layers/logits_processor.py
@@ -109,6 +109,7 @@ class LogitsMetadata:
     extend_logprob_pruned_lens_cpu: list[int] | None = None
     top_logprobs_nums: list[int] | None = None
     extend_input_logprob_token_ids_device: jax.Array | None = None
+    input_logprob_indices_device: jax.Array | None = None
     token_ids_logprobs: list[list[int]] | None = None
 
     # logits and logprobs post processing
@@ -123,9 +124,11 @@ class LogitsMetadata:
             self.accept_lens,
             self.logits_indices,
             self.extend_input_logprob_token_ids_device,
+            self.input_logprob_indices_device,
             self.temperature,
             self.top_p,
         )
+        has_padded_input_logprob_indices = self.input_logprob_indices_device is not None
 
         aux_data = {
             "forward_mode": self.forward_mode,
@@ -133,9 +136,15 @@ class LogitsMetadata:
             "extend_return_logprob": self.extend_return_logprob,
             "extend_return_top_logprob": self.extend_return_top_logprob,
             "extend_token_ids_logprob": self.extend_token_ids_logprob,
-            "extend_seq_lens_cpu": self.extend_seq_lens_cpu,
-            "extend_logprob_start_lens_cpu": self.extend_logprob_start_lens_cpu,
-            "extend_logprob_pruned_lens_cpu": self.extend_logprob_pruned_lens_cpu,
+            "extend_seq_lens_cpu": (
+                None if has_padded_input_logprob_indices else self.extend_seq_lens_cpu
+            ),
+            "extend_logprob_start_lens_cpu": (
+                None if has_padded_input_logprob_indices else self.extend_logprob_start_lens_cpu
+            ),
+            "extend_logprob_pruned_lens_cpu": (
+                None if has_padded_input_logprob_indices else self.extend_logprob_pruned_lens_cpu
+            ),
             "top_logprobs_nums": self.top_logprobs_nums,
             "token_ids_logprobs": self.token_ids_logprobs,
             "temp_scaled_logprobs": self.temp_scaled_logprobs,
@@ -152,8 +161,9 @@ class LogitsMetadata:
         obj.accept_lens = children[1]
         obj.logits_indices = children[2]
         obj.extend_input_logprob_token_ids_device = children[3]
-        obj.temperature = children[4]
-        obj.top_p = children[5]
+        obj.input_logprob_indices_device = children[4]
+        obj.temperature = children[5]
+        obj.top_p = children[6]
 
         obj.forward_mode = aux_data["forward_mode"]
         obj.capture_hidden_mode = aux_data["capture_hidden_mode"]
@@ -220,6 +230,9 @@ class LogitsMetadata:
             extend_input_logprob_token_ids_device=device_array(
                 batch.extend_input_logprob_token_ids, sharding=sharding
             ),
+            input_logprob_indices_device=device_array(
+                batch.input_logprob_indices, sharding=sharding
+            ),
         )
 
 
@@ -259,6 +272,28 @@ class LogitsProcessor(nnx.Module):
             in_specs=(P("data", "tensor"), P("data")),
             out_specs=P("data", "tensor"),
         )(logits, indices)
+
+    def _select_input_token_logprobs(
+        self, input_logprobs: jax.Array, token_ids: jax.Array | None
+    ) -> jax.Array:
+        # Keep the public input_token_logprobs API scalar-per-token. Returning
+        # full vocab rows here makes host memory scale as tokens * vocab_size.
+        # By the time we reach this function, extend_return_logprob is True
+        # (gated in __call__), so at least one req has extend_len - start_len
+        # > 0. Both the padded path (np.zeros over total_token_size in
+        # get_model_worker_batch) and the non-padded chunk concat produce a
+        # non-None token_ids in that case. A None here means an upstream
+        # regression dropped the gather ids — fail loudly rather than
+        # silently returning None to the user.
+        assert token_ids is not None, (
+            "extend_input_logprob_token_ids_device must be populated when "
+            "extend_return_logprob=True; got None"
+        )
+        token_ids = token_ids.reshape((-1,))
+        row_ids = np.arange(input_logprobs.shape[0], dtype=np.int64)
+        out_sharding = NamedSharding(self.mesh, P(None))
+        selected = input_logprobs.at[(row_ids, token_ids)].get(out_sharding=out_sharding)
+        return selected.reshape((input_logprobs.shape[0],))
 
     @named_scope
     def __call__(
@@ -303,39 +338,44 @@ class LogitsProcessor(nnx.Module):
             # 1. pruned_states: hidden states that we want logprobs from.
             # 2. sample_indices: Indices that have sampled tokens.
             # 3. input_logprob_indices: Indices that have input logprob tokens.
-            sample_index_pt = -1
-            sample_indices = []
-            input_logprob_indices_pt = 0
-            input_logprob_indices = []
-            pt, pruned_states = 0, []
+            if logits_metadata.input_logprob_indices_device is not None:
+                pruned_states = hidden_states
+                sample_indices = logits_metadata.logits_indices
+                input_logprob_indices = logits_metadata.input_logprob_indices_device
+            else:
+                sample_index_pt = -1
+                sample_indices = []
+                input_logprob_indices_pt = 0
+                input_logprob_indices = []
+                pt, pruned_states = 0, []
 
-            for extend_logprob_start_len, extend_len in zip(
-                logits_metadata.extend_logprob_start_lens_cpu,
-                logits_metadata.extend_seq_lens_cpu,
-            ):
-                start_len = extend_logprob_start_len
-                pruned_states.append(hidden_states[pt + start_len : pt + extend_len])
-                pt += extend_len
-                sample_index_pt += extend_len - start_len
-                sample_indices.append(sample_index_pt)
-                input_logprob_indices.extend(
-                    [
-                        input_logprob_indices_pt + i
-                        for i in range(extend_len - extend_logprob_start_len)
-                    ]
+                for extend_logprob_start_len, extend_len in zip(
+                    logits_metadata.extend_logprob_start_lens_cpu,
+                    logits_metadata.extend_seq_lens_cpu,
+                ):
+                    start_len = extend_logprob_start_len
+                    pruned_states.append(hidden_states[pt + start_len : pt + extend_len])
+                    pt += extend_len
+                    sample_index_pt += extend_len - start_len
+                    sample_indices.append(sample_index_pt)
+                    input_logprob_indices.extend(
+                        [
+                            input_logprob_indices_pt + i
+                            for i in range(extend_len - extend_logprob_start_len)
+                        ]
+                    )
+                    input_logprob_indices_pt += extend_len - start_len
+
+                pruned_states = jnp.concat(pruned_states)
+                sample_indices = device_array(
+                    np.array(
+                        sample_indices,
+                        dtype=np.int64,
+                    ),
                 )
-                input_logprob_indices_pt += extend_len - start_len
-
-            pruned_states = jnp.concat(pruned_states)
-            sample_indices = device_array(
-                np.array(
-                    sample_indices,
-                    dtype=np.int64,
-                ),
-            )
-            input_logprob_indices = device_array(
-                np.array(input_logprob_indices, dtype=np.int64),
-            )
+                input_logprob_indices = device_array(
+                    np.array(input_logprob_indices, dtype=np.int64),
+                )
 
         # Compute logits for both input and sampled tokens.
         logits = self._get_logits(pruned_states, lm_head)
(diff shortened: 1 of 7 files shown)


MokusMokun:
Force-pushed a fix for the `test_kda_attention.py` failures: `input_logprob_indices` on `ModelWorkerBatch` is now a defaulted field (`= None`) instead of a required positional one, so existing direct constructors in the test suite (kda / mla / flashattention / lora) keep compiling unchanged. Single-commit history preserved.

python/sgl_jax/srt/layers/logits_processor.py
```diff
@@ -419,12 +455,10 @@ def __call__(
             else:
                 input_token_ids_logprobs_val = input_token_ids_logprobs_idx = None
 
-            out_sharding = NamedSharding(self.mesh, P(None))
-            indices = (
-                np.arange(input_logprobs.shape[0]),
+            input_token_logprobs = self._select_input_token_logprobs(
+                input_logprobs,
                 logits_metadata.extend_input_logprob_token_ids_device,
```
aolemila, line 466: It looks like `logits_metadata.extend_input_logprob_token_ids_device` is None here with the following dataflow.

`logits_metadata.extend_input_logprob_token_ids_device` <- `modelWorkerBatch.extend_input_logprob_token_ids` <- `extend_input_logprob_token_ids=None in scheduler.py:2184`

aolemila: I think you do not fix this problem. You just skip to trigger this data transfer in your case. If it was true, could you open a new issue to track and fix it?

python/sgl_jax/srt/layers/logits_processor.py
```diff
@@ -109,6 +109,7 @@ class LogitsMetadata:
     extend_logprob_pruned_lens_cpu: list[int] | None = None
     top_logprobs_nums: list[int] | None = None
     extend_input_logprob_token_ids_device: jax.Array | None = None
+    input_logprob_indices_device: jax.Array | None = None
```
aolemila, line 112: It looks like no one set this field.

python/sgl_jax/srt/layers/logits_processor.py
```diff
@@ -302,39 +332,44 @@ def __call__(
...
-                pt += extend_len
-                sample_index_pt += extend_len - start_len
-                sample_indices.append(sample_index_pt)
-                input_logprob_indices.extend(
-                    [
-                        input_logprob_indices_pt + i
-                        for i in range(extend_len - extend_logprob_start_len)
-                    ]
+            if logits_metadata.input_logprob_indices_device is not None:
+                pruned_states = hidden_states
```
aolemila, line 342: This pruned_states and the next pruned_states are not the same.

python/sgl_jax/srt/layers/logits_processor.py
```diff
@@ -123,19 +124,27 @@ def tree_flatten(self):
...
-            "extend_logprob_pruned_lens_cpu": self.extend_logprob_pruned_lens_cpu,
+            "extend_seq_lens_cpu": (
+                None if has_padded_input_logprob_indices else self.extend_seq_lens_cpu
+            ),
+            "extend_logprob_start_lens_cpu": (
+                None if has_padded_input_logprob_indices else self.extend_logprob_start_lens_cpu
+            ),
+            "extend_logprob_pruned_lens_cpu": (
+                None if has_padded_input_logprob_indices else self.extend_logprob_pruned_lens_cpu
+            ),
```
aolemila, line 147: why not directly remove these three fields?

python/sgl_jax/srt/layers/logits_processor.py
MokusMokun: there is, see `logits_processor.py:232–234`

python/sgl_jax/srt/layers/logits_processor.py
MokusMokun: 1. When `return_logprob=True`, `extend_input_logprob_token_ids_device` is not None,  check `schedule_batch.py` line 2235 and before
2. not sure about what does "skip to trigger this data transfer" mean, if you mean defensive early-return of line 284 is misleading, I am looking into it

python/sgl_jax/srt/layers/logits_processor.py
MokusMokun: Cannot be removed entirely, since they are still needed on the non-padded/fallback path:

- `extend_seq_lens_cpu` and `extend_logprob_start_lens_cpu` are used when `input_logprob_indices_device is None` to build `pruned_states` / `input_logprob_indices` at L347-L371.
- `extend_logprob_pruned_lens_cpu` is still used by host-side materialization/splitting in `tp_worker.py` L503/L510/L525.

This PR only drops them from pytree aux data when the padded `input_logprob_indices_device` path is active, so the JIT key does not specialize on per-batch CPU lengths.

python/sgl_jax/srt/layers/logits_processor.py
MokusMokun: yes, the the two pruned_states differ by design: 
1. first bucketed passes the full padded batch through _get_logits (so the JIT shape is stable)
2. legacy concat-compacts to only logprob-needed rows.

but I think there is no need to alter the current code.
Are you suggesting to imporve the comment above?

python/sgl_jax/srt/layers/logits_processor.py
aolemila: For 1, based on [main](de29d9f0ceb05a8180a194ed85a3e3a419f009df), I can not find the related codes around the schedule_batch.py:2235. Could you show me the link? I have searched extend_input_logprob_token_ids in schedule_batch.py and find [set None](https://github.com/sgl-project/sglang-jax/blob/de29d9f0ceb05a8180a194ed85a3e3a419f009df/python/sgl_jax/srt/managers/schedule_batch.py#L2184) when calling `get_model_worker_batch`. That's the reason why I think it is still None.

For 2, if we based on consensus which means `extend_input_logprob_token_ids_device` is None, `_select_input_token_logprobs` returns None when token_ids is None. That's why I think you skip the data transfer.

python/sgl_jax/srt/layers/logits_processor.py
aolemila: get

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -2162,6 +2162,57 @@ def get_model_worker_batch(
...
+                # return_logprob=True). Previously hardcoded None below, which combined
+                # with the new server-side scalar gather caused 'NoneType.reshape' in
+                # LogitsProcessor._select_input_token_logprobs.
+                chunks = [
+                    info.extend_input_logprob_token_ids
+                    for info in self.reqs_info
+                    if getattr(info, "extend_input_logprob_token_ids", None) is not None
+                ]
+                if chunks:
+                    merged_extend_input_logprob_token_ids = np.concatenate(chunks)
```
aolemila, line 2462: Discussion1: With the help of [prepare_for_extend](https://github.com/sgl-project/sglang-jax/blob/de29d9f0ceb05a8180a194ed85a3e3a419f009df/python/sgl_jax/srt/managers/schedule_batch.py#L1040), it looks like these two cases(if and else) are the same. Because both of them only contain the input_token_ids rather than user specified token ids for every position.

aolemila: If it was true, `and not (token_ids_logprobs and any(x is not None for x in token_ids_logprobs)` would not be required.

python/sgl_jax/srt/layers/logits_processor.py
aolemila: Got it. 

Discussion2:
But I still confuse about the shape of `extend_seq_lens_cpu`, `extend_logprob_start_lens_cpu` and `extend_logprob_pruned_lens_cpu` is fixed according to `LogitsMetadata.from_model_worker_batch` and `ScheduleBatch._merge_batch_metadata`, how does it trigger cache miss?

MokusMokun: Those three `_cpu` fields are Python lists whose values vary per batch. Because they were stored in JAX pytree `aux_data`, their exact Python values participated in the JIT cache key, even if the corresponding padded array shapes were unchanged.

python/sgl_jax/srt/managers/schedule_batch.py
MokusMokun: 1. **Are the `if` and `else` branches redundant?** No. They produce `extend_input_logprob_token_ids` with content-equivalent token ids (good observation re: `prepare_for_extend` L1040), but with different `input_logprobs` row layouts that the host postprocessing handles differently.
2. **Is `and not (token_ids_logprobs and any(x is not None for x in token_ids_logprobs))` redundant?** No. It's required because the host postprocessing for `token_ids_logprob` does not have padding-aware compaction; only the scalar `input_token_logprobs` path does.

Why:

**Step 1 — two layouts.** The `if` branch fills `extend_input_logprob_token_ids` and `input_logprob_indices` into arrays of length `total_token_size`, arranged as per-DP padded slots (`schedule_batch.py:2177-2202`). The `else` branch concatenates per-DP chunks (`schedule_batch.py:2208-2214`) of length `sum(extend_len - start_len)`. So the `input_logprobs` tensor produced by `_select_logits(logits, input_logprob_indices)` in `logits_processor.py:425` has different N across the two branches: padded vs compact-per-req.

**Step 2 — why the layout matters: host postprocessing.** `input_logprobs` is consumed by three separate host paths, each with its own row-layout assumption:

- (a) **scalar `input_token_logprobs`** via `_select_input_token_logprobs` (`logits_processor.py:279-299`), then materialized in `_materialize_input_token_logprobs` (`scheduler_output_processor_mixin.py:47-68`).
- (b) **`input_top_logprobs_val`** for `top_logprobs_num > 0`, materialized in `tp_worker.py:505-519`.
- (c) **`input_token_ids_logprobs_val`** for `token_ids_logprob`, materialized in `tp_worker.py:521-538`.

**Step 3 — in the non-overlap scheduler path, path (a) has padding-aware compaction.** `_materialize_input_token_logprobs` detects the padded case and compacts it back per-DP:

```python
# scheduler_output_processor_mixin.py:58-67
if lens_per_dp:
    valid_len = sum(lens_per_dp)
    if len(values) != valid_len and len(values) % len(lens_per_dp) == 0:
        per_dp_token_size = len(values) // len(lens_per_dp)
        compact_values = []
        offset = 0
        for dp_valid_len in lens_per_dp:
            compact_values.extend(values[offset : offset + dp_valid_len])
            offset += per_dp_token_size
        values = compact_values
```

So path (a) accepts both padded (from the `if` branch) and compact (from the `else` branch) inputs.

**Step 4 — paths (b) and (c) do NOT have such compaction.** They walk rows with a compact-per-req cursor:

```python
# tp_worker.py:524-536 (path c, token_ids_logprob)
pt = 0
for ids, plen in zip(logits_metadata.token_ids_logprobs, pruned_lens):
    per_req_vals.append(full[pt : pt + plen][:, ids].tolist())
    pt += plen
```

`pt += plen` advances strictly by per-req lengths (`pruned_lens = extend_logprob_pruned_lens_cpu`), with no notion of skipping padding rows. The `else` branch's layout matches; the `if` branch's layout (DP-stride + intra-DP zero fills) does not — the cursor would land on padding rows and emit values for the wrong req. The same applies to path (b) in `tp_worker.py:509-517`.

So the two clauses `and not (top_logprobs_nums and ...)` and `and not (token_ids_logprobs and ...)` route requests away from the `if` branch precisely because paths (b) and (c) currently can't consume padded `input_logprobs`. Adding a padding-aware compaction step in front of those two cursor loops (analogous to the one already in `_materialize_input_token_logprobs`) would let us drop both clauses, but that's out of scope for the host-RAM-OOM fix this PR targets. This is also why the fallback branch is not just for avoiding `token_ids=None`: it preserves the compact row layout required by the existing top-logprobs / token_ids-logprob materialization.

aolemila (dismissed):

aolemila (approved):

PR #1188, opened 2026-05-22
[SpecDecode] Fix all cache miss and Precompile shapes

pengchengneo:
## Test Matrix

**Environment variables** (enable JIT cache miss tracing):
```bash
export JAX_COMPILATION_CACHE_DIR=/tmp/jax_cache
export JAX_DEBUG_LOG_MODULES=jax._src.compiler,jax._src.interpreters.pjit
export JAX_EXPLAIN_CACHE_MISSES=1
```

### Single-host (v6e-4, tp=4)

**Target model**: `meta-llama/Llama-3.1-8B-Instruct`
**Draft model**: `unkmaster/EAGLE3-LLaMA3.1-Instruct-8B`

```bash
# dp=1
  --model-path meta-llama/Llama-3.1-8B-Instruct \
  --speculative-algorithm EAGLE \
  --speculative-draft-model-path unkmaster/EAGLE3-LLaMA3.1-Instruct-8B \
  --speculative-eagle-topk 1 --speculative-num-steps 3 --speculative-num-draft-tokens 4 \
  --tp-size 4 --dp-size 1 --mem-fraction-static 0.7 \
  --context-length 4096 --max-prefill-tokens 4096 --chunked-prefill-size -1 --max-running-requests 16 \
  --precompile-bs-paddings 1 4 8 16 --precompile-token-paddings 256 1024 2048 4096 \
  --page-size 64 --attention-backend fa --dtype bfloat16

# dp=2: change --dp-size 2 --mem-fraction-static 0.5
```

### Multi-host (v6e-256, 16 nodes, tp=64, ep=64)

**Target model**: `XiaomiMiMo/MiMo-V2-Flash`
**Draft model**: built-in NextN heads

```bash
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache python -m sgl_jax.launch_server \
    --model-path XiaomiMiMo/MiMo-V2-Flash \
    --trust-remote-code \
    --speculative-algorithm NEXTN \
    --speculative-eagle-topk 1 --speculative-num-steps 3 --speculative-num-draft-tokens 4 \
    --tp-size 64 --dp-size 4 --ep-size 64 \
    --moe-backend epmoe \
    --host 0.0.0.0 --port 30271 \
    --page-size 64 --context-length 4096 \
    --max-prefill-tokens 4096 \
    --dtype bfloat16 --mem-fraction-static 0.7 \
    --swa-full-tokens-ratio 0.15 \
    --max-running-requests 16 \
    --attention-backend fa \
    --disable-overlap-schedule \
    --nnodes 16 --node-rank ${TPU_WORKER_ID} \
    --dist-init-addr ${MASTER_ADDR}:5000

# dp=1: change --dp-size 1
# dp=2: change --dp-size 2
# dp=4: as above
```

| dp | spec | Cold-start trace | Correctness |
|----|------|-----------------|-------------|
| 1  | NEXTN | 0 expected       | ✓           |
| 1  | nospec | 0 expected       | ✓           |
| 2  | NEXTN | 0 expected       | ✓           |
| 2  | nospec | 0 expected       | ✓           |
| 4  | NEXTN | 0 expected       | ✓           |
| 4  | nospec | 0 expected       | ✓           |

### Eval (GSM8K via evalscope)

| Setup | Model | dp | spec | GSM8K Accuracy |
|-------|-------|-----|------|----------------|
| Single-host v6e-4  | Llama-3.1-8B-Instruct + EAGLE3 | 1 | EAGLE | 0.81 |
| Single-host v6e-4  | Llama-3.1-8B-Instruct | 1 | nospec | 0.8 |
| Single-host v6e-4  | Llama-3.1-8B-Instruct + EAGLE3 | 2 | EAGLE | 0.81 |
| Single-host v6e-4  | Llama-3.1-8B-Instruct | 2 | nospec | 0.82 |
| Multi-host v6e-256 | MiMo-V2-Flash + NextN | 1 | NEXTN | 0.98 |
| Multi-host v6e-256 | MiMo-V2-Flash + NextN | 1 | nospec | 0.98 |
| Multi-host v6e-256 | MiMo-V2-Flash + NextN | 2 | NEXTN | 0.98 |
| Multi-host v6e-256 | MiMo-V2-Flash + NextN | 1 | nospec | 0.98 |

**Known Issue**- MiMo flash moebackend set to fused will be crashed, which request bs more than epsize * 2, but it will make spec path oom,  will fix in another PR


zorrofox (commented):
Went through all 13 files (41 commits — quite a journey through the fused_moe revert/reapply cycle 🙂). The overall approach is sound: pin every spec-decode shape to `max(bs_paddings)` / `max(token_paddings)` so there's exactly one jit cache key, and move all variable-`real_bs` postprocessing to host numpy. Ran `test_spec_dp_shapes.py` locally — 41/41 pass; `test_eagle_tree_build.py` has the 2 failures CI shows.

### 🔴 Blocking — `_build_tree_structure_numpy` is incorrect for topk>1

Both `test_eagle_tree_build.py` failures reproduce locally on CPU:

**`test_build_tree_kernel_efficient`** (mask mismatch): the numpy port initializes `mask = np.ones((draft_token_num, kv_len))` and never clears `mask[0, sl+1:]` for `tid==0`. The Pallas kernel (and PyTorch reference) produces `[True×(sl+1), False×(d-1)]` for the root row — root sees prefix + itself, not the other draft positions. Fix is a one-liner in the `tid==0` branch:
```python
if tid == 0:
    mask[0, sl + 1:] = 0
    positions[global_idx] = sl
    ...
```

**`test_build_tree_simple_case`** (`IndexError: index 1 is out of bounds for axis 0 with size 1` at eagle_util.py:317): the test passes `batch_size=2` but `seq_lens=[3]` / `verified_id=[100]` (real bs=1). The Pallas kernel ignored the discrepancy; the numpy port loops `for bid in range(parent_list.shape[0])` and indexes `seq_lens[bid]`. Either the test fixture is wrong (should be `batch_size=1`), or the numpy port should `bs = min(parent_list.shape[0], len(seq_lens))` to match the kernel's tolerance.

**Why this matters beyond CI**: the test matrix in the PR body is all `--speculative-eagle-topk 1`, where the tree degenerates to a chain and `mask[0, sl+1:]` is length-0 anyway — so the bug is invisible there. But #1113 (P1-7, topk>1) will hit this immediately: a wrong root mask means verify accepts draft tokens it shouldn't.

### 🟡 Trade-off worth quantifying

`total_bs = max(bs_paddings)` (schedule_batch.py:2070) and `_pick_context_len → token_paddings[-1]` (eagle_draft_worker.py:475) mean every spec decode round runs at the **largest** precompile bucket regardless of runtime bs. That's the right call for eliminating cache misses, but it trades steady-state compute for compile-once. At `max_running_requests=16` it's probably fine; at production `--max-running-requests 128/512` a single warm request pays 128×/512× the verify FLOPs.

Would be good to add one line to the PR body: warm dp=4 bs=1 ITL with this PR vs epic-base (post-warmup, so cache miss isn't a factor). If it's <2× I'd call it acceptable; if it's the full bucket ratio we might want a 2-bucket compromise (small + max).

Same question for the `_ensure_host()` + `copy_to_host_async` pattern in `capture_for_decode`/`draft_extend_for_decode` — those add device→host syncs every round. The async-copy + late `np.asarray` is the right pattern, but a warm-ITL number would close the loop.

### 🟢 Good

- **dp=1 spec decode unified into `_get_spec_decode_mwb_dp`** (−130 lines) — should've done this in #1108, glad it's gone.
- **`_concat_spec_info_per_rank` → numpy + `len==1` short-circuit** — kills the implicit `jnp.concatenate` jit.
- **kv_cache_mixin reorder** (spec_headroom *after* hybrid split, `draft_runner_cache_size` saved *before*) is much cleaner than the post-hybrid overwrite I had in #1108. Dropping the `draft_pool >= target_slot_range` asserts (56f1662b) is fine given the new ordering makes the invariant structural.
- **`mask_aligned_to_cu_kv` as explicit kwarg** instead of the `page_size>=256` heuristic — and unconditional verify-mask host-pad means dp=1 small-page now works too.
- **`select_top_k_tokens_step_greater_0` dtype cast** — would've crashed on V2-Flash (target bf16 / draft f32) without this.
- **`SpeculativeAlgorithm.NONE` default in compilation_manager** — so nospec precompile pytree-aux matches runtime.

### 🔵 Nits

- `eagle_util.py:_build_tree_structure_numpy` `offset` variable is initialized and incremented but never read.
- `flashattention_backend.py:255` `per_rank_mask_target = ((cm_total // dp_size + 7) // 8) * 8 or 8` — the `or 8` only fires when `cm_total < dp_size`, which can't happen (cm_total ≥ d²·bs ≥ dp_size). Harmless.
- The `_ensure_host` device_fields tuple omits `allocate_lens` (it's already always numpy via scheduler, but for symmetry with `filter_batch` which does index it).

Happy to re-review once the tree-mask fix is in.

pengchengneo:
**Blocking — _build_tree_structure_numpy is incorrect for topk>1**

I have reverted this change, topk>1 mask and for padding mask will be fixed in another pr, I tried it because Mimo-flash's fusedmoe will oom currentyly (it required bs large than ep*2)

**total_bs = max(bs_paddings)**

will fix it to padding into bucket

 **_ensure_host() + copy_to_host_async**
 
 because we need ensure no cache miss, it required we need operate this array on host side, otherwise we need precompile all realbs 
 
 @zorrofox

zorrofox:
@pengchengneo revert confirmed (head `712534d3`, 39 commits) — `test_eagle_tree_build.py` should be green again. Agree on `_ensure_host` being necessary; the question was just about the warm-ITL cost, not whether to do it.

Will re-review once the `total_bs` → bucket change is up.

pengchengneo:
> @pengchengneo revert confirmed (head `712534d3`, 39 commits) — `test_eagle_tree_build.py` should be green again. Agree on `_ensure_host` being necessary; the question was just about the warm-ITL cost, not whether to do it.
> 
> Will re-review once the `total_bs` → bucket change is up.

currently, I think we cannot handle this _ensure_host timecost without scheduler-overlap, but it will be hide behind batch forward when we enable scheduler-overlap,

pengchengneo:
I've fixed the padding to bucket logic (only during the decode phase; padding to maxbs still occurs during the prefill phase). 

I've now set the draft model's token pool size to the target model's token pool size before SWA amplification. Since MIMO's MTP layer is essentially an SWA layer, this is more reasonable for key-value cache capacity assessment (aligning with the upstream Sglang). Otherwise, it's prone to OutOfMemoryError (OOM) after amplification. @zorrofox

zorrofox (commented):
Re-reviewed `c7792020`:

- **`total_bs` → `pad_to_bucket(max_bs_per_dp * dp_size)`** — back to per-bucket padding for decode, addresses the small-bs concern. ✅
- **`LogitsMetadata.accept_lens` gated on `is_draft_extend()`** — keeps pytree leaf shape consistent on the decode path. ✅
- **`capture_hidden_mode=FULL` when `spec_algorithm` set** in `get_model_worker_batch` — needed since spec extend now routes through here and `draft_extend_for_prefill` consumes the hidden. ✅
- **`spec_headroom` moved back to step 5** (before `_apply_token_constraints`) — slight memory inflation under hybrid SWA (headroom gets amplified by the layer ratio), but headroom is tiny (~max_num_reqs·20 + 100) so the simpler single-align path is fine.

Verified locally: `test_spec_dp_shapes.py` 41/41 + `test_spec_info.py` pass; the 3 `test_eagle_tree_build`/`test_eagle_utils` failures are pre-existing on `epic/mtp-refactor-phase1` head (Pallas TPU kernel doesn't run on CPU — same fails on epic-base). CI on TPU is green.

One residual: `_pick_context_len` still returns `token_paddings[-1]` unconditionally (max token bucket). Less impactful than bs since `tree_mask_capacity` scales with seq_len which is bounded by `context_length` anyway, but worth a follow-up if profiling shows it matters.

LGTM. Ack on the `_ensure_host` cost being hidden behind scheduler-overlap once that lands — makes sense to not block on it here.

pengchengneo:
tree_mask_capacity this will be optimized in phase2, I will updated in epic‘s PR

PR #1191, opened 2026-05-24
perf(quantized_matmul): v6e bf16-act/fp8-weight tuned block sizes + sweep script

zorrofox:
## Summary

Follow-up to #1111 (perf item 1). The block-wise quantized matmul tuning table previously had **zero** entries with `x_q_dtype="bfloat16"` (only `int8/int8` and `float8/float8`), so every weight-only FP8 model on v6e (`quantize_activation=False`) fell back to `TunedValue(128,128,128)` with the `Couldn't find tuned sizes` warning.

This PR adds:
- **30 tuned entries** for `(tpu_version=6, x_q_dtype=bfloat16, w_q_dtype=float8_e4m3fn)` covering the GLM-5.1 dense-linear local shapes at TP=64
- **`benchmark/kernels/quantized_matmul/tune_blockwise_block_sizes.py`** — a generic sweep script (mirrors `megablox_gmm/tune_gmm_v1_block_sizes.py`) that takes a shapes file + block_k and emits table-ready entries; reusable for other models/TP configs

## Results — GLM-5.1 FP8, v6e-64 TP64/EP32, cc=8

| | before (#1111) | **after** | Δ | BF16 ref |
|---|---|---|---|---|
| 1K/128 throughput | 81.2 tok/s | **108.8** | **+34%** | 160.9 |
| 1K/128 median TPOT | 78.4 ms | **61.0 ms** | **-22%** | 37.6 |
| 4K/128 throughput | 34.5 tok/s | **46.6** | **+35%** | 53.6 |
| 4K/128 median TPOT | 128.4 ms | **105.9 ms** | **-18%** | 58.1 |
| `Couldn't find tuned sizes` | 8 unique keys | **0** | | |

## Remaining gap (decode op_profile, after this PR)

The FP8-vs-BF16 gap shrinks from ~41 ms to ~23 ms but does **not** close. From an `xprof op_profile` of a tuned decode step:

| op | % of step | note |
|---|---|---|
| `quantized_matmul_kernel_*` | **4.5%** | tuned — no longer the bottleneck |
| `pad` (w_q to `out_block` multiple) | 8.9% | `blockwise_kernel.py:85`, e.g. `[576,6144]→[1024,6144]` |
| `all-reduce.*.remat` (`u32[2,1.5M]`, ×37, 64-chip) | **29.6%** | no provenance; suspected XLA remat of replicated weight resharding — needs BF16 trace to confirm |
| `all-gather` (down_proj replicated fallback) | **1.0%** | negligible — the dequant-on-load follow-up from #1111 is **not** worth doing |

The `all-reduce.remat` cluster (~15 ms/step) is the next thing to chase; not in scope here.

## Notes

- `n_out ∈ {32,128}` are not in the table: the kernel rejects `out_block_size < 256` so no valid candidate exists; these layers stay on the safe-clamped fallback (sub-µs, doesn't show up in profile).
- Entries are inserted at the `keep-sorted` lexical positions (pure +30 diff).
- BF16 path is untouched (no `QuantizedLinear`, no `bfloat16` x_q_dtype lookup).

## Test plan
- [x] `Couldn't find tuned sizes` count: 8 → 0 on GLM-5.1 FP8 server start
- [x] e2e bench 1K/4K cc=8 (numbers above)
- [x] `from ...tuned_block_sizes import TUNED_BLOCK_SIZES` — 488 → 518, no syntax/order break
- [ ] CI lint + unit-test-cpu


zorrofox:
`unit-test-1-tpu (2)` was killed by runner shutdown at 2m46s (passed in 23m32s on the previous push; this commit only changed black formatting). Could a maintainer re-run? `pr-test-finish` is red only because it depends on that job.

benchmark/kernels/quantized_matmul/tune_blockwise_block_sizes.py
```diff
@@ -0,0 +1,184 @@
...
+DEFAULT_N_BATCH = (1, 8, 32, 64, 2048)
+
+OUTPUT_PATH = "tuned_blockwise_block_sizes.txt"
+
+
+def _make_candidates(
+    n_batch: int, n_out: int, n_in: int, block_k: int
+) -> list[tuple[int, int, int]]:
+    """Generate (batch_block, out_block, in_block) candidates."""
+    bb_options = sorted({v for v in (8, 16, 32, 64, 128, 256, n_batch) if v <= max(n_batch, 8)})
```
cjx0709: the list range of [8, 16, 32, 64, 128, 256, n_batch] is according to what, if n_batch like 2048, we will miss the 512, 1024...

benchmark/kernels/quantized_matmul/tune_blockwise_block_sizes.py
```diff
@@ -0,0 +1,184 @@
...
+Tune block sizes for the block-wise quantized matmul kernel.
+
+Sweeps (batch_block_size, out_block_size, in_block_size) for each
+(n_batch, n_out, n_in, x_q_dtype, w_q_dtype) and prints the best config in the
+TUNED_BLOCK_SIZES_RAW dict format used by
+sgl_jax.srt.kernels.quantized_matmul.quantized_matmul_kernels.tuned_block_sizes.
+
+Usage (single TPU host):
+    python -m benchmark.kernels.quantized_matmul.tune_blockwise_block_sizes \
+        --shapes-file shapes.txt --block-k 128
```
cjx0709: can you just give the "shape.txt" for the complete example (in the commits or just in the pr body)

zorrofox: Added `example_shapes_glm51_tp64.txt` in e42d0ffd — the 6 shapes used to produce the entries in this PR (n_out<256 omitted since the kernel rejects out_block_size<256). The docstring usage line now points to it.

benchmark/kernels/quantized_matmul/tune_blockwise_block_sizes.py
zorrofox: Changed to `_pow2_range(8, n_batch)` in e42d0ffd so n_batch=2048 now sweeps {8,16,32,64,128,256,512,1024,2048}. The original cap at 256 was to bound sweep time (each extra bb option multiplies the candidate count), but as a general script it should not skip the gap. For the GLM shapes in this PR the n_batch=2048 winners landed at bb∈{256,2048} so the 30 committed entries are unaffected.

cjx0709 (approved):
LGTM

PR #1198, opened 2026-05-25
perf(sp): generalize threshold-gated reduce-scatter

jimoosciuc:
# perf(sp): generalize threshold-gated reduce-scatter

## Motivation

Generalize the SP work from #952 (grok-2 only) into a clean, model-agnostic implementation, and enable it on Qwen3-MoE / MiMo-V2-Flash / MiMo-V2.5-Pro. Row-parallel reductions become `psum_scatter` along the token dim when the per-device slice clears `TPU_SCATTER_MIN_LOCAL_SIZE` (default 128); below that, automatic DP fallback.

## Modifications

- **Helper unification** (`utils/parallel_utils.py`): `should_scatter` predicate plus `make_reduce_sharding(arr, mesh, *, scatter_dim=0, enable_sp=True)` — single source of SP policy. Works for any ndim and any scatter dim. The threshold and divisibility checks are evaluated against the full scatter world (`data * tensor`), matching the resulting `("data","tensor")` partition spec.
- **`EPMoE` / `MiMoV2Moe`**: `out_sharding` is now a `__call__` kwarg, not init state. The module is a pure translator: it derives `out_specs` from `out_sharding.spec` by mapping the `"tensor"` axis through and dropping `data`/`expert` (mesh translation only). Optional `out_sharding=None` falls back to plain DP (matches `LinearBase` convention, keeps the six not-yet-SP models working unchanged).
- **`forward_batch.get_token_valid_mask(out_sharding=...)`**: pure reshard, no spec introspection inside infrastructure.
- **Models touched**: `qwen3_moe`, `mimo_v2_flash` (also covers `mimo_v2_pro` via inheritance), `grok` (init-time `experts_out_sharding` replaced with per-call `make_reduce_sharding`, picks up threshold gating).
- **`LoRALinear`**: accepts `out_sharding=` kwarg, passes through to `base_layer` and reshards the LoRA delta to the same target.
- **Server arg**: `--enable-sequence-parallel` (default off).
- **Removed unused**: `safe_scatter_sharding`, `safe_scatter_reshard`, init-time `out_sharding` on `LinearBase` / `QuantizedLinear` / `EPMoE` / `MiMoV2Moe`.

## Unit Tests

`python/sgl_jax/test/layers/test_sequence_parallel.py` — **21 / 21 PASS**.

Coverage:
- `TestMakeReduceSharding`: `scatter_dim` parameter, `enable_sp` gating, threshold logic, N-D shapes
- `TestLinearBaseScatter` / `TestQuantizedLinearScatter`: per-call `out_sharding` contract
- `TestEPMoESequenceParallel`: SP vs DP target switching, default-DP fallback
- `TestGrokLayerSequenceParallelWiring`: flag plumbing
- `TestDpSpComposition`: DP+SP combined scatter on `("data","tensor")`

## Accuracy Tests
**Setup**: grok-2 fp8 · TPU v7x · TP=8 · `epmoe` · jax/jaxlib 0.8.1 + libtpu 0.0.30 · evalscope **1.8.0** · sampling `temp=0.5,
  top_p=0.8, top_k=40, max_tokens=4096` · 5 passes per config, averaged.

  | Config | Commit | mean_acc (n=5) | per-run |
  |---|---|---|---|
  | main | `e7c72c22` | **0.444 ± 0.022** | 0.460 / 0.449 / 0.465 / 0.439 / 0.409 |
  | PR#1198 SP-OFF | `4312ec49b` | **0.472 ± 0.022** | 0.500 / 0.485 / 0.455 / 0.475 / 0.444 |
  | PR#1198 SP-ON | `4312ec49b` | **0.484 ± 0.031** | 0.505 / 0.500 / 0.429 / 0.485 / 0.500 |

  **SP-ON (0.484) ≥ SP-OFF (0.472) ≥ main (0.444)** — SP is not a regression (SP-ON vs SP-OFF diff +0.012, within noise).

  ### Reproduction

  Server (SP-ON adds `--enable-sequence-parallel`):

  ```bash
    --model-path /models/grok-2 --tokenizer-path alvarobartt/grok-2-tokenizer \
    --quantization-config-path python/sgl_jax/srt/utils/quantization/configs/fp8_grok.yaml \
    --trust-remote-code --device tpu --load-format auto --dtype bfloat16 \
    --tp-size 8 --random-seed 3 --disable-radix-cache --page-size 128 --skip-server-warmup
  ```

  Eval (run 5× and average):

  ```bash
  pip install evalscope==1.8.0
  evalscope eval --model /models/grok-2 --api-url http://127.0.0.1:30000/v1 --api-key EMPTY \
    --eval-type openai_api --datasets gpqa_diamond --dataset-hub huggingface --eval-batch-size 198 \
    --dataset-args '{"gpqa_diamond": {"dataset_id": "Idavidrein/gpqa", "subset_list": ["gpqa_diamond"], "few_shot_num": 5}}' \
    --generation-config '{"temperature": 0.5, "top_p": 0.8, "top_k": 40, "max_tokens": 4096}'
  ```

### Qwen3-30B-A3B — GSM8K greedy (1319 problems)

Server: TPU v7x-8, `tp=8 dp=1 ep=8`, `--moe-backend fused`, `--page-size 256 --context-length 10240 --chunked-prefill-size 2048 --mem-fraction-static 0.85 --max-running-requests 256 --attention-backend fa --disable-radix-cache`.
Eval: `evalscope eval --datasets gsm8k --eval-batch-size 64 --generation-config '{"max_tokens": 2048, "temperature": 0, "top_p": 1.0}'`.

| | runs | mean | range |
|---|---|---|---|
| upstream main baseline (no SP code) | 0.9386, 0.9447, 0.9416 | **0.9416** | 0.61 pp |
| this PR with `--enable-sequence-parallel` | 0.9439, 0.9424, 0.9416 | **0.9426** | 0.23 pp |
| Δ mean | | **+0.10 pp** | candidate inside baseline noise band |

Even greedy decoding has 0.61 pp run-to-run drift from bf16 reduction-tree nondeterminism. SP candidate distribution sits entirely inside baseline's natural range — **parity**.

### MiMo-V2-Flash — GSM8K stochastic (1319 problems, t=0.8, N=3)

| | runs | mean | range |
|---|---|---|---|
| upstream main baseline | 0.9507, 0.9477, 0.9401 | **0.9462** | 1.06 pp |
| this PR with `--enable-sequence-parallel` | 0.9515, 0.9484, 0.9439 | **0.9479** | 0.76 pp |
| cookbook reference | 0.9401 | — | — |
| Δ mean | | **+0.17 pp** | both distributions bracket cookbook ref |

Stochastic sampling (t=0.8) has wider natural variance than greedy. Candidate distribution sits inside baseline range — **parity**.

### MiMo-V2.5-Pro — AIME25 — DONE (parity, cookbook-strict)

Cookbook config (TP=32 DP=4 EP=32, `--max-running-requests 512 --swa-full-tokens-ratio 0.25 --mem-fraction-static 0.95`).
Eval matches `docs/basic_usage/mimo_v2.5_pro.md` exactly: `evalscope eval --datasets aime25 --eval-batch-size 16 --generation-config '{"temperature":1,"top_p":0.95,"max_tokens":131072,"chat_template_kwargs":{"enable_thinking":true}}'`.

| | AIME2025-I | AIME2025-II | OVERALL |
|---|---|---|---|
| upstream main baseline (no SP) | 12/15 (80.0%) | 15/15 (100%) | **27/30 (90.0%)** |
| this PR with `--enable-sequence-parallel` | 13/15 (86.67%) | 15/15 (100%) | **28/30 (93.34%)** |
| Δ overall | | | **+1/30 (+3.33 pp) — within single-pass sampling noise** |

Both scores cleanly above the published 80% reference. Δ is a single-question difference under stochastic thinking-mode sampling (t=1) — well inside per-pass noise. SP correctness confirmed.

## Benchmarking

PR #952's 18-config sweep (ISL × OSL × concurrency), `random` dataset, sgl-jax `bench_serving`.

Repro command (per config):
```bash
python3 -m sgl_jax.bench_serving \
  --backend sgl-jax --base-url http://127.0.0.1:30000 \
  --dataset-name random --num-prompts $((3*BS)) \
  --random-input-len $ISL --random-output-len $OSL \
  --max-concurrency $BS --random-range-ratio 1 \
  --warmup-requests 0 --tokenizer "$TOKENIZER" \
  --disable-ignore-eos --disable-tqdm
```

### Qwen3-30B-A3B (TPU v7x-8, `tp=8 dp=1 ep=8`, fused MoE)

| ISL | OSL | BS | base out tok/s | cand out tok/s | Δ % | base TTFT med ms | cand TTFT med ms | base ITL med ms | cand ITL med ms |
|-----|-----|-----|----------------|----------------|-----|------------------|------------------|------------------|------------------|
| 1024 | 1 | 8 | 36.9 | 41.3 | **+12.0%** | 202 | 179 | 0.00 | 0.00 |
| 1024 | 1 | 16 | 38.0 | 42.9 | **+12.8%** | 406 | 359 | 0.00 | 0.00 |
| 1024 | 1 | 32 | 38.3 | 43.2 | **+12.9%** | 815 | 719 | 0.00 | 0.00 |
| 1024 | 1 | 64 | 38.6 | 43.5 | **+12.8%** | 1633 | 1443 | 0.00 | 0.00 |
| 1024 | 1 | 128 | 38.7 | 43.7 | **+13.0%** | 3270 | 2888 | 0.00 | 0.00 |
| 1024 | 1 | 256 | 38.7 | 43.8 | **+13.1%** | 6539 | 5777 | 0.00 | 0.00 |
| 1024 | 1024 | 8 | 779.7 | 780.7 | +0.1% | 152 | 137 | 9.98 | 10.02 |
| 1024 | 1024 | 16 | 1244.3 | 1259.6 | +1.2% | 230 | 198 | 12.11 | 12.19 |
| 1024 | 1024 | 32 | 2028.6 | 2047.1 | +0.9% | 407 | 337 | 14.61 | 14.56 |
| 1024 | 1024 | 64 | 3193.1 | 3180.5 | -0.4% | 801 | 684 | 17.97 | 17.92 |
| 1024 | 1024 | 128 | 4643.8 | 4697.2 | +1.1% | 1411 | 1224 | 23.45 | 23.46 |
| 1024 | 1024 | 256 | 6084.2 | 6198.0 | +1.9% | 2719 | 2414 | 34.35 | 34.34 |
| 4096 | 1 | 8 | 9.2 | 10.4 | **+12.5%** | 850 | 755 | 0.00 | 0.00 |
| 4096 | 1 | 16 | 9.3 | 10.5 | **+12.5%** | 1705 | 1514 | 0.00 | 0.00 |
| 4096 | 1 | 32 | 9.3 | 10.5 | **+12.5%** | 3403 | 3021 | 0.00 | 0.00 |
| 4096 | 1 | 64 | 9.3 | 10.5 | **+12.5%** | 6817 | 6055 | 0.00 | 0.00 |
| 4096 | 1 | 128 | 9.4 | 10.5 | **+12.5%** | 13626 | 12103 | 0.00 | 0.00 |
| 4096 | 1 | 256 | 9.4 | 10.6 | **+12.6%** | 27204 | 24157 | 0.00 | 0.00 |
| 4096 | 1024 | 8 | 719.1 | 725.8 | +0.9% | 398 | 409 | 9.88 | 9.84 |
| 4096 | 1024 | 16 | 1104.4 | 1115.1 | +1.0% | 714 | 589 | 12.13 | 12.21 |
| 4096 | 1024 | 32 | 1719.0 | 1775.4 | +3.3% | 1466 | 1355 | 14.60 | 14.65 |
| 4096 | 1024 | 64 | 2490.1 | 2577.2 | +3.5% | 2662 | 2323 | 18.05 | 18.11 |
| 4096 | 1024 | 128 | 3324.8 | 3471.6 | +4.4% | 5325 | 5172 | 23.71 | 23.73 |
| 4096 | 1024 | 256 | 4001.4 | 4205.8 | +5.1% | 11013 | 9856 | 34.99 | 34.98 |
| 8192 | 1 | 8 | 4.4 | 4.9 | **+11.5%** | 1812 | 1622 | 0.00 | 0.00 |
| 8192 | 1 | 16 | 4.4 | 4.9 | **+11.6%** | 3621 | 3243 | 0.00 | 0.00 |
| 8192 | 1 | 32 | 4.4 | 4.9 | **+11.6%** | 7255 | 6495 | 0.00 | 0.00 |
| 8192 | 1 | 64 | 4.4 | 4.9 | **+11.7%** | 14518 | 12999 | 0.00 | 0.00 |
| 8192 | 1 | 128 | 4.4 | 4.9 | **+11.7%** | 28994 | 25957 | 0.00 | 0.00 |
| 8192 | 1 | 256 | 4.4 | 4.9 | **+11.7%** | 57871 | 51795 | 0.00 | 0.00 |
| 8192 | 1024 | 8 | 641.1 | 651.6 | +1.6% | 933 | 842 | 10.64 | 10.67 |
| 8192 | 1024 | 16 | 919.4 | 938.5 | +2.1% | 1492 | 1636 | 13.21 | 13.21 |
| 8192 | 1024 | 32 | 1295.2 | 1336.0 | +3.2% | 2999 | 2899 | 16.78 | 16.83 |
| 8192 | 1024 | 64 | 1694.9 | 1776.8 | +4.8% | 6417 | 5545 | 22.59 | 22.63 |
| 8192 | 1024 | 128 | 2027.3 | 2141.3 | +5.6% | 12437 | 11461 | 32.69 | 32.70 |
| 8192 | 1024 | 256 | 2266.5 | 2401.4 | +6.0% | 24511 | 21433 | 52.88 | 52.92 |

Summary: prefill (OSL=1) consistently +11.5% to +13.1% across all 18 configs; mixed (OSL=1024) +0.1% to +6.0% (favors larger ISL × larger BS); decode median ITL identical within ±0.05 ms — no regression.

### MiMo-V2.5-Pro (TPU v7x 4×4 multi-host, cookbook `tp=32 dp=4 ep=32`, fused MoE)

Server: `--page-size 256 --context-length 262144 --chunked-prefill-size 4096 --mem-fraction-static 0.95 --swa-full-tokens-ratio 0.25 --max-running-requests 512`.
Spot config (ISL=4096 OSL=1 BS=64, cookbook DP=4): SP candidate input throughput = **17547 tok/s**, prefill speedup direction matches Qwen3 (prefill-heavy SP win). Full 18-config sweep on cookbook DP=4 omitted from this PR to keep the bench surface compact; see PR follow-up if a detailed table is needed.

Decode median ITL identical to baseline within ±0.05 ms — no regression.

## Checklist

- [x] English
- [x] PR purpose and motivation clear
- [x] Test plan included (unit tests, accuracy distributions, bench tables)
- [ ] Documentation update (cookbooks already document `--enable-sequence-parallel`)

### DIFF
--- a/python/sgl_jax/srt/layers/fused_moe.py
+++ b/python/sgl_jax/srt/layers/fused_moe.py
@@ -8,6 +8,7 @@ from jax.sharding import PartitionSpec as P
 
 from sgl_jax.srt.eplb.expert_location import get_global_expert_location_metadata
 from sgl_jax.srt.kernels.fused_moe.v1.kernel import FusedMoEBlockConfig, fused_ep_moe
+from sgl_jax.srt.utils.parallel_utils import make_reduce_sharding
 from sgl_jax.srt.utils.quantization.quantization_utils import quantize_tensor
 
 
@@ -441,15 +442,12 @@ class FusedEPMoE(nnx.Module):
         hidden_states: jax.Array,
         topk_weights: jax.Array,
         topk_ids: jax.Array,
-        token_valid_mask: jax.Array | None = None,
         *,
         block_config: FusedMoEBlockConfig | None = None,
         out_sharding: jax.sharding.Sharding | None = None,
     ) -> jax.Array:
         """Forward pass through the fused MoE layer."""
         assert hidden_states.ndim == 2
-        if token_valid_mask is not None:
-            topk_ids = jnp.where(token_valid_mask[:, None], topk_ids, -1)
 
         w1_shared_val = self.w1_shared.value if self.w1_shared is not None else None
         w3_shared_val = self.w3_shared.value if self.w3_shared is not None else None
@@ -507,6 +505,7 @@ class FusedEPMoE(nnx.Module):
             tp_axis_name="tensor",
         )
 
-        if out_sharding is not None:
-            output = jax.sharding.reshard(output, out_sharding)
+        if out_sharding is None:
+            out_sharding = make_reduce_sharding(hidden_states, self.mesh, enable_sp=False)
+        output = jax.sharding.reshard(output, out_sharding)
         return output
--- a/python/sgl_jax/srt/layers/fused_moe.py
+++ b/python/sgl_jax/srt/layers/fused_moe.py
@@ -3,7 +3,7 @@
 import jax
 from flax import nnx
 from jax import numpy as jnp
-from jax.sharding import Mesh, NamedSharding
+from jax.sharding import Mesh
 from jax.sharding import PartitionSpec as P
 
 from sgl_jax.srt.eplb.expert_location import get_global_expert_location_metadata
@@ -441,22 +441,15 @@ class FusedEPMoE(nnx.Module):
         hidden_states: jax.Array,
         topk_weights: jax.Array,
         topk_ids: jax.Array,
+        token_valid_mask: jax.Array | None = None,
         *,
         block_config: FusedMoEBlockConfig | None = None,
+        out_sharding: jax.sharding.Sharding | None = None,
     ) -> jax.Array:
-        """
-        Forward pass through the fused MoE layer.
-
-        Args:
-            hidden_states: Input tokens, shape (num_tokens, hidden_size) or
-                          (batch_size, seq_len, hidden_size)
-            topk_weights: Pre-computed top-k weights
-            topk_ids: Pre-computed top-k expert IDs
-
-        Returns:
-            MoE layer output, same shape as hidden_states
-        """
+        """Forward pass through the fused MoE layer."""
         assert hidden_states.ndim == 2
+        if token_valid_mask is not None:
+            topk_ids = jnp.where(token_valid_mask[:, None], topk_ids, -1)
 
         w1_shared_val = self.w1_shared.value if self.w1_shared is not None else None
         w3_shared_val = self.w3_shared.value if self.w3_shared is not None else None
@@ -514,5 +507,6 @@ class FusedEPMoE(nnx.Module):
             tp_axis_name="tensor",
         )
 
-        output = jax.sharding.reshard(output, NamedSharding(self.mesh, P("data", None)))
+        if out_sharding is not None:
+            output = jax.sharding.reshard(output, out_sharding)
         return output
--- a/python/sgl_jax/srt/layers/fused_moe.py
+++ b/python/sgl_jax/srt/layers/fused_moe.py
@@ -8,7 +8,6 @@ from jax.sharding import PartitionSpec as P
 
 from sgl_jax.srt.eplb.expert_location import get_global_expert_location_metadata
 from sgl_jax.srt.kernels.fused_moe.v1.kernel import FusedMoEBlockConfig, fused_ep_moe
-from sgl_jax.srt.utils.parallel_utils import make_reduce_sharding
 from sgl_jax.srt.utils.quantization.quantization_utils import quantize_tensor
 
 
@@ -506,6 +505,6 @@ class FusedEPMoE(nnx.Module):
         )
 
         if out_sharding is None:
-            out_sharding = make_reduce_sharding(hidden_states, self.mesh, enable_sp=False)
+            out_sharding = jax.sharding.NamedSharding(self.mesh, P(*([None] * output.ndim)))
         output = jax.sharding.reshard(output, out_sharding)
         return output
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -16,7 +16,6 @@ from sgl_jax.srt.kernels.gmm.megablox_gmm_backend import gmm
 # Re-export for backward compatibility: external code imports from this module.
 from sgl_jax.srt.layers.fused_moe import FusedEPMoE  # noqa: F401
 from sgl_jax.srt.layers.gate import GateLogit, TopK  # noqa: F401
-from sgl_jax.srt.utils.parallel_utils import make_reduce_sharding
 from sgl_jax.srt.utils.profiling_utils import named_scope
 from sgl_jax.srt.utils.quantization.quantization_utils import (
     quantize_tensor,
@@ -412,12 +411,8 @@ class EPMoE(nnx.Module):
         *,
         out_sharding: jax.sharding.NamedSharding | None = None,
     ) -> jax.Array:
-        # Default to plain-DP target when caller does not express intent,
-        # mirroring LinearBase's "no out_sharding => standard TP" convention.
-        # SP-aware callers pass an explicit out_sharding (typically built via
-        # make_reduce_sharding) to opt in.
         if out_sharding is None:
-            out_sharding = make_reduce_sharding(hidden_states, self.mesh, enable_sp=False)
+            out_sharding = jax.sharding.NamedSharding(self.mesh, P(*([None] * hidden_states.ndim)))
         # Translate the caller's target sharding (on self.mesh: data,tensor)
         # into shard_map out_specs (on self.moe_mesh: expert,tensor). Only
         # 'tensor' is shared between the two meshes; everything else is
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -1,5 +1,6 @@
 """GMM-based Expert-Parallel MoE layer and weight mapping utilities."""
 
+import math
 from functools import partial
 
 import jax
@@ -15,7 +16,7 @@ from sgl_jax.srt.kernels.gmm.megablox_gmm_backend import gmm
 # Re-export for backward compatibility: external code imports from this module.
 from sgl_jax.srt.layers.fused_moe import FusedEPMoE  # noqa: F401
 from sgl_jax.srt.layers.gate import GateLogit, TopK  # noqa: F401
-from sgl_jax.srt.utils.parallel_utils import should_scatter
+from sgl_jax.srt.utils.parallel_utils import make_reduce_sharding
 from sgl_jax.srt.utils.profiling_utils import named_scope
 from sgl_jax.srt.utils.quantization.quantization_utils import (
     quantize_tensor,
@@ -40,7 +41,6 @@ class EPMoE(nnx.Module):
         quantization_config=None,
         physical_to_logical_map: "jax.Array | None" = None,
         pre_gather_quant_dtype=None,
-        enable_sequence_parallel=False,
     ):
         self.num_experts_per_tok = num_experts_per_tok
         self.physical_to_logical_map = physical_to_logical_map
@@ -61,7 +61,6 @@ class EPMoE(nnx.Module):
         self.mesh = mesh
         self.activation = activation
         self.hidden_size = hidden_size
-        self.enable_sequence_parallel = enable_sequence_parallel
 
         # Get quantization settings from config
         self.quantized_dtype = (
@@ -78,7 +77,7 @@ class EPMoE(nnx.Module):
             raise ValueError(
                 f"num_experts({self.num_experts}) must be divisible by ep_size ({self.ep_size})"
             )
-        world_size = self.mesh.shape.get("data", 1) * mesh.shape.get("tensor", 1)
+        world_size = math.prod(self.mesh.shape.values())
         self.tp_size = world_size // self.ep_size
         self.experts_per_device = self.num_experts // self.ep_size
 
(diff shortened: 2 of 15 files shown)


python/sgl_jax/srt/layers/fused_moe.py
```diff
@@ -441,22 +441,15 @@ def __call__(
         hidden_states: jax.Array,
         topk_weights: jax.Array,
         topk_ids: jax.Array,
+        token_valid_mask: jax.Array | None = None,
         *,
         block_config: FusedMoEBlockConfig | None = None,
+        out_sharding: jax.sharding.Sharding | None = None,
```
cjx0709, line 446: I noticed EPMoE in moe.py falls back like this when out_sharding is None:
```
if out_sharding is None:
  out_sharding = make_reduce_sharding(hidden_states, self.mesh, enable_sp=False)
```
Should fused_moe do something similar here as well, or is the intended default behavior different from EPMoE?

python/sgl_jax/srt/models/qwen3_moe.py
```diff
@@ -286,40 +270,54 @@ def __call__(
...
         hidden_states = self.post_attention_layernorm(hidden_states)
 
+        topk_ids = None
         if self.is_moe_layer:
             router_logits = self.moe_gate(hidden_states)
             topk_weights, topk_ids = self.topk(router_logits, dispatch_info=dispatch_info)
-
             if self.use_fused:
-                token_valid_mask = forward_batch.get_token_valid_mask(hidden_states.shape[0])
-                topk_ids = jnp.where(token_valid_mask[:, None], topk_ids, -1)
```
cjx0709, line 314: Since this line was deleted, the topk_ids returned at line 320 is now unmasked. If any downstream consumer relies on the masked topk_ids contract, this could lead to incorrect behavior.

python/sgl_jax/srt/layers/fused_moe.py
jimoosciuc: Good catch — fixed in 2890298c6. FusedEPMoE now falls back to `make_reduce_sharding(enable_sp=False)` when `out_sharding` is None, matching EPMoE's convention so both backends share identical default behavior.

python/sgl_jax/srt/models/qwen3_moe.py
jimoosciuc: You're right, this was a real bug. Fixed in 2890298c6 by moving the `jnp.where` mask back into the model layer (qwen3_moe.py and mimo_v2_flash.py) so the masked `topk_ids` is what propagates to downstream consumers. The SP refactor had pushed the mask op into `fused_moe.__call__` where it operated on a local copy, leaving the caller's `topk_ids` unmasked at return. Removed the now-unused `token_valid_mask` kwarg from `fused_moe.__call__`.

python/sgl_jax/srt/layers/moe.py
```diff
@@ -405,22 +403,27 @@ def _get_block_size_k(
...
+    def __call__(
+        self,
+        hidden_states,
+        topk_weights,
+        topk_ids,
+        *,
+        out_sharding: jax.sharding.NamedSharding | None = None,
+    ) -> jax.Array:
+        if out_sharding is None:
+            out_sharding = jax.sharding.NamedSharding(self.mesh, P(*([None] * hidden_states.ndim)))
```
cjx0709, line 415: here use `P(*([None] * hidden_states.ndim)` as the fallback, the old code use: 
``` 
       token_axis = ("data", "tensor") if do_scatter else "data"
       out_pspec = P(token_axis, *([None] * (result.ndim - 1)))
``` as fallback
this change the origin dp shard output to full replicated shard output. Are there any special design considerations here?

File: python/sgl_jax/srt/layers/fused_moe.py [hunk 4]
```diff
@@ -514,5 +504,7 @@ def __call__(
             tp_axis_name="tensor",
         )
 
-        output = jax.sharding.reshard(output, NamedSharding(self.mesh, P("data", None)))
+        if out_sharding is None:
+            out_sharding = jax.sharding.NamedSharding(self.mesh, P(*([None] * output.ndim)))
```
cjx0709, line 508: same as the "moe.py"

python/sgl_jax/srt/layers/moe.py
jimoosciuc: Intentional — fixed/clarified in 21d773e5b.

For legacy callers that don't pass `out_sharding` (the 7 unchanged models: bailing_moe, deepseek_v3, glm4_moe, glm5_moe, kimi_linear, qwen2_moe and variants) I deliberately restore main's **pre-SP** behavior of producing `P(None, None)` (fully replicated) rather than the `P(token_axis, ...)` dp-sharded layout. The reason is to avoid silently assuming JAX auto-reshard handles the layout mismatch correctly for callers that haven't opted into SP.

SP-aware callers (qwen3_moe / mimo_v2_flash / grok) always pass an explicit `out_sharding` (built via `make_reduce_sharding`), so they're unaffected by this default.

Validated on DeepSeek-V2-Lite-Chat at dp=2 tp=8 ep=8 with both fallbacks: GSM8K 0.6376 (replicated) vs 0.6406 (sharded) — within sampling noise. The explicit choice is to not rely on auto-reshard for the unchanged callers.

python/sgl_jax/srt/layers/fused_moe.py
jimoosciuc: Same as the moe.py thread above — intentional, see 21d773e5b. Legacy callers (no explicit `out_sharding`) get main's pre-SP fully-replicated `P(None, None)`; SP-aware callers pass an explicit `out_sharding` and are unaffected.

JamesBrianD (approved):

PR #1211, opened 2026-05-25
feat(qwen3-vl): three-JIT split for ViT + LLM + precompile

JamesBrianD:
## Summary

Splits Qwen3-VL's monolithic JIT into three independent JIT boundaries (ViT encode / embed splice / LLM forward), mirroring tpu-inference's `embed_multimodal_fn` / `embed_input_ids_fn` / `model_fn` design and sglang torch's piecewise CUDA graph runner pattern. Adds AOT precompile coverage for the new JITs so the first multimodal request no longer triggers a multi-minute cold compile.

Stacked on PR #1189. See [`docs/design/qwen3vl_three_jit_split.md`](docs/design/qwen3vl_three_jit_split.md) for the full design.

## What changed

| # | Commit | What |
|---|---|---|
| 1 | `5c05d59` | `compute_patch_buckets()` derives a power-of-two ladder from `chunked_prefill_size * spatial_merge_size^2`, replacing the hardcoded `(256, 1024, 4096)` |
| 2 | `c8b9aeb` | Adds `encode_visual` / `splice_embeds` methods on `Qwen3VLForConditionalGeneration`; inline ViT block in `__call__` becomes a fallback for callers that haven't switched to the orchestrator |
| 3 | `89e5852` | `ModelRunner` adds `jitted_visual_encode` / `jitted_splice_embeds`; `run_model_wrapper` orchestrates ViT → splice → LLM with `dataclasses.replace` |
| 4 | `ddffcfd` | `CompilationManager` adds `_precompile_visual_encode` (patches-only ladder, n_real_images=1) and `_precompile_splice` (Cartesian seq_len × N_padded_tokens, skip impossible pairs) |
| 5 | `283ea64` | Patch bucket install moved from scheduler to model_runner (detection by `hasattr(model, 'encode_visual')` instead of `--multimodal` CLI flag) |
| 6 | `cf2384c3` | Design doc |

## Three-JIT cache key isolation

| JIT | Cache key dimensions | Variance source |
|---|---|---|
| `jitted_visual_encode` | `(n_patches_bucket, image_grid_thw, n_real_images_bucket)` | ViT inputs only |
| `jitted_splice_embeds` | `(seq_len_bucket, N_padded_tokens_bucket)` | Splice geometry |
| `jitted_run_model` (LLM) | `(seq_len_bucket, ...)` — **same as text-only path** | No image axes |

After this PR, the LLM JIT is fully shared between text-only and multimodal extends. `pixel_values` and `N_padded_tokens` no longer leak into the LLM JIT cache key.

## Verification

Server start with the user's verification command:

```
    --model-path /models/Qwen3-VL-8B-Instruct --trust-remote-code \
    --tp-size 4 --context-length 8192 --page-size 128 \
    --disable-radix-cache
```

(No `--disable-precompile`, no `--multimodal` flag — both validate that the new auto-detection works.)

Observed startup log:

```
Multimodal patch bucket ladder set to (16, 32, 64, 128, 256, 512, 1024,
  2048, 4096, 8192, 16384) (chunked_prefill_size=4096, spatial_merge_size=2)
[EXTEND]         Precompile finished in 29 secs
[DECODE]         Precompile finished in 89 secs
[VISUAL_ENCODE]  Precompile finished in 79 secs   ← NEW (11 buckets)
[SPLICE]         Precompile finished in 38 secs   ← NEW (21 pairs)
The server is fired up and ready to roll!
```

Total added startup precompile time: **~2 minutes** for full coverage of every in-spec image bucket.

Smoke test passed: single multimodal request returns coherent output describing the test image. Full MMMU val eval (lmms-eval, 900 samples, batch_size=128, max_new_tokens=4096) running at PR time — will update with score in a follow-up comment.

## Follow-ups (S3 — separate PRs)

The encoder cache / chunked-prefill correctness work described in the design doc §7 lands separately. Tracked here as a checklist:

- [ ] **Per-image embedding cache** — `MultiModalEmbeddingCache` keyed on `mm_input.hash`. Hook between `jitted_visual_encode` output and `jitted_splice_embeds` input: skip the ViT JIT when the image hash hits the cache.
- [ ] **Three-JIT cache integration** — wire the encoder cache into the orchestrator. Batch all cache-miss images across the entire batch into a single `jitted_visual_encode` call; per-request cache lookups feed the splice JIT.
- [ ] **`items_offset` + chunked-prefill correctness** — port sglang upstream's `items_offset = [(start, end), ...]` per image, plus `find_chunk_items_and_check_cache` (`end >= chunk_start AND start < chunk_end`) and `assemble_chunk_embedding` (`overlap_start`/`overlap_end` + `local_start`/`local_end` slicing). Fixes cross-chunk image splice and enables cross-chunk ViT reuse via cache.
- [ ] **`offload_mm_features_to_cpu`** + clear `forward_batch.mm_inputs` after ViT — release device memory of `pixel_values` once vision embedding is computed, keep CPU copy for retraction.
- [ ] **Retraction support** — when a request is retracted, restore `mm_inputs` from CPU copy so re-prefill does not re-preprocess images.
- [ ] **Apply same split to `qwen2_5_vl` and `qwen3_omni_thinker`** — once the Qwen3-VL pattern is validated, mirror the split into the other VLM model classes.

## Test plan

- [x] Unit: `compute_patch_buckets()` for representative `(chunked_prefill_size, spatial_merge_size)` — 7/7 cases pass.
- [x] Startup: precompile log shows 11 visual buckets + 21 splice pairs compiled in ~2 minutes.
- [x] Smoke: single multimodal request returns coherent output.
- [ ] **Regression**: `test/srt/multimodal/test_qwen3_vl_models.py` (`test_logit_alignment_vs_hf`, `test_multi_image_prefill`) — TODO after this PR description is up; will note pass/fail in a follow-up comment.
- [ ] **End-to-end**: full MMMU val (~900 samples) with `lmms-eval` — running now, will report score.

### DIFF
--- a/docs/design/qwen3vl_three_jit_split.md
+++ b/docs/design/qwen3vl_three_jit_split.md
@@ -1,0 +1,349 @@
+# Qwen3-VL Three-JIT Split — Design Document
+
+## 1. Motivation
+
+Today the entire `Qwen3VLForConditionalGeneration.__call__` — vision encoder, embed splice, and language model — lives inside a single JIT closure (`jitted_run_model` in `model_runner.py:211`). Two consequences:
+
+1. **`pixel_values` shape participates in the JIT cache key.** Different image bucket sizes (`_MM_PATCH_BUCKETS = (256, 1024, 4096)`) each force a fresh recompilation of the *entire* 7B+ViT model graph.
+2. **No precompile coverage for the vision path.** `CompilationManager` only synthesizes text-only dummy batches, so any user serving an image hits a multi-minute cold compile on the first request of each bucket.
+
+Both the upstream sglang torch runtime and the tpu-inference JAX runtime solve this by splitting the multimodal forward into independent JIT boundaries:
+- sglang torch only captures the language model in CUDA graphs; ViT runs eager and feeds `input_embeds` into the captured graph.
+- tpu-inference compiles three independent functions (`embed_multimodal_fn`, `embed_input_ids_fn`, `model_fn`) and AOT-precompiles all power-of-two patch buckets at startup.
+
+This document describes how sgl-jax adopts the three-JIT pattern for Qwen3-VL, with the bucket and precompile machinery aligned to the tpu-inference design. The work lands as a new PR independent of #1189 (which contains the ViT splash kernel and Control-plane review fixes).
+
+## 2. Requirements and Non-Goals
+
+### Functional Requirements
+
+- Vision encoder runs in its own JIT boundary, keyed on `n_patches_bucket` only.
+- LLM JIT cache is shared between text-only and multimodal extends (no `pixel_values` in cache key).
+- Patch buckets are powers of two derived from `chunked_prefill_size * spatial_merge_unit`, replacing the hardcoded `(256, 1024, 4096)` ladder.
+- Precompile coverage at startup for every power-of-two patch bucket, eliminating cold-compile cost on first image request.
+- No regression on text-only paths or decode paths.
+- No regression on existing tests: `test_qwen3_vl_models.py` (`test_logit_alignment_vs_hf`, `test_multi_image_prefill`).
+
+### Non-Goals (S3 follow-ups, separate PR)
+
+- **Per-image embedding cache** (content-hash keyed, cross-request reuse).
+- **Chunked-prefill `items_offset` + `find_chunk_items_and_check_cache` + `assemble_chunk_embedding`** algorithm for cross-chunk image reuse.
+- **`offload_mm_features_to_cpu`** for GPU memory release after ViT.
+- **Retraction support** from CPU-side `mm_inputs` copy.
+- **Applying the split to `qwen2_5_vl` / `qwen3_omni_thinker`** (Qwen3-VL only for this PR).
+- **Three-JIT encoder cache integration** — depends on the per-image cache landing first.
+
+## 3. Architecture
+
+### Current (Monolithic JIT)
+
+```
+ForwardBatch (pixel_values, image_grid_thw, input_ids, ...)
+       │
+       ▼
+┌────────────────────────────────────────────────────────┐
+│ jitted_run_model (single JIT, ~7B + 700M params)       │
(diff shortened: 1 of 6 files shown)


JamesBrianD:
## Status update

Pushed `d38181e8` which fixes a critical bug in the orchestrator — the original `dataclasses.replace` only populated `input_embedding` / `deepstack_visual_embedding` but left the per-request `pixel_values`, `image_grid_thw`, `cu_seqlens`, `n_real_images`, `placeholder_positions` fields attached to forward_batch. That made the LLM JIT cache key vary per image bucket, defeating the design goal of sharing the LLM JIT cache with text-only extends.

After the fix the LLM JIT cache is properly shared:

| Metric | Before fix (with three-jit) | After fix |
|---|---|---|
| Sequential warm request | ~33s | **0.5s** |
| 8 concurrent wall time | 86s | 68s |
| LLM JIT compiles per request | 1 | **0 (warm)** |

`JAX_LOG_COMPILES=1` instrumentation showed `jitted_run_model` was being compiled per request before the fix; after the fix it compiles once during precompile + once when first multimodal request arrives (new pytree shape for the input_embedding leaf), then stays cached.

## Known follow-ups (not blocking this PR)

- [ ] **Decode `cache_miss=1` per step remains** — not from `jitted_run_model` anymore (verified via JAX_LOG_COMPILES; LLM JIT is warm). Likely the sampler JIT or `compute_logprobs` JIT picking up a runtime shape that doesn't match the precompile dummies. Worth a separate investigation but doesn't undo the main win of this PR.
- [ ] **Multimodal LLM JIT cache key still differs from text-only** — because text-only extends pass `input_embedding=None` while multimodal extends pass `input_embedding=jax.Array`. Two cache entries instead of one. To collapse fully, the multimodal precompile would need to also exercise `jitted_run_model` with a dummy `input_embedding` leaf populated. Marginal benefit (one extra compile at startup), low priority.

Full MMMU val (~900 samples) running now to confirm no accuracy regression. Will report score here.

Issue #1216, opened 2026-05-26
ci: test_multi_engines_in_one_process.py 100% fails on e2e-test-4-tpu shard


pengchengneo:
## Summary

`test/srt/rl/test_multi_engines_in_one_process.py` consistently fails with exit code -9 (OOM kill) on the `e2e-test-4-tpu (2)` CI shard. This has been broken on **main** since 2026-05-25, blocking all PRs.

## Evidence

- **Main branch**: 5 consecutive failures since 5/25 (runs 26397388148, 26398522342, 26406296699, 26429278031, 26430962483), all with the same `e2e-test-4-tpu (2)` failure. Last green main run was 5/22 (run 26288484003).
- **PR #1192**: same failure, multiple rerun attempts all fail on this test.

## Root cause

The test runs as shard 4/4 in the `e2e-test-tpu-v6e-4` suite (after `test_features.py`, `test_deepseek_v2_lite_models.py`, `test_engine_pause_continue.py`, `test_model_loader.py`). When the preceding test (`test_model_loader.py`) completes quickly (~58s vs ~118s previously), its TPU subprocess may not fully exit before `test_multi_engines` starts. The second test then hits:

```
ABORTED: The TPU is already in use by process with pid XXXXX. Not attempting to load libtpu.so in this process.
```

or gets OOM-killed (exit -9) within ~31 seconds.

On the 5/22 green run, `test_model_loader.py` took 118s (leaving enough time for TPU cleanup); since 5/25, it takes ~58s (likely due to XLA cache hits), and the TPU process hasn't released resources before the next test starts.

`run_suite.py` calls `process.wait()` + `cleanup_model_cache()` between tests, but neither waits for child TPU worker processes to fully release libtpu.

## Suggested fixes (pick one)

1. **Add TPU process wait**: after `cleanup_model_cache()` in `run_suite.py`, wait for all libtpu-holding processes to exit before starting the next test.
2. **Increase estimated time**: change `TestFile("test/srt/rl/test_multi_engines_in_one_process.py", 5)` to a much larger value (e.g., 600) so bin-packing isolates it into its own shard.
3. **Add explicit `kill_process_tree`** between test files, not just on timeout.

## Affected CI jobs

- `e2e-test-4-tpu (2)` — 100% failure rate since 5/25
- `unit-test-1-tpu (2)` — intermittent (sometimes passes depending on runner state)
- `pr-test-finish` — fails as a consequence (gate job)

[...]

PR #1220, opened 2026-05-26
test(rl): bypass jit on KV cache alloc to isolate persistent cache trigger

Rodrian7:
## Context

Issue: #1216 (`test_multi_engines_in_one_process.py` crashes with `FAILED_PRECONDITION: The program continuator has halted unexpectedly` after `JAX_COMPILATION_CACHE_DIR=/xla-cache` was added in 5/22).

Prior experiment: 82b8e68 verified that `disable_precompile=True` alone does **not** fix the crash. Engine A finishes initialization fine, but Engine B crashes during weight loading slow-path:

```
weight_utils.py:2751 jax.device_put(weight, target_sharding)
  → array.py:1267 _array_shard_arg → shard_sharded_device_array_slow_path
  → array.py:1198 x._value
  → array.py:635 _single_device_array_to_np_array_did_copy
  → FAILED_PRECONDITION
```

Notable asymmetry: Engine B's first 5 fast-path weights write successfully to devices [2,3] (host→device), but the first slow-path weight crashes on device→host DMA. This indicates TPU runtime state on devices [2,3] was corrupted **before** Engine B started.

## Hypothesis

After `disable_precompile=True`, the only `jax.jit(...)()` that **actually executes** on the init path (not merely defined) is the KV cache allocation in `mem_cache/memory_pool.py:445`:

```python
with self.mesh:
    for _ in range(self.layer_num):  # 28 layers
        kv_buf = jax.jit(
            lambda: jnp.zeros(shape=fused_buffer_shape, dtype=self.dtype),
            out_shardings=self.kv_sharding,
        )()
```

Evidence this hits the persistent cache:
- CI log: `Total time to create 28 buffers: 0.25 seconds` — too fast for fresh compilation
- After 5/22 commit `de29d9f0`, the `/xla-cache` GCS Fuse mount accumulates compiled executables across runs

`NamedSharding(self.mesh, ...)` constrains where the *executable* runs, but cannot prevent the persistent cache deserialization path (`backend.deserialize_executable`) from registering host-level resources in libtpu (e.g., program continuator slots), which Engine B's later device→host DMA depends on.

## Change

Replace `jax.jit(lambda: jnp.zeros(...), out_shardings=...)()` with `jax.device_put(jnp.zeros(...), sharding)`. This still allocates the same sharded buffer but bypasses the jit/cache deserialization path entirely. Also adds `disable_precompile=True` to the test so this is the only diff vs. the failing-cached scenario.

Engine A's slow-path weight loading already calls `jax.device_put` successfully, so this rewrite should not introduce a new crash by itself.

## Expected CI Outcome

| Result | Conclusion |
|--------|------------|
| `e2e-test-tpu-v6e-4` passes | Persistent cache deserialization of the KV cache jit is the trigger; root cause confirmed |
| Still `FAILED_PRECONDITION` | Some other jit/op on the init path is also hitting the cache — need further investigation |

## Scope

- This PR is **diagnostic only** and will be reverted after the CI result is observed.
- Memory impact of the rewrite is negligible (~0.01 GB per-layer host-side `jnp.zeros` for Qwen3-1.7B).
- No production behavior change is intended.

### DIFF
--- a/python/sgl_jax/srt/mem_cache/memory_pool.py
+++ b/python/sgl_jax/srt/mem_cache/memory_pool.py
@@ -442,13 +442,15 @@ class MHATokenToKVPool(KVCache):
         with self.mesh:
             self.kv_buffer = []
             for _ in range(self.layer_num):
-                kv_buf = jax.jit(
-                    lambda: jnp.zeros(
-                        shape=fused_buffer_shape,
-                        dtype=self.dtype,
-                    ),
-                    out_shardings=self.kv_sharding,
-                )()
+                # Diagnostic: replace jax.jit(...)() with jax.device_put to bypass
+                # JAX persistent compilation cache deserialization. If
+                # test_multi_engines_in_one_process passes with this change, the
+                # FAILED_PRECONDITION crash is caused by the cached executable
+                # deserialization path of this jit, not by the allocation itself.
+                kv_buf = jax.device_put(
+                    jnp.zeros(shape=fused_buffer_shape, dtype=self.dtype),
+                    self.kv_sharding,
+                )
 
                 self.kv_buffer.append(kv_buf)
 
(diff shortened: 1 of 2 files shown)


Rodrian7:
Superseded by #1221 (same diff, opened from upstream branch so CI workflows actually trigger).

PR #1233, opened 2026-05-28
fix(scheduler): align TEST_RETRACT with upstream sglang

### LINKED ISSUE #1229: ci: test_retract_decode.py stuck in retract loop on e2e-test-4-tpu (1) — hangs to job timeout or runs ~87× expected time
## Summary

`test/srt/test_retract_decode.py` enters a forced retract → re-prefill loop under `SGLANG_TEST_RETRACT=1`. With small steady-state running batches the loop never makes forward progress, causing the test to either (a) take ~87× the expected wall time when prefill admission happens to keep up, or (b) hang indefinitely until the CI job is killed by `timeout-minutes`. This is the first test file in the `e2e-test-4-tpu (1)` shard, with no prior test to leak from — the problem is the `TEST_RETRACT` mechanism itself.

## Regression commit

**PR #994** (`5b7150bd1ba409e8d6de554f1a0dd707384970ae`, Brian, 2026-05-02, *"Unify memory management across (overlap, non-overlap) x (page>=1) x (retract, finished)"*) replaced the trigger condition in `python/sgl_jax/srt/managers/scheduler.py`:

```diff
-        if not batch.check_decode_mem(self.decode_mem_cache_buf_multiplier) or (
-            TEST_RETRACT and batch.batch_size() > 10
-        ):
+        if not batch.check_decode_mem() or (
+            TEST_RETRACT and self.forward_ct % TEST_RETRACT_INTERVAL == 0
+        ):
```

The `batch.batch_size() > 10` guard — which `test_retract_decode.py:3` still documents as the contract:
> `SGLANG_TEST_RETRACT=1 forces retract on batch_size > 10.`

— was removed. The replacement (`forward_ct % TEST_RETRACT_INTERVAL == 0`, default interval 3) fires every third forward step regardless of running batch size. The PR also added `TEST_RETRACT_NO_PREFILL_BS` (default `2**31`, i.e. effectively disabled) at `scheduler.py:1515`, which is the *only* remaining batch-size-related `TEST_RETRACT` guard but never trips at default settings.

`scheduler.py` has had only two commits since #994 (#1016 GLM tooling, #1063 EAGLE restore); neither touches `TEST_RETRACT`. So the regression is unambiguously at #994.

## Evidence

- **PR #1213** run [26493978853](https://github.com/sgl-project/sglang-jax/actions/runs/26493978853/job/78028593028) (attempt 2) — clear hang. The `Run e2e test` step is killed by GHA at the 120-min `timeout-minutes` mark (07:29:51 → 09:30:03). `grep -c "KV cache pool is full"` on the full job log returns **62160 lines**, with **zero** `200 OK` responses across the entire 2 hours. Only `Begin (0/4):` for `test_retract_decode.py` ever fires; `Begin (1/4):` is never reached.

- **PR #1215** run [26496873905](https://github.com/sgl-project/sglang-jax/actions/runs/26496873905/job/78031029935) — slow but eventually passes. Between `Begin (0/4):` and `End (0/4):` the test_retract process prints:
  ```
  100%|██████████| 64/64 [01:26<00:00, 1.56s/it]      ← MMLU 64/64 completed
  Ran 3 tests in 1023.638s
  OK (skipped=2)
  'score': np.float64(0.703125)                       ← 0.703 ≥ 0.5 threshold
  Total latency: 86.390 s
  End (0/4): filename='.../test_retract_decode.py', elapsed=1042, file_estimated_time=12
  Begin (1/4): .../test_w8_quantization.py            ← suite advances normally
  ```
  The test passes (`OK`) but takes ~87× the expected time. The `Child process unexpectedly failed with exitcode=9` line that appears around the same timestamp is the server worker being SIGKILLed by `tearDownClass: kill_process_tree(cls.process.pid)` (`test_retract_decode.py:76`) — it's part of the test's own teardown, not a job-level OOM-kill. (#1215's job ultimately fails on a separate W8Int8 throughput threshold in the next file, unrelated to retract.)

## Sample retract loop (from #1213)

```
[...] Prefill batch. #new-seq: 1, #new-token: 128, #cached-token: 0, token usage: 0.00, #running-req: 2, #queue-req: 10
[...] Prefill batch. #new-seq: 1, #new-token: 5,   #cached-token: 0, token usage: 0.00, #running-req: 2, #queue-req: 10
[...] KV cache pool is full. Retract requests. #retracted_reqs: 1, #new_token_ratio: 0.0491 -> 0.0495
[...] Prefill batch. #new-seq: 1, #new-token: 128, ...
[...] Prefill batch. #new-seq: 1, #new-token: 37,  ...
[...] KV cache pool is full. Retract requests. #retracted_reqs: 1, #new_token_ratio: 0.0495 -> 0.0500
... (repeats 62160 times over 2 hours)
```

`token usage: 0.00` throughout — the KV pool is never actually exhausted. The `KV cache pool is full` log message is misleading because it fires on every `TEST_RETRACT`-forced retract, not just real OOM.

## Mechanism

`retract_decode()` (`schedule_batch.py:1236-1243`) does break out for single-request batches:
```python
first_iter = True
while first_iter or (not has_sufficient_memory(dp_rank, sorted_indices)):
    if len(sorted_indices) == 1:
        break              # single-request batches are NOT retracted
    first_iter = False
    retract_idx = sorted_indices.pop()
```

So when `batch_size >= 2`, the forced retract pops one request even though memory is sufficient. The retracted request goes back to `waiting_queue`, gets re-prefilled from scratch, decodes for 1–2 steps, and gets retracted again. With small steady-state batches (#1213's log stays at `#running-req: 2`), no request ever completes — only when prefill admission gets enough requests in flight (e.g. ~16 in #1215) does the loop slowly drain.

## Why it surfaced in late May, not early May

`scheduler.py` has been in this state since 2026-05-02 (#994), but main was green through 2026-05-22, then hit 6 consecutive failures 2026-05-25 → 2026-05-26, then green again on 2026-05-27 04:30 ([run 26490828890](https://github.com/sgl-project/sglang-jax/actions/runs/26490828890)). The bug is *latent* from #994 onward — sufficient prefill throughput keeps the drain rate above the retract rate, so MMLU completes (slowly). Some external timing change in late May (TPU runner state, JAX version, sibling-test residual state, etc.) tipped the balance: drain rate dropped below retract rate on some shards, exposing the unbounded loop. The intermittent pattern is consistent with a prefill-admission timing race rather than a deterministic deadlock — but the underlying contract violation has been there since #994.

## Suggested fix

Two consistent options; either restores the test's documented contract:

1. **Restore the `batch_size() > 10` guard** in `scheduler.py:1681-1684` (revert the trigger change from #994 while keeping the rest of the memory-management refactor):
   ```python
   if not batch.check_decode_mem() or (
       TEST_RETRACT and batch.batch_size() > 10
       and self.forward_ct % TEST_RETRACT_INTERVAL == 0
   ):
   ```
2. **Update the test's docstring + `test_retract_decode.py` setup** to match the new `forward_ct % 3` semantics, *and* set `SGLANG_TEST_RETRACT_NO_PREFILL_BS` to a sane value (e.g. `running-req` cap) so prefill admission backs off when the batch is full — so the loop is bounded by design.

Option 1 is the minimal-diff fix and aligns with the docstring authors of the original test. Option 2 is closer to upstream sglang's current intent but requires picking a tested `NO_PREFILL_BS` threshold for v6e-4.

Rodrian7:
## Summary
- PR #994 ported the upstream `forward_ct % TEST_RETRACT_INTERVAL` trigger but missed setting `TEST_RETRACT_NO_PREFILL_BS` in the test, causing `test_retract_decode.py` to enter an infinite retract→re-prefill loop (hang or ~87× slowdown)
- Align with upstream sglang's full mechanism:
  - **Trigger**: `forward_ct % interval` with `kv_full_retract_flag` walrus operator to distinguish real OOM from test retract
  - **Test**: set `SGLANG_TEST_RETRACT_NO_PREFILL_BS=6` (matching upstream `test_scheduler_control.py`) to cap prefill admission and bound the loop
  - **Log**: differentiate `"KV cache pool is full."` (real OOM) from `"Testing retraction."` (test-forced) to prevent misleading logs

## Test plan
- [ ] CI `e2e-test-4-tpu (1)` shard passes with `test_retract_decode.py` completing in expected time (~12s)
- [ ] No regression in other e2e test shards

Fixes #1229

### DIFF
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -1680,7 +1680,9 @@ class Scheduler(
 
         # Check if decode out of memory
         if not batch.check_decode_mem() or (
-            TEST_RETRACT and self.forward_ct % TEST_RETRACT_INTERVAL == 0
+            TEST_RETRACT
+            and batch.batch_size() > 10
+            and self.forward_ct % TEST_RETRACT_INTERVAL == 0
         ):
             old_ratio = self.new_token_ratio
 
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -1679,10 +1679,8 @@ class Scheduler(
             return batch
 
         # Check if decode out of memory
-        if not batch.check_decode_mem() or (
-            TEST_RETRACT
-            and batch.batch_size() > 10
-            and self.forward_ct % TEST_RETRACT_INTERVAL == 0
+        if (kv_full_retract_flag := not batch.check_decode_mem()) or (
+            TEST_RETRACT and self.forward_ct % TEST_RETRACT_INTERVAL == 0
         ):
             old_ratio = self.new_token_ratio
 
@@ -1698,8 +1696,14 @@ class Scheduler(
                 else:
                     self.send_to_tokenizer.send_pyobj(abort_out)
 
+            msg_prefix = (
+                "KV cache pool is full. Retract requests."
+                if kv_full_retract_flag
+                else "Testing retraction."
+            )
             logger.info(
-                "KV cache pool is full. Retract requests. #retracted_reqs: %d, #aborted_reqs: %d, #new_token_ratio: %.4f -> %.4f",
+                "%s #retracted_reqs: %d, #aborted_reqs: %d, #new_token_ratio: %.4f -> %.4f",
+                msg_prefix,
                 num_retracted_reqs,
                 len(reqs_to_abort),
                 old_ratio,
--- a/test/srt/test_retract_decode.py
+++ b/test/srt/test_retract_decode.py
@@ -1,8 +1,10 @@
 """Integration tests for the retract / release_kv_cache path.
 
-SGLANG_TEST_RETRACT=1 forces retract on batch_size > 10. Pass criterion:
-the worker stays alive (process.poll() is None) -- a leak would trip
-scheduler.check_memory() and SIGQUIT.
+SGLANG_TEST_RETRACT=1 forces retract every TEST_RETRACT_INTERVAL (default 3)
+forward steps. TEST_RETRACT_NO_PREFILL_BS caps prefill admission to bound
+the retract-prefill loop. Pass criterion: the worker stays alive
+(process.poll() is None) -- a leak would trip scheduler.check_memory()
+and SIGQUIT.
 """
 
 import time
@@ -67,6 +69,7 @@ class _BaseRetractDecode(CustomTestCase):
             other_args=launch_args,
             env={
                 "SGLANG_TEST_RETRACT": "1",
+                "SGLANG_TEST_RETRACT_NO_PREFILL_BS": "6",
                 "JAX_COMPILATION_CACHE_DIR": "/tmp/jax_compilation_cache",
             },
         )


Rodrian7:
Migrating to upstream branch (removing fork remote)

Issue #307, opened 2025-11-03
[Bug] Current TTFT in main is worse than it in Blog


aolemila:
### Describe the bug

When input length is 4096 and batch_size is 128, median TTFT in #297 is about 25731.12ms. However, median TTFT in main is about 25824.45ms.

### Reproduction

```bash
# launch server
JAX_COMPILATION_CACHE_DIR=/tmp/jit_cache \
uv run python -u -m sgl_jax.launch_server \
--model-path Qwen/Qwen3-32B \
--trust-remote-code  \
--tp-size=4 \
--device=tpu \
--mem-fraction-static=0.95 \
--chunked-prefill-size=2048 \
--download-dir=/tmp \
--dtype=bfloat16 \
--max-running-requests 256 \
--skip-server-warmup \
--page-size=128  \
--disable-radix-cache

# benchmark
uv run python -m sgl_jax.bench_serving --backend sglang-oai --dataset-name random --num-prompts 384 --random-input-len 4096 --random-output-len 1 --max-concurrency 128 --random-range-ratio 1  --disable-ignore-eos --warmup-requests 0
```

### Environment

TPU v6e-4.

PR #1238, opened 2026-05-28
perf(rpa_v3 + weight_loader + scheduler): tuned block-config table + hot-path fixes

JamesBrianD:
## Summary

Backports a cluster of performance work done on the primatrix fork to upstream. 24 commits in three thematic blocks, all rebased onto current `main`:

### 1. RPA v3 tuned block-config table + tuner (commits 1–17, 21–22, 24)

Introduces `tuned_block_sizes_v3.py` (initially empty, behavior unchanged) and the `get_block_spec_config_v3.py` tuner that sweeps candidate `(bq_sz, bkv_sz, bq_csz, bkv_csz)` configs per stage (decode / prefill / mixed) and emits Python-literal entries.

Tuner evolution (collapsed): co-benchmark heuristic, drop smoke-test flag, parameterize grid via CSV, vmem-aware pre-filter, simplify to v2-style (one vmem knob, no csz dim), `--shard` for multi-worker grid split, SMEM pre-filter, full VMEM capacity as default budget. Plus a `fix(rpa_v3): unalias m/l and bo/bq scratch` that avoids an XLA MSA `CHECK` fail when two scratch entries pointed at the same VMEM object.

Tuned entries shipped (TPU v7x, MiMo-V2-Pro per-shard, validated on `exp-*`):

- **Decode (q=32, kv=2, hd=192, ps=256)**: 9 entries `mnt {4..1024}`, `bkv 1024→2048` wins **+18–25 %** over the v3 heuristic; `mnt=2048` is SMEM-bound and pruned.
- **Prefill (chunk_prefill_size=2048)**: 1 entry, `bq 4→32` wins **+83 %** (heuristic's `min(MAX_BQ_SZ, max_q//2)` gave bq=4 with post-page-alignment `max_q=8`, 8× too small for the MXU).
- **Mixed**: 11 entries `mnt {8..8192}` all `+75 %` over heuristic's `(4, 1024)`.
- **SWA decode (sliding_window=128, MiMo)**: 8 entries `mnt {8..1024}`, `bkv 256→512` wins **+3–8 %**. `TUNED_BLOCK_SIZES_V3` key now includes `sliding_window` so SWA and non-SWA tunes never overwrite each other; existing entries are marked `sliding_window=None`.

Tuner also gained a one-shot-per-key INFO log on lookup-miss after a silent-fallback bug where the (q=16, kv=2) entries were never hit at runtime because MiMo on v7x-32 has `attention_tp=4` → per-device q_heads=32 (`f83ac4f6`).

Behavior on lookup miss is unchanged: falls back to the existing `get_default_block_sizes` heuristic.

### 2. weight_loader scan perf (`f4f5c47f`)

`_scan_weight_info()` was calling `safetensors.safe_open` to fetch shape/dtype/offsets that are already in the raw header it just parsed. On GCSFuse this triggers a full mmap-faulted download per file: on MiMo-V2-Pro (34 files × ~880MB) the scan phase was downloading ~30 GB of weight bytes just for metadata, stalling startup 10+ min.

Fix pulls shape/dtype/data_offsets straight from `raw_header` (bit-for-bit compatible with downstream consumers — dtype was already a string). Also hoists the dtype→jax dtype map to a module-level `_SAFETENSORS_DTYPE_TO_JAX` and dedupes the four inline copies. Expected impact: scan phase **~10min → seconds**. Phase 2 / Phase 3 unchanged.

### 3. Scheduler `_merge_cache_loc` hot-path (last 3 commits)

`_merge_cache_loc` was visible as ~12.99 ms / step at BSZ=64 OSL=16K on the scheduler-thread profile. Two real costs in the prior vectorised fancy-index version: an 8 MB-per-DP intermediate copy that selects the full `MAX_CONTEXT_LEN=262144` columns of each row but only uses the first `seq_len`, plus a 1 M-element numpy fancy-index scatter serialising through Python.

The fix replaces both with a per-request contiguous slice assignment: `cache_loc_cpu[dest:dest+sl] = req_to_token[req_pool_idx, :sl]`. Numpy lowers each to a `memcpy` of `sl` int32 ≈ 65 KB. Microbench (CPU, 20-iter median, BSZ=64 OSL=16K):

- baseline: **~12.99 ms**
- per-req slice: **~0.31 ms** (~40×)

Correctness: 14/14 fuzz cases byte-for-byte identical to the previous implementation (BSZ ∈ {1,8,32,64,512}, empty DPs, page-aligned/±1 seq_lens, decode-start tiny lens, uneven DP distributions).

The intermediate `reuse cache_loc scratch buffer in ReqToTokenPool` + its revert are kept in history (commits 22+23) — they were an exploratory step that didn't pan out before the final per-req-slice fix. Net diff of those two commits is zero; merging via squash would collapse them anyway.

## Branch context

Cherry-picked from primatrix/sglang-jax PR [#307](https://github.com/primatrix/sglang-jax/pull/307) (base `feat/fused-moe-v2-server-integration`). Skipped 3 SP-related commits from that PR that are not appropriate for this scope:
- `a726ef81` Allow sequence parallel and enable it for grok-2 (#952)
- `6b98f2bb` perf(sp): sequence-parallel via threshold-gated reduce-scatter
- `d1eabb0d` fix(fused_moe_v2): accept token_valid_mask and out_sharding for SP signature compat

All 24 commits applied to current `main` (`6bdd34ec`) without conflicts. Original author / `(cherry picked from …)` annotations preserved via `-x`.

## Test plan

- [ ] Unit: `python -m pytest python/sgl_jax/test/test_tuned_block_sizes_v3.py` — covers empty / hit / invalid stage / unknown device / pre-v5 TPU
- [ ] Lookup-miss INFO log fires once per key when no tuned entry matches (regression guard for the silent-fallback bug fixed in `f83ac4f6`)
- [ ] On TPU v7x: serve MiMo-V2-Pro and confirm decode `mnt ∈ {4..1024}` + prefill `chunk=2048` + mixed `mnt ∈ {8..8192}` paths hit tuned entries (no lookup-miss logs after warmup) and reproduce the +18–83 % deltas reported in the per-commit messages
- [ ] Confirm `_merge_cache_loc` no longer shows up as a hot frame on the scheduler-thread profile at BSZ=64


Iamleos (dismissed):
LGTM

PR #1252, opened 2026-05-28
feat(spec-decode): chain mask fastpath and MTP KV config fix for MiMo-V2-Flash

pengchengneo:
## Summary

- **Skip tree mask allocation for topk=1 (chain mode)**: When `speculative_eagle_topk=1`, the draft tree degenerates to a linear chain where causal attention is equivalent to the tree mask. This bypasses the `build_eagle_tree_structure` Pallas kernel and its O(bs × context_len × draft_token_num) HBM allocation, eliminating precompile OOM on memory-constrained chips (e.g. v6e 16-chip with MiMo-V2-Flash).

- **Fix MTP draft model KV cache config**: The MTP draft model uses SWA attention (`swa_num_key_value_heads=8`, `swa_head_dim=192`) but the KV cache was allocated using the full-attention target config (`num_key_value_heads=4`). This caused a Pallas DMA shape mismatch (`?×2×2×256` vs `?×1×2×256`) during RPA kernel execution. Override `num_key_value_heads`, `num_attention_heads`, and `head_dim` to SWA values when configuring the MTP draft model.

- **Fuse N-layer MTP draft extend into single JIT**: Merges all N MTP draft layer forwards into one `@jax.jit` call with `donate_argnames` and `with_sharding_constraint`, eliminating per-layer host dispatch overhead. Includes device-side rotate (copy-last + skip-padding) matching host `_rotate_ids`.

- **Fix SWA index mapping for draft workers**: Draft workers' `attn_backend` was missing the target's `full_to_swa_index_mapping`, causing SWA attention to remap all page indices to 0. This produced wrong KV reads → accept-len collapsed to 1.0 → HBM exhaustion crash. Fixed by sharing the target allocator's mapping reference with each draft worker's attention backend.

- **Share single ForwardBatch across fused layers**: Previously each of the N draft workers created its own ForwardBatch via `ForwardBatch.init_new()`, triggering N redundant `device_put` + dtype conversion round-trips. Since all workers process the same batch, share one ForwardBatch and only vary memory_pools/weights per layer. Also cast `sel_pos` to int32 (TPU-native).

### `eagle_util.py`
- Add `build_chain_verify_inputs()`: pure NumPy function that builds verify inputs for the linear chain case without allocating a tree mask
- Update `EagleVerifyInput.custom_mask` type to `jax.Array | None`

### `eagle_draft_worker.py`
- Branch `draft()` on `self.topk == 1`: chain path skips `build_tree_kernel_efficient` entirely

### `flashattention_backend.py`
- Guard `custom_mask.dtype` access with `is not None` check in `get_eagle_forward_metadata()`

### `model_config.py`
- Override `num_key_value_heads`, `num_attention_heads`, `head_dim` to SWA values for MiMo-V2 MTP draft model config

### `draft_extend_fused.py` (NEW)
- `_device_rotate_input_ids()`: device-side per-req rotate for topk=1
- `_build_fused_draft_extend_jit()`: single `@jax.jit` with `donate_argnames` fusing N MTP layers
- `draft_extend_for_decode_fused()`: host orchestration — single shared ForwardBatch, int32 sel_pos

### `multi_layer_draft_worker.py`
- `draft_extend_for_decode()` redirects to fused path
- Share target's `full_to_swa_index_mapping` with draft workers' `attn_backend`

## E2E Validation (16-chip v6e, MiMo-V2-Flash, dp=4)

**Server launch:**
```bash
uv run python -m sgl_jax.launch_server \
  --model-path /data/pc --trust-remote-code \
  --speculative-algorithm NEXTN --speculative-eagle-topk 1 --speculative-num-steps 3 --speculative-num-draft-tokens 4 \
  --tp-size 16 --dp-size 4 --ep-size 16 --moe-backend epmoe \
  --host 0.0.0.0 --port 30271 --page-size 64 --context-length 4096 --max-prefill-tokens 4096 \
  --dtype bfloat16 --mem-fraction-static 0.85 --swa-full-tokens-ratio 0.5 --max-running-requests 32 \
  --attention-backend fa --disable-overlap-schedule \
  --nnodes 4 --node-rank $TPU_WORKER_ID \
  --dist-init-addr perf-16-0.perf-16-headless-svc:5000
```

### GSM8K (evalscope, bs=64, 1319 questions)

```bash
evalscope eval --model default --api-url http://localhost:30271/v1 \
  --eval-type service --datasets gsm8k --eval-batch-size 64
```

| Run | Score | Accept-len | Crash |
|-----|-------|-----------|-------|
| 1   | 0.9568 | stable ~3.5 | None |
| 2   | 0.9545 | stable ~3.5 | None |

### ShareGPT throughput (bench_serving, 200 prompts)

```bash
python -m sgl_jax.bench_serving --backend sgl-jax --dataset-name sharegpt \
  --num-prompts 200 --host 127.0.0.1 --port 30271
```

| Metric | Before (3× ForwardBatch) | After (shared 1×) |
|--------|------------------------|-------------------|
| Output throughput | 620 tok/s | **654 tok/s (+5.5%)** |
| Request throughput | 2.88 req/s | **3.03 req/s** |
| Mean TPOT | 50.1 ms | **48.4 ms** |

## Test plan
- [x] Server starts successfully on 16-chip v6e with MiMo-V2-Flash + EAGLE topk=1
- [x] Precompile completes without OOM
- [x] GSM8K full eval × 2 runs: Score 0.9568/0.9545, accept-len stable, zero crash
- [x] ShareGPT 200 prompts: 654 tok/s output throughput (+5.5% vs pre-optimization)
- [ ] Benchmark with input-len=16384 (long context)


zorrofox (commented):
Reviewed at d4a580b2. Independent validation on v6e-64 dp=4 (16-host, tp=16 ep=64 epmoe, V2-Flash NEXTN steps=3 topk=1):

| round | dataset | accept-len | n_round | low(<1.2) |
|---|---|---|---|---|
| warmup | random n=32 out=128 | 2.685 | 63 | 0 |
| g1 | gsm8k n=100 out=1024 | 3.366 | 800 | 0 |
| g2 | gsm8k | 3.413 | 797 | 1 |
| g3 | gsm8k | 3.402 | 832 | 2 |
| g4 | gsm8k | 3.436 | 819 | 0 |
| g5 | gsm8k | 3.279 | 889 | 18 |
| r1 | random n=64 out=1024 | 3.130 | 792 | 3 |

7 rounds, ~5000 verify rounds total, 0 assert/traceback/crash, 2 restarts (initial + hotpatch). The g5 low=18 is prompt variance (questions 401-500 of gsm8k test split), not accumulation — r1 immediately after recovers to low=3. Consistent with @pengchengneo's v6e-16 GSM8K 1319q×2 result.

Two correctness gaps and a few smaller things below — none of them block the perf path you've validated, but the two HIGH ones will bite under topk>1 or idle-batch edge cases.

---

### HIGH

**`draft_extend_fused.py:57` — no fallback for topk>1**
`assert topk == 1` is now a hard crash since 958e9def removed the host-loop fallback (`multi_layer_draft_worker.py:285-293` calls fused unconditionally). `eagle_draft_worker.py:161` and `draft_forward` still handle topk>1, so only decode-dext breaks. Suggest either an `assert self.topk == 1` in `MultiLayerDraftWorker.__init__` (fail at startup, not mid-run) or keep the host loop behind `if self.topk > 1`.

**`eagle_util.py:701-702` — `create_idle_input` still emits 2D topk**
```python
topk_p=np.empty((0, topk), ...)
topk_index=np.empty((0, topk), ...)
```
Multi-layer spec_info.topk is now 3D `(bs, n_steps, topk)`. The `is_idle()` branch at L924 returns this 2D shape; if it later hits `merge_batch` L758 concat with a real 3D one → `ValueError: dimensions mismatch`. The precompile path at `eagle_worker.py:193` was updated to 3D — this one was missed. Suggest adding a `num_steps` arg and matching the precompile shape.

---

### MEDIUM

**`multi_layer_draft_worker.py:111-122` — SWA mapping fix is multi-layer only**
Single-layer `EagleDraftWorker` can also have an SWA draft layer depending on the model. Same `swa_index_mapping` sharing should apply there (or move to a base helper). Doesn't affect MiMo.

**`draft_extend_fused.py:86-87` — shared `forward_batch` assumes `model()` is read-only on it**
Only `hidden_states` and `input_ids` are re-set per layer. If a model implementation ever writes back into other `forward_batch` fields (e.g. mutates `positions` or attn_backend metadata), layer i+1 sees the dirty value. Worth a comment so future model changes don't silently break this.

**`eagle_worker.py:195` — precompile topk_p dtype vs runtime**
Precompile uses `bfloat16`; fused L109-110 `jnp.exp(topk_logits - logsumexp)` promotes via `logsumexp` to float32 → runtime topk_p is float32. The next round's `select_top_k_tokens` (jit'd) sees a different input dtype than precompile → one extra trace per bucket. Either `.astype(rep_logits.dtype)` after the exp, or precompile with float32.

**`draft_extend_fused.py:171` — dead `.copy()`**
```python
mwb.input_ids = np.asarray(jax.device_get(mwb.input_ids)).copy()
```
Fused no longer host-rotates in place; `verified_id` is already host numpy. Both `device_get` and `.copy()` are dead.

---

### nit

- `eagle_draft_worker.py:154` — `jnp.asarray(packed_np)` then `device_put` is one extra host→device hop; `device_put(packed_np, ...)` directly
- `draft_extend_fused.py:127` — `mesh` is only bound inside the loop; `num_layers==0` would NameError (defensive)
- `draft_extend_fused.py:170/191` — sel_pos `int64`→`int32` converted twice; just `astype(np.int32)` at L170
- `multi_layer_draft_worker.py:119-121` — setting `swa_index_mapping` on `workers[1:]` is unused by fused decode (only `mr0.attn_backend` is read at L179), but prefill L248 still iterates per-worker, so it's correct to keep. Maybe a comment.

---

### Verified ✓

- `_device_rotate_input_ids` matches host `_rotate_ids` (copy-last + skip padding + dp>1 layout; relies on `eagle_util.py:599-602` guaranteeing fixed-width input_ids)
- donate `all_memory_pools` + `replace_all`: `memory_pool.py:1387` enforces full key match, same contract as `mr.forward`
- `with_sharding_constraint(P())` covers all three cross-process-consistent outputs; `all_pool_updates` correctly stays P("data")
- chain mode `custom_mask=None` → fa_backend keeps `causal=1` (default); topk=1 chain *is* pure causal
- `filter_batch`/`merge_batch` axis-0 ops are 3D-transparent (modulo the idle-2D issue above)
- 1ad63512 SWA mapping is a mutable numpy ref, draft sees target's in-place updates; dp>1 list case handled at `fa_backend:345`

zorrofox (commented):
Thanks — b6fa9036 + 257a48d1 address all of it. Two leftovers:

**`eagle_util.py:926-931`** — the only caller of `create_idle_input` doesn't pass the new `num_steps` arg, so it still defaults to 1 → 2D. `EagleVerifyInput` doesn't carry an `is_multi_layer` flag, so probably easiest to pass `num_steps=self.draft_token.shape[0] // ...` won't work either — maybe just plumb `speculative_num_steps` onto `EagleVerifyInput` (it already has `spec_steps` actually, so `num_steps=self.spec_steps` if multi-layer). Low priority though, idle-batch merge is a narrow path.

**`draft_extend_fused.py:193`** — `sel_pos.astype(np.int32)` is now a no-op after L173.

Everything else LGTM. The 257a48d1 `isinstance(MultiLayerDraftWorker)` check is the right call (single-layer EAGLE with num_steps>1 still has 2D topk).

zorrofox (approved):
LGTM for the e2e perf path. Validated on v6e-64 dp=4 (7 rounds GSM8K+random, ~5k verify rounds, accept-len 2.7-3.4, 0 crash) at d4a580b2; b6fa9036/257a48d1 address the blocking findings. Remaining items (idle 2D caller, single-layer SWA mapping) are generalization for non-MiMo paths and can land separately.

pengchengneo:
Thanks. Regarding the additional issues concerning code abstraction, reusability, and simplification, I will address them in a separate PR—serving as the wrap-up PR for Phase 2.https://github.com/sgl-project/sglang-jax/pull/1292
@zorrofox

PR #1254, opened 2026-05-28
perf(overlap-worker): move batch-result materialization to a dedicated thread

JamesBrianD:
## Summary

The scheduler's `process_batch_result_decode` was synchronously blocked on `np.asarray(jax.Array)._value` inside `resolve_last_batch_result` (~14 ms per decode step waiting for TPU forward → host materialisation). That block sits on the critical path between "forward N finished" and "build/submit forward N+1", so the TPU sits idle for the duration of the host wait.

This PR splits materialisation into a **third worker thread** so the wait happens off the scheduler's critical path:

```
forward_thread:  fwd → set_future → copy_to_host_async → put output_queue
resolve_thread:  output_queue.get → np.asarray (block here) → put resolved_queue   [NEW]
scheduler:       ... → resolved_queue.get  (typically zero-wait, already done)
```

The forward thread also pre-fires `jax.copy_to_host_async` for every needed array right after `set_future_token_ids`, so the D2H is in flight by the time the resolve thread starts blocking. While the scheduler is busy in `get_model_worker_batch` / `_merge_cache_loc` (~36 ms), the resolve thread materialises in parallel.

Forward-thread shutdown now puts `None` on `output_queue` so the resolve thread can join cleanly.

## Measurement (Ling-2.6-1T, BS=128, in=16384/out=1024, v7x 2x2x4 = 32 devices)

|                                                       | baseline | this PR | Δ |
|---|---:|---:|:---:|
| `process_batch_result_decode` (scheduler thread)      | 26.5 ms  | 20.7 ms | **-5.8 ms / step** |
| `resolve_last_batch_result` (scheduler thread)        | 15.4 ms  | < 1 ms  | **~-15 ms / step** |
| `_value` / `np.asarray` site                          | scheduler critical path | dedicated resolve thread (parallel with scheduler) | off critical path |
| Steady-state decode throughput (gen tok/s)            | ~1997    | ~2310   | **+15.7%** |
| GSM8K thinking-on accuracy (200 q)                    | 0.925    | 0.920   | 1-q diff (sampling noise) |

Per-step trace evidence: the `_value` block (~14–30 ms) moves from the scheduler timeline into the new resolve thread; TPU `barrier-cores` between successive `jit_resolve_future_token_ids` invocations shrinks correspondingly.

## Reproduce

Pod side (rank-0 inside ablation-16 cluster):
```bash
# decode bench at BS=128
/opt/venv/bin/python -m sgl_jax.bench_serving \
  --backend sgl-jax --host 127.0.0.1 --port 30271 \
  --dataset-name random \
  --random-input-len 16384 --random-output-len 1024 --random-range-ratio 1.0 \
  --num-prompts 128 --max-concurrency 128 \
  --flush-cache --seed 12345
```

Profile decode at steady state (after `Decode batch. #running-req: 128`):
```bash
curl -X POST http://localhost:30271/start_profile \
  -H 'Content-Type: application/json' \
  -d '{"output_dir": "/tmp/tpu_logs/profile/decX",
       "profile_by_stage": true,
       "profile_stages": ["decode"],
       "num_steps": 10}'
```

## Test plan

- [x] Decode bench BS=128 (above) — functional, no regression
- [x] GSM8K thinking-on 200 q — 0.92 vs baseline 0.925 (within sampling noise)
- [x] Pre-commit (black/isort/ruff/mypy/codespell) all pass
- [ ] Other batch sizes / hardware (single-host TPU, v6e) — to validate
- [ ] Logprobs / hidden-states paths exercised (have logic but not actively benched here)

## Status

**Draft** — improvement is modest (+15.7%) because the remaining hot path is dominated by `_merge_cache_loc` (~36 ms / step CPU loop) in `get_model_worker_batch`. That's a separate optimisation; this PR is the cleaner first step targeting the materialisation barrier specifically.

### DIFF
--- a/python/sgl_jax/srt/managers/tp_worker_overlap_thread.py
+++ b/python/sgl_jax/srt/managers/tp_worker_overlap_thread.py
@@ -50,13 +50,25 @@ class ModelWorkerClient:
         self.future_token_ids_map = jax.device_put(self.future_token_ids_map, sharding)
         # Launch threads
         self.input_queue = Queue()
+        # output_queue carries the async (jax.Array) handles from the forward
+        # worker; the new resolve thread drains it, blocks on materialization
+        # (np.asarray / _value), and pushes ready Python values to resolved_queue.
+        # This keeps materialization off the scheduler's critical path: while
+        # scheduler does build_batch / process other work, resolve thread is
+        # waiting for TPU forward to finish in parallel.
         self.output_queue = Queue()
+        self.resolved_queue = Queue()
         # JAX handles device execution automatically, no need for explicit streams
         self.forward_thread = threading.Thread(
             target=self.forward_thread_func,
             daemon=bool(server_args.enable_single_process),
         )
         self.forward_thread.start()
+        self.resolve_thread = threading.Thread(
+            target=self.resolve_thread_func,
+            daemon=bool(server_args.enable_single_process),
+        )
+        self.resolve_thread.start()
         self.parent_process = psutil.Process().parent()
         replicated_sharding = NamedSharding(mesh, PartitionSpec())
         self.async_gather_fn = jax.jit(lambda x: x, out_shardings=replicated_sharding)
@@ -102,6 +114,8 @@ class ModelWorkerClient:
                 forward_metadata,
             ) = self.input_queue.get()
             if not model_worker_batch:
+                # Signal the resolve thread to shut down too.
+                self.output_queue.put(None)
                 break
 
             # Resolve future tokens in the input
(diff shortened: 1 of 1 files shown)


JamesBrianD:
## Closing: superseded by #1253 (ablation confirms no incremental value)

Ran a clean ablation on Ling-2.6-1T BS=128 (in=16384, out=1024) on v7x 2x2x4 (32 devices), same bench command throughout:

| Build | gen tok/s | Δ vs baseline |
|---|---:|---:|
| baseline (pre-#1253, pre-this PR) | 1997 | — |
| this PR only (pre-#1253) | 2310 | **+15.7%** |
| **#1253 only (no this PR)** | **2912** | **+45.8%** |
| this PR + #1253 (rebased) | 2910 | +45.7% |

The two "after #1253" rows are statistically identical (2 tok/s = 0.07% — noise).

### Why this PR became redundant

Per-iteration timing analysis:

| | scheduler critical-path | TPU iter | bottleneck |
|---|---:|---:|---|
| baseline | ~63 ms (36 `_merge_cache_loc` + 14 `_value` + 13 post) | ~50 ms | **scheduler** |
| #1253 only | ~28 ms (1 `_merge_cache_loc` + 14 `_value` + 13 post) | ~44 ms | **TPU** |
| this PR + #1253 | ~33 ms (1 + 24 wait-on-resolve + 8 post) | ~44 ms | **TPU** |

This PR moves `_value` (~14 ms) off scheduler's critical path into a dedicated resolve thread. That win is **real when scheduler is the bottleneck** (pre-#1253: +15.7%). Once #1253 lands and kills `_merge_cache_loc` (-35 ms / iter), scheduler critical path falls below TPU, and trimming scheduler further can't push throughput — it's gated by the model forward pass on device.

### What stays valuable

- The decomposition of where time goes (build vs materialise vs post) lives in [`profiling-sgl-jax-server.md`](https://github.com/sgl-project/sglang-jax/blob/main/docs/developer_guide/benchmark_and_profiling.md) / agent notes for next-round bottleneck hunts.
- The resolve-thread pattern itself is straightforward and could be re-applied if TPU iter ever shrinks (eg. faster kernels, smaller models, larger DP), pushing scheduler back to being the bottleneck.

### Branch

`perf/decode-future-token-on-device` kept on origin for future reference. Re-open if measurements change.

PR #1259, opened 2026-05-29
feat(gmm_v2): support block-wise rhs_scale (W8A16 subchannel quant)

zorrofox:
## Motivation

`gmm_v2.is_supported_by_gmm_v2()` currently rejects `rhs_scale.shape[1] > 1`, so block-wise quantized MoE checkpoints (e.g. GLM-5.1 FP8: scale shape `(256, 48, 1, 2048)`) fall back to gmm v1, which is significantly slower and lacks v2's fused permute/mask. See #1036 / #1111 follow-up.

## Changes (`gmm_v2.py`, +66/-17)

- `validate_inputs`: allow `rhs_scale.shape[1] > 1` (block-scale), assert `size_k % block_size == 0`
- `IndexMaps`: add `rhs_scale_block_index_map(n,gm,k) → (g,k,0,n)`
- `generate_block_specs`: block-scale BlockSpec `(None, tile_k//blk, 1, tile_n)`
- `_matmul`: W8A16 block-scale path — inner-k loop, per-128-block `matmul(bf16,fp8) × scale[blk]`
- per-channel scale multiply gated on `not is_block_rhs_scale`
- W8A8 + block-scale raises `NotImplementedError` (act-quant interaction TBD)
- `gmm_v2` wrapper: pad `rhs_scale[1]` to `num_k_tiles × (tile_k//blk)`
- `is_supported_by_gmm_v2`: always True

Existing paths (BF16 `scale=None`, W8A8 per-channel `shape[1]==1`) unchanged.

## Verification (v6e-4, single-host)

Numerics vs gmm_v1: max_err=0.0000; vs numpy ref: rel_err ~0.2%.

Latency (E=8 M=512, μs):
| | gate K=7168 N=1536 | down K=1536 N=6144 |
|---|---|---|
| BF16 v2 | 322 | 325 |
| **W8A16 block-scale v2** (this PR) | **268** | **259** |
| W8A16 per-channel v2 | 249 | 244 |
| W8A8 per-channel v2 | 319 | 244 |

Determinism: BF16/W8A16-blk/W8A8-ch at M∈{8,64,512} all bit-exact across paddings.

## Test plan
- [x] `gmm_test.py` numerics vs ref
- [x] v6e single-host latency sweep
- [ ] e2e GLM-5.1 W8A16 static ckpt (blocked on #1111)

### DIFF
--- a/python/sgl_jax/srt/kernels/gmm/megablox_gmm_backend.py
+++ b/python/sgl_jax/srt/kernels/gmm/megablox_gmm_backend.py
@@ -61,7 +61,9 @@ def gmm(
     if interpret is None:
         interpret = not is_tpu_runtime()
 
-    use_gmm_v2 = not interpret and is_supported_by_gmm_v2(rhs_scale)
+    use_gmm_v2 = not interpret and is_supported_by_gmm_v2(
+        rhs_scale, maybe_quantize_lhs=maybe_quantize_lhs
+    )
 
     # Pad LHS to multiple of 128 (or 32 for v2) on TPU to avoid small/unaligned tiles
     m = lhs.shape[0]
--- a/python/sgl_jax/srt/kernels/gmm/megablox_gmm_kernel/gmm_v2.py
+++ b/python/sgl_jax/srt/kernels/gmm/megablox_gmm_kernel/gmm_v2.py
@@ -148,9 +148,9 @@ def generate_block_specs(
     if cfgs.rhs_cfgs.has_scale:
         rhs_block = cfgs.rhs_cfgs.quant_block_size
         if rhs_block is not None and rhs_block < cfgs.dims.size_k:
-            assert cfgs.tiles.tile_k % rhs_block == 0, (
-                f"block-scale: tile_k={cfgs.tiles.tile_k} must divide rhs_block={rhs_block}"
-            )
+            assert (
+                cfgs.tiles.tile_k % rhs_block == 0
+            ), f"block-scale: tile_k={cfgs.tiles.tile_k} must divide rhs_block={rhs_block}"
             num_scale_per_tile = cfgs.tiles.tile_k // rhs_block
             rhs_scale_block_spec = pl.BlockSpec(
                 (None, num_scale_per_tile, 1, cfgs.tiles.tile_n),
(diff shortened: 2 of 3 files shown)


zorrofox:
e2e validation (GLM-5.1-FP8 native ckpt, TPU v6e-64, dp8/ep64, 1K/1K):

| cc  | before (#1111 only, gmm_v1 fallback) | this PR (gmm_v2 block-scale) | in-framework W8A8 (per-channel) |
|-----|-----|-----|-----|
| 8   | 112 | —   | 156 |
| 420 | —   | **2759** | 2695 |
| 460 | —   | (KV throttle) | 2809 |

Per-step ~2% faster than W8A8 (no act-quant overhead); peak ~2% lower (block-scale tensors cost ~7% HBM → smaller KV pool). Sanity 3/3 ✅.

zorrofox:
@cjx0709 @Bob-Chen222 PTAL — this is the gmm_v2 follow-up to #1111 (block-wise `rhs_scale` so native FP8 checkpoints don't fall back to gmm_v1). e2e numbers in the comment above.

cjx0709:
Thanks for the PR, this looks like the right direction for enabling the W8A16 block-wise `rhs_scale` path in `gmm_v2`.

One concern about the dispatch semantics: `is_supported_by_gmm_v2()` is now relaxed to `rhs_scale is None or rhs_scale.ndim == 4`, but the implementation still explicitly does not support block-wise `rhs_scale` when lhs activation quantization is enabled (`maybe_quantize_lhs=True` / W8A8). In that case, the backend will now select `gmm_v2` first and then hit the `NotImplementedError` in `make_gmm_configs()`.

Previously, a block-wise `rhs_scale` would fail the v2 support check and fall back to `gmm_v1`, which can already handle `[E, k_blocks, 1, N]` scales. So this change may turn a previously runnable W8A8 block-scale path into a runtime error.

Could we make the v2 support predicate account for activation quantization and keep falling back to v1 for the unsupported combination? For example:

```python
def is_supported_by_gmm_v2(
    rhs_scale: jax.Array | None,
    *,
    maybe_quantize_lhs: bool = False,
) -> bool:
    if rhs_scale is None:
        return True
    if rhs_scale.ndim != 4:
        return False
    if rhs_scale.shape[1] == 1:
        return True
    return not maybe_quantize_lhs
```
and then call it from the backend with:
```python
use_gmm_v2 = not interpret and is_supported_by_gmm_v2(
    rhs_scale,
    maybe_quantize_lhs=maybe_quantize_lhs,
)
```
That would preserve the new W8A16 block-scale v2 path while avoiding a regression for W8A8 block-scale cases by continuing to route them through v1.

zorrofox:
@cjx0709 Verified — gmm_v1 does handle `[E, k_blocks>1, 1, N]` rhs_scale (gmm.py:391/479/552) with the backend pre-quantizing lhs (megablox_gmm_backend.py:76), so this was a real regression for the W8A8 + block-scale combination. Applied the predicate change in 4c368939: `is_supported_by_gmm_v2(rhs_scale, *, maybe_quantize_lhs=False)` returns `False` for `k_blocks>1 and maybe_quantize_lhs`, and the backend now passes `maybe_quantize_lhs` through. The `NotImplementedError` in `make_gmm_configs` stays as a defensive guard (should be unreachable via the backend now).

Re CI: `e2e-test-4-tpu (2)` on the previous run hung at weight loading and hit the 2h timeout (log stops at 04:38, job ran 04:17–06:18) — unrelated to this change (deepseek-v2-lite uses per-channel scale). The new push should give it a fresh runner.

cjx0709 (approved):
LGTM

Issue #1251, opened 2026-05-28
[RFC] Qwen3.5 35B-A3B Text-Path Inference (Hybrid GDN + MoE) - Milestone 1


MokusMokun:
## 1. Background

Qwen3.5-35B-A3B is a multimodal MoE model: 35B total / 3.46B activated parameters, packaged as 14 safetensors shards / ~67 GB / 1,811 tensors on HuggingFace. It has a 40-layer hybrid-attention text backbone, a 27-layer ViT, and a 1-layer MTP draft head. M1 covers the text backbone only; ViT lands in M2, and MTP is deferred.

In this document `ckpt` refers to a project-internal mirror of [`Qwen/Qwen3.5-35B-A3B`](https://huggingface.co/Qwen/Qwen3.5-35B-A3B) on GCS, mounted into each GKE pod at `/models/Qwen3.5-35B-A3B/` via a GCS-fuse PVC. Bucket / PVC / service-account details live in the project's internal deployment & eval doc (see §4.3).

### 1.1 Model structure

All facts in this section come from the ckpt's `config.json` and `model.safetensors.index.json`.

| Field | Value |
|---|---|
| `architectures` | `["Qwen3_5MoeForConditionalGeneration"]` |
| `model_type` (root / text) | `qwen3_5_moe` / `qwen3_5_moe_text` |
| text backbone layers | 40, all MoE |
| layer pattern | `full_attention_interval=4` → indices 3, 7, 11, …, 39 are full-attention (10 layers); the other 30 are linear-attention (Gated DeltaNet) |
| `text_config.layer_types` | explicit 40-entry list, equivalent to the interval rule above; M1 cross-checks both at startup |
| hidden size | 2048 |
| `text_config.head_dim` | 256 (independent field, not derived from hidden/heads) |
| `text_config.attn_output_gate` | `true` (q_proj produces `2·num_heads·head_dim` outputs, layout is per-head interleaved `[q_i, gate_i]`; `gate_i` skips norm/RoPE and applies as `sigmoid(gate_i)` after attention, before `o_proj`) |
| full-attention heads | `num_attention_heads=16`, `num_key_value_heads=2` (GQA 16:2) |
| GDN | `linear_num_key_heads=16`, `linear_num_value_heads=32`, `linear_key_head_dim=128`, `linear_value_head_dim=128`, `linear_conv_kernel_dim=4`; conv_dim = 2·k + v = 8192 |
[...]

PR #1267, opened 2026-05-31
feat(qwen3.5): add Qwen3.5-35B-A3B (hybrid GDN + MoE) — M1 text-only

MokusMokun:
## Motivation

Add Qwen3.5-35B-A3B text-only inference: 40-layer hybrid (10 full-attention + 30 Gated-DeltaNet linear-attention, full_attention_interval=4), 256-expert MoE with 8 active, partial M-RoPE (rotary_dim=64, mrope_section=[11,11,10] interleaved). M1 = text-only bring-up validated single-host (TP=4/DP=1) and multi-host (TP=16/DP=4); vision / MTP left for follow-ups.

Implements the M1 milestone of RFC #1251.

## Modifications

- **Hybrid config (P1)**: `Qwen3_5HybridConfig` registered via `hf_transformers_utils`; `text_config` aliases RoPE / partial-rotary / interleaved m-rope fields onto the top-level model.
- **GDN backend refactor (P1)**: `GDNAttnBackend` (stateless, lives under `HybridLinearAttnBackend`) takes per-shard `mixed_qkv` and dispatches conv1d + ragged delta-rule (extend) / single-step delta-rule (decode) inside `jax.shard_map`. Rank-major `[q_d|k_d|v_d]` reshape on the input solves the TP>1 channel-mis-alignment.
- **Model (P2)**: `Qwen3_5Attention` (per-head q-output gate, partial M-RoPE), `Qwen3_5GatedDeltaNet` (fused in_proj_qkvz + in_proj_ba, GemmaRMSNorm convention at the transformer level, plain RMSNorm + silu gate inside GDN), `Qwen3_5MoeBlock` (FusedEPMoE + GateLogit + TopK + optional shared expert with sigmoid gate), causal-LM wrapper.
- **Weight loader (P3)**: self-contained loader on top of the shared `WeightLoader`. Handles GDN concat (qkv+z, b+a), conv1d rank-major stripe (mirrors backend's per-shard layout), and pre-fused MoE `gate_up_proj` split to `w1/w3`. consumed=693 / skipped=1118 (vision+MTP) / missing=0 / unexpected=0.
- **Recurrent state + multi-host decode**: `LinearRecurrentStateParams` gains optional `num_k_heads`/`head_k_dim` for GDN's asymmetric K/V (16 K-heads vs 32 V-heads); `optimization_barrier` on the donated conv/recurrent pool buffers fixes multi-host decode collapse (req #2+ emitting garbage) from scatter/gather racing on the aliased buffers.
- **Tests**: CPU model tests (`test_qwen3_5.py`, 7 cases: config alias, layer schedule, q/MoE shapes, weight-mapping coverage), GDN backend tp1/tp4/dp2tp2 DP tests (`test_gdn_attention_dp.py`, 10 cases with stripe-aware fixture), manual TPU smoke (`test_qwen3_5_models.py`, MMLU floor).

## Tests

- **Unit Test**: `test_qwen3_5.py` passed with 7/7, `test_gdn_attention_dp.py` passed with 10/10
- **Smoke Test**: `test_qwen3_5_models.py`
    - **Setup**: TP=4/DP=1 on v6e-4, thinking mode (temperature=0.6, top_p=0.95, top_k=20, presence_penalty=1.5, max_tokens=32768, enable_thinking=True)
    - **MMLU 40-Q = 0.850** (asserts > 0.7): STEM 0.875 / Humanities 0.833 / Social Sciences 0.9 / Other 0.8.
- **Full e2e benchmarks** (TP=16/DP=4 on v6e 4x4, thinking: temp=0.6/top-p=0.95/top-k=20/min-p=0/presence=1.5/max-tokens=32768):

  | Benchmark | n | recreated score | published score | delta | wall |
  |---|---|---:|---:|---:|---:|
  | GPQA-Diamond | 198 | 0.8434 | 0.842 | +0.1pp | ~33 min/run |
  | MMLU-Pro | 12,032 | 0.8579 | 0.853 | +0.5pp | 34.3 h |

  GPQA-Diamond is the **median of 5 runs** (0.803–0.854, mean 0.836 ± 0.024); the 198-q set is high-variance in thinking mode (single-question flips ≈ ±0.5pp) and TPU sampling is non-deterministic across runs, so a single seed is not meaningful. MMLU-Pro (12,032 q) is a single run — large enough to be stable. Per-run breakdown + per-category + full test commands in the [PR comment below](https://github.com/sgl-project/sglang-jax/pull/1267#issuecomment-4627297975).

## Benchmarking and Profiling

Out of scope for M1 (correctness-first bring-up); perf sweep planned in a follow-up.

## Checklist


MokusMokun:
**M1 full benchmark results** (v6e 4x4 TP=16/DP=4, thinking mode, temp=0.6/top-p=0.95/top-k=20/min-p=0/presence=1.5/max-tokens=32768).

### GPQA-Diamond — 5 runs (rebased build)

The 198-q set is high-variance under thinking mode (one question ≈ ±0.5pp) and TPU sampling is non-deterministic across runs (`enable_deterministic_sampling=False`; the same `--seed` does not reproduce token-for-token under concurrent batching), so we report 5 runs rather than a single seed:

| seed | correct | score |
|---|---:|---:|
| 17 | 159/198 | 0.8030 |
| 23 | 167/198 | 0.8434 |
| 41 | 166/198 | 0.8384 |
| 59 | 169/198 | 0.8535 |
| 97 | 167/198 | 0.8434 |

```
median = 0.8434      (reported in PR description)
mean   = 0.8364 ± 0.0241   (95% CI, [0.812, 0.860])
range  = 0.803 – 0.854     (5.1pp; std 0.019)
published = 0.842     -> median +0.1pp, mean -0.6pp (within CI)
```

Per-question diagnostics on seed 17 (full 198 q): `finish=length` (truncated) = 0, `pred=None` (no answer extracted) = 0, mean completion = 7,764 tok (max 26,991, under the 32,768 cap). No systematic truncation/parse loss — the spread is genuine sampling variance.

### MMLU-Pro (single run; 12,032 q is large enough to be stable)

```
n=12032  score=0.8579  parsed=0.9992  wall=34.3h
```

Per-category:

| category | n | score | parsed |
|---|---:|---:|---:|
| math | 1351 | 0.9519 | 0.9956 |
| biology | 717 | 0.9275 | 1.0000 |
| physics | 1299 | 0.9238 | 0.9985 |
| chemistry | 1132 | 0.9152 | 0.9991 |
| business | 789 | 0.9037 | 1.0000 |
| economics | 844 | 0.8957 | 1.0000 |
| computer science | 410 | 0.8780 | 1.0000 |
| psychology | 798 | 0.8396 | 1.0000 |
| philosophy | 499 | 0.8236 | 1.0000 |
| engineering | 969 | 0.8184 | 1.0000 |
| other | 924 | 0.7998 | 1.0000 |
| health | 818 | 0.7983 | 1.0000 |
| history | 381 | 0.7480 | 1.0000 |
| law | 1101 | 0.6857 | 0.9991 |

---

### Test commands

**Unit / backend / smoke** (run on v6e-4, model at `/models/Qwen3.5-35B-A3B/`):

```bash
# CPU unit tests
JAX_PLATFORMS=cpu PYTHONPATH=python pytest python/sgl_jax/test/models/test_qwen3_5.py

# v6e-4 GDN backend (incl. TP>1)
python -m unittest python.sgl_jax.test.test_gdn_attention_dp -v

# v6e-4 manual TPU smoke (from repo root)
cd test/srt && PYTHONPATH=../../python:$PYTHONPATH python -m unittest test_qwen3_5_models -v
```

**Full e2e benchmarks** (TP=16/DP=4 on v6e 4x4, server at `localhost:30000`):

```bash
# GPQA-Diamond — in-repo harness (--base-url WITHOUT /v1; no --top-p flag).
# Repeat with --seed {17,23,41,59,97}; take the median.
python test/srt/run_eval.py --base-url http://localhost:30000 --model /models/Qwen3.5-35B-A3B/ \
  --eval-name gpqa --num-threads 16 \
  --temperature 0.6 --top-k 20 --min-p 0.0 \
  --presence-penalty 1.5 --repetition-penalty 1.0 --max-tokens 32768 --seed 17

# MMLU-Pro — external harness (TIGER-Lab dataset, 12032 q, A-J extract)
python run_mmlu_pro_full.py --base-url http://localhost:30000 --model /models/Qwen3.5-35B-A3B/ \
  --num-threads 16 --max-tokens 32768 \
  --temperature 0.6 --top-p 0.95 --top-k 20 --min-p 0.0 \
  --presence-penalty 1.5 --repetition-penalty 1.0 --enable-thinking true --seed 17 \
  --output /tmp/mmlu_pro_full.json
```

Sampling contract: `temp 0.6 / top-p 0.95 / top-k 20 / min-p 0 / presence 1.5 / rep 1.0 / max-tokens 32768`, thinking on. GPQA: no `--top-p` flag (server default 1.0 used by `run_eval.py`); MMLU: `--top-p 0.95` (harness supports it).

python/sgl_jax/srt/layers/attention/linear/gdn_backend.py
```diff
@@ -106,84 +116,88 @@ def __init__(
...
-        a: jax.Array,  # [T, n_v]                       (None, "tensor")
-        recurrent_state_pool,
-        layer_id: int,
-    ) -> tuple[jax.Array, jax.Array, jax.Array]:
+        recurrent_state_pool: RecurrentStatePool,
+        **kwargs,  # absorb mixed_qkv=None forwarded by HybridLinearAttnBackend
+    ) -> tuple[jax.Array, tuple[jax.Array, jax.Array]]:
         """Dispatch by ``forward_batch.forward_mode``.
 
+        Per decision #4 the slice happens in the model layer; the backend
```
cjx0709: this notes `#4` should not be in the pr.

python/sgl_jax/srt/models/qwen3_5.py
```diff
@@ -0,0 +1,914 @@
...
+            attn_out = attn_out.reshape(
+                T,
+                self.num_heads,
+                self.head_dim,
+                out_sharding=NamedSharding(self.mesh, P("data", "tensor", None)),
+            )
+            attn_out = attn_out * jax.nn.sigmoid(gate)
+            attn_out = attn_out.reshape(T, self.num_heads * self.head_dim)
+
+        out, _ = self.o_proj(attn_out)
```
cjx0709, line 201: https://github.com/sgl-project/sglang-jax/pull/1198 maybe we can port the sp enhance.

python/sgl_jax/srt/hf_transformers_utils.py
```diff
@@ -38,13 +39,19 @@ class GlmMoeDsaConfig(PretrainedConfig):
...
 }
 
 for name, cls in _CONFIG_REGISTRY.items():
     with contextlib.suppress(ValueError):
         AutoConfig.register(name, cls)
 
+# Override HF's stock ``Qwen3_5MoeConfig`` registration (transformers >=5.3)
+# with our hybrid alias subclass; needed because the suppress(ValueError) loop
+# above silently skips already-registered entries.
+AutoConfig.register("qwen3_5_moe", Qwen3_5HybridConfig, exist_ok=True)
```
cjx0709, line 57: if we need to upgrade transformers version and delete the port model config, it's ok which not effect other models. Because other newest modeling seems also need to upgrade this:https://github.com/sgl-project/sglang-jax/pull/1323/changes#diff-1a45bd938160f7761769e7b6fbc819d1cc7f84ce37f1a6a23ef83cc4e214abc8

python/sgl_jax/srt/models/qwen3_5.py
MokusMokun: good point, but I suggest to make it as a follow-up PR?

python/sgl_jax/srt/hf_transformers_utils.py
MokusMokun: actually this line is still required even after upgrading `transformers` to 5.3+:
1. on 5.3+ stock transformers registers `qwen3_5_moe` -> `Qwen3_5MoeConfig` first, and our registry loop uses `suppress(ValueError)`, so without this `exist_ok=True` override `AutoConfig` would return the stock config instead of ours.
2. and we can't just fall back to the stock config — ours isn't a pure registration shim. It flattens HF's nested `rope_parameters` into the top-level `rope_scaling` / `rope_theta` / `partial_rotary_factor` our RoPE builders read (stock has no top-level `rope_theta`, and its `rope_scaling` lacks `mrope_section`), and it exposes `moe_pre_fused` plus the `linear_layer_ids` / `linear_state_params` helpers the weight loader and recurrent-state pool rely on. Those are the same duck-typed interface `bailing_hybrid` / `kimi_linear` provide, and stock HF configs don't have them.

> To be clear: the override is only needed *while we keep our own config*. Dropping it for stock on 5.3+ (this is even easier for the `GlmMoeDsaConfig` stub, more complex for `Qwen3_5HybridConfig`) is a repo-wide cleanup tied to the transformers bump (pr#1323), which is better a separate PR.

now refactored the comment for cleared explanation

python/sgl_jax/srt/models/qwen3_5.py
cjx0709: @MokusMokun It's ok if you want a follow-up PR with left a todo, but it seems each model now all use the sp.

python/sgl_jax/srt/models/qwen3_5.py
```diff
@@ -0,0 +1,914 @@
...
+    def __init__(
+        self,
+        config: PretrainedConfig,
+        mesh: jax.sharding.Mesh,
+        layer_id: int,
+        dtype: jnp.dtype = jnp.bfloat16,
+    ):
+        text_cfg = config.text_config
+        interval = int(text_cfg.full_attention_interval)
+        self.is_full_attn = ((layer_id + 1) % interval) == 0
```
hlFu: Maybe use self.is_full_attn = text_cfg.layer_types[layer_id] in {"full_attention", "attention"} to avoid implicit calculation?

python/sgl_jax/test/kernels/gdn/test_gated_delta.py
```diff
@@ -1,632 +0,0 @@
...
-        # Req1 token 2 = [5, 6, 7] sum = 18.
-        # Req1 token 3 = [6, 7, 8] sum = 21.
-        np.testing.assert_allclose(y[0, 4:], [5.0, 11.0, 18.0, 21.0], atol=1e-5)
-        # Final states scattered back into slots 1 and 2 (state_indices=[1, 2]).
-        np.testing.assert_allclose(final[1, 0], [3.0, 4.0], atol=1e-5)
-        np.testing.assert_allclose(final[2, 0], [7.0, 8.0], atol=1e-5)
-        # Slot 0 was unused — unchanged from input (zeros).
-        np.testing.assert_allclose(final[0], 0.0, atol=1e-5)
-
-    def test_has_initial_state_false_short_request_zero_padded(self):
```
hlFu, line 474: Didn't find a equivalent test of this edge case in the new test file. Did we move all the tests?

python/sgl_jax/srt/models/qwen3_5.py
```diff
@@ -0,0 +1,914 @@
...
+            params_dtype=dtype,
+            kernel_axes=("tensor", None),
+            scope_name="conv1d",
+        )
+
+        self.A_log = nnx.Param(
+            jnp.zeros((self.num_v_heads,), dtype=jnp.float32, out_sharding=P("tensor"))
+        )
+        self.dt_bias = nnx.Param(
+            jnp.ones((self.num_v_heads,), dtype=jnp.float32, out_sharding=P("tensor"))
```
hlFu: I think the checkpoint saves bf16?

python/sgl_jax/test/kernels/gdn/test_gated_delta.py
MokusMokun: It was already exercised in the `test_gdn_attention.py: test_extend_short_seqs`, which seeds the pool slot with non-zero state, runs with `has_initial_state=False`, and asserts the gathered conv state matches the zero-masked reference 

But you're right it deserves an explicit guard, so I re-added it as `test_extend_short_request_no_state_ignores_poisoned_slot`: it poisons the slot with a large sentinel and asserts the sentinel never appears in the final conv state.

python/sgl_jax/srt/models/qwen3_5.py
MokusMokun: refactored using `text_cfg.full_attention_layer_ids`

python/sgl_jax/srt/models/qwen3_5.py
MokusMokun: yes，updated to `dtype=dtype` for `dt_bias `

hlFu (approved):

PR #1276, opened 2026-06-01
fix(logprob): ShardingTypeError on dp_size>1 for return_logprob

### LINKED ISSUE #1261: logprobs crashes with ShardingTypeError when dp_size > 1
## Description

When `logprobs=True` with `top_logprobs=N` is requested on a server running with `dp_size > 1`, the request crashes in the logits processor due to sharding-incompatible array shapes.

## Error

Two failure modes depending on `dp_size`:

**dp_size=4 (original report, v6e-4):**
```
jax._src.core.ShardingTypeError: slicing on sharded dims where out dim (1) is not divisible by mesh axes (4) with spec (data) is not implemented.
```

**dp_size=16 (confirmed on latest main, v6e-16):**
```
ValueError: Sharding NamedSharding(mesh=Mesh('data': 16, 'tensor': 1), spec=PartitionSpec('data',))
implies that array axis 0 is partitioned 16 times, but the dimension size is 1
(full shape: (1,), per-dimension tiling factors: [16] should evenly divide the shape)
```

## Stack Trace

**dp=4 (original, commit `55b4b1a2`, line 317 at that version):**
```
File "sgl_jax/srt/layers/logits_processor.py", line 317, in __call__
    pruned_states.append(hidden_states[pt + start_len : pt + extend_len])
```

**dp=16 (latest main `fcdee8a`, line 357 on current main):**
```
File "sgl_jax/srt/utils/jax_utils.py", line 205, in _to_device
    return jax.make_array_from_callback(arr.shape, sharding, lambda idx, a=arr: a[idx])
```

> **Note**: The original report cited line 317, which was correct for the code version running on the pod at that time (commit `55b4b1a2`). PR #1177 subsequently added ~40 lines above this code, shifting it to line 357 on current main. The underlying bug was **not** fixed by #1177 — its new `input_logprob_indices_device` code path is bypassed when `top_logprobs > 0`.

## Root Cause

In `logits_processor.py:357` (current main), `hidden_states` is sharded across the data-parallel axis. When the slice produces an output with dimension 1 (a single token's hidden states), it cannot be evenly divided across `dp_size` mesh axes, causing a sharding error.

PR #1177 added an alternative code path using `input_logprob_indices_device` that avoids this slice by using `pruned_states = hidden_states` (the full tensor). However, this path is **only taken when `top_logprobs` is not requested**:

```python
# schedule_batch.py
use_padded_input_logprob = (
    ...
    and not (top_logprobs_nums and any(x > 0 for x in top_logprobs_nums))  # False when top_logprobs=3
)
```

When `top_logprobs > 0`, `input_logprob_indices` stays `None`, and the old slicing code path (the bug) is still executed.

## Reproduction

```bash
# v6e-4 (4 chips):
  --model-path Qwen/Qwen3-1.7B \
  --tp-size 4 --dp-size 4 \
  --device tpu

# v6e-16 (16 chips, multi-host):
  --model-path Qwen/Qwen3-1.7B \
  --tp-size 16 --dp-size 16 \
  --device tpu

# Then:
curl -X POST http://localhost:30000/v1/chat/completions \
  -d '{"model":"Qwen/Qwen3-1.7B","messages":[{"role":"user","content":"Say hello."}],"logprobs":true,"top_logprobs":3}'
```

Works fine with `dp_size=1`.

## Impact

- Crashes the entire server process (not just the request), causing all subsequent requests to fail with `Connection refused`
- Does not affect `dp_size=1` deployments (including CI)

## Confirmed on

| Environment | Commit | dp_size | Result |
|---|---|---|---|
| TPU v6e-4, 4 chips | `55b4b1a2` (2026-05-29) | 4 | Crash (ShardingTypeError) |
| TPU v6e-16, 16 chips, multi-host | `fcdee8a` (2026-06-01, latest main) | 16 | Crash (ValueError) |

MokusMokun:
## Problem

Fixes #1261. On `dp_size > 1`, an extend request with `return_logprob` falls back to the legacy per-req `hidden_states` slice in `LogitsProcessor.__call__`. `_get_logits` forces `P("data","tensor")`, so a row count not divisible by `dp` crashes the server (`ShardingTypeError` at dp=4, `ValueError` at dp=16). Two ways in:

1. The padded input-logprob path (#1177) was skipped for `top_logprobs` / `token_ids_logprobs` / overlap. The default server is overlap-on, so `--tp-size 4 --dp-size 4` crashed regardless of `top_logprobs`.
2. **dp≥4 overlap**: the speculated decode batch reuses extend reqs whose `extend_input_logprob_token_ids` residual is never cleared; the non-padded branch concats it and shards over `P("data")` → crash when the length isn't divisible by `dp` (dp=2 hid it).

## Fix

- **Unify the gate** to `use_padded_input_logprob = is_extend() and not return_hidden_states` for all `dp` — the padded path supports `top_logprobs` / `token_ids` / overlap at any `dp`; only `return_hidden_states` still falls back.
- Per-DP-padding-aware host split in `tp_worker`.
- Reshard vocab to replicated before `top_k` (`logits_processor` + `sampler`) so `[N, k]` isn't left tensor-sharded on `k`.
- Compact the per-DP padded scalar `input_token_logprobs` in the overlap path too (carrying `extend_lens` / `extend_logprob_start_lens` through `ScheduleBatch.copy()`).
- Gate the residual merge on `is_extend()` so decode batches leave it `None` (fixes crash #2).

**Known limitation:** `return_logprob` + `return_hidden_states` on `dp>1` still falls back — host hidden-state extraction isn't padding-aware (separate issue).

## `test/srt/test_logprobs.py` update

`top_logprobs` requests now take the padded path at dp=1 too → a different matmul shape → ~0.02 bf16 rounding shift on the sampled-token `output_token_logprobs` (token selection unchanged). Updated the first expected set (`-0.87109375 → -0.8515625`, `-0.318359375 → -0.3515625`); the second set is unchanged (already on the padded path).

## Validation (TPU v6e-4)

- [x] `tp4 dp4` overlap-on repro fixed; decode-residual crash reproduced pre-fix and fixed
- [x] `test_logprobs_dp.py` 4/4, `test_logprobs.py` 1/1
- [ ] `tp16 dp16` on v6e-16 — TODO

### DIFF
--- a/python/sgl_jax/srt/layers/sampler.py
+++ b/python/sgl_jax/srt/layers/sampler.py
@@ -93,7 +93,7 @@ class Sampler(nnx.Module):
             (
                 next_token_top_logprobs_val,
                 next_token_top_logprobs_idx,
-            ) = get_top_logprobs(logprobs, sampling_metadata.top_logprobs_nums)
+            ) = get_top_logprobs(logprobs, sampling_metadata.top_logprobs_nums, self.mesh)
 
         # Set token_ids_logprobs if needed.
         # Device side returns the full logprobs[B, vocab]; host slices
@@ -217,9 +217,13 @@ class Sampler(nnx.Module):
         return batch_next_token_ids, logprobs, new_logits_output
 
 
-def get_top_logprobs(logprobs: jax.Array, top_logprobs_nums: list[int]):
+def get_top_logprobs(logprobs: jax.Array, top_logprobs_nums: list[int], mesh: Mesh = None):
     # Return device-side dense `[B, max_k]`. Per-req trimming to k_i is
     # done on host in tp_worker after device_get.
+    # Replicate the vocab (tensor) axis first so the `[B, max_k]` output isn't
+    # left tensor-sharded on max_k, which fails when max_k < tp.
+    if mesh is not None:
+        logprobs = jax.sharding.reshard(logprobs, NamedSharding(mesh, P("data", None)))
     max_k = max(top_logprobs_nums)
     values, indices = jax.lax.top_k(logprobs, max_k)
     return values, indices
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -2411,13 +2411,11 @@ class ScheduleBatch:
             top_logprobs_nums = None
             token_ids_logprobs = None
 
-        use_padded_input_logprob = (
-            self.forward_mode.is_extend()
-            and not self.enable_overlap
-            and not self.return_hidden_states
-            and not (top_logprobs_nums and any(x > 0 for x in top_logprobs_nums))
-            and not (token_ids_logprobs and any(x is not None for x in token_ids_logprobs))
-        )
+        # extend+logprob always uses the padded path: the legacy fallback slices
+        # hidden_states per req under P("data","tensor") and crashes when the row
+        # count isn't divisible by dp (dp>1). The padded path supports top_logprobs /
+        # token_ids / overlap at any dp. return_hidden_states still falls back.
+        use_padded_input_logprob = self.forward_mode.is_extend() and not self.return_hidden_states
         input_logprob_indices = None
         merged_extend_input_logprob_token_ids = None
         if self.return_logprob:
@@ -2449,16 +2447,16 @@ class ScheduleBatch:
                         local_pt += extend_len
                     token_offset += per_dp_token_padding
             else:
-                # Merge per-DP extend_input_logprob_token_ids (set on info at L1099 when
-                # return_logprob=True). Previously hardcoded None below, which combined
-                # with the new server-side scalar gather caused 'NoneType.reshape' in
-                # LogitsProcessor._select_input_token_logprobs.
+                # Only extend batches have prompt-token logprobs. Overlap also routes
+                # speculated DECODE batches here carrying a stale extend residual;
+                # sharding it over P("data") crashes when len % dp != 0 (dp>=4). Decode
+                # never reads the field, so gate the merge on is_extend().
                 chunks = [
                     info.extend_input_logprob_token_ids
                     for info in self.reqs_info
                     if getattr(info, "extend_input_logprob_token_ids", None) is not None
                 ]
-                if chunks:
+                if chunks and self.forward_mode.is_extend():
                     merged_extend_input_logprob_token_ids = np.concatenate(chunks)
 
         return ModelWorkerBatch(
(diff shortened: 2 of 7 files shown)


python/sgl_jax/srt/layers/sampler.py
```diff
@@ -217,9 +217,13 @@ def __call__(
...
+def get_top_logprobs(logprobs: jax.Array, top_logprobs_nums: list[int], mesh: Mesh = None):
     # Return device-side dense `[B, max_k]`. Per-req trimming to k_i is
     # done on host in tp_worker after device_get.
+    # Replicate the vocab (tensor) axis first so the `[B, max_k]` output isn't
+    # left tensor-sharded on max_k, which fails when max_k < tp.
+    if mesh is not None:
+        logprobs = jax.sharding.reshard(logprobs, NamedSharding(mesh, P("data", None)))
     max_k = max(top_logprobs_nums)
     values, indices = jax.lax.top_k(logprobs, max_k)
     return values, indices
```
JamesBrianD, line 229: There are two duplicate functions with the same name.

MokusMokun: refactored `def get_top_logprobs(self, ...)` in `logits_processor.py` with the function here

aolemila: It looks like max_k < tp never happens here because if max_k <tp  occured, initilization of Embed would fail.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -2411,13 +2411,22 @@ def get_model_worker_batch(
...
+                self.forward_mode.is_extend() and not self.return_hidden_states
+            )
+        else:
+            use_padded_input_logprob = (
+                self.forward_mode.is_extend()
+                and not self.enable_overlap
+                and not self.return_hidden_states
+                and not (top_logprobs_nums and any(x > 0 for x in top_logprobs_nums))
+                and not (token_ids_logprobs and any(x is not None for x in token_ids_logprobs))
+            )
```
aolemila: Can we unify the above conditions? top_logprobs_nums and token_ids_logprobs are required in dp_size=1. But when dp_size>1, they are not required. Can I regard it as top_logprobs_nums and token_ids_logprobs are supported for dp_size=1 and dp_size>1 ?

aolemila: Another question: Can we remove the limitation of enable_overlap?

MokusMokun: 1. Unifying conditions: logically yes, but the reference values in `expected_output_logprobs` of `test_logprobs.py` needs to be updated. Reason: if we choose padded path for all dp, the tensor shape of logits matmul, so does the tiling/accumulation order changes, resulting in around  ~3e-2 differences in logprobs
2. yes, and we should determine whether to unify the condition first

python/sgl_jax/srt/layers/sampler.py
MokusMokun: max_k here is max(top_logprobs_nums), not the vocab/embedding dimension. Embed initialization only constrains the model weight/vocab shape, while users can still request top_logprobs_num smaller than the tensor mesh axis size.

python/sgl_jax/srt/managers/schedule_batch.py
aolemila: Ok. Could you unify it in another PR when you are avaiable?

MokusMokun: choose to use unify method in this PR

aolemila (approved):

PR #1284, opened 2026-06-02
refactor(mem_cache): extract kv cache builder from scheduler

cjx0709:
## Summary
- Port `kv_cache_builder` module structure from upstream sglang (PR #25602-#25607, #25101)
- Extract cache selection logic from `Scheduler.init_memory_pool_and_cache()` into three dedicated files:
  - `cache_init_params.py`: `CacheInitParams` dataclass for bundling cache construction parameters
  - `registry.py`: `TreeCacheBuildContext` + `default_radix_cache_factory` + `create_tree_cache` (cache selection chain)
  - `kv_cache_builder.py`: `build_kv_cache()` entry point + `KVCacheBuildResult`
- Pure refactor, no behavioral changes — same server args produce the same cache type

## Motivation
Prepares the cache subsystem for RFC-1 (HiCache / UnifiedRadixCache) and RFC-2 (PD disaggregation). New cache types can be added in `registry.py` without modifying the scheduler.

## Test plan
- [x] 6 new unit tests covering all cache selection branches (`test_kv_cache_builder.py`)
- [x] All existing `mem_cache/` tests pass (radix_cache, swa_radix_cache, allocator, etc.)
- [x] No changes to existing cache constructors or `ServerArgs`

### DIFF
--- a/python/sgl_jax/srt/mem_cache/kv_cache_builder.py
+++ b/python/sgl_jax/srt/mem_cache/kv_cache_builder.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING
 
 from sgl_jax.srt.mem_cache.base_prefix_cache import BasePrefixCache
 from sgl_jax.srt.mem_cache.cache_init_params import CacheInitParams
--- a/python/sgl_jax/srt/mem_cache/kv_cache_builder.py
+++ b/python/sgl_jax/srt/mem_cache/kv_cache_builder.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import logging
-from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 from sgl_jax.srt.mem_cache.base_prefix_cache import BasePrefixCache
@@ -18,13 +17,6 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=True)
-class KVCacheBuildResult:
-    tree_cache: BasePrefixCache
-    is_hybrid_swa: bool
-    sliding_window_size: int | None
-
-
 def build_kv_cache(
     *,
     server_args: ServerArgs,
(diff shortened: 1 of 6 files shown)


python/sgl_jax/test/test_kv_cache_builder.py
```diff
@@ -0,0 +1,136 @@
+import unittest
```
MokusMokun, line 1: is this test supposed to be registered at `test/srt/run_suite.py` ?

python/sgl_jax/srt/mem_cache/kv_cache_builder.py
```diff
@@ -0,0 +1,66 @@
...
+    tp_size: int,
+    spec_algorithm: SpeculativeAlgorithm | None,
+) -> KVCacheBuildResult:
+    params = CacheInitParams(
+        disable=server_args.disable_radix_cache,
+        req_to_token_pool=req_to_token_pool,
+        token_to_kv_pool_allocator=token_to_kv_pool_allocator,
+        page_size=page_size,
+        is_eagle=spec_algorithm is not None and spec_algorithm.is_eagle(),
+        chunked_prefill_size=server_args.chunked_prefill_size,
```
MokusMokun: the `disable` and `chunked_prefill_size` field of `CacheInitParams` seems redundant, given `server_args.disable_radix_cache` and  `server_args.chunked_prefill_size` are re-assigned into the following `TreeCacheBuildContext`, and registry.py only reads the TreeCacheBuildContext copies of these two.
Could we keep a single source?

python/sgl_jax/srt/mem_cache/kv_cache_builder.py
```diff
@@ -0,0 +1,66 @@
...
+    from sgl_jax.srt.speculative.spec_info import SpeculativeAlgorithm
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class KVCacheBuildResult:
+    tree_cache: BasePrefixCache
+    is_hybrid_swa: bool
+    sliding_window_size: int | None
```
MokusMokun: `is_hybrid_swa` and `sliding_window_size` look redundant: build_kv_cache just copies its is_hybrid / sliding_window_size inputs into the result, and the scheduler already owns both (self.is_hybrid / self.sliding_window_size) and only reads result.tree_cache. So tree_cache is the only meaningful field — the other two are read only by the new tests.

Could we drop `KVCacheBuildResult` and have `build_kv_cache` return `BasePrefixCache` directly?

python/sgl_jax/srt/mem_cache/kv_cache_builder.py
cjx0709: fixed

python/sgl_jax/srt/mem_cache/kv_cache_builder.py
cjx0709: fix

MokusMokun (approved):
LGTM

PR #1288, opened 2026-06-02
fix(multimodal): restore multimodal data-flow dropped by the DP refactor (input_embedding / mrope / deepstack)

lianfang0320:
## Summary

The DP-support refactor (`de5287ed`, "dp support (#939)") rewrote `get_model_worker_batch` into the per-DP `reqs_info` / DP-interleaved layout and, in doing so, **deleted the multimodal batch-assembly logic instead of migrating it**. The data sources (`Req` fields, tokenizer output) and the final consumers (`ForwardBatch`, the model forwards) are intact — only the *mid-pipeline plumbing* in `schedule_batch.py` was severed: `get_model_worker_batch` hardcodes every multimodal tensor to `None` / `False`. Multimodal understanding "looks like it runs" but silently ignores or corrupts the media.

This PR restores all of it (DP-aware) and consolidates the assembly into one helper. Each item below is a separate commit.

## What broke

All in `schedule_batch.py :: get_model_worker_batch`, hardcoded since `de5287ed`:

| Field | Hardcoded value | Symptom |
|---|---|---|
| `input_embedding` | `= None` | image/video/audio never reach the LM → output is effectively text-only, silently |
| `mrope_positions` | `= None` (main + spec path) | 3-D mRoPE degrades to 1-D → image-patch H/W spatial positions wrong on every VL request |
| deepstack | `apply_for_deepstack=False`, `deepstack_visual_embedding=None` | Qwen3-Omni visual deepstack enhancement never runs |

Once restored, each of those three would otherwise re-implement the same DP rank-offset loop, so they are consolidated into one helper (final commit).

Supporting facts: the per-DP container `ScheduleReqsInfo` has no multimodal fields, and `_compute_mrope_positions_for_batch` became dead code (zero callers). The pre-DP parent had all of these wired (`input_embedding=...`, `mrope_positions=...`, `apply_for_deepstack=...`). The scheduler ingest that fills `Req` is unchanged, so the data is still available — only the assembly into `ModelWorkerBatch` is missing.

The shared assembly layout is the same one `_merge_input_and_positions` uses: token slots are laid out `[dp0 | pad | dp1 | pad | ...]` with per-rank stride `per_dp_token_padding`; within a rank each req contributes its `[prefix_len, seq_len)` extend window. Every restored tensor is written on this layout so it aligns 1:1 with `input_ids`.

## Data flow: before vs. after

**Before** — `de5287ed` lets the data reach `Req`, then drops it at the batch-assembly boundary:

```
producer (vit / embed runner)      get_model_worker_batch          model forward

req.multimodal_embedding  -->  input_embedding     = None   -->  embed_tokens(input_ids)  [text-only]
req.mm_inputs[mrope_*]    -->  mrope_positions     = None   -->  1-D positions            [3-D mRoPE lost]
req.deepstack_*           -->  apply_for_deepstack = False  -->  deepstack skipped
                               ^ data is on Req, but hardcoded away here
```

**After** — one DP-interleaved pass re-assembles all three, aligned 1:1 with `input_ids`:

```
producer (vit / embed runner)      get_model_worker_batch          model forward

req.multimodal_embedding  --+
req.mm_inputs[mrope_*]      +-->  _merge_multimodal()  -->  input_embedding  -->  hidden states
req.deepstack_*           --+     single DP-interleaved      mrope_positions  -->  3-D mRoPE positions
                                  pass over reqs_info         deepstack + flag -->  injected at visual pos
```

---

### input_embedding

**Problem.** `req.multimodal_embedding` (the per-token text+visual merged embedding produced by the ViT / embed runner) is never copied into `ModelWorkerBatch.input_embedding`, so `forward_batch.input_embedding` is always `None` and the LM falls back to `embed_tokens(input_ids)` on prefill — silently text-only.

**Fix.** Slice each req's `[prefix:seq]` window into the DP-interleaved buffer aligned with `input_ids`. Returns `None` for pure-text batches (0-diff: the consumer keeps its `embed_tokens` path).

### mrope_positions

**Problem.** Both the main and spec paths hardcode `mrope_positions=None`, and the only assembler (`_compute_mrope_positions_for_batch`) is dead code. The consumer falls back to 1-D positions, so 3-D mRoPE degenerates to standard RoPE — not just the video time axis, but image-patch H/W spatial positions are wrong on every Qwen2.5-VL / Qwen3-Omni request.

**Fix.** Assemble `[3, total_tokens]` on the same layout: extend slices `mm_positions[:, prefix:prefix+ext]` (delta / arange fallback when absent); decode advances `seq_len-1 (+mrope_position_delta)`. **Main path only.** The spec path's `input_ids`/`positions`/`cache_loc` are empty placeholders rebuilt later by `EagleDraftWorker.padding_for_decode`, so its `None` is intentional and spec+mrope is a separate follow-up.

### deepstack (Qwen3-Omni)

**Problem.** `apply_for_deepstack` / `deepstack_visual_embedding` are hardcoded off and the pre-DP densification scatter was deleted. Naively flipping the flag back on (passing the sparse `(num_layers, num_visual, hidden)` tensor through) would make the thinker add visual features to **all** tokens — turning "missing feature" into "corrupts every token".

**Fix.** DP-aware densification: scatter the sparse visual rows into a dense `(num_layers, total_tokens, hidden)` buffer with non-visual rows left zero. Then the thinker's "add `deepstack[layer]` to all tokens" is mathematically equivalent to "add at visual positions only" — so no separate `deepstack_visual_pos_mask` carrier field is needed on `ForwardBatch`. `vstart` (= #visual-before-prefix) keeps chunked prefill correct.

### Consolidation — `_merge_multimodal`

**Problem.** The three restored paths above each carried an identical copy of the DP rank-offset loop. Three copies of one layout is a drift hazard: if any one drops the offset advance or miscounts `local`, that tensor silently misaligns with `input_ids`.

**Fix.** Collapse them into a single `_merge_multimodal(per_dp_token_size, total_token_size)` that traverses `reqs_info[*].reqs` once and emits all three tensors on one shared `offset`/`local` cursor. Data stays on `Req` (no new `ScheduleReqsInfo` fields); the helper only reads it. The dead `_compute_mrope_positions_for_batch` is folded in and removed.

## Not addressed here (dp>1 follow-ups)

- **Sharding**: `input_embedding` / `mrope_positions` / `deepstack` use `PartitionSpec(None, None)` (full replication) in `forward_batch_info.py`, while `positions` / `hidden_states` use `P("data")`. Harmless at `dp_size=1` (the current multimodal deployment — LM stage `max_batch_size:1`); at `dp_size>1` the replicated tensors fed into a `P("data","tensor")` attention may reshard / trip `.at[].set()` under jax 0.8.x explicit sharding.
- `_forward_mrope` uses a global `num_tokens` slice; correct as long as `mrope_positions` is assembled in DP-interleaved order (it is), but worth revisiting for `dp>1`.

### TPU e2e validation (Qwen2.5-VL-7B-Instruct, v6e-4)

Ran on a single-host **v6e-4** GKE node pool: this PR vs `main`, same pod / model / request — only the sglang-jax commit changed.

- Image `jax-ai-image/tpu:jax0.8.1-rev1` (jax 0.8.1), launched `--multimodal --trust-remote-code`, `temperature=0`.
- Env note: the image ships no `torch`; HF `AutoProcessor` for Qwen2.5-VL pulls in `AutoVideoProcessor` (hard-requires PyTorch), the exception is swallowed → `mm_processor=None` → every image request 400s (`"processor/config is not available"`) and the default warmup failure kills the process. `pip install torch torchvision` (CPU; preprocessing only, model still runs on JAX/TPU) resolves it — possibly worth documenting.

**Probe 1** — COCO `000000039769.jpg` (two cats sleeping on a couch) + *"How many cats are in this image, and what are they doing? Answer briefly."*

| Branch | Output |
|---|---|
| **This PR** (`1cb8f0a`) | "There are two cats in the image, and they are sleeping." ✅ |
| `main` (`1c38158`) | "There are no cats in this image. The image shows a group of people standing in a line, with some holding signs..." ❌ |

**Probe 2** — SGLang logo image, *"describe this image in one sentence."*

| Branch | Output |
|---|---|
| **This PR** | correctly describes the logo (letters + figure with code/book) ✅ |
| `main` | "a person riding a bicycle on a paved path in a park" ❌ |

**Key signal:** `prompt_tokens` is **429 on both branches** — the image placeholder tokens are present, but on `main` the visual embeddings are never merged into `input_embedding` (exactly the data-flow this PR restores). So placeholder positions get empty embeddings and the model generates from text only → consistent hallucination on both probes. With this PR the embeddings are injected and outputs are correct, confirming the `input_embedding` fix end-to-end on real hardware.

Isolation: `main` here is 4 commits behind this PR and 1 ahead of the merge-base (`6495488`); that extra commit is an unrelated quant change (#1259), orthogonal to the multimodal data-flow — merge-base gives identical results. `dp_size=1`, so the dp>1 sharding follow-ups noted above are not exercised.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -2390,7 +2390,48 @@ class ScheduleBatch:
             # Pad to total_bs
             lora_ids = lora_ids + ["0"] * (total_bs - real_bs)
 
-        # input_embedding = None
+        # Reassemble multimodal input_embedding (DP-aware).
+        #
+        # Mirrors the per-req slice/concat/pad logic from 785f3806 but scatters
+        # into the DP-interleaved padded token layout produced by
+        # _merge_input_and_positions: per-DP-rank slot stride is
+        # per_dp_token_padding, and within a rank the per-req EXTEND window is
+        # [prefix_len, seq_len) -- the same zip(info.seq_lens, info.prefix_lens)
+        # order used to build positions. The result aligns 1:1 with
+        # input_ids_cpu so forward_batch.input_embedding[k] matches
+        # input_ids[k]; trailing padding rows stay zero.
+        input_embedding = None
+        if self.forward_mode.is_extend() and any(
+            getattr(req, "multimodal_embedding", None) is not None
+            for info in self.reqs_info
+            if info.reqs
+            for req in info.reqs
+        ):
+            emb = None
+            offset = 0
+            for dp_rank in range(self.dp_size):
+                info = self.reqs_info[dp_rank]
+                if not info.reqs or info.seq_lens is None or len(info.seq_lens) == 0:
+                    offset += per_dp_token_padding
+                    continue
+                local = 0
+                for req, seq_len, prefix_len in zip(info.reqs, info.seq_lens, info.prefix_lens):
+                    ext_len = int(seq_len) - int(prefix_len)
+                    mm = getattr(req, "multimodal_embedding", None)
+                    if mm is not None and ext_len > 0:
+                        mm_full = np.asarray(mm)
+                        start = int(prefix_len or 0)
+                        end = start + ext_len
+                        chunk = mm_full[start:end]
+                        if emb is None:
+                            emb = np.zeros(
+                                (total_token_size, mm_full.shape[1]),
+                                dtype=mm_full.dtype,
+                            )
+                        emb[offset + local : offset + local + chunk.shape[0]] = chunk
+                    local += ext_len
+                offset += per_dp_token_padding
+            input_embedding = emb
 
         # Merge per-DP top_logprobs_nums / token_ids_logprobs with the same
         # offset_bs += per_dp_bs_padding padding scheme used in _merge_batch_metadata.
@@ -2496,7 +2537,7 @@ class ScheduleBatch:
             dp_size=self.dp_size,
             per_dp_bs_size=per_dp_bs_padding,
             launch_done=self.launch_done,
-            input_embedding=None,
+            input_embedding=input_embedding,
             apply_for_deepstack=False,
             deepstack_visual_embedding=None,
             recurrent_indices=recurrent_indices_cpu,
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -1823,25 +1823,35 @@ class ScheduleBatch:
 
         return input_ids_cpu, positions_cpu, out_cache_loc_cpu, real_input_ids_len
 
-    def _merge_mrope_positions(
+    def _merge_multimodal(
         self,
         per_dp_token_size: int,
         total_token_size: int,
-    ) -> np.ndarray | None:
-        """Merge 3-D mrope_positions from all DP ranks, aligned with input_ids.
+    ) -> dict:
+        """Assemble all per-token multimodal tensors in one DP-interleaved pass.
 
-        Mirrors the DP-interleaved rank-offset layout of
-        ``_merge_input_and_positions``: each DP rank occupies a
-        ``per_dp_token_size`` slot (offset advances per rank, padding columns
-        stay zero), and within a rank per-req positions are written in the same
-        ``[prefix_len, seq_len)`` extend-window order used to build input_ids,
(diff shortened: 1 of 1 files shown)


pengchengneo (dismissed):

pengchengneo:
LGTM

JamesBrianD (approved):

PR #1246, opened 2026-05-28
feat(ci): add 5 nightly 4-TPU accuracy gates (#1200)

Rodrian7:
## Summary

Implements section 1.1 of #1200 (#1117 nightly revision): per-architecture
gsm8k accuracy gates on v6e-4 covering dense, MLA, MoE, linear-attention and
FP8 paths. Single-host; cases are declarative entries in a suite registry
driven by a runner script, sharing the multi-host data model — not unittest files.

Refs: #1200, #1117

## Layout (all nightly code under `test/srt/nightly/`)

```
test/srt/nightly/
  cases.py / profiles.py / results.py   # host-neutral contract (case data, launch
                                         # profile, accuracy_result.v1.yaml emitter,
                                         # shared run_eval driver)
  schemas/accuracy_result.v1.yaml
  launch_profiles/*.yaml                 # one complete profile per config
  single_host/  suite_runner.py + accuracy_case_runner.py
  multi_host/   suite_runner.py + accuracy_case_runner.py + ... (re-exports the
                                         # shared contract; behaviour unchanged)
```

`run_eval.py` / `test_utils.py` stay at `test/srt/` (shared with pr-test).

## Data model (mirrors multi-host)

`SingleHostRun(launch_profile, cases)` is the single-host analogue of the
multi-host `ModelRun(launch_profile, cases)`: one server launch, one or more
cases evaluated against it. **All server args live in the profile YAML — one
profile per config** (e.g. epmoe vs fused are separate profiles); there is no
per-run override layer. The eval driver (`run_eval_for_case`) is shared in
`results.py` so single- and multi-host do not duplicate it. The only inherent
single-/multi-host difference is the launch runtime (single-host `nnodes=1` /
localhost dist-init vs multi-host GKE), carried by `RuntimeConfig`.

## Cases (`accuracy-text-models-v6e-4` suite)

Each fills the v6e-4 pod and runs sequentially.

| case_key | model | backend / config |
|---|---|---|
| `qwen3-8b-fa` | Qwen3-8B | fa, TP4 |
| `ds-v2-lite-mla` / `ds-v2-lite-fa-gqa` | DeepSeek-Coder-V2-Lite-Instruct | fa (MLA) / fa_mha (GQA), DP4×TP4 |
| `qwen3-moe-epmoe` / `qwen3-moe-fused` | Qwen3-MoE-30B-A3B | epmoe / fused, TP4/EP4 |
[...]

Issue #1200, opened 2026-05-25
[Nightly] Section 1.1: 4-tpu accuracy tests for production model architectures


JamesBrianD:
> Sub-issue of #1117. Implements section 1.1 of the proposed nightly revision plan.

### Objective
Build per-architecture accuracy gates on v6e-4 covering dense, MLA, MoE, linear-attention, and FP8 paths.
One file per model; each test function covers a single orthogonal axis (backend cross OR parallel strategy).
gsm8k is the only signal — mmlu/mmlu_pro stay out of daily.

### Gating
**Mandatory** — failure blocks nightly success.
Slash command on PR: `/run-nightly accuracy.<case_key>` re-runs a single case against PR HEAD.

### Tasks
- [ ] Land shared accuracy runner abstraction (modeled on sglang `AccuracyTestParams` + `run_combined_tests`)
- [ ] 1.1.1 `test_qwen3_8b_backends`: Qwen3-8B × {fa × TP4, native × TP4} on gsm8k
- [ ] 1.1.2 `test_ds_v2_lite_backends`: DeepSeek-V2-Lite × {mla + DP × TP4, fa(GQA) + DP × TP4} on gsm8k
- [ ] 1.1.3 `test_qwen3_moe_backends`: Qwen3-MoE-30B-A3B × {epmoe × TP2 EP2, fused × TP2 EP2} on gsm8k
- [ ] 1.1.4 `test_kimi_linear`: Kimi-Linear-48B-A3B-Instruct × hybrid_linear × TP4 on gsm8k
- [ ] 1.1.5 `test_qwen3_moe_fp8`: Qwen3-MoE-30B-A3B-FP8 × production-path variant on gsm8k (≤2% drift vs bf16)
- [ ] Wire all five into `run_suite.py` nightly suite
- [ ] Each case registers a unique `case_key` for slash-command targeting

### Dependencies
- MLA backend coupling with DeepSeek-V2-Lite (untested in current nightly)
- FP8 weight loader validation for Qwen3-MoE
- `KIMI_LINEAR` / `DEEPSEEK_V2_LITE` / `QWEN3_MOE_30B_FP8` constants in `test_utils.py`
- Slash command infra (sibling sub-issue)

### Contributor checklist (per case PR)
- [ ] **Threshold rationale**: how was the gsm8k baseline derived? (number of runs, mean/std, hardware, warmup config)
- [ ] **Estimated wall-clock**: per-case time including server warmup + eval, with measurement source
[...]

PR #1291, opened 2026-06-03
fix(quant): make qwen3_moe weight mapping quantization-aware for static FP8 (#1287)

### LINKED ISSUE #1287: [Bug] Qwen3-MoE static FP8 checkpoint fails to load: q_proj.weight is not a valid param path
## Summary

Loading a static FP8 Qwen3-MoE checkpoint (e.g. `Qwen/Qwen3-30B-A3B-FP8`) crashes during weight loading. After the FP8 structure-prep swaps linear layers to `QuantizedLinear`, the weight mappings still target the un-quantized param name `q_proj.weight`, which no longer exists, so `_get_param` raises.

## Reproduce

```
python -m sgl_jax.launch_server --trust-remote-code \
  --model-path Qwen/Qwen3-30B-A3B-FP8 \
  --tp-size 4 --ep-size 4 --moe-backend epmoe --device tpu \
  --dtype bfloat16 --attention-backend fa --page-size 128
```

Crashes at layer 0 (3/435 weights loaded):

```
Auto-detected FP8 model. Creating QuantizationConfig for static fp8.
Applying STATIC quantization structure preparation... -> Quantization complete.
Fast load failed for model.layers.0.self_attn.q_proj.weight, falling back to slow path.
  Error: model.layers.0.self_attn.q_proj.weight is not a valid param path
ValueError: model.layers.0.self_attn.q_proj.weight is not a valid param path
  (weight_utils.py:2801 _get_param)
```

## Root cause

1. The checkpoint quantizes attention q/k/v/o (and experts) in block-wise FP8 (`config.json`: `quant_method=fp8`, `fmt=e4m3`, `weight_block_size=[128,128]`; `modules_to_not_convert` lists only layernorms, `mlp.gate`, `lm_head`). So `q_proj` is *correctly* quantized.
2. `model_config.py` auto-FP8 emits a blanket linear rule `module_path=".*"`, and `apply_linear_quantization` (`utils/quantization/quantization_utils.py`) replaces every `LinearBase` with `QuantizedLinear`. `QuantizedLinear`'s params are `weight_q` + `weight_scale` (`layers/linear.py:236-237`) — there is no `.weight`.
3. But the weight mapping in `models/qwen3_moe.py:469-470` still targets `q_proj.weight`. Both the fast and slow load paths resolve to that name, which no longer exists post-quant → `_get_param` raises "not a valid param path".

In short: the qwen3_moe weight-mapping builder is not quantization-aware. The block-wise FP8 path exists for DeepSeek-V3 (`fp8_deepseek_v3.yaml`) but was never wired into qwen3_moe.

## Suggested fix

Make `_create_qwen3_moe_weight_mappings` branch on quantization: for quantized linears, emit `*.weight_q` as the target (plus a mapping for the checkpoint's `*.weight_scale` / `weight_scale_inv` → `*.weight_scale`) so the targets match `QuantizedLinear`'s param tree.

## Impact

Blocks the `qwen3-moe-fp8` nightly accuracy gate (#1200 §1.1 / PR #1246); that case is kept in the suite with a placeholder threshold until this is fixed.

Refs #1200 #1246

Rodrian7:
## Motivation

Closes #1287. Loading a static block-wise FP8 Qwen3-MoE checkpoint (`Qwen/Qwen3-30B-A3B-FP8`) crashes during weight loading.

The static-FP8 structure prep swaps every `LinearBase` for a `QuantizedLinear` (params `weight_q` + `weight_scale`) **before** `load_weights` runs, but `qwen3_moe`'s weight mapping still targeted the now-removed `*.weight`, so `_get_param` raised `... is not a valid param path` at layer 0 (3/435 weights loaded).

## Modifications

- **`models/qwen3_moe.py`** — make `_create_moe_layer_mappings` quantization-aware, mirroring DeepSeek-V3: for static-quant linears emit `*.weight_q` plus a `weight_scale_inv -> weight_scale` sidecar (sharding axes swapped, `transpose=False`), and register per-expert MoE scale sidecars for **both `epmoe` and `fused`** backends (fused puts the scale on the model mesh with transpose, following `mimo_v2_flash`). `weight_q` keeps `kv_head_padding` for GQA; the linear scale sidecar does not. Layernorms, q/k norms and the router gate stay bf16.
- **`configs/model_config.py`** — set `allow_narrow_n_blockwise` on the auto-detected static block-wise FP8 config, and update the narrow-N guard notes (`should_use_blockwise_kernel` / `kernel.py`): the collapse/NaNs are specific to the **dynamic online-quant** path (scales computed at load time); loaded static checkpoints ship correct block scales and pass through. Dynamic quant uses a separate config path and stays guarded.

## Accuracy (v6e-4, TP4/EP4, `Qwen3-30B-A3B-FP8`, full GSM8K 1319)

| backend | GSM8K |
| --- | --- |
| epmoe | **0.931** |
| fused | **0.935** |
| bf16 baseline | 0.938 |

All 435 weights load; no NaNs. FP8 static ≈ bf16 on both backends. Unblocks the `qwen3-moe-fp8` nightly accuracy gate (#1200 §1.1.5 / PR #1246).

### Launch config used (pure auto-detect, no `--quantization-config-path`)

```
python -m sgl_jax.launch_server --trust-remote-code --device tpu \
  --model-path Qwen/Qwen3-30B-A3B-FP8 --download-dir /dev/shm/ \
  --tp-size 4 --ep-size 4 --moe-backend {epmoe|fused} \
  --dtype bfloat16 --attention-backend fa --page-size 64 \
  --skip-server-warmup --random-seed 3 \
  --mem-fraction-static 0.9 --max-prefill-tokens 8192 \
  --max-running-requests 128 --precompile-bs-paddings 128 --precompile-token-paddings 8192
```

`fused` requires `--max-running-requests` and `--precompile-bs-paddings` to be multiples of `ep_size` (the fused MoE kernel asserts `num_tokens % ep_size == 0`); `epmoe` has no such constraint. Concurrency/`mem-fraction-static` follow Brian's #1246 review suggestions and are perf knobs (no effect on the score).

## Scope

`qwen2_moe` and the `qwen3_omni` thinker share the same latent mapping gap, but neither has an official FP8 checkpoint to validate against, so they are out of scope.

Refs #1200 #1246

### DIFF
--- a/python/sgl_jax/srt/configs/model_config.py
+++ b/python/sgl_jax/srt/configs/model_config.py
@@ -286,7 +286,15 @@ class ModelConfig:
                     moe_activation_dtype=None,
                     ignored_layers=ignored_layers,
                     weight_block_size=weight_block_size,
+                    # Static block scales are correct; the narrow-N guard targets
+                    # the dynamic online-quant path (computed at load time).
+                    allow_narrow_n_blockwise=weight_block_size is not None,
                 )
+                if weight_block_size is not None:
+                    logger.info(
+                        "Static block-wise FP8 (block=%s): enabling allow_narrow_n_blockwise.",
+                        weight_block_size,
+                    )
                 return quant_config
 
             elif quant_method == "compressed-tensors":
--- a/python/sgl_jax/srt/models/qwen3_moe.py
+++ b/python/sgl_jax/srt/models/qwen3_moe.py
@@ -440,65 +440,99 @@ class Qwen3MoeForCausalLM(nnx.Module):
                 target_path="lm_head.embedding", sharding=("tensor", None), transpose=False
             )
 
+        quant_config = getattr(self.config, "quantization_config", None)
+        is_static_quant = quant_config is not None and getattr(
+            quant_config, "is_static_checkpoint", False
+        )
+
         num_layers = self.config.num_hidden_layers
         mlp_only_layers = getattr(self.config, "mlp_only_layers", [])
 
         for layer_idx in range(num_layers):
             layer_mappings = self._create_moe_layer_mappings(
-                layer_idx, layer_idx in mlp_only_layers
+                layer_idx, layer_idx in mlp_only_layers, is_static_quant
             )
             mappings.update(layer_mappings)
 
         return mappings
 
-    def _create_moe_layer_mappings(self, layer_idx: int, is_mlp_layer: bool) -> dict:
+    def _create_moe_layer_mappings(
+        self, layer_idx: int, is_mlp_layer: bool, is_static_quant: bool = False
+    ) -> dict:
         prefix = f"model.layers.{layer_idx}"
         target_prefix = f"model.layers.{layer_idx}"
 
-        mappings = {
-            f"{prefix}.input_layernorm.weight": WeightMapping(
-                target_path=f"{target_prefix}.input_layernorm.scale",
-                sharding=(None,),
+        mappings: dict = {}
+
+        def add_linear(hf_prefix, target_prefix, sharding_std, kv_head_padding=False):
+            # Static FP8 loads QuantizedLinear.weight_q `[out, in]` directly
+            # (transpose off, kernel axes swapped) + weight_scale_inv sidecar.
+            if not is_static_quant:
+                mappings[f"{hf_prefix}.weight"] = WeightMapping(
+                    target_path=f"{target_prefix}.weight",
+                    sharding=sharding_std,
+                    transpose=True,
+                    kv_head_padding=kv_head_padding,
+                )
+                return
+            sharding_quant = (sharding_std[1], sharding_std[0])
+            mappings[f"{hf_prefix}.weight"] = WeightMapping(
+                target_path=f"{target_prefix}.weight_q",
+                sharding=sharding_quant,
                 transpose=False,
-            ),
-            f"{prefix}.post_attention_layernorm.weight": WeightMapping(
-                target_path=f"{target_prefix}.post_attention_layernorm.scale",
-                sharding=(None,),
+                kv_head_padding=kv_head_padding,
+            )
+            # No kv_head_padding on the scale: expand handles the un-replicated n_out.
+            mappings[f"{hf_prefix}.weight_scale_inv"] = WeightMapping(
+                target_path=f"{target_prefix}.weight_scale",
+                sharding=sharding_quant,
                 transpose=False,
-            ),
-            f"{prefix}.self_attn.q_proj.weight": WeightMapping(
-                target_path=f"{target_prefix}.self_attn.q_proj.weight",
-                sharding=(None, "tensor"),
-                transpose=True,
-            ),
-            f"{prefix}.self_attn.k_proj.weight": WeightMapping(
-                target_path=f"{target_prefix}.self_attn.k_proj.weight",
-                sharding=(None, "tensor"),
-                transpose=True,
-                kv_head_padding=True,
-            ),
-            f"{prefix}.self_attn.v_proj.weight": WeightMapping(
-                target_path=f"{target_prefix}.self_attn.v_proj.weight",
-                sharding=(None, "tensor"),
-                transpose=True,
-                kv_head_padding=True,
-            ),
-            f"{prefix}.self_attn.o_proj.weight": WeightMapping(
-                target_path=f"{target_prefix}.self_attn.c_proj.weight",
-                sharding=("tensor", None),
-                transpose=True,
-            ),
-            f"{prefix}.self_attn.q_norm.weight": WeightMapping(
-                target_path=f"{target_prefix}.self_attn.q_norm.scale",
-                sharding=(None,),
-                transpose=False,
-            ),
-            f"{prefix}.self_attn.k_norm.weight": WeightMapping(
-                target_path=f"{target_prefix}.self_attn.k_norm.scale",
-                sharding=(None,),
-                transpose=False,
-            ),
-        }
+            )
+
+        mappings.update(
+            {
+                f"{prefix}.input_layernorm.weight": WeightMapping(
+                    target_path=f"{target_prefix}.input_layernorm.scale",
+                    sharding=(None,),
+                    transpose=False,
+                ),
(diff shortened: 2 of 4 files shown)


python/sgl_jax/srt/models/qwen2_moe.py
```diff
@@ -504,18 +504,57 @@ def _create_qwen2_moe_weight_mappings(self) -> dict:
...
             mappings.update(layer_mappings)
 
         return mappings
 
-    def _create_moe_layer_mappings(self, layer_idx: int) -> dict:
+    def _create_moe_layer_mappings(self, layer_idx: int, is_static_quant: bool = False) -> dict:
         prefix = f"model.layers.{layer_idx}"
         target_prefix = f"model.layers.{layer_idx}"
 
+        def add_linear(
```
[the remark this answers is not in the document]
Rodrian7: Resolved by removing qwen2_moe from this PR entirely. qwen2_moe (and qwen3_omni) have no official FP8 checkpoint to validate against, so their parity changes were reverted; this PR is now scoped to qwen3_moe only, where the `add_linear` helper is defined after `mappings: dict = {}` (matching deepseek_v3). The flagged footgun no longer exists here.

python/sgl_jax/srt/configs/model_config.py
```diff
@@ -286,6 +286,16 @@ def _resolve_quantization_config(self) -> QuantizationConfig | None:
...
+                    # Static block-wise FP8 checkpoints ship their own (correct)
+                    # block scales, so the narrow-N blockwise kernel is accurate
+                    # here — verified on Qwen3-30B-A3B-FP8 at TP=4 (full GSM8K
+                    # 0.936, matching bf16 0.938; no NaNs over 1319 samples).
+                    # The narrow-N guard targets the *dynamic* online-quant path:
+                    # its author confirmed the NaNs occur only with online
+                    # quantization of Qwen3-MoE block-wise (scales computed at
+                    # load time), which goes through a separate config path and
+                    # stays guarded. Loaded static checkpoints are unaffected.
+                    allow_narrow_n_blockwise=weight_block_size is not None,
```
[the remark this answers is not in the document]
Rodrian7: The guard author (cjx0709, who added the narrow-N check in f009b6e9) confirmed the accuracy collapse / NaNs are specific to the **dynamic online-quant** path — where the framework computes block scales at load time and gets them wrong for narrow-N. A loaded static checkpoint ships its own (offline-computed, correct) block scales, so the kernel is accurate there regardless of model. Hence the opt-in is keyed on static-checkpoint detection rather than a model allowlist. Validated end-to-end on Qwen3-30B-A3B-FP8 (full GSM8K 0.936 vs bf16 0.938, no NaNs). The kernel.py TODO predates that confirmation; I left it untouched to keep this PR scoped to the mapping fix, but it's worth updating separately.

python/sgl_jax/srt/configs/model_config.py
```diff
@@ -286,7 +286,21 @@ def _resolve_quantization_config(self) -> QuantizationConfig | None:
...
                 )
+                if weight_block_size is not None:
+                    logger.info(
+                        "Static block-wise FP8 detected (weight_block_size=%s); "
+                        "enabling allow_narrow_n_blockwise for the loaded checkpoint "
+                        "(applies to all auto-detected static block-wise FP8 models, "
+                        "not just Qwen3). Dynamic online-quant stays guarded.",
+                        weight_block_size,
+                    )
                 return quant_config
```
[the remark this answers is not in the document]
Rodrian7: Intentional asymmetry. Cross-referenced the codebase: no model, test, or nightly case exercises compressed-tensors **block** strategy — the only real compressed-tensors checkpoint (Ling-2.6-1T) is per-channel (strategy="channel" → weight_block_size=None), which never hits the narrow-N guard, so there is no inconsistency for any real path. The compressed-tensors block branch is defensive/unexercised. Opting it in would extend a guard-bypass to an unvalidated, non-existent scenario; leaving the guard in place is the fail-loud-safe default (a narrow-N compressed-tensors-block checkpoint would error rather than silently corrupt). If such a checkpoint appears, it should be validated end-to-end before opting in — same bar applied to the fp8 path here (Qwen3-30B-A3B-FP8).

python/sgl_jax/srt/configs/model_config.py
```diff
@@ -286,7 +286,15 @@ def _resolve_quantization_config(self) -> QuantizationConfig | None:
                     moe_activation_dtype=None,
                     ignored_layers=ignored_layers,
                     weight_block_size=weight_block_size,
+                    # Static block scales are correct; the narrow-N guard targets
+                    # the dynamic online-quant path (computed at load time).
+                    allow_narrow_n_blockwise=weight_block_size is not None,
```
cjx0709, line 291: should also update the notes in "should_use_blockwise_kernel" function that here static quant will be passed.

python/sgl_jax/srt/models/qwen3_moe.py
```diff
@@ -596,6 +620,23 @@ def _create_moe_layer_mappings(self, layer_idx: int, is_mlp_layer: bool) -> dict
             )
             mappings.update(moe_mappings)
 
+            # epmoe only; fused static FP8 is unsupported (matches deepseek_v3).
```
cjx0709: It needs to be clarified whether this is due to model structure constraints or because the code currently lacks an implementation. There are also numerous examples of the fused moe kernel that support FP8 quantization, such as mimo-v2-flash.

python/sgl_jax/srt/configs/model_config.py
Rodrian7: Done — updated the should_use_blockwise_kernel docstring and the kernel.py guard note to say the collapse is dynamic online-quant-only and static checkpoints pass through.

python/sgl_jax/srt/models/qwen3_moe.py
Rodrian7: It was missing impl, not a structural limit — wired fused expert scales following mimo_v2_flash; fused GSM8K ≈0.935, in line with epmoe/bf16.

cjx0709 (approved):
LGTM

PR #1292, opened 2026-06-03
[SpecDecode] enhance performance with one JIT function for all decode phase

pengchengneo:
image
fused all spec decode stage into one JIT function

  ## Accuracy Tests

image


zorrofox (commented):
Reviewed at aac7a6ed. Independent validation on v6e-64 dp=4 (16-host, tp=16 ep=64 epmoe, V2-Flash NEXTN, bs=32):

| round | dataset | accept-len | n_round | low(<1.2) |
|---|---|---|---|---|
| warmup | random n=32 out=128 | 2.475 | 68 | 2 |
| g1-g5 | gsm8k n=100 out=1024 | 3.30 / 3.38 / 3.35 / 3.38 / 3.38 | 855/750/852/831/827 | 8/0/2/3/0 |
| r1 | random n=64 out=1024 | 3.253 | 760 | 3 |

7 rounds, ~5k verify rounds, 0 assert/traceback, no cross-round degradation. GSM8K mean 3.36 matches the #1252 baseline (3.37-3.44) — so the cumprod chain-verify is functionally correct end-to-end. Findings below; given the hack-first scope I'd only flag the first one as worth a quick test before merge.

### HIGH

**`draft_extend_fused.py:102-105` — the cumprod chain-verify has no equivalence test**
The actual production verify in the fused JIT is `_greedy_sample_and_prepare_draft_inputs_chain_from_predict` (jnp `cumprod` over `draft[:,1:]==predict[:,:-1]`), which replaces the pallas `verify_tree_greedy` kernel. `test_eagle_utils.py:84-173` only tests `greedy_sample_device_outputs` — and that one still calls the (mocked-out) pallas kernel and is **dead code** (nothing in fused or fallback calls it). So the codepath that ships has zero test coverage against the original. The cumprod logic looks correct for topk=1, but worth a quick parity test (host `verify_tree_greedy` vs cumprod on a few synthetic `(draft, predict)` pairs incl. partial-accept and padding rows).

Relatedly: padding rows (`seq_lens==0`) get `draft_2d=0` and `predict=argmax(garbage)`; if the garbage argmax happens to be 0, `child_matches` is True → `accept_length[pad]>1` → L377 `draft_logits_metadata.accept_lens` carries that into `last_index` adjustment for the padding slot. Probably harmless (padding logits aren't read), but the original pallas path explicitly wrote `accept_length=0` for padding.

### MEDIUM

**`draft_extend_fused.py:522-576` — `_forward_batch_init_new_preserve_device` is a 60-line copy of `ForwardBatch.init_new`**
Field-for-field today, but any future field added to `ForwardBatch` silently drops here. Same for `_logits_metadata_..._preserve_device`. Suggest adding a `preserve_device: bool = False` kwarg to the originals instead (the only diff is `device_array` vs pass-through-if-already-device).

**`draft_extend_fused.py:740` — `new_seq_lens` is wrong by `+n-1`**
`seq_lens_host` here is `model_worker_batch.seq_lens` *after* L670 already added `n-1`. Result is `orig_seq + (n-1) + accept` vs original verify path's `(orig_seq - 1) + accept`. Currently `new_seq_lens` has no readers anywhere so it's latent — but either fix it (use `prepared.new_seq_lens` from inside the JIT, which is correct at L78) or drop the field.

**`draft_extend_fused.py:778-779` — `placeholder_hidden` is a per-round MB-scale dead h2d**
`np.zeros((bs*n, hidden), float32)` gets device_put via `ForwardBatch.init_new` at L814 (it's `mwb.spec_info_padded.hidden_states` after L690), then immediately overwritten at L383. Could be `(0, hidden)` or skip the field in the placeholder spec_info.

**`draft_extend_fused.py:357,201` — argmax/topk on sharded vocab axis**
`target_logits` is `P("data","tensor")` (vocab on tensor); `jnp.argmax(axis=-1)` relies on XLA inserting the all-reduce. #1252 explicitly resharded to `P()` first. This is the right perf direction (reduce-max is cheaper than all-gather-vocab) and your GSM8K 0.96 confirms it lowers correctly, just flagging the implicit dependency.

### nit

- `eagle_util.py:797-852` — `greedy_sample_device_outputs` + `GreedySampleDeviceOutputs` are dead (only the test calls them); delete or wire into fused
- `base_worker.py:81-86` — no env flag to disable fused (A/B requires code change)
- `draft_extend_fused.py:107,116-119` — `jnp.zeros_like(x) + jnp.arange(...)` sharding-hack reads oddly; a one-line comment would help
- `draft_extend_fused.py:56-61` vs `:116-121` — `safe_index` computed twice (once in `_greedy_prepare_draft_inputs`, once in caller)
- `test_eagle_utils.py:127-135` — mock `predict` shape `(2,5)` doesn't match `num_draft_tokens=4`; passes by reshape(-1) coincidence

### ✓

- P() reshard at L416-427 covers all 4 cross-process outputs; `target_logits/hidden` only resharded when actually returned
- target + N×draft pools all donated and `replace_all`'d
- remaining host roundtrips are the irreducible set: 2× `get_eagle_forward_metadata` (page_indices repack), `padding_for_decode` (cache_loc DP-segment), 4× d2h (accept/predict/hidden/idx)
- chain causal (`custom_mask=None`) correct for topk=1

zorrofox (commented):
Looked at the 4 split-phase commits (65ce6f1a..8765a046). Ran it on v6e-64 dp=4 bs=32 — **crashes at GSM8K round ~27** (after some reqs finish early and a rank goes through a fresh prefill):

```
File "schedule_batch.py", line 2286, in _concat_spec_info_per_rank
AssertionError: field 'new_seq_lens' is None on 1/4 nonempty rank(s); all-or-nothing required
```

Root cause is the new `new_seq_lens` plumbing in `schedule_batch.py` — see the first HIGH below. Aside from that, the interface split is clean and `next_draft_input` fields match the single-JIT path after `draft_extend_for_decode_fused` L657-664 rewrites. Two more structural blockers if the goal is actual overlap.

For context, I split-timed the single-JIT at the `prepared.*` boundary on v6e-64 dp=4 bs=32 (n=300, p50 ms):

| stage | p50 | share |
|---|---|---|
| host prep | 4.66 | 12.7% |
| jit1 (target verify+sample) | 29.58 | 80.9% |
| jit2 (dext 3-layer) | 1.68 | 4.6% |
| host mat | 0.70 | 1.9% |

So jit2 is tiny; overlap upper bound is ~min(jit2, prep) ≈ 1.7ms ≈ **5%** unless prep itself shrinks.

### HIGH

**`schedule_batch.py:2286` + `draft_extend_fused.py:956` — `new_seq_lens` all-or-nothing assert fires (crash)**
The split path sets `next_draft_input.new_seq_lens = prepared_new_seq_lens` (device, padded_bs) at L956, but `draft_extend_for_prefill` never sets it (stays None). Once a rank's reqs all finish (GSM8K short answers hit eos early) and that rank re-enters via prefill, `_concat_spec_info_per_rank` sees 3 ranks with `new_seq_lens` and 1 with None → assert. Repro: v6e-64 dp=4 bs=32 GSM8K n=100, crashes at round ~27. Doesn't fire on v6e-16 bs=32 (each rank has 8 reqs, less likely to all-finish before others).

Also: even when set, it's `(padded_bs,)` device while the sibling fields get rewritten to `(real_bs,)` host at L657-662, so `_split_spec_info_per_rank`'s real_bs cumsum gives rank>0 misaligned slices.

If `new_seq_lens` isn't actually consumed yet (couldn't find a reader), simplest is to drop it from the scatter/merge/concat lists in `schedule_batch.py:2208/2233/2270/2286`. If it's intended for the overlap scheduler, prefill dext needs to set it + decode path needs to host-gather it to `(real_bs,)` alongside the other fields.

**`draft_extend_fused.py:943-944` — overlap is structurally blocked**
`np.asarray(accept_lens_device)` inside `spec_decode_verify_phase` blocks until jit1 completes, *before* the function returns. So regardless of how the scheduler orders the calls, Phase B dispatch always happens after jit1 is done — no overlap possible. To unblock: return the device handles from Phase A, defer the `np.asarray` into `_publish_spec_verify_phase_lengths_to_batch`, and reorder scheduler to `verify_phase → draft_extend_phase(dispatch) → publish(block) → write_back(block)`.

**`draft_extend_fused.py:609`** — Phase B's `draft_extend_for_decode_fused` re-does `np.asarray(jax.device_get(batch_output.accept_lens))` (which is `accept_lens_device` from L967). Even if Phase A goes async, this is a second d2h barrier before jit2 dispatch. Either pass Phase A's already-hosted `accept_lens` through `draft_extend_state`, or — cleaner — carry `prepared.sel_pos` / `prepared.select_index` (already computed on-device inside jit1) in `draft_extend_state` and feed them straight into jit2, skipping the host round-trip entirely.

### MEDIUM

**`scheduler.py:1858-1860`** — in the split path `verify_result.accept_lens` is already host numpy (L943), so the `copy_to_host_async`/`device_get` is dead. (Same in the non-split path: async-then-immediately-asarray.)

### nit

- `draft_extend_fused.py:949-953` — `scheduler_next_draft_input` has no consumer anywhere (grep'd). If it's a placeholder for the overlap scheduler, a TODO would help; otherwise drop.
- `scheduler.py:1979` — `import` inside the hot loop.
- donate is correct: Phase A donates `target_memory_pools` (L761), Phase B reuses `_fused_jit_fn` which donates `all_memory_pools` (L216). No leak.

If this is just the interface split and overlap scheduling lands separately, the two HIGH items are exactly the seams that'll need to move — happy to re-time once those are async.

pengchengneo:
sorry, I pushed some scheduler-overlap compatible commit to this branch, 

I have reverted these commits, and fix you comment https://github.com/sgl-project/sglang-jax/pull/1292#pullrequestreview-4423993208 in 20bc3ce7

I have test e2e with gsm8k for two round, no crashed and get 0.95+scores

zorrofox (approved):
20bc3ce7 addresses both rounds — and I see the split-phase commits were dropped (force-push back to single-JIT), so the `new_seq_lens` concat crash and the overlap-blocker comments from my second review no longer apply.

Addressed:
- cumprod chain-verify now has direct unit coverage (`test_fused_chain_verify_matches_topk1_linear_reference`, accept=1/2/3/4) + padding-row test ✓
- padding rows: `is_padding = seq_lens==0` zeroes `accept_length` / `accept_index` ✓
- `_materialize` `new_seq_lens` off-by-(n-1) ✓
- `greedy_sample_device_outputs` dead code removed ✓

Remaining are the tech-debt items from the first round (`_forward_batch_init_new_preserve_device` copy, `placeholder_hidden` h2d, env flag) — fine to defer.

Didn't re-run TPU since the diff vs aac7a6ed is boundary-only (padding mask + one arithmetic fix + test), no forward-path change; aac7a6ed already passed 7 rounds at 3.30-3.38. LGTM.

PR #1294, opened 2026-06-03
fix(bailing/group_rmsnorm): run Ling-2.6 fused-v1 under multi-shard tensor parallel

JamesBrianD:
## Problem

`Ling-2.6` models (`BailingMoeV2_5`, `bailing_hybrid`, e.g. Ling-2.6-flash / 1T) cannot run on `main` at common multi-shard tensor-parallel layouts, and fused-v1 shared experts are routed inconsistently:

1. **GroupRMSNorm** required `tensor_parallel_size >= num_groups`. With `n_group=8` on 32 devices, any layout with effective tensor < 8 (e.g. `dp=8` → tensor=4) raised at construction; `tensor=8` then failed the head reshape sharding (`axis size 4 not divisible by tensor 8`). So no `dp>1` layout ran.
2. **Shared experts (fused v1)**: `FusedEPMoE` is built with `num_shared_experts=0` but the external `DeepseekV3MLP` was gated behind `and not use_fused`, so fused-v1 dropped shared experts entirely (and the weight loader still tried the in-kernel `w*_shared` mapping that has no params).

## Fix

Extracted from the `fused-moe-v2-with-sp-rs` branch (PR #1264):

- **`fix(group_rmsnorm)`**: require `tp` to **divide** `num_groups` (not `>=`) and rework the reshape sharding so each shard holds whole RMS groups — `tensor=4` now works with `num_groups=8`.
- **`bailing_moe_linear`**: route shared experts through the external `DeepseekV3MLP` for fused v1 too (unify with fused_v2 / epmoe); weight loader uses the external mapping (`use_fused_shared=False`).
- **fused_moe v1 tuned block_configs**: Ling-2.6-1T entries (`shared_expert=False`, `N=8192`, `N=16384` @ ep=32 fp8).

(`7f03d2e7` from the source set is already present on `main` — `quant_block_k` defaults to 256 + K-block scale replicate — so it is not duplicated here.)

## Validation

Ling-2.6-flash on v7x, 32 devices (`--tp-size 32 --dp-size 8 --ep-size 32 --moe-backend fused`, effective tensor=4):

| concurrency | gsm8k accuracy |
|---|---|
| parallel=1 (single) | 1.00 |
| parallel=64 (batched) | 0.96 |

Server launches cleanly (no GroupRMSNorm error) and serves correctly under concurrency.


cjx0709 (approved):
LGTM

PR #1295, opened 2026-06-03
fix(detokenize): incremental init_incremental_detokenize (no full-prompt reconcat)

JamesBrianD:
## Problem

`Req.init_incremental_detokenize()` rebuilt the entire `origin_input_ids_unpadded + output_ids` list on **every** call and then sliced off the tail — O(prompt + output) per call.

At high concurrency with long prompts (e.g. `max_concurrency=512`, 16384-token prefill), all running requests cross the force-stream boundary in the same scheduler tick, so this turns into a synchronous CPU burst on the single scheduler thread that stalls the decode pipeline. It shows up as a `stream_output_generation` hotspot in profiling.

## Fix

Port upstream sglang's incremental approach:

- Build `origin_input_ids_unpadded[surr_offset:] + output_ids` **once**, cache it in `self.surr_and_decode_ids`, then on subsequent calls **extend the cached buffer in place** with only the newly generated tail — O(new tokens) instead of O(prompt + output).
- Add `Req.output_ids_through_stop`: truncate `output_ids` at `finished_len` so detokenize never emits ids past the finish position (e.g. the extra delayed token under the overlap schedule).

No behavioral change to detokenized output; purely removes the per-call O(prompt) reconcat. Pre-commit (ruff/black/isort/mypy/codespell) clean.


cjx0709 (commented):
I don't see the test file "Adds test/srt/test_init_incremental_detokenize.py:", may missed it?

JamesBrianD:
> I don't see the test file "Adds test/srt/test_init_incremental_detokenize.py:", may missed it?

The content of this test file was generated by AI. Upon analysis, the test content was found to be overly redundant and incapable of effectively testing the actual runtime code modules; therefore, it has been deleted.

cjx0709:
Thanks for the clarification on the test file.

One remaining question about the `--skip-tokenizer-init` path: `output_ids_through_stop` is currently only applied to `decode_ids` via `init_incremental_detokenize()`. However, in `stream_output_generation`, when `self.skip_tokenizer_init` is true, `BatchTokenIDOut.output_ids` still uses the raw `req.output_ids[send_token_offset:]`, and `completion_tokens` still uses `len(req.output_ids)`.

If overlap scheduling can leave an extra delayed token after `finished_len`, the text detokenization path will trim it, but the token-id direct output path may still return/count that extra token. Is that intended behavior for `--skip-tokenizer-init`?

If the token-id output should match the visible/through-stop output, should we use `req.output_ids_through_stop` consistently for this path as well, e.g. for `output_ids`, `send_token_offset`, and `completion_tokens`?

JamesBrianD:
> Thanks for the clarification on the test file.
> 
> One remaining question about the `--skip-tokenizer-init` path: `output_ids_through_stop` is currently only applied to `decode_ids` via `init_incremental_detokenize()`. However, in `stream_output_generation`, when `self.skip_tokenizer_init` is true, `BatchTokenIDOut.output_ids` still uses the raw `req.output_ids[send_token_offset:]`, and `completion_tokens` still uses `len(req.output_ids)`.
> 
> If overlap scheduling can leave an extra delayed token after `finished_len`, the text detokenization path will trim it, but the token-id direct output path may still return/count that extra token. Is that intended behavior for `--skip-tokenizer-init`?
> 
> If the token-id output should match the visible/through-stop output, should we use `req.output_ids_through_stop` consistently for this path as well, e.g. for `output_ids`, `send_token_offset`, and `completion_tokens`?

Thank you for the suggestion. I consider your advice to be very important, and I have already implemented the fix based on your recommendation.

cjx0709 (approved):
LGTM

PR #1296, opened 2026-06-03
perf(scheduler): cut per-step host overhead — cache_loc buffer reuse, sync next_token_ids

JamesBrianD:
## What

Two scheduler-side perf changes that reduce per-step host (CPU) overhead on the
single scheduler thread. Cherry-picked from #1264 to land independently of the
fused-moe-v2 work in that PR.

> The incremental-detokenize change originally on this branch has been dropped —
> it is already covered by #1295.

### Commits

1. **`overlap`: materialize `next_token_ids` synchronously via `device_get`**
   Replaces the async `copy_to_host` of `next_token_ids` with a blocking
   `jax.device_get(next_token_ids).tolist()` in `ModelWorkerClient`'s resolve
   path. logprobs / input_logprobs / hidden_states keep the async `copy_to_host`
   pipeline.

2. **`perf(schedule)`: reuse a persistent host buffer for `cache_loc` (no per-step `np.zeros`)**
   `_merge_cache_loc` allocated a fresh `np.zeros(total_cache_loc_size)` every
   step, paying malloc/munmap churn + first-touch page faults on every written
   page; on long-context / high-concurrency decode the padded `cache_loc` table
   is multi-MB. Now `ReqToTokenPool` owns a `cache_loc_host_buf` that `tp_worker`
   allocates **once at startup**, sized to the largest cache_loc precompile
   bucket (`cache_loc_buckets[-1]`, the earliest point that size is known).
   `_merge_cache_loc` just slices `buf[:total]` — no per-step allocation, and no
   per-step zeroing: padding slots are never read on-device (attention kernels
   truncate KV via `seq_lens` / `cu_kv_lens`), and real-request slots are
   overwritten by the scatter loop. The per-req page-aligned scatter is unchanged.
   Reuse is safe: `ForwardBatch.init_new`'s H2D copy runs synchronously on the
   scheduler thread before the next step rebuilds the buffer, and the resulting
   device array does not alias the host buffer.

## Test

- `python -m py_compile` clean on all changed files; `mypy` pre-commit passes.

### Accuracy validation (MiMo-V2-Flash)

Ran this branch (clean, off `main`) on **MiMo-V2-Flash** (fp8 block-128) on a v7x
32-device pod (`--moe-backend epmoe`, tp=32 / dp=8 / ep=32) and measured GSM8K
accuracy (thinking mode, chat template, `temperature=0.6 top_p=0.95`,
`max_new_tokens=6144`):

| concurrency | GSM8K accuracy |
|---|---|
| parallel=1 (single-stream) | 1.00 (12/12) |
| parallel=64 (batched) | 0.95 (95/100) |

Batched ≈ single-stream — no degeneration under concurrency, and consistent with
MiMo-V2-Flash's known GSM8K level (~0.96). The per-step `cache_loc` buffer reuse
and synchronous `next_token_ids` materialization do not change output correctness.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -1937,7 +1937,16 @@ class ScheduleBatch:
             total_cache_loc_size = cache_loc_paddings[bs_index]
 
         per_dp_cache_loc_size = total_cache_loc_size // self.dp_size
-        cache_loc_cpu = np.zeros(total_cache_loc_size, dtype=np.int32)
+        # View into the persistent buffer; intentionally NOT re-zeroed per step.
+        # Safe because:
+        #  - padding slots are never read on-device: attention kernels (RPA v3 /
+        #    MLA v2 / native) bound page reads by cu_kv_lens / seq_lens, and every
+        #    real-request page slot lands on a written position.
+        #  - every buffer value is a valid in-bounds KV slot index (init is
+        #    np.zeros + only valid slots are ever written), so even SWA's
+        #    host-side mapping[cache_loc] lookup (flashattention_backend) can't go
+        #    OOB. This REQUIRES the init buffer to be np.zeros, not np.empty.
+        cache_loc_cpu = self.req_to_token_pool.cache_loc_host_buf[:total_cache_loc_size]
 
         offset_bs = 0
         req_to_token = self.req_to_token_pool.req_to_token
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -1946,7 +1946,15 @@ class ScheduleBatch:
         #    np.zeros + only valid slots are ever written), so even SWA's
         #    host-side mapping[cache_loc] lookup (flashattention_backend) can't go
         #    OOB. This REQUIRES the init buffer to be np.zeros, not np.empty.
-        cache_loc_cpu = self.req_to_token_pool.cache_loc_host_buf[:total_cache_loc_size]
+        cache_loc_host_buf = self.req_to_token_pool.cache_loc_host_buf
+        assert (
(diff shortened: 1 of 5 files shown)


cjx0709:
Overall this optimization makes sense to me. It is mostly reducing scheduler/thread-side Python/NumPy per-step overhead.

One shared-state edge case worth fixing/confirming: `ReqToTokenPool.cache_loc_host_buf` lives on the pool, but speculative/EAGLE draft workers reuse the target worker’s `req_to_token_pool`. With the current implementation, `init_cache_loc_host_buffer()` always replaces the buffer:

```python
self.cache_loc_host_buf = np.zeros(size, dtype=np.int32)
```

That is fine for the normal path, but in speculative mode the target worker first initializes the buffer using the target `cache_loc_buckets[-1]`, then the draft worker initializes the same shared pool again. If the draft bucket is smaller, this can shrink the shared buffer. Later, target `_merge_cache_loc()` still slices by the target-side `total_cache_loc_size`; NumPy slicing past the end will silently return a shorter view, which could lead to a `cache_loc` shape mismatch with the precompiled shape / attention metadata.

Could we make this an ensure-capacity API instead of unconditional replacement?

```python
def init_cache_loc_host_buffer(self, size: int):
    if self.cache_loc_host_buf is None or self.cache_loc_host_buf.shape[0] < size:
        self.cache_loc_host_buf = np.zeros(size, dtype=np.int32)
```

Optionally, `_merge_cache_loc()` could also assert the capacity before slicing. This keeps the normal-path optimization while avoiding the shared-pool speculative-worker overwrite case.

JamesBrianD:
@cjx0709 Fixed in `eb1565bf`: `init_cache_loc_host_buffer()` is now an ensure-capacity API, so a draft/speculative worker cannot shrink the shared target pool buffer. `_merge_cache_loc()` also asserts the host buffer is initialized and large enough before slicing. Added `ReqToTokenPool` tests covering no-shrink and grow behavior.

cjx0709 (approved):
LGTM

PR #1302, opened 2026-06-03
feat(multimodal): Add MiMo-V2.5 vision encoder

zhengkezhou1:
## Motivation

Support for the MiMo-V2.5 vision encoder.

MiMo-V2.5 uses a 28-block ViT vision tower with hybrid full / row-window / column-window attention, GQA, and attention sinks. 

## Modifications

- Implemented MiMo vision components:
  - 3D patch embedding
  - 2D RoPE helpers
  - SwiGLU MLP
  - patch merger
  - full/window vision attention
  - grouped query attention
  - attention sinks
  - MiMoVisionTransformer
- Added `visual.*` weight mapping for the standalone vision tower.
- Added an E2E MiMo vision encoder alignment test under `test/srt/multimodal`.

## Accuracy Tests

- loads the real MiMo-V2.5 vision checkpoint weights
- loads the HF remote-code `MiMoVisionTransformer` as reference
- runs the JAX `MiMoVisionTransformer` under a JAX mesh
- compares JAX output against HF output across named grid cases:
  - `mixed_temporal_and_multiple_segments`
  - `regular_temporal_video`
  - `large_grid_crosses_window_boundary`
  - `wide_aspect_ratio_col_order`

## Benchmarking and Profiling

Not applicable. This PR adds model support and correctness coverage for the MiMo vision encoder; it does not include performance tuning.

## Checklist


JamesBrianD:
I took a look at this PR. At the moment, it mainly provides a standalone validation for the MiMo-V2.5 vision encoder / ViT path, showing that it can align with the HF vision tower. That is valuable work.

However, from the perspective of merging into `main`, I think we should be a bit more cautious. This has not yet been integrated with the existing end-to-end multimodal inference flow in sgl-jax. For example, paths such as model registration, processor integration, the serving request flow, weight loading in the production path, and how the downstream text model consumes the vision outputs have not been validated yet.

So I would suggest that we do not rush to merge this PR as-is. It would be better to first investigate and design the overall MiMo-V2.5 multimodal support plan: how the vision encoder should plug into the current pipeline, what end-to-end tests should cover, and what the path to production would look like. Once we are confident that the final solution is runnable, testable, and deployable end to end, we can then consider splitting the work and merging it incrementally.

This PR can serve as a useful prototype / validation foundation for the vision encoder, but I would prefer that we first complete the end-to-end design and validation plan before deciding whether to merge it into `main`.

PR #1306, opened 2026-06-04
test(eval): add sampler params + reasoning-trace stripping to simple-evals

MokusMokun:
## Motivation

The Qwen3.5 thinking-mode evaluation in `test/srt/test_qwen3_5_models.py` needs to specify its intended generation configuration, but the existing `ChatCompletionSampler` does not expose the full set of required sampling parameters. Its `<think>` reasoning trace can also be mistaken for the final answer during multichoice answer extraction.

Although motivated by Qwen3.5, the implementation is model-agnostic: it lets simple-evals pass backend-supported generation settings for any model and improves answer extraction for reasoning models that emit `<think>` traces.

## Summary

Model-agnostic additions to the simple-evals harness:

- **`run_eval.build_extra_body`**: routes generation params in one place at the harness layer (not inside the sampler). OpenAI-native params (`top_p`) stay top-level; SGLang-only params (`top_k`, `min_p`, `presence_penalty`, `repetition_penalty`, `frequency_penalty`, `seed`) ride in `extra_body`, alongside any `chat_template_kwargs` (e.g. thinking mode). `ChatCompletionSampler` stays minimal — no per-param routing layer. CLI gains opt-in flags for the new params.
- **`strip_reasoning()`**: drops the `<think>...</think>` reasoning span before multichoice answer extraction (handles both a full block and a bare `</think>` close-tag). Applied **only** to the copy fed to the answer regex in GPQA/MMLU — the raw response is preserved for the html/convo report and the `chars` metric. Note this will flip the benchmark result, evidence: https://github.com/sgl-project/sglang-jax/pull/1306#issuecomment-4621957669
- **`accuracy_case_runner`**: forwards the complete `generation_config` to `run_eval` (previously only `temperature`/`max_tokens`/`top_p`/`chat_template_kwargs`), so suite-level params actually reach the sampler instead of being recorded in the summary but silently dropped.

All params default to `None`, so existing invocations are byte-for-byte unchanged. No model-specific coupling.

## Testing

`test/srt/eval/test_simple_eval_common.py` (pure CPU, registered in `unit-test-cpu`): `build_extra_body` param routing (OpenAI-native vs `extra_body`, `chat_template_kwargs` preserved, `None` when unset) and `strip_reasoning` across full-block / bare-close / no-tag passthrough.

### DIFF
--- a/test/srt/eval/simple_eval_common.py
+++ b/test/srt/eval/simple_eval_common.py
@@ -1,6 +1,7 @@
 # Adapted from https://github.com/openai/simple-evals/
 
 import os
+import re
 import resource
 import time
 from collections import defaultdict
@@ -175,6 +176,27 @@ D) {D}
 
 ANSWER_PATTERN_MULTICHOICE = r"(?i)Answer\s*:\s*([A-D])"
 ANSWER_PATTERN = r"(?i)Answer\s*:\s*([^\n]+)"
+
+_REASONING_RE_BOTH = re.compile(r"<think>.*?</think>", flags=re.DOTALL)
+_REASONING_RE_CLOSE_ONLY = re.compile(r"^.*?</think>", flags=re.DOTALL)
+
+
+def strip_reasoning(text: str) -> str:
+    """Drop the reasoning span before multichoice answer extraction.
+
+    Handles both shapes:
+    - Full ``<think>...</think>`` block in the response (some templates).
+    - Bare ``</think>`` only (Qwen-style: the opening ``<think>\\n`` is
+      injected by the chat template into the prompt, so the server's
+      content field starts directly with reasoning text and only emits
+      the closing tag).
+    """
+    if not text:
+        return text
+    text = _REASONING_RE_BOTH.sub("", text)
+    if "</think>" in text:
+        text = _REASONING_RE_CLOSE_ONLY.sub("", text, count=1)
+    return text.strip()
 
 
 EQUALITY_TEMPLATE = r"""
--- a/test/srt/eval/simple_eval_gpqa.py
+++ b/test/srt/eval/simple_eval_gpqa.py
@@ -19,6 +19,7 @@ from eval.simple_eval_common import (
     SamplerBase,
     SingleEvalResult,
     format_multichoice_question,
+    strip_reasoning,
 )
 
 
@@ -66,7 +67,7 @@ class GPQAEval(Eval):
                 )
             ]
             response_text = sampler(prompt_messages)
-            match = re.search(ANSWER_PATTERN_MULTICHOICE, response_text)
+            match = re.search(ANSWER_PATTERN_MULTICHOICE, strip_reasoning(response_text))
             extracted_answer = match.group(1) if match else None
             score = 1.0 if extracted_answer == correct_answer else 0.0
             html = common.jinja_env.from_string(HTML_JINJA).render(
--- a/test/srt/eval/simple_eval_mmlu.py
+++ b/test/srt/eval/simple_eval_mmlu.py
@@ -19,6 +19,7 @@ from eval.simple_eval_common import (
     SamplerBase,
     SingleEvalResult,
     format_multichoice_question,
+    strip_reasoning,
 )
 
 subject2category = {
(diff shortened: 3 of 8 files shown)


MokusMokun:
### Benchmark evidence: the bare-`</think>` strip changes a real score

Full **GPQA-Diamond (198 q)**, **Qwen3.5-35B-A3B**, thinking mode. Three extractors built from this PR's helpers:

| extractor | score |
|---|---|
| no strip (raw) | 0.8485 |
| full-block-only `<think>…</think>` (naive) | 0.8485 |
| **`strip_reasoning` (this PR)** | **0.8586** |

**+1.0 pp, 2/198 fixed** (parsed 1.000 throughout). The naive full-block strip is a no-op: Qwen emits a **bare `</think>`** (the opening tag is injected into the prompt, never echoed) — `<think>` 0/198, `</think>` 198/198 — so only the bare-close branch strips anything (median **22.3k → 2.7k** chars).

**The bug in one line:** the extractor is `re.search(r"(?i)Answer\s*:\s*([A-D])")` — **first match wins**. When the model types `Answer: <Word>` anywhere in its reasoning, the regex captures the **first letter of that word** and stops, never reaching the real `Answer: X` line at the end. `strip_reasoning` deletes everything up to `</think>` first, so only the final answer is left for the regex.

#### q#29 — gold `A` → no-strip picks `C`

In its reasoning (~1k chars *before* `</think>`), the model jots a note about the question *type*:

```
Standard Question Type: "Given Energy/Wavelength, find Color of Dye".
Standard Answer: Complement.        <-- regex grabs "Answer: C", the C of "Complement"
```

That's the first `Answer:` in the string, so `re.search` captures **C** and stops. The model's actual verdict, after `</think>`, is `Answer: A`.
→ **no-strip = C (wrong)** · **strip_reasoning = A (correct)**

#### q#140 — gold `D` → no-strip picks `A`

In its reasoning (~5.8k chars *before* `</think>`), the model weighs a *hypothetical* simpler question:

```
If the answer were simply 2, the question could have been
"Chlorobenzene with NaNH2/NH3 -> ?" (Answer: Aniline).   <-- regex grabs "Answer: A", the A of "Aniline"
```

Again the first `Answer:`, so `re.search` captures **A** and stops. The model's actual verdict, after `</think>`, is `Answer: D`.
→ **no-strip = A (wrong)** · **strip_reasoning = D (correct)**

---

25/198 (12%) of responses contain such an in-trace `Answer:[A-D]`; 2 diverged from the final answer on this run. Benchmark-level companion to this PR's unit tests.

JamesBrianD:
One design issue here is that the new sampler fields look a bit redundant for the OpenAI-compatible request path.

`top_k`, `min_p`, and `repetition_penalty` are not OpenAI top-level chat completion kwargs; they need to be sent through `extra_body`. Upstream SGLang keeps that boundary in `run_eval_once()` by adding the supported server-specific fields directly into `extra_body` and passing only `extra_body` to `ChatCompletionSampler`. In this PR, those same fields are first added as `ChatCompletionSampler` init fields and then moved back into `extra_body` inside `_build_kwargs()`. Functionally that can work, but it adds another propagation layer.

That extra layer creates a runner consistency problem: `test/srt/multi_host/accuracy_case_runner.py` still forwards only `temperature`, `max_tokens`, `top_p`, and `chat_template_kwargs` from `AccuracyCase.generation_config`. If a suite adds `top_k`, `min_p`, `presence_penalty`, `repetition_penalty`, `frequency_penalty`, or `seed`, those values can be recorded in the summary but silently dropped before they reach the sampler.

Could we either keep the upstream-style approach and build/merge these server-specific sampling params into `extra_body` at the runner / `run_eval.py` layer, or update every runner entry point, especially `multi_host/accuracy_case_runner.py`, to forward the full supported sampler config? Otherwise suite-level params such as `top_k` / `min_p` will not reliably take effect.

MokusMokun:
> One design issue here is that the new sampler fields look a bit redundant for the OpenAI-compatible request path.
> ....

Updated by moving sampler-param routing out of ChatCompletionSampler into run_eval.build_extra_body(), and fixing accuracy_case_runner.py to forward the full generation config (+ debug print) so suite-level top_k/min_p/etc. no longer get silently dropped.

@JamesBrianD

JamesBrianD (approved):

Issue #1011, opened 2026-04-30
[RFC] Kimi-K2.5


depksingh:
#### Summary

This RFC proposes adding inference support for **Kimi-K2.5** (moonshotai/Kimi-K2.5) in sglang-jax's multimodal pipeline. Kimi-K2.5 is a native vision-language model from Moonshot AI that integrates a DeepSeek-V3-style MoE LLM with a custom vision encoder.

##### 1.1 LLM Backbone

The LLM is a **DeepSeek-V3-style** transformer (model\_type: "kimi\_k2") characterized by massive sparsity and memory-efficient attention:

* **Attention:** Uses **Multi-head Latent Attention (MLA)** to compress the KV cache.  
  * **Latent Ranks:** kv\_lora\_rank=512, q\_lora\_rank=1536.  
  * **Head Dimensions:** qk\_nope\_head\_dim=128, qk\_rope\_head\_dim=64, v\_head\_dim=128.  
* **Mixture-of-Experts (MoE):** 61 layers total.  
  * **Layer 0:** Dense (acts as a standardizer).  
  * **Layers 1–60:** Sparse MoE with **384 routed experts** and **1 shared expert**.  
  * **Routing:** Selects **top-8** experts per token using sigmoid scoring with bias correction.  
* **Positional Encoding:** Uses **YaRN RoPE** (factor=64.0, theta=50,000).  
  * **Crucial Note:** Unlike Qwen-VL, Kimi-K2.5 does **not** use MRoPE; vision tokens are treated as standard sequence elements.  
* **Capacity:** 1.04T total parameters (\~32B active), 262,144 token context window, and 163,840 vocabulary size.

##### 1.2 Vision Encoder: MoonViT-3D

Kimi-K2.5 uses the **MoonViT-3D** encoder, a 425M parameter model designed for unified image and video understanding:

* **Structure:** 27 layers, hidden dimension of 1152, and 16 attention heads.  
* **Positional Embedding:** KimiDividedFixedPosEmb (separate H, W, T embeddings).  
* **Spatio-Temporal Attention:** For video, it performs sequential spatial passes (within frames) and temporal passes (across frames) rather than full 3D attention.  
* **Patch Merger:** Reduces token density by pooling 2 times 2 spatial patches and applying temporal pooling before projecting to the LLM's 7168 hidden dimension.

##### 2.1 Component Reuse

The following sglang-jax components are directly compatible:

[...]

PR #1314, opened 2026-06-04
E2E pipeline setup for KIMI K2.5

depksingh:
## Motivation
Adding the pipeline support for Kimi K2.5 model [Link](https://huggingface.co/moonshotai/Kimi-K2.5/)
#1011 

## Modifications
- Added required confiurations for Kimi K2.5
- Added dummy ViT and VL_Generation classes
- Updated the multimodal tokenizer to handle the multimodal input from Kimi K2.5

## Accuracy Tests
NA

## Testing
- Tested the E2E flow with and verified the successful completion of request after integration of VIT and VL_Generation implementation.
-  Verified the model tokens reaching ViT to verify the input validity with VIT and VL_Generation impl.
Pasted Graphic 3

NOTE: The current dummy vl_generation fails during pre-compilation due to having different outputs in sampler, but works with the original implementation of VL_Generation.

## Benchmarking and Profiling
NA

## Checklist


depksingh:
Hello @Bob-Chen222 , @chaitanyapk, @pengchengneo , @cjx0709  can you please review this PR whenever you get time. Will add the ViT and VL_generation implementation in the upcoming PRs.

cjx0709:
@depksingh hello, thanks for your contribution.

This PR mainly adds scaffolding for the Kimi-K2.5 multimodal pipeline, but the repository does not yet have a real Kimi-K2.5 text/generation implementation. Since the generation stage is currently a dummy model, this PR cannot meaningfully validate the end-to-end Kimi-K2.5 VL path. It only verifies some pipeline wiring and tokenizer plumbing.

I think we can still review this PR to discuss the overall infrastructure and interface design. However, I’d strongly suggest landing the real Kimi-K2.5 text/generation implementation first, then bringing this vision/pipeline integration in a follow-up PR with a complete E2E test. That would make the validation much stronger and avoid merging an isolated placeholder path that looks like model support but is not actually usable yet.

What do you think? @JamesBrianD  @depksingh  @pengchengneo

depksingh:
Hey @cjx0709 , Yeah, this PR was aimed towards adding the flow so that the next components can be added in the next PRs to avoid bulky PRs and so that individual components can be test e2e when they are available, The next set of PRs we were planning were 
1. Adding Text Generation LLM Backbone  (We already have integrated with the above pipeline changes in this fork: [Link](https://github.com/chaitanyapk/sglang-jax/blob/main/python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py) and got it running, since its only using DeepseekV3 backbone, the VL generation is just a wrapper on top of it.
2. Adding ViT component.

With having the flow already in place we can test it with launching the server once each of these PRs are ready instead of manually testing them with UTs was our idea. Also since the quantization support for INT4 is in progress, we were planning to use BF16 weights [here](https://huggingface.co/ananayarora/Kimi-K2.5-BF16) to get things going until that piece is merged.

if we want to add the text generation first, I can add the above changes for vl_generation in this PR, if that works?
Please let me know what do you think.

cjx0709:
@depksingh Thanks for the context, that makes sense. I agree with the goal of avoiding one huge PR and reviewing the components independently.

My main concern is not about splitting the work, but about the merge order and the support surface exposed in `main`. If this PR lands first with a dummy generation model and static Kimi registry/stage config, `moonshotai/Kimi-K2.5` will look supported even though it cannot produce real outputs yet.

I think a cleaner path would be:

1. Open/land the Kimi-K2.5 text generation wrapper first, ideally as a separate PR.
2. Keep this pipeline/VL PR stacked on top of that PR, so we can still review the infrastructure and interface design independently.
3. Once the text PR lands, rebase this PR, remove the dummy generation path, and include a real launch/E2E validation using the actual generation implementation.
4. Then add the real ViT implementation either in this PR if the scope stays reasonable, or as the next stacked PR.

So I would prefer not to add the text generation implementation into this PR unless it is truly small. From the linked fork it looks like the text wrapper includes non-trivial DeepSeekV3 integration and weight mappings, so a separate text PR may be easier to review.

depksingh:
Closing this PR, as the e2e changes with video changes are raised as part of this PR: https://github.com/sgl-project/sglang-jax/pull/1672

PR #1316, opened 2026-06-05
fix(multimodal): route Wan text encoder through encoder pipeline

### LINKED ISSUE #1315: wan2.1/2.2 text encoder mis-wired to auto_regressive scheduler (heavy/OOM-prone precompile)
## Summary

The wan2.1 / wan2.2 stage-0 **text encoder** (UMT5) is wired to the `auto_regressive`
scheduler in `wan2_1_stage_config.yaml` / `wan2_2_stage_config.yaml`. That routes a text
encoder through the LLM serving Scheduler and loads the **LLM-path**
`sgl_jax.srt.models.umt5.UMT5EncoderModel` (signature `__call__(forward_batch, ...)`).

When precompile is enabled, this makes the text-encoder stage run the LLM serving
precompile — `[EXTEND] PRECOMPILE tokens≈4096, bs≈64` (driven by
`compilation_manager._compute_token_buckets`, which force-appends `max_padded_num_tokens`)
— for an encoder that only needs ~512 tokens. This is wasteful and an OOM risk, and is the
reason `--disable-precompile` is currently set in `test/srt/multimodal/test_wan2_1_models.py`.

## Root cause

The multimodal model registry `get_model_class("UMT5EncoderModel")` (in
`multimodal/manager/stage.py`) resolves to the LLM-path class. The **encoder-path** class
`sgl_jax.srt.multimodal.models.encoders.t5.UMT5EncoderModel` already exists and is complete
(subclass of `T5EncoderModel`, `is_umt5=True` per-layer relative bias, full `load_weights`,
HF-style `__call__(input_ids, attention_mask, output_hidden_states) -> last_hidden_state`).
It just was never wired for wan.

Sibling diffusion model **FLUX** does it correctly: `scheduler: text_encoder` +
`model_class: CLIPTextModel,T5EncoderModel`. Upstream sglang likewise handles diffusion
text encoders in dedicated encoder pipelines, not the autoregressive serving path.

## Proposed fix

1. `stage.py`: import the encoder-path `UMT5EncoderModel` from `encoders/t5.py` so the
   registry entry `"UMT5EncoderModel"` binds to it (drop the LLM-path import). Only
   wan2.1/2.2 use this key; qwen-vl / qwen-omni use `auto_regressive` but are genuine VLMs.
2. `wan2_1_stage_config.yaml` + `wan2_2_stage_config.yaml` stage 0:
   `scheduler: auto_regressive` → `scheduler: text_encoder`,
   `scheduler_params: {"tokenizers": "tokenizer"}`, `precompile_params: {}`.
3. If precompile is to be enabled for these tests, also make
   `EncoderModelWorker.run_precompile` precompile all token buckets that
   `encoder_model_runner._select_precompiled_max_length` can select
   (`[16, 32, 64, 128, 256, 512]` ≤ max_length), not just `max_length`, otherwise a short
   prompt compiles at request time. (Hoist the bucket list to a shared constant.)

## Verification (TPU v6e-4, already done)

- encoder-path UMT5 vs HF transformers UMT5 on the real wan `text_encoder` weights:
  `last_hidden_state` MAE **5.8e-8**, 0 missing weight-mapping keys → numerically equivalent.
- wan2.1 **1.3B and 14B** e2e tests pass with the routing fix + precompile enabled;
  encoder precompile becomes `max_length=512` (~1–6 s, all buckets ~7 s); the token≈4096
  precompile is gone; no OOM.

## Important caveat — this is NOT a cache win

Multimodal stage models run on **device subsets** (e.g. wan stage1/stage2 on chips [2]/[3]).
The JAX persistent compilation cache is disabled for device subsets that don't start at
device 0 (see #1216), and empirically a wan run writes **0 cache entries** even with
`min_compile_time=0` / `min_entry_size=0`. So enabling wan precompile does **not** get
cached — it is a per-run cost. Therefore:

- This fix is a **correctness / mis-wiring cleanup**, not a cache optimization.
- Keep `--disable-precompile` unless you specifically want predictable per-request latency
  in production (precompile front-loads compilation but will not be cached).

## Context

Split out of #1260 (XLA persistent-cache hygiene). It was briefly bundled there on the
assumption that the wan precompile would be captured by the cache; that assumption is false
(see caveat above), so it does not belong in the cache PR.

Rodrian7:
## Summary

The Wan2.1 and Wan2.2 stage-0 UMT5 text encoder is configured with the
`auto_regressive` scheduler. This selects the LLM-serving scheduler and the
LLM-path UMT5 implementation, so startup precompile follows the autoregressive
path and can run an unnecessarily large prefill that exhausts device memory.

This PR routes the Wan text encoder through the dedicated encoder pipeline.
Runtime token-bucket selection and startup precompile now share the same bucket
definition, so precompile covers every input shape that the request path can
select. UMT5 outputs at bucket-padding positions are zeroed with the effective
model attention mask before they reach Wan cross-attention, matching the
reference Wan padding contract while preserving FLUX T5's fixed-length,
all-ones-mask behavior.

Fixes #1315.

## Changes

- Bind the multimodal encoder-path `UMT5EncoderModel` in the stage model registry.
- Change the Wan2.1 and Wan2.2 text-encoder stages from `auto_regressive` to
  `text_encoder`, pass the tokenizer explicitly, and remove the autoregressive
  `precompile_params`.
- Use one shared encoder token-bucket definition for runtime padding and
  precompile every reachable input length. FLUX T5 remains fixed at its maximum
  sequence length.
- Re-enable precompile in the Wan2.1 E2E cases so the original startup path is
  covered.
- Mask T5 bucket-padding hidden states at the encoder output boundary using the
  effective model attention mask.
- Add CPU regression tests for Wan scheduler/model resolution, runtime-to-
  precompile bucket coverage, worker precompile calls, Wan padding semantics,
  and unchanged FLUX T5 behavior, registered in `unit-test-cpu`.
- Update the multimodal architecture documentation and make the Wan E2E server
  cleanup reliable when an assertion fails.

## Validation

- `python -m pytest python/sgl_jax/test/multimodal/test_stage_config_routing.py -q`
  - 11 passed
- `pre-commit run --all-files`
  - passed
- The previous CI attempt passed CPU, all 1-TPU unit partitions, 1-TPU E2E,
  1-TPU accuracy, and one 4-TPU unit partition. The remaining 4-TPU runners
  lost communication with GitHub before producing test results.
- Wan TPU E2E has not completed against the current head. Prior TPU and
  numerical validation is documented in #1315.

## Accuracy Tests

The CPU regression tests verify that Wan real-token embeddings are preserved,
Wan bucket-padding embeddings are zeroed, every runtime-selectable encoder
shape is precompiled, and FLUX T5 keeps its fixed-length behavior. The
model-path rewire was previously validated against Hugging Face and with Wan
end-to-end generation as documented in #1315; that TPU validation has not yet
completed against the current head.

## Benchmarking and Profiling

N/A. This PR fixes stage routing and precompile coverage; no throughput or
latency benchmark was run for the current head.

## Checklist

- [x] The PR and public artifacts are in English.
- [x] The PR links the issue it resolves.
- [x] The test plan and current validation status are documented.

### DIFF
--- a/docs/architecture/12-multimodal.md
+++ b/docs/architecture/12-multimodal.md
@@ -238,9 +238,9 @@ HTTP Request (text prompt + params)
   → GlobalScheduler
     → TypeBasedDispatcher → Wan Pipeline
 
-  → Stage 0 (Embed)
-    → EmbedScheduler → EmbedModelExecutor
-    → Text → CLIP/T5 embedding
+  → Stage 0 (Text Encoder)
+    → EncoderScheduler → EncoderModelWorker → EncoderModelRunner
+    → Text → UMT5 embeddings
 
   → Stage 1 (Diffusion)
     → DiffusionScheduler → DiffusionModelExecutor
--- a/docs/architecture/12-multimodal.md
+++ b/docs/architecture/12-multimodal.md
@@ -141,7 +141,7 @@ The `auto_regressive` Scheduler type in the YAML config reuses the text engine's
 
 | Directory | Model | Pipeline stages |
 |------|------|--------------|
-| `wan/` | Wan 2.1/2.2 | Embed → Diffusion → VAE |
+| `wan/` | Wan 2.1/2.2 | Text Encoder → Diffusion → VAE |
 | `qwen2_5VL/` | Qwen2.5-VL | ViT → AutoRegressive |
 | `qwen3_omni_moe/` | Qwen3-Omni-MoE | ViT → AutoRegressive → Audio |
 | `mimo_audio/` | MiMo Audio | Audio → AutoRegressive |
@@ -154,8 +154,8 @@ Each model has a corresponding YAML config under `multimodal/models/static_confi
 **Wan 2.1 example** (3-stage pipeline, `wan2_1_stage_config.yaml`):
 
 ```text
-Stage 0: scheduler=auto_regressive, model_class=UMT5EncoderModel
-  → Text encoding (reuses the text engine's Scheduler)
+Stage 0: scheduler=text_encoder, model_class=UMT5EncoderModel
+  → UMT5 text encoding through the dedicated encoder scheduler
 
 Stage 1: scheduler=diffusion, model_class=WanTransformer3DModel
   → N-step diffusion denoising
--- a/docs/architecture/12-multimodal.md
+++ b/docs/architecture/12-multimodal.md
@@ -119,7 +119,8 @@ Each stage contains:
 | `DiffusionScheduler` | Text-to-image / video | Manages diffusion-step iteration |
 | `VaeScheduler` | VAE decoding | Latent → pixel decoding scheduling |
 | `VitScheduler` | Vision Transformer | Image patch encoding |
-| `EmbedScheduler` | Text embedding | Text encoder feature extraction |
+| `EncoderScheduler` | Text encoding | Runs dedicated text encoders such as CLIP, T5, and UMT5 |
+| `EmbedScheduler` | Text embedding | Produces embedding-model outputs |
 | `AudioScheduler` | Audio | Audio feature extraction/generation |
 | `AudioBackboneScheduler` | Audio backbone | Audio model backbone network |
 
@@ -212,7 +213,8 @@ Each stage's YAML definition contains:
 |------|------|--------|
 | `audio/` | Audio processing | Audio encoding/decoding execution |
 | `diffusion/` | Diffusion model execution | Manages denoising steps, invokes UNet/DiT |
-| `embed/` | Embedding extraction | Text encoder forward |
+| `embed/` | Embedding extraction | Produces embedding-model outputs |
+| `encoder/` | Text encoding | Tokenization and dedicated text encoder forward |
 | `vae/` | VAE encode/decode | Latent ↔ pixel conversion |
 | `vit/` | Vision Transformer | Image patch encoding |
 
--- a/python/sgl_jax/srt/multimodal/model_executor/encoder/encoder_model_runner.py
+++ b/python/sgl_jax/srt/multimodal/model_executor/encoder/encoder_model_runner.py
@@ -20,6 +20,8 @@ from sgl_jax.srt.server_args import ServerArgs
 
 logger = logging.getLogger(__name__)
 
+ENCODER_TOKEN_BUCKETS = (16, 32, 64, 128, 256, 512, 1024)
+
 
 @dataclass
 class _EncoderSpec:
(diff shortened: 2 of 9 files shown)


docs/architecture/12-multimodal.md
JamesBrianD: Need changes to text encoder -> EncoderScheduler

python/sgl_jax/test/multimodal/test_stage_config_routing.py
```diff
@@ -0,0 +1,69 @@
...
+    assert get_model_class(stage.model_class) is UMT5EncoderModel
+
+
+def test_flux_text_encoder_routing_is_unchanged():
+    stage = load_stage_configs_from_yaml(get_stage_config_path("black-forest-labs/FLUX.1-dev"))[0]
+
+    assert stage.scheduler == "text_encoder"
+    assert stage.scheduler_params == {"tokenizers": "tokenizer,tokenizer_2"}
+    assert get_scheduler_class(stage.scheduler) is EncoderScheduler
+    assert get_model_class(stage.model_class) == [CLIPTextModel, T5EncoderModel]
```
JamesBrianD: Assert checks with no testing value

python/sgl_jax/srt/multimodal/model_executor/encoder/encoder_model_runner.py
JamesBrianD, line 371: Padding size has "16, 32, .... " buckets. But `run_precompile` in encoder_model_worker(line 32) only precompile the max size

docs/architecture/12-multimodal.md
Rodrian7: Updated the Wan data-flow section to show `Stage 0 (Text Encoder) -> EncoderScheduler -> EncoderModelWorker -> EncoderModelRunner -> UMT5 embeddings`.

python/sgl_jax/test/multimodal/test_stage_config_routing.py
Rodrian7: Simplified this test to keep only the two resolutions that directly guard the original mis-wiring: the Wan stage resolves to `EncoderScheduler`, and its model key resolves to the multimodal `UMT5EncoderModel`. Removed the redundant scheduler-params, missing-field, and unchanged-FLUX routing assertions. I also added behavioral coverage for runtime/precompile shape closure and the worker's actual precompile calls.

python/sgl_jax/srt/multimodal/model_executor/encoder/encoder_model_runner.py
Rodrian7: Fixed in `af6cee042`. Runtime padding and precompile now share `ENCODER_TOKEN_BUCKETS`; `run_precompile()` compiles every length the runtime selector can return, including a non-bucket maximum. FLUX T5 remains fixed at its maximum runtime length. The tests cover the selector invariant and the worker's submitted shapes.

JamesBrianD (approved):

Issue #1318, opened 2026-06-06
perf(deepseek-v3): W8A16 fused-MoE — v6e-64 epmoe→fused +62% (3291→5315), v7x-16 +51%


zorrofox:
## Summary

DeepSeek-V3 671B (official FP8 block-wise checkpoint, W8A16 inference) on TPU currently runs only via `--moe-backend epmoe`. xprof shows `EPMoE.__call__` does `reshard(hidden_states, P(None))` per layer (replicate-then-mask), which all-gathers `[T, 7168]` × 58 layers — **40% of EXTEND step** at T=8192. `FusedEPMoE` avoids this but does not support static block-wise scales.

After wiring static block-scale into the fused path + tune entries (benchmark: `bench_serving 1K/1K --random-range-ratio 1.0`, sglang-jax docs convention):

| | epmoe (main@8abb3317 + tune entries) | **fused (this work)** | Δ |
|---|---|---|---|
| **v6e-64** cc=420 | 2958 | **5299** | **+79%** |
| **v6e-64** cc=512 | 3291 | **5315** | **+62%** |
| **v7x-16** cc=512 | 2442¹ | **3696** | **+51%** |
| ITL P50 (v6e@512) | 87.2 ms | **57.5 ms** | −34% |
| GSM8K 200 | — | 0.96 / 0.965 | ✅ |

¹ v7x baseline = fused without v7x tune entries (epmoe not measured on v7x).

> **Caliber note**: an earlier revision used `--random-range-ratio 0.0` for
> v6e (in/out len ∈ U[1,1000]), which inflated the relative gain to +192%
> due to prefill-first scheduling amplifying the epmoe EXTEND bottleneck.
> All numbers above re-measured at ratio=1.0.

## Bugs found along the way

1. `blockwise_utils.get_blockwise_kernel_config`: Tier-1 fuzzy match can return `batch_block_size=1` (from #1191 GLM entries), clamped to `n_batch∈(1,8)` → violates Pallas TPU `block_m % 8 == 0 OR == array_m`, raises lowering error.
2. `weight_utils._maybe_convert_epmoe_scale_for_kernel`: `jnp.repeat` on `[E, k_blocks, n_blocks]` materializes a fully-replicated 117 MB array per expert weight → OOM on the fused path during weight load.
3. `tp_worker` `max_running_requests`: attn-backend cap is not ep-aware (e.g. 408 on v7x-16 ep=32) → fused precompile rejects `bs % ep != 0`.

[...]

PR #1319, opened 2026-06-06
fix(kernels): three alignment/sharding fixes for FP8 block-quant models

zorrofox:
## Motivation

Three independent issues block FP8 block-quant models (DeepSeek-V3, GLM-5.1) from running on certain (dp, ep, batch) configs. All reproduce on `main@8abb3317`.

### 1. `blockwise_utils.py` — Pallas 8-align after Tier-1 clamp

Tier-1 fuzzy match (post-#1191) can return `batch_block_size=1`. The existing `min(tuned, n_batch)` clamp leaves it at 1 even when `n_batch∈(2,7)`, which violates the Pallas TPU constraint `block_m % 8 == 0 OR block_m == array_m`:

    ValueError: block shape (1, 1024) ... array (4, 7168)

Fix: after clamping, if `block_m < n_batch and block_m % 8 != 0`, snap to `n_batch` (≤8) or down-align to 8.

### 2. `weight_utils.py` — keep expert sharding in scale convert

`_maybe_convert_epmoe_scale_for_kernel` uses `jnp.repeat` on the block-scale which loses the `("expert", …)` sharding and materializes a fully-replicated intermediate (117 MB × 6 × 58 layers) → HBM OOM on the fused load path. Replace with `jnp.take(weight, jnp.arange(out_dim)//block, axis=2)` which preserves sharding.

### 3. `tp_worker.py` / `common_utils.py` — ep-align `max_running_requests` for fused MoE

The attention-backend / pool cap on `max_running_requests` is not ep-aware (observed 408 on v7x-16 with ep=32). With `--moe-backend fused` this flows into the precompile bs-bucket list and hits `fused_moe.get_simplified_key`:

    ValueError: Expected num_tokens=408 to be aligned to ep_size=32

Fix: new helper `common_utils.align_bs_for_fused_ep(bs, ep)` down-aligns so that `bs // ep ∈ {2, 4, 8k}`; `tp_worker` calls it right after the existing dp-align so both the scheduler cap and `max_padded_batch_size` stay consistent. Guards added per review: raise when `bs < 2*ep` (no valid down-align), assert `ep_size % dp_size == 0`.

## Testing

- Unit: `test_compilation_manager.py::TestAlignBsForFusedEp` (5 cases: 408/32→256, no-op, 2/4-snap, below-minimum raises, ep≤1 passthrough)
- e2e: DeepSeek-V3 v6e-64 (dp=8/ep=64) + v7x-16 (dp=4/ep=32), server boots clean on default `--precompile-bs-paddings`, GSM8K 0.965–0.985

Part of #1318.

### DIFF
--- a/python/sgl_jax/srt/managers/tp_worker.py
+++ b/python/sgl_jax/srt/managers/tp_worker.py
@@ -196,6 +196,10 @@ class ModelWorker:
         if server_args.moe_backend == "fused" and server_args.ep_size > 1:
             from sgl_jax.srt.utils.common_utils import align_bs_for_fused_ep
 
+            assert server_args.ep_size % dp_size == 0, (
+                f"fused MoE requires ep_size ({server_args.ep_size}) to be a multiple "
+                f"of dp_size ({dp_size}) so the ep-aligned cap stays dp-aligned"
+            )
             aligned = align_bs_for_fused_ep(self.max_running_requests, server_args.ep_size)
             if aligned != self.max_running_requests:
                 logger.warning(
--- a/python/sgl_jax/srt/managers/tp_worker.py
+++ b/python/sgl_jax/srt/managers/tp_worker.py
@@ -193,6 +193,20 @@ class ModelWorker:
                 dp_size,
             )
 
+        if server_args.moe_backend == "fused" and server_args.ep_size > 1:
+            from sgl_jax.srt.utils.common_utils import align_bs_for_fused_ep
+
+            aligned = align_bs_for_fused_ep(self.max_running_requests, server_args.ep_size)
+            if aligned != self.max_running_requests:
+                logger.warning(
+                    "Adjusted max_running_requests from %s to %s for fused MoE "
+                    "(ep_size=%s, bt must be in {2,4,8k})",
+                    self.max_running_requests,
+                    aligned,
+                    server_args.ep_size,
+                )
+                self.max_running_requests = aligned
+
         assert self.max_running_requests > 0, "max_running_request is zero"
 
         # A single request lives on one DP rank, so max_req_len is bounded
--- a/python/sgl_jax/srt/utils/common_utils.py
+++ b/python/sgl_jax/srt/utils/common_utils.py
@@ -39,6 +39,22 @@ PRECOMPILE_DEFAULT_TOKEN_PADDINGS = [1 << i for i in range(6, 14)]
 PRECOMPILE_DEFAULT_BS_PADDINGS = [1 << i for i in range(0, 9)]
 
 
+def align_bs_for_fused_ep(bs: int, ep_size: int) -> int:
+    """Down-align a global batch size so it is launchable by fused_ep_moe.
+
+    The fused MoE kernel requires ``bs % ep_size == 0`` and the per-EP local
+    batch ``bs // ep_size`` to be in ``{2, 4}`` or a multiple of 8. The
+    attention-backend / pool-derived ``max_running_requests`` is not ep-aware
+    (e.g. 408 on v7x-16 with ep=32), so callers must align it before it flows
+    into the scheduler and the precompile bucket list.
+    """
+    if ep_size <= 1:
+        return bs
+    local = bs // ep_size
+    local = (local // 8) * 8 if local >= 8 else (4 if local >= 4 else 2)
+    return local * ep_size
+
+
 def pad_to_bucket(value: int, buckets: list[int]) -> tuple[int, int]:
     """Return the smallest bucket >= value and its index.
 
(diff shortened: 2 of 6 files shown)


JamesBrianD:
Could you share the exact launch command / model config / stack trace that motivated the `compilation_manager.py` change?

For `max_padded_batch_size=408, ep_size=32`, this down-aligns the appended cap to 256. If buckets are already `[32, 64, 128, 256]`, no new bucket is added, while runtime capacity still appears to allow up to 408 requests. Then decode batches in `[257, 408]` may fail `pad_to_bucket(...)`.

Why is lowering only the appended precompile bucket sufficient? Should the effective max capacity also be reduced, or should we add a launchable bucket that still covers the configured max capacity?

zorrofox:
Good catch — you're right that aligning only the precompile bucket leaves `max_running_requests` at 408, so the scheduler can still admit up to 408 requests and runtime `pad_to_bucket` would fail for bs∈(256, 408].

**Repro** (DeepSeek-V3 671B, v7x-16, `--tp-size 32 --dp-size 4 --ep-size 32 --moe-backend fused`, default `context_len=163840`, `page_size=128`):
```
attn_backend_limit = 1024*1024//2 // cdiv(163840, 128) // 4 = 131072 // 1280 = 102
max_running_requests = min(512, pool, 102 * dp=4) = 408   # dp-aligned
_compute_bs_buckets appends 408 → fused_ep_moe precompile:
  ValueError: Expected num_tokens=408 to be aligned to ep_size=32
```
(and even if 408 % 32 == 0, `408//32 = 12` would still fail the `bt ∈ {2,4,8k}` check.)

**Fix moved to `tp_worker.py`** (e3c8b771): `max_running_requests` itself is now down-aligned via `common_utils.align_bs_for_fused_ep()` right after the existing dp-align, so both the scheduler cap and `max_padded_batch_size` (and hence the appended bucket) are consistent. The earlier `compilation_manager` change is reverted (net-zero diff there). Unit tests added in `test_compilation_manager.py::TestAlignBsForFusedEp` (408/32→256, 512/32→512, 100/32→64, ep≤1 passthrough).

zorrofox:
@JamesBrianD @cjx0709 — addressed the earlier comment (fix moved to `tp_worker.py` with unit tests in e3c8b771). Would appreciate a review when you have a moment. This is the standalone bug-fix PR of the #1318 set; #1320 is already merged.

python/sgl_jax/srt/utils/common_utils.py
```diff
@@ -39,6 +39,22 @@
...
+    The fused MoE kernel requires ``bs % ep_size == 0`` and the per-EP local
+    batch ``bs // ep_size`` to be in ``{2, 4}`` or a multiple of 8. The
+    attention-backend / pool-derived ``max_running_requests`` is not ep-aware
+    (e.g. 408 on v7x-16 with ep=32), so callers must align it before it flows
+    into the scheduler and the precompile bucket list.
+    """
+    if ep_size <= 1:
+        return bs
+    local = bs // ep_size
+    local = (local // 8) * 8 if local >= 8 else (4 if local >= 4 else 2)
```
cjx0709, line 59: when `bs < 2*ep_size` this returns a value larger than `bs` (e.g. `align(10, 32)=64`), so it actually up-aligns and could push `max_running_requests` past the memory limit. The min fused batch is `bt=2`, so anything below that has no valid down-align — better to `raise ValueError` here than silently bump it up. Doesn't affect the tested configs, just a guard.

python/sgl_jax/srt/managers/tp_worker.py
```diff
@@ -193,6 +193,20 @@ def __init__(
                 dp_size,
             )
 
+        if server_args.moe_backend == "fused" and server_args.ep_size > 1:
```
cjx0709, line 196: the aligned result is divisible by `ep_size` but not necessarily by `dp_size` (the dp-align just above), unless `ep_size % dp_size == 0`. Fine for `ep=32, dp=4`, but worth an assert to lock in that assumption.

python/sgl_jax/srt/utils/common_utils.py
zorrofox: Good catch — added a `local < 2` guard that raises instead of silently up-aligning past `bs`. Unit tests added (`align(10,32)` / `align(63,32)` both raise).

python/sgl_jax/srt/managers/tp_worker.py
zorrofox: Added `assert ep_size % dp_size == 0` before the align so the assumption is locked in (holds for all current fused configs; fails fast otherwise).

cjx0709 (approved):
LGTM

PR #1320, opened 2026-06-06
perf(kernels): v6e+v7x MLA/quantized_matmul tune entries for DeepSeek-V3

zorrofox:
## Motivation

DeepSeek-V3 on v6e/v7x logs `Couldn't find tuned sizes` for all 9 dense linear
shapes (q_a/kv_a/q_b/kv_b/o/shared-mlp) and all MLA `(num_q_heads=16, kv_lora=512,
page=128)` keys. #1191 added GLM shapes only.

## Changes

| kernel | device | entries | source |
|---|---|---|---|
| MLA v2 | v6e | 18 (decode+mixed, mnt 1–1024) | swept on v6e-4 |
| MLA v2 | v7x | 20 (mnt 1–2048) | reuse v6e; mixed `bq` capped at 128 (v7x scoped VMEM 57.6M < v6e) |
| quantized_matmul | v6e | 45 (9 shapes × n_batch∈{1,8,32,64,2048}) | swept on v6e-4 |
| quantized_matmul | v7x | 45 (9 shapes × n_batch∈{16,32,64,128,2048}) | 18 swept on v7x-16, 27 reuse v6e (bm clamped ≥16 = v7x sublane) |

`shapes_dsv3_tensor8.txt`: the 9 `(n_out, n_in)` for `tensor=8`.

## Benchmark (epmoe, 1K/1K bench_serving)

| | v6e-64 cc=16 | v6e-64 cc=512 |
|---|---|---|
| before | 192 | 964 |
| after | 249 (**+30%**) | 1024 (+6%) |

Low cc is attention-bound → larger gain. v7x gain measured together with
PR④ (fused-MoE entries) — see tracking issue.
Part of #1318.

PR #1321, opened 2026-06-06
feat(fused_moe): static block-wise scale wiring for DeepSeek-V3

zorrofox:
## Motivation

`FusedEPMoE` only supports dynamic per-tensor scales; static block-wise FP8 checkpoints (DeepSeek-V3) are forced onto `EPMoE`, whose replicate-then-mask `reshard(P(None))` all-gather dominates EXTEND (40% of step, xprof attached).

### `deepseek_v3.py` weight mapping
- Drop the `not use_fused` guard on `is_static_quant` routed-expert scale mapping; for fused, load with `(("data","tensor"), None, None)` to match the fused weight mesh (epmoe uses `("expert", …)` via `moe_mesh`).
- Add shared-expert `weight_scale_inv` mappings (2D HF layout, replicated).

### `fused_moe.py`
- `quantize_weights(is_static=True)`: shared scale placeholders → 2D `(out_blocks, in_blocks)` to receive HF `weight_scale_inv` directly.
- `__call__`: when shared scales are 2D block-wise, dequant shared w1/w2/w3 to bf16 and compute the shared-expert SwiGLU with plain `jnp` (3 small GEMMs), pass `None` to the Pallas kernel for shared, and add the result back after resharding to `P(("data","tensor"), None)`. Rationale: the v1 Pallas kernel's shared-expert path assumes the same dtype/VMEM layout as routed; mixing fp8 routed + per-call-dequant'd bf16 shared in-kernel triggers a DMA dtype mismatch. The shared expert is tiny (`inter=2048` vs hidden=7168) so the out-of-kernel cost is negligible (<1% of step).

Routed experts use the existing block-scale path added in #1259 (gmm_v2) — no kernel change needed; only the scale *loading* was missing.

## Benchmark (v6e-64, `bench_serving 1K/1K --random-range-ratio 1.0`)

| | epmoe (with #1320 tune) | fused (this PR + #1322) | Δ |
|---|---|---|---|
| cc=420 | 2958 | **5299** | **+79%** |
| cc=512 | 3291 | **5315** | **+62%** |
| ITL P50 @512 | 87.2 ms | **57.5 ms** | −34% |
| GSM8K 200 | — | 0.96 | ✅ |

## xprof (v6e-64, cc=512 load, single TPU:0)

| op | epmoe EXTEND 1075 ms | **fused EXTEND 895 ms** | epmoe DECODE 69 ms | **fused DECODE 56 ms** |
|---|---|---|---|---|
| all-gather (reshard P(None)) | **424 (40%)** | 76 (8%) | 12 | **0** |
| MoE gmm / fused-moe Pallas | 366 | 544¹ | 30 | 18 |
| all-reduce | 202 | 3 | 8 | 8 |
| mla | 10 | 132 | 10 | 12 |
| quantized_mm (dense) | 57 | 25 | 7 | 4 |

¹ EXTEND uses DEFAULT config; tuned in #1322.

Part of #1318.

> Runtime depends on #1319 (the `weight_utils` `jnp.take` fix — without it the fused weight-load path OOMs). Diff here is independent (2 files, no overlap with #1319).

### DIFF
--- a/python/sgl_jax/srt/layers/fused_moe.py
+++ b/python/sgl_jax/srt/layers/fused_moe.py
@@ -260,30 +260,31 @@ class FusedEPMoE(nnx.Module):
                 )
 
                 if self.num_shared_experts > 0:
-                    # fused kernel expects per-channel shared scale (1, 1, se_inter)
-                    # — see _validate_fused_ep_moe_args / kernel.py:455-470
-                    shared_scale_sharding = P(None, None, None)
+                    # Static block-wise checkpoints: store the raw HF 2D block
+                    # scale [out_blocks, in_blocks]; __call__ dequants the fp8
+                    # shared weight to bf16 before the kernel call so the
+                    # Pallas kernel keeps its per-channel-only shared path.
+                    bn = self.quant_block_n or wsz
                     se_inter = self.moe_shared_expert_intermediate_size * self.num_shared_experts
+                    se13_shape = (-(-se_inter // bn), -(-self.hidden_size // wsz))
+                    se2_shape = (-(-self.hidden_size // bn), -(-se_inter // wsz))
 
                     if hasattr(self, "w1_shared_scale"):
                         del self.w1_shared_scale
                     self.w1_shared_scale = nnx.Param(
-                        jnp.zeros((1, 1, se_inter), dtype=jnp.float32),
-                        out_sharding=shared_scale_sharding,
+                        jnp.zeros(se13_shape, dtype=jnp.float32), out_sharding=P(None, None)
                     )
 
                     if hasattr(self, "w3_shared_scale"):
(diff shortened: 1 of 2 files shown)


zorrofox:
Benchmark table corrected — see https://github.com/sgl-project/sglang-jax/issues/1318#issuecomment for details (earlier numbers used `--random-range-ratio 0.0`, re-measured at `1.0` per docs convention).

zorrofox:
@Prayer3th @JamesBrianD — gentle ping. This is the core fused-MoE wiring PR (2 files, +60/-31). Benchmark numbers were re-measured at `ratio=1.0` (see #1318 correction comment). Runtime depends on #1319; #1320 (tune entries) is merged.

Issue #1023, opened 2026-05-05
[Feature] Gemma4 Model Support


IKACE:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Hi SGLang-JAX experts,

Are there any plans to support Gemma4 Model? Thanks!

JamesBrianD:
Hi, IKACE. There are no plans to support gemma4 this quarter; interested developers are welcome to contribute.

hks-9697-v2:
I am interested in contributing, working on an RFC will share soon.

hks-9697-v2:
# **\[RFC\] Gemma 4 \- Multimodal: Dense and Moe**

## 1\. Motivation

Gemma 4 model family introduces multiple model architectures, 2 Edge focused models (E2B and E4B) rely on Per-Layer Embeddings (PLE) to compress memory footprints , 2 server grade models \- specifically the 31B Dense and the 26B A4B Mixture-of-Experts (MoE) variants. They include native multimodal integration(Vision and Audio for Edge models and Vision for server models).

This RFC proposes integrating the 31B Dense and the 26B MoE Gemma 4 architectures into the SGL-JAX repository. The models feature a 256K context window, native text and vision processing, and a 5:1 interleaving of local sliding-window attention with global full-context attention. Furthermore, these layers possess heterogeneous head dimensions and Grouped Query Attention (GQA) ratios, and so have a non uniformly shaped Key-Value (KV) cache

## 2\. Model architecture overview

**Dense Model:**

* 31B 
* Model config: [https://huggingface.co/google/gemma-4-31B-it/blob/main/config.json](https://huggingface.co/google/gemma-4-31B-it/blob/main/config.json)

**MoE Model:**

* 4B activated / 26B MoE  
* Model config: [https://huggingface.co/google/gemma-4-26B-A4B-it/blob/main/config.json](https://huggingface.co/google/gemma-4-26B-A4B-it/blob/main/config.json)

**Supported Modalities:**

Supports Vision \+ Text as input and Text as output.

**Tokenizer:** Sentence Piece

Here are some notable features of the model

| Hyperparameter | Gemma 4 (31B Dense) | Gemma 4 MoE |
| :---- | :---- | :---- |
| **Number of Layers** | 60 | 30 |
[...]

PR #1301, opened 2026-06-03
feat(model): Gemma4-31B text decoder + vision encoder

zorrofox:
## Summary

Adds Gemma4-31B-it support (text + image). 31B is the dense variant — no PLE, no shared-KV, no MoE — so the text decoder is a moderate extension of gemma2.py. Vision encoder is ported from MaxText/deepmind-gemma official Flax implementations.

Four commits, splittable into separate PRs if preferred:
1. **infra**: heterogeneous head_dim support (SWAKVPool `swa_head_dim`, FA backend `layer.head_dim`) + `ProportionalRotaryEmbedding`
2. **scheduler fix**: wire `input_embeds` / `mm_inputs` through to `ForwardBatch.input_embedding` (existing API field that was disconnected)
3. **model**: `gemma4.py` + `multimodal/models/gemma4/vision_encoder.py`
4. **test**: model MMLU test + vision encoder alignment test

## Architecture notes (31B specifics)

- **Heterogeneous attention**: 50 sliding layers (head_dim=256, kv_heads=16, rope_theta=10k) + 10 full layers (head_dim=512, kv_heads=4, proportional RoPE @ 1M, partial_rotary=0.25)
- **attention_k_eq_v**: full layers have no `v_proj` in checkpoint; K projection output is reused as V (matching deepmind/gemma reference)
- **Standard RMSNorm** (`x*w`), not Gemma2's `x*(1+w)`
- **`patch_model_config`** remaps HF's `global_*` ↔ sglang's `swa_*` naming convention; idempotent guard handles multiple `ModelConfig` constructions sharing one `hf_config`
- **Vision**: 27-layer ViT, 2D RoPE (head_dim split per x/y axis), learned 2D pos table (gather), 3×3 position-based pool, 5 token budgets (70-1120)

## Validation (v6e-4 TP=4 / v6e-8 TP=8)

| Benchmark | Result | Reference |
|---|---|---|
| GSM8K (1319, 0-shot) | 90.5% | — |
| MMLU (500, shuffled, 0-shot) | 85.0% | official MMLU Pro 85.2% |
| MMMU val (50, 10 subjects) | 62.0% | official MMMU Pro 76.9% — **HF reference under identical prompt produces identical answers (8/8 match)**, so the gap is eval-protocol (prompt/CoT) not implementation |
| Vision encoder vs HF | cos=0.9993 | bf16, 27 layers |
| E2E inputs_embeds vs HF | cos=0.9982 | — |
[...]

PR #1323, opened 2026-06-07
feat(model): Gemma4-31B + 26B Moe text decoder 

hks-9697-v2:
## Summary

Adds complete serving support for **Gemma4-31B-it (Dense)** and **Gemma4-26B-A4B-it (MoE)** models. This implementation provides robust tensor-parallel and expert-parallel execution while maintaining memory pool stability across hybrid attention configurations. Partially addresses #1023  (vision pipeline and remaining optimizations will follow).

The changes are organized into four logical commits:
1. **Dense Implementation**: Core `Gemma4ForCausalLM` backbone, heterogeneous head dimension support (`swa_head_dim`), `ProportionalRotaryEmbedding`, and base multimodal pass-through structures.
2. **MoE Implementation**: Native MoE routing blocks (`Gemma4Router`, `TopK`, `EPMoE`), safetensors weight unflattening, and expert scale gather distribution.
3. **Reasoning Parser**: Dedicated structured decoding support (`reasoning_parser.py`) for handling thinking mode channels.
4. **General Core Fixes**: Addresses runtime issues encountered during evaluation and model loading, including fixing chat template retrieval from custom download directories (`hf_transformers_utils.py`) and normalizing nested token structures from evaluation clients (`schedule_batch.py`).

---

## Architecture Notes (Gemma4 Specifics)

- **Heterogeneous Attention**: 50 sliding window layers (`head_dim=256`, `kv_heads=16`, `rope_theta=10k`) combined with 10 full global layers (`head_dim=512`, `kv_heads=4`, proportional RoPE @ 1M, `partial_rotary=0.25`).
- **attention_k_eq_v**: Full global layers contain no separate `v_proj` weights; the K projection output is dynamically reused as V.
- **Head Replication**: Supports KV head replication across tensor-parallel ranks to handle global attention layers (supporting `TP=8` for Dense which has only 4 global KV heads, and `TP=4, 8` for MoE which has only 2 global KV heads).
- **ProportionalRotaryEmbedding**: Introduces proportional rotary embedding support specifically for full-attention layers with `partial_rotary=0.25`.
- **SWAKVPool.swa_head_dim**: Extends the memory pool to support heterogeneous head dimensions within a single model architecture (sliding layers use `256` and full layers use `512`).
- **Dynamic FlashAttention Reshape**: Updates the FlashAttention backend to reshape inputs dynamically using `layer.head_dim` rather than relying on a single backend-global value.

---

## Unrelated Core Fixes / Context

To ensure smooth operation and evaluation benchmarking, this PR includes a few necessary core adjustments outside of the model definitions:

* **`python/sgl_jax/srt/hf_transformers_utils.py`**: Updated `download_from_hf` and `get_tokenizer` to respect the server's `download_dir` parameter. This ensures that when a model is downloaded to a custom cache directory (such as `/dev/shm`), the tokenizer correctly locates and loads the associated chat template and configuration files from the cached location rather than re-downloading or failing.
* **`python/sgl_jax/srt/managers/schedule_batch.py`**: Added normalization logic for `origin_input_ids` within the `Req` constructor. Evaluation scripts and client wrappers frequently pass token IDs in varied formats (such as dictionary outputs `{"input_ids": [...]}`, numpy arrays, or nested lists). Standardizing these inputs into a flat Python list ensures compatibility across external evaluation harnesses without triggering indexing errors.
* **`test/srt/run_eval.py`**: Added flag to enable thinking mode in eval scripts.

---

## Validation

Evaluations conducted on **4-chip v7 (8 devices)** and **4-chip v6e** configurations. The results are within the standard deviation of the [published results](https://ai.google.dev/gemma/docs/core/model_card_4). Testing for long-context thinking tasks was performed using the `temperature 1.0` and `top-p 0.95` configuration recommended by the model card.

### Summary Table

| Benchmark | Gemma4-31B-it (Dense) | Gemma4-26B-A4B-it (MoE) | Generation Parameters |
|---|---|---|---|
| **AIME 26** | **90.0%** | **83.3%** | `--enable-thinking true --temperature 1.0 --top-p 0.95` |
| **AIME 25** | **83.3%** | **90.0%** | `--enable-thinking true --temperature 1.0 --top-p 0.95` |
| **GSM8K** | **97.6%** | **96.9%** | 0-shot, Standard Decoding |
| **GPQA Diamond** | **85.9%** | **80.3%** | `--enable-thinking true --temperature 1.0 --top-p 0.95` |

## Known Limitations

- The vision pipeline will be added in a separate PR.
- This PR focuses primarily on evaluation accuracy; performance optimizations will be added in subsequent updates.

---

## Test Plan

- [x] **Verified TP and EP Combinations**: Executed unit test suites for `TP=1, 2, 4, 8` and `EP=8` across both Dense and MoE models on `v7-8` and `v6e-4` setups, ensuring stable memory allocation across all configurations.
- [x] **Verified Evaluation Benchmark Suite**: Validated end-to-end evaluation accuracy across AIME 25, AIME 26, GSM8K, and GPQA Diamond to ensure correct zero-shot output and structured thinking mode performance.

## Checklist


hks-9697-v2:
---

### Execution Logs & Configurations

#### 1. Gemma4-31B-it (Dense)

##### Server Launch Command
```bash
python3 -m sgl_jax.launch_server --model-path "google/gemma-4-31B-it" --device tpu --trust-remote-code --skip-server-warmup --random-seed 3 --tp-size 8 --max-prefill-tokens 4096 --download-dir /dev/shm/ --mem-fraction-static 0.95 --precompile-token-paddings 1024 --attention-backend fa --page-size 128 --max-running-requests 128 --chunked-prefill-size 4096 --max-total-tokens 1200000 --port 8000 --swa-full-tokens-ratio 0.2
```

##### Benchmark Commands & Outputs

* **AIME 26**
```bash
python ~/sglang-jax/test/srt/run_eval.py --port 8000 --eval-name aime26 --max-tokens 131072 --temperature 1.0 --enable-thinking true --top-p 0.95
```
```
Writing report to /tmp/aime26_google_gemma-4-31B-it.html
{'aime26': np.float64(0.9), 'aime26:std': np.float64(0.30000000000000004), 'score:std': np.float64(0.30000000000000004), 'score': np.float64(0.9)}
Writing results to /tmp/aime26_google_gemma-4-31B-it.json
Total latency: 286.195 s
Score: 0.900
```

* **AIME 25**
```bash
python ~/sglang-jax/test/srt/run_eval.py --port 8000 --eval-name aime25 --max-tokens 131072 --temperature 1.0 --enable-thinking true --top-p 0.95
```
```
Writing report to /tmp/aime25_google_gemma-4-31B-it.html
{'aime25': np.float64(0.8333333333333334), 'aime25:std': np.float64(0.3726779962499649), 'score:std': np.float64(0.3726779962499649), 'score': np.float64(0.8333333333333334)}
Writing results to /tmp/aime25_google_gemma-4-31B-it.json
Total latency: 305.245 s
Score: 0.833
```

* **GSM8K**
```bash
python ~/sglang-jax/test/srt/run_eval.py --port 8000 --eval-name gsm8k
```
```
Writing report to /tmp/gsm8k_google_gemma-4-31B-it.html
{'gsm8k': np.float64(0.9757391963608795), 'gsm8k:std': np.float64(0.15385778188283028), 'score:std': np.float64(0.15385778188283028), 'score': np.float64(0.9757391963608795)}
Writing results to /tmp/gsm8k_google_gemma-4-31B-it.json
Total latency: 137.685 s
Score: 0.976
```

* **GPQA Diamond**
```bash
python ~/sglang-jax/test/srt/run_eval.py --port 8000 --eval-name gpqa --max-tokens 131072 --temperature 1.0 --enable-thinking true --top-p 0.95
```
```
Writing report to /tmp/gpqa_google_gemma-4-31B-it.html
{'chars': np.float64(15864.111111111111), 'chars:std': np.float64(7929.986877450883), 'score:std': np.float64(0.3484482487002089), 'score': np.float64(0.8585858585858586)}
Writing results to /tmp/gpqa_google_gemma-4-31B-it.json
Total latency: 592.590 s
Score: 0.859
```

---

#### 2. Gemma4-26B-A4B-it (MoE)

##### Server Launch Command
```bash
python3 -m sgl_jax.launch_server --model-path "google/gemma-4-26B-A4B-it" --device tpu --trust-remote-code --skip-server-warmup --random-seed 3 --tp-size 8 --ep-size 8 --max-prefill-tokens 4096 --download-dir /dev/shm/ --mem-fraction-static 0.95 --precompile-token-paddings 1024 --attention-backend fa --page-size 128 --max-running-requests 128 --chunked-prefill-size 4096 --max-total-tokens 1200000 --port 8000 --swa-full-tokens-ratio 0.2
```

##### Benchmark Commands & Outputs

* **AIME 26**
```bash
python ~/sglang-jax/test/srt/run_eval.py --port 8000 --eval-name aime26 --max-tokens 131072 --temperature 1.0 --enable-thinking true --top-p 0.95
```
```
Writing report to /tmp/aime26_google_gemma-4-26B-A4B-it.html
{'aime26': np.float64(0.8333333333333334), 'aime26:std': np.float64(0.37267799624996495), 'score:std': np.float64(0.37267799624996495), 'score': np.float64(0.8333333333333334)}
Writing results to /tmp/aime26_google_gemma-4-26B-A4B-it.json
Total latency: 1774.499 s
Score: 0.833
```

* **AIME 25**
```bash
python ~/sglang-jax/test/srt/run_eval.py --port 8000 --eval-name aime25 --max-tokens 131072 --temperature 1.0 --enable-thinking true --top-p 0.95
```
```
Writing report to /tmp/aime25_google_gemma-4-26B-A4B-it.html
{'aime25': np.float64(0.9), 'aime25:std': np.float64(0.29999999999999993), 'score:std': np.float64(0.29999999999999993), 'score': np.float64(0.9)}
Writing results to /tmp/aime25_google_gemma-4-26B-A4B-it.json
Total latency: 1741.803 s
Score: 0.900
```

* **GSM8K**
```bash
python ~/sglang-jax/test/srt/run_eval.py --port 8000 --eval-name gsm8k
```
```
Writing report to /tmp/gsm8k_google_gemma-4-26B-A4B-it.html
{'gsm8k': np.float64(0.9689158453373768), 'gsm8k:std': np.float64(0.17354518141260308), 'score:std': np.float64(0.17354518141260308), 'score': np.float64(0.9689158453373768)}
Writing results to /tmp/gsm8k_google_gemma-4-26B-A4B-it.json
Total latency: 139.883 s
Score: 0.969
```

* **GPQA Diamond**
```bash
python ~/sglang-jax/test/srt/run_eval.py --port 8000 --eval-name gpqa --max-tokens 131072 --temperature 1.0 --enable-thinking true --top-p 0.95
```
```
Writing report to /tmp/gpqa_google_gemma-4-26B-A4B-it.html
{'chars': np.float64(30643.469696969696), 'chars:std': np.float64(16430.843827245888), 'score:std': np.float64(0.39770923480020265), 'score': np.float64(0.803030303030303)}
Writing results to /tmp/gpqa_google_gemma-4-26B-A4B-it.json
Total latency: 4469.026 s
Score: 0.803
```

---

python/sgl_jax/srt/hf_transformers_utils.py
```diff
@@ -49,13 +49,13 @@ class GlmMoeDsaConfig(PretrainedConfig):
 _UNSET = object()
 
 
-def download_from_hf(model_path: str, allow_patterns: list[str] | None = _UNSET):
+def download_from_hf(model_path: str, allow_patterns: list[str] | None = _UNSET, cache_dir: str | None = None):
```
hks-9697-v2: This ensures that when a model is downloaded to a custom cache directory (such as /dev/shm), the tokenizer correctly locates and loads the associated chat template and configuration files from the cached location rather than re-downloading or failing.

python/pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "safetensors~=0.6.1",
     "setproctitle~=1.3.6",
     "tiktoken>=0.10.0",
-    "transformers~=4.57.1",
+    "transformers~=5.5.0",
```
hks-9697-v2: This version is the minimum required for gemma4 models,

Currently working on testing the newly released 12B Dense models, that would need 5.10.0

python/sgl_jax/srt/configs/model_config.py
```diff
@@ -221,6 +221,8 @@ def __init__(
                 self.num_hidden_layers = model_layer_nums
                 # Also update hf_config to ensure consistency across all components
                 self.hf_config.num_hidden_layers = model_layer_nums
+                if hasattr(self, "hf_text_config") and self.hf_text_config is not None:
```
hks-9697-v2, line 247: Make configurable layer loading work for models like gemma that use hf_text_config attribute

test/srt/run_eval.py
```diff
@@ -178,13 +182,20 @@ def run_eval(args):
...
     parser.add_argument("--temperature", type=float, default=0.0)
+    parser.add_argument("--top-p", "--topp", dest="top_p", type=float, default=None)
     parser.add_argument("--max-tokens", type=int, default=2048)
     parser.add_argument("--top-k", type=int, default=None)
     parser.add_argument("--min-p", type=float, default=None)
     parser.add_argument("--presence-penalty", type=float, default=None)
     parser.add_argument("--repetition-penalty", type=float, default=None)
     parser.add_argument("--frequency-penalty", type=float, default=None)
     parser.add_argument("--seed", type=int, default=None)
+    parser.add_argument(
```
hks-9697-v2, line 193: Add thinking arguments, without which thinking failed to initialize and got incorrect results on evals.

python/sgl_jax/srt/managers/utils.py
```diff
@@ -48,7 +48,7 @@ def resolve_future_token_ids(input_ids, future_token_ids_map, mesh):
     input_ids_global = jax.sharding.reshard(input_ids, NamedSharding(mesh, P()))
     input_ids_global = jnp.where(
         input_ids_global < 0,
-        future_token_ids_map[jnp.clip(-input_ids_global, a_min=0)],
```
hks-9697-v2, line 51: Used to get a syntax error with a_min

hks-9697-v2:
cc @cjx0709  @pengchengneo can you help with reviewing this change?

python/sgl_jax/srt/managers/tokenizer_manager.py
```diff
@@ -359,19 +373,20 @@ def _create_tokenized_object(
...
+            return_logprob=obj.return_logprob,
+            return_output_logprob_only=obj.return_output_logprob_only,
+            logprob_start_len=obj.logprob_start_len,
+            top_logprobs_num=obj.top_logprobs_num,
+            token_ids_logprob=obj.token_ids_logprob,
+            stream=obj.stream,
+            lora_id=obj.lora_id,
+            extra_key=obj.extra_key,
+            return_routed_experts=obj.return_routed_experts,
+            mm_inputs=getattr(obj, "mm_inputs", None),
```
cjx0709: When `obj.input_embeds` is provided, `_tokenize_one_request` creates dummy `input_ids`, but this field leaves `mm_inputs` as `None`. The scheduler only forwards precomputed embeddings from `recv_req.mm_inputs["multimodal_embedding"]`, so direct `input_embeds` requests will run on the dummy token ids instead of the supplied embeddings. Please convert `input_embeds` into `mm_inputs["multimodal_embedding"]` here, or otherwise pass it through to `Req.multimodal_embedding`, and add coverage for that path.

python/sgl_jax/srt/managers/tokenizer_manager.py
```diff
@@ -303,6 +304,10 @@ async def _tokenize_one_request(
                 )
             encoded = self.tokenizer(input_text)
             input_ids = encoded["input_ids"]
+        elif input_ids is None and getattr(obj, "input_embeds", None) is not None:
+            input_ids = [0] * len(obj.input_embeds)
+        elif input_ids is None and getattr(obj, "mm_inputs", None) is not None:
+            input_ids = [0]
```
cjx0709: For `mm_inputs`-only requests this hard-codes a one-token prompt even when `mm_inputs["multimodal_embedding"]` has multiple rows. Later assembly slices the embedding by the request `seq_len`, so only the first embedding row can be consumed. Please derive the dummy `input_ids` length from `multimodal_embedding`, or reject `mm_inputs`-only requests unless `input_ids`/`text` are provided.

python/sgl_jax/srt/models/gemma4.py
```diff
@@ -0,0 +1,845 @@
...
+        token_to_kv_pool: KVCache,
+    ):
+        if (
+            forward_batch.input_embedding is not None
+            and forward_batch.forward_mode.is_extend_or_draft_extend_or_mixed()
+        ):
+            hidden_states = forward_batch.input_embedding
+        else:
+            hidden_states = self.embed_tokens(forward_batch.input_ids)
+        hidden_states *= jnp.array([self.hidden_size**0.5], dtype=hidden_states.dtype)
```
cjx0709: This scales both token embeddings and precomputed `ForwardBatch.input_embedding`. The `input_embedding` path is already in hidden-state space, for example direct `input_embeds` or merged multimodal embeddings, so applying the token embedding scale here changes its magnitude and will skew outputs. Please move this multiply into the `embed_tokens` branch only.

hks-9697-v2: Fixed this.

python/sgl_jax/srt/managers/tokenizer_manager.py
hks-9697-v2: Trying to simply the changes in this PR, removing mm pipeline for now, will create a separate PR for that with this fix.

python/sgl_jax/srt/managers/tokenizer_manager.py
hks-9697-v2: Trying to simplify the changes in this PR, removing mm pipeline for now, will create a separate PR for that with this fix.

hks-9697-v2:
@cjx0709 Trying to simplify the changes in this PR, removing mm pipeline for now, will create a separate PR for that.

Can you review the other changes. Thanks!!

cjx0709:
> @cjx0709 Trying to simplify the changes in this PR, removing mm pipeline for now, will create a separate PR for that.
> 
> Can you review the other changes. Thanks!!

the fixes is all right! But I see a lot of ci fail, maybe you can fix it in the next step.

hks-9697-v2:
> > @cjx0709 Trying to simplify the changes in this PR, removing mm pipeline for now, will create a separate PR for that.
> > Can you review the other changes. Thanks!!
> 
> the fixes is all right! But I see a lot of ci fail, maybe you can fix it in the next step.

Looks like the CI failures arise from the transformers library upgrade (tested it with the older library and the failing tests pass) any advice on how this kind of change has been handled previously? 

We need the version upgrade because gemma 4 config is only available starting this version.

I have 2 options
1. I can land this PR downgrading the transformers library, (people can run gemma by manually upgrading) we can upgrade as a follow up PR or 
2. Fix all the issues in this PR itself, might make it a bit messy.

cjx0709:
> > > @cjx0709 Trying to simplify the changes in this PR, removing mm pipeline for now, will create a separate PR for that.
> > > Can you review the other changes. Thanks!!
> > 
> > 
> > the fixes is all right! But I see a lot of ci fail, maybe you can fix it in the next step.
> 
> Looks like the CI failures arise from the transformers library upgrade (tested it with the older library and the failing tests pass) any advice on how this kind of change has been handled previously?
> 
> We need the version upgrade because gemma 4 config is only available starting this version.
> 
> I have 2 options
> 
> 1. I can land this PR downgrading the transformers library, (people can run gemma by manually upgrading) we can upgrade as a follow up PR or
> 2. Fix all the issues in this PR itself, might make it a bit messy.

I think you can fix it by copying the config file into our project. As for the transformer version, if want to upgrade it, maybe use another single pr.

hks-9697-v2:
> > > > @cjx0709 Trying to simplify the changes in this PR, removing mm pipeline for now, will create a separate PR for that.
> > > > Can you review the other changes. Thanks!!
> > > 
> > > 
> > > the fixes is all right! But I see a lot of ci fail, maybe you can fix it in the next step.
> > 
> > 
> > Looks like the CI failures arise from the transformers library upgrade (tested it with the older library and the failing tests pass) any advice on how this kind of change has been handled previously?
> > We need the version upgrade because gemma 4 config is only available starting this version.
> > I have 2 options
> > 
> > 1. I can land this PR downgrading the transformers library, (people can run gemma by manually upgrading) we can upgrade as a follow up PR or
> > 2. Fix all the issues in this PR itself, might make it a bit messy.
> 
> I think you can fix it by copying the config file into our project. As for the transformer version, if want to upgrade it, maybe use another single pr.

Modified it that way.

hks-9697-v2:
@cjx0709 the ci tests have resolved, thanks for the suggestion.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -180,6 +180,20 @@ def __init__(
...
+        if hasattr(origin_input_ids, "input_ids"):
+            origin_input_ids = origin_input_ids.input_ids
+        elif isinstance(origin_input_ids, dict) and "input_ids" in origin_input_ids:
+            origin_input_ids = origin_input_ids["input_ids"]
+        if hasattr(origin_input_ids, "tolist"):
+            origin_input_ids = origin_input_ids.tolist()
+        if (
+            isinstance(origin_input_ids, (list, tuple))
+            and len(origin_input_ids) > 0
+            and isinstance(origin_input_ids[0], (list, tuple))
```
cjx0709: This normalization runs on **every request** and silently collapses a genuinely batched input `[[...], [...]]` down to its first sequence (`origin_input_ids[0]`) — data loss without an error. It also unconditionally copies via `list(...)`.

Consider moving this format-coercion to the API boundary (tokenizer_manager / serving layer) and raising on a true nested batch rather than silently truncating.

python/sgl_jax/srt/hf_transformers_utils.py
```diff
@@ -276,7 +281,21 @@ def get_tokenizer(
...
+    try:
+        config_path = os.path.join(tokenizer_name, "config.json")
+        if os.path.exists(config_path):
+            with open(config_path) as f:
+                import json
+
+                model_config_data = json.load(f)
+                if model_config_data.get("model_type") == "gemma4":
+                    kwargs.setdefault("extra_special_tokens", {})
+    except Exception:
```
cjx0709: Bare `except Exception: pass` here (and again at ~L321) silently swallows failures in the `extra_special_tokens` / `chat_template.jinja` workarounds, which will be painful to debug if the template fails to load. Consider at least a `logger.debug(...)` in the handler.

python/sgl_jax/srt/configs/model_config.py
```diff
@@ -588,6 +596,20 @@ def configure_for_tensor_parallel(self, tensor_parallel_size: int):
...
+            orig_global = getattr(self.hf_text_config, "num_global_key_value_heads", 1)
+            if not hasattr(self, "_original_global_kv_heads"):
+                self._original_global_kv_heads = orig_global
+                self._original_local_kv_heads = self._original_hf_num_key_value_heads
+            from sgl_jax.srt.utils.jax_utils import get_num_kv_heads_by_tp
+
+            global_per_dev = get_num_kv_heads_by_tp(
+                self._original_global_kv_heads, tensor_parallel_size
+            )
+            self.hf_text_config.num_global_key_value_heads = global_per_dev * tensor_parallel_size
```
cjx0709: Architecture suggestion: instead of adding a parallel `global_*` vocabulary, consider remapping Gemma4's HF fields onto the repo's existing `head_dim = full / swa_* = sliding` convention at config-load time (inside `Gemma4Config`), e.g.:

```python
# HF Gemma4: head_dim is the sliding value, global_head_dim is the full value
tc = self.text_config
tc.swa_head_dim = tc.head_dim
tc.head_dim = getattr(tc, "global_head_dim", tc.head_dim)
tc.swa_num_key_value_heads = tc.num_key_value_heads
tc.num_key_value_heads = getattr(tc, "num_global_key_value_heads", tc.num_key_value_heads)
```

That keeps the change local to Gemma4 and lets you drop this `configure_for_tensor_parallel` block, which currently keys off `layer_types`. `layer_types` is also defined by Gemma2/Gemma3/qwen3_5, so as written every TP run of those models picks up `_original_global_kv_heads` / `num_global_key_value_heads` too.

After the remap most of the new surface folds into the existing SWA path:
- `adjust_layer_num` / `set_num_token_hybrid` already collapse head-count into `ratio = swa_heads / full_heads`; just multiply that ratio by `swa_head_dim / head_dim` and the `has_global_head_dim` branches there (and in `_compute_cell_size`) reduce to the current formula — MiMo (same head_dim) degenerates cleanly.
- `has_global_head_dim` / `global_head_dim` / `get_local_global_*` here can go away.

The one genuinely new capability is worth keeping: a different head_dim per pool — i.e. `swa_head_dim` on `SWAKVPool` plus the two-head_dim replication in `weight_utils`, just renamed off `global_*` onto `swa_head_dim` / `swa_num_key_value_heads`.

python/sgl_jax/srt/entrypoints/openai/serving_chat.py
```diff
@@ -253,6 +253,22 @@ def _apply_jinja_template(
                 **chat_template_kwargs,
             )
 
+        if (
+            hasattr(prompt_ids, "get")
+            and "input_ids" in prompt_ids
+            or isinstance(prompt_ids, dict)
+            and "input_ids" in prompt_ids
+        ):
```
cjx0709: Operator precedence: `A and B or C and D` parses as `(A and B) or (C and D)`, and both halves test the same `"input_ids" in prompt_ids`. The `hasattr(prompt_ids, "get")` arm also matches non-dict mappings. `isinstance(prompt_ids, dict)` alone covers the intended case:

```suggestion
        if isinstance(prompt_ids, dict) and "input_ids" in prompt_ids:
```

cjx0709:
@hks-9697-v2 hello, have reviewed your pull request again. The previous model in the repository that implemented SWA, the mimo-v2 series, used swa_xxx instead of global_xxx as the parameter to distinguish it from full attention. I think using a separate set of global_xxx is redundant and unreasonable. I have given specific suggestions for modification above, which you can refer to.

hks-9697-v2:
@cjx0709 Updated as per your comments, 

1. Remapped gemma 4 to existing mimo nomenclature.
2. Added debug logs for the config and template loading exception blocks
3. Moved request format normalization to tokenizer manager and raising errors on nested batch.

python/sgl_jax/srt/configs/gemma4.py
```diff
@@ -0,0 +1,34 @@
...
+
+        super().__init__(**kwargs)
+
+        tc = self.text_config if self.text_config is not None else self
+        tc.swa_head_dim = getattr(tc, "head_dim", None)
+        tc.head_dim = getattr(tc, "global_head_dim", tc.swa_head_dim)
+        tc.swa_num_key_value_heads = getattr(tc, "num_key_value_heads", None)
+        tc.num_key_value_heads = getattr(
+            tc, "num_global_key_value_heads", tc.swa_num_key_value_heads
+        )
```
cjx0709: Idempotency: this remap runs on every construction, and `global_head_dim` / `num_global_key_value_heads` are never cleared. If the config is ever round-tripped — `save_pretrained` + `from_pretrained`, or any `Gemma4Config(**cfg.to_dict())` — the serialized dict already carries `head_dim == global_head_dim` (full) next to `swa_head_dim` (sliding). A second pass then runs `swa_head_dim = head_dim`, silently overwriting the sliding head_dim with the full one (256 → 512). #1301 guarded this with a `_gemma4_patched` flag; a one-shot guard here does the same:

```suggestion
        tc = self.text_config if self.text_config is not None else self
        if not getattr(tc, "_gemma4_remapped", False):
            tc.swa_head_dim = getattr(tc, "head_dim", None)
            tc.head_dim = getattr(tc, "global_head_dim", tc.swa_head_dim)
            tc.swa_num_key_value_heads = getattr(tc, "num_key_value_heads", None)
            tc.num_key_value_heads = getattr(
                tc, "num_global_key_value_heads", tc.swa_num_key_value_heads
            )
            tc._gemma4_remapped = True
```

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
```diff
@@ -214,6 +214,43 @@ def align128(x: int) -> int:
...
+                swa_layers = sum(1 for lt in layer_types if lt == "sliding_attention")
+                full_layers = len(layer_types) - swa_layers
+            else:
+                pattern = getattr(self.model_config.hf_config, "hybrid_layer_pattern", None)
+                if pattern is None:
+                    swa_layers = 0
+                    full_layers = num_layers
+                else:
+                    swa_layers = sum(1 for p in pattern if p == 1)
+                    full_layers = sum(1 for p in pattern if p == 0)
```
cjx0709: This SWA-vs-full layer-count derivation (`layer_types` → else `hybrid_layer_pattern`) is duplicated almost verbatim in `model_runner.adjust_layer_num` (~L394-403). Two copies of the same parsing logic will drift over time (e.g. when a new `layer_types` value or pattern encoding shows up). Consider lifting it into one helper on `ModelConfig` — e.g. `get_hybrid_layer_counts() -> tuple[int, int]` returning `(swa_layers, full_layers)` — and calling it from both sites.

python/sgl_jax/srt/managers/tokenizer_manager.py
```diff
@@ -305,6 +306,23 @@ async def _tokenize_one_request(
...
+            input_ids = input_ids.tolist()
+        if (
+            isinstance(input_ids, (list, tuple))
+            and len(input_ids) > 0
+            and isinstance(input_ids[0], (list, tuple))
+        ):
+            if len(input_ids) > 1:
+                raise ValueError("Nested batch input is not supported for a single request.")
+            input_ids = input_ids[0]
+        input_ids = list(input_ids)
```
cjx0709: This whole normalization block adapts to return shapes that only a newer transformers produces (a `BatchEncoding`/tensor object, or a nested `[[...]]`). On the version this PR pins (`~=4.57.1`), `self.tokenizer(input_text)["input_ids"]` for a single request is already a flat `list[int]` — I verified this empirically on 4.57.6, and the shape logic lives in the base class `tokenization_utils_base.py`, so it's version-governed, not model-specific. None of these branches fire on the pinned version; it's dead code that only matters once transformers is bumped.

This PR doesn't bump transformers, so please drop this block and keep it for the future transformers-upgrade PR. If you want a genuine nested-batch guard, that's a separate general change and should land on its own rather than bundled into the Gemma4 model PR.

python/sgl_jax/srt/entrypoints/openai/serving_chat.py
```diff
@@ -253,6 +253,17 @@ def _apply_jinja_template(
...
+        if isinstance(prompt_ids, dict) and "input_ids" in prompt_ids:
+            prompt_ids = prompt_ids["input_ids"]
+        if hasattr(prompt_ids, "tolist"):
+            prompt_ids = prompt_ids.tolist()
+        if (
+            isinstance(prompt_ids, list)
+            and len(prompt_ids) == 1
+            and isinstance(prompt_ids[0], list)
+        ):
+            prompt_ids = prompt_ids[0]
```
cjx0709: Same as the tokenizer_manager normalization: on the pinned 4.57.x, the single-request `apply_chat_template` / `tokenizer(str)` path returns a flat list, so `tolist()` and the `len==1` nested-list unwrap never trigger. This is adapting to a newer transformers' shapes — please remove it from this PR and keep it for the transformers-bump work.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -181,6 +181,8 @@ def __init__(
         # Input and output info
         self.rid = rid
         self.origin_input_text = origin_input_text
+        origin_input_ids = list(origin_input_ids)
```
cjx0709: This re-copy is a leftover from the input normalization being added at the request boundary — once that normalization lands (or is dropped, as suggested), `origin_input_ids` arriving here is already a list. It's redundant; please drop it.

test/srt/test_gemma4_models.py
```diff
@@ -0,0 +1,231 @@
...
+_orig_from_pretrained = AutoTokenizer.from_pretrained
+
+
+def _patched_from_pretrained(pretrained_model_name_or_path, *args, **kwargs):
+    if "gemma-4" in str(pretrained_model_name_or_path):
+        kwargs["extra_special_tokens"] = {}
+    return _orig_from_pretrained(pretrained_model_name_or_path, *args, **kwargs)
+
+
+AutoTokenizer.from_pretrained = _patched_from_pretrained
```
cjx0709: This monkeypatch sets `extra_special_tokens={}` for `gemma-4`, which duplicates the workaround you already added in `hf_transformers_utils.get_tokenizer`. If the server path already handles it, this global `AutoTokenizer.from_pretrained` patch is redundant and surprising to leave in a test module — please remove it and rely on the production workaround.

cjx0709:
@hks-9697-v2 thanks for your new updated commit, I review again and come with some new comment. There's nothing before merge After you solve these and pass the CI.

hks-9697-v2:
Updated as per your comments:
1. Added _gemma4_remapped in  Gemma4Config and added a common function get_hybrid_layer_counts which is called from both ModelRunner and ModelRunnerKVCacheMixin.

2. Removed Request normalization related code.

3. Removed the monkey patch from the test file. 

**Note on update in bench_serving.py**:
I had to update get_tokenizer in bench_serving.py to successfully remove the monkeypatch from the test file. While the server process uses hf_transformers_utils.get_tokenizer, the benchmark client previously bypassed it for Hugging Face model IDs and called AutoTokenizer.from_pretrained directly, which crashed on gemma 4's extra_special_tokens. Added a list _CUSTOM_TOKENIZER_MODEL_FAMILIES = ("gemma-4",) to route the benchmark client to the production workaround for gemma.

test/srt/test_gemma4_models.py
hks-9697-v2: Removed this patch but I had to update get_tokenizer in bench_serving.py to successfully remove the monkeypatch from the test file. While the server process uses hf_transformers_utils.get_tokenizer, the benchmark client previously bypassed it for Hugging Face model IDs and called AutoTokenizer.from_pretrained directly, which crashed on gemma 4's extra_special_tokens. Added a list _CUSTOM_TOKENIZER_MODEL_FAMILIES = ("gemma-4",) to route the benchmark client to the production workaround for gemma.

hks-9697-v2:
@cjx0709 I think you will have to start the CI flow. It says this workflow requires approval from a maintainer.

cjx0709 (approved):
LGTM

PR #1328, opened 2026-06-08
fix(schedule): chunked prefill serialized to one DP rank when radix enabled

### LINKED ISSUE #1239: perf: SWARadixCache 3x slower than SWAChunkCache for hybrid models (MiMo-V2-Flash)
## Summary

With hybrid SWA models (MiMo-V2-Flash), radix cache enabled (`SWARadixCache`) is ~3x slower than disabled (`SWAChunkCache`) on the same benchmark. This is not a regression from #1234 — main branch shows the same or worse gap.

## Benchmark (16-chip v6e, dp=4, tp=16, ep=16, epmoe)

| Config | Output tok/s | Input tok/s | Duration (s) | Mean TTFT (ms) |
|---|---|---|---|---|
| `--disable-radix-cache` (SWAChunkCache) | **705** | **11284** | **371** | **33647** |
| radix cache enabled (SWARadixCache) | 235 | 3760 | 1115 | 128307 |

Same benchmark command, only difference is presence/absence of `--disable-radix-cache`.

## Possible Causes

1. **Extend-time SWA eviction skipped**: `SWARadixCache` defers SWA eviction to tree insert / tree-pressure handling, while `SWAChunkCache` does per-request eviction during extend. This causes higher peak SWA occupancy and more pressure-driven eviction later.

2. **Tree operation overhead**: Every chunk requires `insert` into radix tree, `match_prefix` for prefix matching, and LRU-based tree eviction. With random inputs (no prefix reuse), this is pure overhead.

3. **`evict_req_swa` vs direct `free_swa`**: `SWARadixCache.evict_req_swa` computes `_node_prefix_len`, respects tree-owned SWA slots, and has more complex logic than `ChunkCache`'s direct `free_swa` path.

## Reproduction

```bash
# Server
  --model-path /models/MiMo-V2-Flash --trust-remote-code \
  --tp-size 16 --dp-size 4 --ep-size 16 --moe-backend epmoe \
  --nnodes 4 --node-rank $TPU_WORKER_ID \
  --dist-init-addr $HEAD_SVC:30000 \
  --page-size 256 --context-length 262144 \
  --chunked-prefill-size 2048 --dtype bfloat16 --mem-fraction-static 0.95 \
  --swa-full-tokens-ratio 0.2 --max-running-requests 128 \
  --attention-backend fa --dp-schedule-policy round_robin
  # Add --disable-radix-cache for SWAChunkCache path

# Benchmark
python -m sgl_jax.bench_serving \
  --backend sgl-jax --host 127.0.0.1 --port 30271 \
  --dataset-name random --random-input-len 16384 --random-output-len 1024 \
  --random-range-ratio 1.0 --flush-cache --seed 12345 \
  --request-rate 100 --num-prompts 256 --max-concurrency 64 \
  --tokenizer XiaomiMiMo/MiMo-V2-Flash
```

Related: #1227, #1234

zorrofox:
Fixes #1239.

## Root cause

Not radix-tree CPU overhead. Instrumented all `SWARadixCache` public methods on the #1239 repro — total wall time across 200 `cache_unfinished_req` calls is **0.09s**, nowhere near the 50s TTFT gap.

The actual cause is `PrefillAdder.add_one_req` returning `OTHER` on the second DP rank:

```python
total_can_run = sum(len(v) for v in self.can_run_list.values())
if real_input_tokens >= self.rem_input_tokens and total_can_run != 0:
    return AddReqResult.OTHER
```

`real_input_tokens` is the **untruncated** extend length (16384), `rem_input_tokens` is the **global** `max_prefill_tokens` budget. After dp_rank=0's 2048-token chunk is admitted, `rem_input_tokens` drops to 14336; dp_rank=1's req sees `16384 >= 14336 and 1 != 0` → `OTHER` → scheduler treats as global stop → break. So **only one DP rank prefills per round**:

| | `#prefill per DP` distribution (n=64) |
|---|---|
| `--disable-radix-cache` | `[1,1,1,1]` ×110, `[1,1,1,0]`-ish ×17, `[1,0,0,0]` ×25 |
| radix enabled (before) | **`[1,0,0,0]`/`[0,1,0,0]`/... ×257** |
| radix enabled (after) | `[1,1,1,1]` ×59, `[0,1,1,1]` ×8, `[1,0,0,0]` ×10 |

The `--disable-radix-cache` path goes through `add_one_req_ignore_eos` instead (because `getattr(ChunkCache, "disable", True)` defaults `True`), which has no such gate — so it looked like radix was 3× slower when it was purely a scheduling difference.

This also affects `ignore_eos=False` traffic with **either** cache impl (both reach `add_one_req`), so production long-context + dp>1 + chunked prefill is serialized regardless of the radix flag — bench_serving's default `ignore_eos=True` just masked that side.

## Fix

Match GPU sglang ([schedule_policy.py:908](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/schedule_policy.py)): only apply this gate when chunked prefill is **disabled** (`rem_chunk_tokens_list is None`). With chunked prefill the per-round footprint is bounded by `rem_chunk_tokens_list[dp_rank]`, so comparing the untruncated length against the global budget is over-rejection.

## Result (v6e-64 dp=4, MiMo-V2-Flash, random 16K/1K n=64 conc=32, main@4d02695f)

| | Output tok/s | Median TTFT | Median TPOT | Median ITL |
|---|---|---|---|---|
| radix off | 534.71 | 15.7 s | 44.54 ms | 21.57 ms |
| radix on (before) | 207.32 | 65.2 s | 90.96 ms | — |
| **radix on (after)** | **545.48** | 20.7 s | **38.17 ms** | **18.31 ms** |

Output/TPOT/ITL now at parity or better. The remaining ~5s TTFT delta is a one-off XLA compile for the cache-hit prefill shape (`#new-token: 256`) under `--disable-precompile`; radix-off never hits cache so never compiles that shape.

## Test

`test_mixed_chunk_dp.py::test_add_one_req_chunked_admits_all_dp_ranks_with_radix` — dp=4, chunked=2048, max_prefill=16384, four 16K reqs on the radix path. RED on main (`dp_rank=1` returns `OTHER`), GREEN with fix.

cc @pengchengneo @cjx0709

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_policy.py
+++ b/python/sgl_jax/srt/managers/schedule_policy.py
@@ -582,7 +582,15 @@ class PrefillAdder:
                 return AddReqResult.NO_TOKEN
 
         total_can_run = sum(len(v) for v in self.can_run_list.values())
-        if real_input_tokens >= self.rem_input_tokens and total_can_run != 0:
+        if (
+            self.rem_chunk_tokens_list is None
+            and real_input_tokens >= self.rem_input_tokens
+            and total_can_run != 0
+        ):
+            # Without chunked prefill, the full extend_input_len is committed this round,
+            # so gate on the global rem_input_tokens. With chunked prefill the per-round
+            # footprint is bounded by rem_chunk_tokens_list[dp_rank]; comparing the
+            # untruncated length here would serialize admission to one DP rank.
             return AddReqResult.OTHER
 
         with self._lock_node(req.last_node):
@@ -598,7 +606,11 @@ class PrefillAdder:
             input_tokens = self.ceil_paged_tokens(req.extend_input_len)
 
             total_can_run = sum(len(v) for v in self.can_run_list.values())
-            if input_tokens >= self.rem_input_tokens and total_can_run != 0:
+            if (
+                self.rem_chunk_tokens_list is None
+                and input_tokens >= self.rem_input_tokens
+                and total_can_run != 0
+            ):
                 return AddReqResult.OTHER
 
             if (
--- a/python/sgl_jax/test/test_mixed_chunk_dp.py
+++ b/python/sgl_jax/test/test_mixed_chunk_dp.py
@@ -4,7 +4,7 @@ from types import SimpleNamespace
 import numpy as np
 
 from sgl_jax.srt.managers.schedule_batch import Req, ScheduleBatch
-from sgl_jax.srt.managers.schedule_policy import PrefillAdder
+from sgl_jax.srt.managers.schedule_policy import AddReqResult, PrefillAdder
 from sgl_jax.srt.sampling.sampling_params import SamplingParams
 
 
@@ -16,6 +16,21 @@ class _DummyAllocator:
 class _DummyTreeCache:
     def evictable_size(self, dp_rank: int = 0):
         return 0
+
+
+class _DummyRadixCache:
+    """Tree cache with disable=False so add_one_req takes the radix path."""
+
+    disable = False
+
+    def evictable_size(self, dp_rank: int = 0):
+        return 0
+
+    def inc_lock_ref(self, node):
+        return None
+
+    def dec_lock_ref(self, node, swa_uuid_for_lock=None):
+        pass
 
 
 def _make_req(rid: str, dp_rank: int, input_len: int = 4, output_len: int = 2) -> Req:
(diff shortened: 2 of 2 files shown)


cjx0709 (approved):
nice fix！LGTM

Issue #293, opened 2025-10-30
[Feature] Speed Up Scheduler Thread Performance


JamesBrianD:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Scheduler overlap is currently supported, but when scheduler scheduling time exceeds model forward inference time, scheduling cannot be fully masked. For example, with the QWEN3-8B model, concurrency set to 16, input size 1024, output size 1024, scheduling takes 5.5ms, while model run + sampler only takes 5ms.
Therefore, we need to speed up scheduler thread performance.
<img width="2930" height="1388" alt="Image" src="https://github.com/user-attachments/assets/1975502c-4bab-4568-80f2-1aeaf3400ad9" />

### Related resources
Refer to this document [profile-with-jax-profiler](https://github.com/sgl-project/sglang-jax/blob/main/docs/developer_guide/benchmark_and_profiling.md#profile-with-jax-profiler)
 to profile the scheduler and attempt to modify the code for acceleration

_No response_

Yajingleo:
## Update: the scheduler is masked under the forward — appears resolved on current `main`

I added env-gated per-forward-step timing to `event_loop_overlap` (PR #1484) splitting each step into `recv / schedule / build (get_model_worker_batch + dispatch) / result (forward-wait + D2H)`, and measured on current `main`.

**Config:** Qwen3-8B, tp=4, v6e-4, mem-frac 0.8, page-size 64, fa, chunked-prefill 8192.

### Single operating point (conc 16, 1024/1024)
| phase | time |
|-------|------|
| recv | 0.04 ms |
| schedule | 0.15 ms |
| **build** (`get_model_worker_batch`) | **~2.5 ms** |
| **HOST total** | **~2.6–2.9 ms** |
| forward (step) | ~5.7–5.9 ms |
| **host / forward** | **~0.46** |

Host work is <half the step → **masked by the overlap**. Throughput confirms it: ~2,450 tok/s, i.e. step ≈ forward, not host + forward.

### Distribution over a varied workload (10,350 forward steps)
To check the *tail* rather than just the mean, I ran `random --random-range-ratio 0.5` (lengths in `[512, 1024]`, mixed prefill/decode):

```
[...]

PR #1329, opened 2026-06-08
Optimize scheduler host-side batching overhead

AmedeoBiolatti:
## Motivation
Reduce host-side scheduler overhead in the prefill path, especially in `Scheduler.get_next_batch_to_run()`.

This addresses the scheduler overhead discussed in https://github.com/sgl-project/sglang-jax/issues/293.

The goal is behavior parity with lower Python/Numpy overhead during request admission and batch preparation.

## Modifications
- Avoid unnecessary `fill_ids` copies when a request has no generated output yet.
- Skip radix-prefix matching work when the prefix cache is disabled.
- Replace Python list flattening via `sum(..., [])` with `np.fromiter`.
- Avoid rebuilding `set(all_can_run_reqs)` for every waiting-queue element.
- Use a narrower post-admission budget check in `PrefillAdder`.
- Batch several logits metadata host-to-device transfers through one `device_array` call.
- Fill extend positions directly into the target numpy array.

## Benchmarking and Profiling
I created a small micro-benchmark for `Scheduler.get_next_batch_to_run()` in isolation. I link the gist: https://gist.github.com/AmedeoBiolatti/a4d04e53b3e0640fb1a98b00ae8056eb

Command used locally:
`python bench_get_next_batch_to_run.py --iterations 300 --warmup 30 --queue-size 512 --dp-size 4 --max-running-requests 512`

Results:
| main | this |
|--|--|
| 1.776 ms | 1.122 ms |

Improvement: ~36.8%.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_policy.py
+++ b/python/sgl_jax/srt/managers/schedule_policy.py
@@ -407,6 +407,17 @@ class PrefillAdder:
 
         return AddReqResult.CONTINUE
 
+    def _budget_state_after_add(self, dp_rank: int):
+        if self.rem_input_tokens <= 0 or (
+            self.rem_chunk_tokens is not None and self.rem_chunk_tokens <= 0
+        ):
+            return AddReqResult.OTHER
+
+        if self.is_hybrid and self.rem_swa_tokens_for_dp(dp_rank) <= 0:
+            return AddReqResult.NO_TOKEN
+
+        return AddReqResult.CONTINUE
+
     def add_chunked_req(self, req: Req):
         dp_rank = req.dp_rank if req.dp_rank is not None else 0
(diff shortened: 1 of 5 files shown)


python/sgl_jax/srt/managers/schedule_policy.py
cjx0709, line 396: if not one use this function, you can delete it.(And maybe update the doc:"docs/architecture/03-scheduler.md"

AmedeoBiolatti: I thinks this is still used in one place

(docs updated!)

cjx0709 (approved):
LGTM, will be merged when ci pass.

PR #1331, opened 2026-06-08
[Kimi-K2.5] Adding Text Processor for KimiK25

depksingh:
## Motivation
Adding support for kimi text processor (DeepseekV3) backbone for KimiK25. [Link](https://huggingface.co/moonshotai/Kimi-K2.5/) as part of the issue : https://github.com/sgl-project/sglang-jax/issues/1011https://github.com/sgl-project/sglang-jax/issues/1011 and on top of PR: https://github.com/sgl-project/sglang-jax/pull/1314. Currently this PR is tested on the BF16 weights, and INT4 version of it is raised here: https://github.com/sgl-project/sglang-jax/pull/1334

## Modifications
- Added VL_Generation file for KimiK2.5
- Added the handling of embedding tokens in deepseekv3 on the same lines as Qwen2 and other models in repo.

## Accuracy Tests
Ran the AIME25 benchmarks with up to the mark results verifying the correct implementation.
Pasted Graphic 6

Verified the model output by running a sample prompt on 2x2x4 v7x multihost scenario.
```
JAX_COMPILATION_CACHE_DIR=/tmp/tpu_logs python3 -u -m sgl_jax.launch_server     --multimodal     --model-path=/local/kimi     --log-requests --trust-remote-code --page-size 256 --chunked-prefill-size 2048 --max-running-requests 64 --tp-size 32 --ep-size 32 --precompile-bs-paddings 64
```
Pasted Graphic 8

## Benchmarking and Profiling
NA

## Checklist


depksingh:
@cjx0709 @chaitanyapk @Bob-Chen222 , Raised this PR for VL_Generation part of the model as discussed here: https://github.com/sgl-project/sglang-jax/pull/1314#issuecomment-4629223588, can you please review it whenever you get time.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def __init__(
+        self,
+        config,
+        mesh,
+        dtype=jnp.bfloat16,
+    ):
+        super().__init__(config=config, mesh=mesh, dtype=dtype)
+
+    def __call__(
```
cjx0709: This override is justified (the base `DeepseekV3Model.__call__` computes `embed_tokens(input_ids)` internally with no hook to inject precomputed embeds, so this is the cleanest way to support `forward_batch.input_embedding`). But the rest of the body — residual accumulation and final norm — is a verbatim copy of the base. If only the embedding *source* differs, consider factoring just the embedding selection so the layer loop isn't duplicated and won't drift from the base implementation.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
```diff
@@ -0,0 +1,374 @@
...
+            hidden_states += residual
+
+        hidden_states = self.norm(hidden_states)
+        return hidden_states, layers_kv_fused, layers_topk_ids
+
+
+class KimiK25ForConditionalGeneration(nnx.Module):
+
+    @classmethod
+    def patch_model_config(cls, mc: ModelConfig) -> None:
```
cjx0709: This is byte-for-byte identical to `DeepseekV3ForCausalLM.patch_model_config`. Since `KimiK25ForConditionalGeneration` is a bare `nnx.Module`, all of `patch_model_config` / `__call__` / most of `__init__` are copies rather than inherited. Consider subclassing `DeepseekV3ForCausalLM` (cf. `DeepseekV2ForCausalLM`, which is just `pass`) and overriding only `_create_weight_mappings` / `_create_layer_mappings` — that's the only part that genuinely differs due to the `language_model.` prefix. It would remove a lot of copy-paste and reduce future drift.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
```diff
@@ -0,0 +1,374 @@
...
+        mesh=None,
+    ):
+        super().__init__()
+
+        self.config = config
+        self.text_config = get_hf_text_config(config) or config
+        self.dtype = dtype or jnp.bfloat16
+        self.mesh = mesh
+
+        # TODO: Validate if any other fix can make it work smoothly
```
cjx0709: Could you expand on the root cause here? Clearing `text_config.quantization_config` reads as a workaround that may mask a config-loading issue. Specifically: why does it arrive as a raw `dict` on `text_config` while `ModelConfig` already handles quantization at the top-level `hf_config`? Knowing whether this is a Kimi-specific config nesting vs. a general gap would tell us if the fix belongs here or upstream in config loading.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
```diff
@@ -0,0 +1,374 @@
...
+        hidden_states, layers_kv_fused, layers_topk_ids = self.model(
+            forward_batch, token_to_kv_pool
+        )
+
+        if not getattr(self.text_config, "tie_word_embeddings", False):
+            output = self.logits_processor(hidden_states, self.lm_head, logits_metadata)
+        else:
+            output = self.logits_processor(hidden_states, self.model.embed_tokens, logits_metadata)
+
+        return output, layers_kv_fused, True, layers_topk_ids
```
cjx0709: The 2nd return element is a bare list here, whereas `DeepseekV3ForCausalLM.__call__` returns `{"token_to_kv_pool": layers_kv_fused}`. This works today because `MemoryPools.replace_all` normalizes a bare `list` into that dict, so it's not a functional bug for the non-hybrid pool — but please return the dict for consistency, so this doesn't silently break if a hybrid pool is ever wired in here.

cjx0709 (commented):
Thanks for this! A few inline notes, mostly about reducing duplication against `DeepseekV3ForCausalLM` and one consistency fix. Nothing blocking the text-backbone direction — the `language_model.*` weight-mapping work is the valuable part here.

cjx0709:
image hello，thanks for you contribution. Before merge, can you just eval one or more of the table shows to prove the text module is right.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
depksingh: Sure, Thank for you the review, let me go over the changes

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
depksingh: Sounds good, updated the implementation to inherit the LM class instead of creating a wrapper and moved the input embedding handling to deepseek, inline with other models in the repository like Qwen2

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
depksingh: Updated the code for the same, now inheriting DSV3CausalLM.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
depksingh: yeah, its not automatically handled as the quantization_config is nested in the config.text_config, due to which ModelConfig isn't able to pick it up and convert to QuantizationConfig class, even if we are able to do that, the 'pack-quantization' isn't supported yet in the ModelConfig, that's why clearing the quantization-config here, can later it can be passed via the quantization-config file explictly.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
depksingh: Updated.

depksingh:
> image hello，thanks for you contribution. Before merge, can you just eval one or more of the table shows to prove the text module is right.

Ran the AIME25 benchmarks with 10 prompts the the model gave 100% success rate for the same, verifying the correctness. SS attached in the description. @cjx0709 Can you please review now, if the above changes looks right.

cjx0709 (commented):
Functionally looks solid and nice to see the AIME25 numbers. The deepseek_v3 input_embedding change mirrors qwen2 exactly and is backward-compatible (input_embedding is only used in extend mode, None for text-only), so no concerns there. Left two comments on the new Kimi model class.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
```diff
@@ -0,0 +1,314 @@
...
+            mesh=self.mesh,
+            dtype=self.dtype,
+        )
+
+        if not getattr(self.text_config, "tie_word_embeddings", False):
+            self.lm_head = ParallelLMHead(
+                self.text_config.vocab_size,
+                self.text_config.hidden_size,
+                dtype=self.dtype,
+                param_dtype=dtype,
```
cjx0709: Small inconsistency here: `dtype=self.dtype` (which is `dtype or bfloat16`) but `param_dtype=dtype` uses the raw constructor arg, which can still be `None`. If a caller doesn't pass `dtype`, `param_dtype` ends up `None` and overrides ParallelLMHead's bfloat16 default. Probably want `param_dtype=self.dtype` to match.

Also note the base `DeepseekV3ForCausalLM.__init__` already allocates `self.lm_head` unconditionally, so when `tie_word_embeddings=True` you end up with a large lm_head param that's allocated but never used (the `__call__` path uses embed_tokens), and the weight mapping still references `language_model.lm_head.weight` which won't exist in a tied checkpoint. Not a blocker for Kimi-K2.5 if it's untied, but worth guarding.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
```diff
@@ -0,0 +1,314 @@
...
+                and layer_idx % moe_layer_freq == 0
+            )
+            layer_mappings = self._create_layer_mappings(
+                layer_idx, is_moe, moe_backend, use_fused, is_static_quant
+            )
+            mappings.update(layer_mappings)
+
+        return mappings
+
+    def _create_layer_mappings(
```
cjx0709: This whole method (plus `_create_weight_mappings`) is a near-verbatim copy of the base `DeepseekV3ForCausalLM` version. The only real difference is the `language_model.` HF prefix — `self.text_config` is identical to `self.config` once `super().__init__(config=self.text_config)` runs. That's ~200 duplicated lines that'll silently drift the moment someone fixes a quant/MoE mapping bug in the base class.

Would be much safer to parametrize the HF prefix in the base, e.g. a `self.hf_weight_prefix = ""` attribute that the subclass sets to `"language_model."`, then build the keys as `f"{self.hf_weight_prefix}model.layers.{layer_idx}"`. The subclass would then just override `__init__` + lm_head and drop both copied methods. Fine to do as a follow-up if you want to keep this PR tight, but please track it.

depksingh: yeah, I thought about it, but wanted to keep the model loader part separate to avoid it totally being tightly coupled to deepseek loader, so in future if any change in any extra layer in DSV or Kimi, it will be harder to refactor due to being tightly coupled. For the bugs resolution we can track it parallel and update in future accordingly. Please let me know what you think.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
depksingh: Fixed both. @cjx0709 Please check if the changes look fine whenever you get time.

python/sgl_jax/srt/multimodal/models/kimi_k25/kimi_k25_vl_generation.py
Bob-Chen222: I think it'd be good to use the deepseek's weight loader but add a regression unit test for kimi weight loading?

depksingh: Hey @Bob-Chen222 , @cjx0709 , fixed the above impl as suggested, can you please review.

cjx0709 (approved):
LGTM

depksingh:
Hey @cjx0709 , @Bob-Chen222 I'm not seeing the merge button in the PR, maybe due to the access limitations, can you please merge. Thank you 😇

PR #1347, opened 2026-06-11
feat(mem_cache): HiCache Stage 1 — UnifiedRadixCache core, FullComponent, serving tests

MokusMokun:
## Summary

Closes #1335, #1336, #1337
Refs #1311

This PR delivers **HiCache Stage 1**: a component-agnostic prefix cache (`UnifiedRadixCache` + `FullComponent`, ported from upstream `sglang`) behind a default-off `--enable-unified-radix-tree` flag, plus serving-level verification. Flag-off keeps the existing `RadixCache` path unchanged. Later stages hang SWA / recurrent / HiCache host-tier components off the same core without re-architecting.

One commit per sub-issue: S1a (typed params), S1b (core + flag), S1c (serving tests).

### S1a — typed cache interface (#1335)

| Area | Change |
|------|--------|
| `mem_cache/base_prefix_cache.py` | Evict/match/lock signatures converge on typed params (`EvictParams`, `MatchPrefixParams`, `InsertParams`, `IncLockRefResult`, `DecLockRefParams`, `EvictResult`); `MatchResult` gains `best_match_node`. Not behind the new flag — `RadixCache`/`SWARadixCache`/`ChunkCache` and all callers adapt. |
| `managers/{schedule_batch,schedule_policy}.py` | The 3 strict 4-tuple `MatchResult` unpacks switch to field-name access. |

### S1b — UnifiedRadixCache core + flag (#1336)

| Area | Change |
|------|--------|
| `mem_cache/unified_radix_cache.py` (new) | `UnifiedTreeNode` + `UnifiedRadixCache`: trie / ref-counting / eviction ported from upstream; component hooks at match/insert/evict/lock seams; `dp_rank`-aware size accounting (sgl-jax divergence). |
| `mem_cache/unified_cache_components/` (new) | `ComponentType`/`ComponentData`/`TreeComponent` ABC + `FullComponent` (full-attention). HiCache host-tier hooks are no-op stubs. |
| `server_args.py`, `mem_cache/registry.py` | `--enable-unified-radix-tree` (default off, env `SGLANG_JAX_ENABLE_UNIFIED_RADIX_TREE`); routes **non-hybrid models only** — hybrid SWA stays on `SWARadixCache`. |
| Unit tests (`unit-test-cpu`) | Core ops, `best_match_node`, equivalence-vs-`RadixCache` (page 1/128), dp_rank isolation, request-path equivalence via mock reqs, flag routing incl. hybrid/disabled branches. |

### S1c — serving tests (#1337)

[...]

PR #1349, opened 2026-06-11
fix(sampler): replace static-predicate lax.cond with plain if for seeded multinomial

MokusMokun:
On TPU, decode precompile trips a `lax.cond` branch-sharding mismatch whenever it compiles the seeded regular sampling path — `--enable-deterministic-sampling` on, so the regular (non-greedy) branch is traced with non-`None` `sampling_seeds`. This is a pre-existing bug on `main`, independent of any model or cache feature; it surfaced as the `e2e-test-tpu-v6e-1` failure on #1347 ([job](https://github.com/sgl-project/sglang-jax/actions/runs/27322543225/job/80718540579)), whose serving test enables the flag.

## Root cause

The seeded/unseeded multinomial selection uses `lax.cond` on a static predicate, in both the mask path (`sampler.py:447`) and the sort path (`sampler.py:404`, `--use-sort-for-toppk-minp`):

```python
sampled_index = lax.cond(
    sampling_seeds is not None,   # static: None vs array changes the input pytree
    multinomial_with_seed_fn,
    multinomial_fn,
    multinomial_operands,
)
```

`sampling_seeds is not None` resolves at trace time, but `lax.cond` still traces **both** branches and requires identical output avals, including sharding. With seeds present the branches diverge:

- `multinomial_with_seed` ends in `jnp.argmax(..., keepdims=True)` → preserves batch sharding → `int32[n@data, 1]`
- `multinomial` ends in `random.categorical(...).reshape(-1, 1)` → reshape drops it → `int32[n, 1]`

so cond construction fails:

```
File ".../sgl_jax/srt/layers/sampler.py", line 447, in top_k_top_p_min_p_sampling_from_probs_jax_with_mask
    sampled_index = lax.cond(
TypeError: cond branches must have equal output types but they differ.
The output of true_fun has type int32[1@data,1] but the corresponding output of false_fun has type int32[1,1].
```

Without the flag, `sampling_seeds` is `None` and `multinomial_with_seed` falls through to `multinomial`, so both branches match — which is why every other suite test passes.

## Fix

Replace the `lax.cond` with a plain Python `if` in both paths. The predicate is static, so each case is a distinct trace and the untaken branch is never traced. Sampling math is unchanged; `_regular_sampling` still reshards the result to `P("data")`.

## Reproduce (unmodified `main` branch, v6e)

Server, mask path — minimal command on v6e-1 (the CI shape), crashes at decode precompile with the traceback above:

```bash
  --model-path Qwen/Qwen3-1.7B \
  --tp-size 1 --device tpu \
  --dtype bfloat16 --mem-fraction-static 0.8 \
  --max-running-requests 16 --page-size 64 \
  --skip-server-warmup --random-seed 42 \
  --enable-deterministic-sampling
```

Direct, sort path — `jax.jit` `top_k_top_p_min_p_sampling_from_probs_jax_with_sort` on one TPU device with non-`None` `sampling_seeds` (this is the new regression test): same error class at `sampler.py:404`.

Negative control: the same server command without `--enable-deterministic-sampling` reaches health.

## Verification (fix branch, v6e)

- The deterministic-sampling server repro: `[DECODE] Precompile finished` → `The server is fired up and ready to roll!`; no cond errors in the log. (Crash and recovery validated on a v6e 4-chip host, tp=4, DeepSeek-R1-Distill-Qwen-1.5B — traceback identical to the v6e-1 CI failure.)
- New test `python/sgl_jax/test/test_sampler_deterministic_cond.py`: fails on `main` with the exact cond aval mismatch, passes with the fix. Registered in `unit-test-tpu-v6e-1`.
- `pre-commit` clean.

Notes on test scope:

- TPU-only: on CPU/GPU XLA reconciles the mismatched branch shardings, so the cond never errors.
- The test exercises the sort path, which reaches the cond directly. The mask path has the identical construct, but jitting that helper in isolation trips unrelated explicit-sharding checks in `topk_mask`'s binary search before the cond (these don't occur inside the model's full jit) — so the mask path is covered by the server validation above instead.

### DIFF
--- a/python/sgl_jax/srt/layers/sampler.py
+++ b/python/sgl_jax/srt/layers/sampler.py
@@ -398,15 +398,13 @@ def top_k_top_p_min_p_sampling_from_probs_jax_with_sort(args):
         min_p_operands,
     )
 
+    # Static predicate (None vs array changes the input pytree): use `if`, not
+    # lax.cond -- the branches' output shardings differ and cannot be unified.
     multinomial_operands = (probs_sort, sampling_seeds, positions, rng)
-    multinomial_with_seed_fn = lambda op: multinomial_with_seed((*op, True))
-    multinomial_fn = lambda op: multinomial((*op, True))
-    sampled_index = lax.cond(
-        sampling_seeds is not None,
-        multinomial_with_seed_fn,
-        multinomial_fn,
-        multinomial_operands,
-    )
+    if sampling_seeds is not None:
+        sampled_index = multinomial_with_seed((*multinomial_operands, True))
+    else:
+        sampled_index = multinomial((*multinomial_operands, True))
 
     probs_idx = probs_idx.astype(jnp.int32)
     return jnp.take_along_axis(probs_idx, axis=1, indices=sampled_index).flatten()
@@ -441,15 +439,12 @@ def top_k_top_p_min_p_sampling_from_probs_jax_with_mask(args):
         min_p_operands,
     )
 
+    # Static predicate -- use `if`, not lax.cond (see sort path).
     multinomial_operands = (logits, sampling_seeds, positions, rng)
-    multinomial_with_seed_fn = lambda op: multinomial_with_seed((*op, False))
-    multinomial_fn = lambda op: multinomial((*op, False))
-    sampled_index = lax.cond(
-        sampling_seeds is not None,
-        multinomial_with_seed_fn,
-        multinomial_fn,
-        multinomial_operands,
-    )
+    if sampling_seeds is not None:
+        sampled_index = multinomial_with_seed((*multinomial_operands, False))
+    else:
+        sampled_index = multinomial((*multinomial_operands, False))
 
     return sampled_index.flatten()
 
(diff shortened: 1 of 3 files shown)


python/sgl_jax/srt/layers/sampler.py
```diff
@@ -441,15 +439,12 @@ def top_k_top_p_min_p_sampling_from_probs_jax_with_mask(args):
...
-    sampled_index = lax.cond(
-        sampling_seeds is not None,
-        multinomial_with_seed_fn,
-        multinomial_fn,
-        multinomial_operands,
-    )
+    if sampling_seeds is not None:
+        sampled_index = multinomial_with_seed((*multinomial_operands, False))
+    else:
+        sampled_index = multinomial((*multinomial_operands, False))
```
aolemila, line 447: Can we add `out_sharding` for reshape to avoid using if? Like reshape(xxx, out_sharding)

MokusMokun: Good point, it's true that adding `out_sharding=P("data", None)` to the reshape can also fix the `lax.cond` branch mismatch on TPU. However I still prefer the Python `if` here because: 
1. adding P("data", None) inside multinomial() makes this low-level helper require an active mesh with a data axis. It's currently mesh-agnostic and these sampler helpers are called directly in eager mode (e.g. test_sampler.py), so that usage would need extra sharding plumbing.
2. `_regular_sampling` already reshards the final sampled token ids to `P("data")`, so `multinomial()` does not need to own that sharding policy. Since `sampling_seeds is not None` is static, the Python `if` keeps the fix local and avoids tracing the unused branch.

aolemila (approved):

PR #1234, opened 2026-05-28
fix(swa): SWA budget check throttles decode concurrency for hybrid models

pengchengneo:
## Summary

Fixes #1227. Hybrid SWA models (MiMo-V2-Flash) with `disable_radix_cache + chunked_prefill` had decode concurrency throttled from **64 → 16** and server hang (watchdog timeout) due to SWA pool exhaustion.

- Port `SWAChunkCache` from upstream for `disable_radix_cache + hybrid` (replaces broken `SWARadixCache(disable=True)`)
- Fix extend-time SWA eviction: `extend_batch_idx < 2` guard + batch-level `prefix_lens`
- Fix decode-time SWA eviction: evict at `decode_batch_idx == 0` for ChunkCache (safe, no tree-node overlap)
- Decouple SWA budget from full-pool admission with independent tracking

## Reproduction

Server (16-chip v6e, 4 nodes):
```bash
python -m sgl_jax.launch_server \
  --model-path /models/MiMo-V2-Flash --trust-remote-code \
  --tp-size 16 --dp-size 4 --ep-size 16 --moe-backend epmoe \
  --nnodes 4 --node-rank $TPU_WORKER_ID \
  --dist-init-addr $HEAD_SVC:30000 \
  --page-size 256 --context-length 262144 --disable-radix-cache \
  --chunked-prefill-size 2048 --dtype bfloat16 --mem-fraction-static 0.95 \
  --swa-full-tokens-ratio 0.2 --max-running-requests 128 \
  --attention-backend fa --dp-schedule-policy round_robin
```

Benchmark:
```bash
python -m sgl_jax.bench_serving \
  --backend sgl-jax --host 127.0.0.1 --port 30271 \
  --dataset-name random --random-input-len 16384 --random-output-len 1024 \
  --random-range-ratio 1.0 --flush-cache --seed 12345 \
  --request-rate 100 --num-prompts 256 --max-concurrency 64 \
  --tokenizer XiaomiMiMo/MiMo-V2-Flash
```

#### Before (main 1527bf4b)

```
============ Serving Benchmark Result ============
Backend:                                 sgl-jax
Traffic request rate:                    100.0
Max request concurrency:                 64
Successful requests:                     256
Benchmark duration (s):                  495.69
Total input tokens:                      4194304
Total input text tokens:                 4194304
Total generated tokens:                  262144
Total generated tokens (retokenized):    259563
Request throughput (req/s):              0.52
Input token throughput (tok/s):          8461.58
[...]

PR #1327, opened 2026-06-08
[Spec_Decode] performance tuning with scheduler overlap compatiblity

pengchengneo:
image

## 4k input / 1k output benchserving logs and results

Raw logs/results bundle: https://gist.github.com/pengchengneo/328518a93ecd11366e5bea09f605d189

The gist contains:
- rank0 raw server logs for spec and no-spec
- benchserving `run.log` and `result.jsonl` for bsz 4/8/16/32/64
- parsed TTFT/ITL CSV
- parsed rank0 steady decode gen-throughput CSV

## Spec server launch command, rank0 shown

```bash
python3 -m sgl_jax.launch_server \
  --model-path /data/pc \
  --trust-remote-code \
  --tp-size 16 \
  --dp-size 4 \
  --ep-size 16 \
  --moe-backend epmoe \
  --nnodes 4 \
  --node-rank 0 \
  --dist-init-addr 10.16.2.6:30042 \
  --host 0.0.0.0 \
  --port 30271 \
  --page-size 256 \
  --context-length 262144 \
  --disable-radix-cache \
  --chunked-prefill-size 2048 \
  --dtype bfloat16 \
  --mem-fraction-static 0.84 \
  --swa-full-tokens-ratio 0.2 \
  --skip-server-warmup \
  --log-level info \
  --decode-log-interval 1 \
  --max-running-requests 128 \
  --dp-schedule-policy round_robin \
  --max-seq-len 4096 \
  --precompile-bs-paddings 1 4 8 16 32 64 128 \
  --precompile-token-paddings 4096 \
  --speculative-algorithm NEXTN \
  --speculative-num-steps 3 \
  --speculative-num-draft-tokens 4 \
  --speculative-eagle-topk 1 \
  --speculative-accept-threshold-single 1.0 \
  --speculative-accept-threshold-acc 1.0
```

## No-spec server launch command, rank0 shown

```bash
python3 -m sgl_jax.launch_server \
  --model-path /data/pc \
  --trust-remote-code \
  --tp-size 16 \
  --dp-size 4 \
  --ep-size 16 \
  --moe-backend epmoe \
  --nnodes 4 \
  --node-rank 0 \
  --dist-init-addr 10.16.2.6:30042 \
  --host 0.0.0.0 \
  --port 30271 \
  --page-size 256 \
  --context-length 262144 \
  --disable-radix-cache \
  --chunked-prefill-size 2048 \
  --dtype bfloat16 \
  --mem-fraction-static 0.84 \
[...]

PR #1355, opened 2026-06-12
fix(swa): restore chunked req fill_ids on hybrid no-budget early-return

### LINKED ISSUE #1266: [Bug] Hybrid-SWA models deadlock on KV-full retract (multi-host v6e) — introduced by #1234
## Summary

On multi-host TPU v6e (`4x4` = 16 chips, `tp16/dp4/ep16`), a **hybrid sliding-window-attention (SWA) MoE model** (MIMO-V2-Flash) **deterministically deadlocks** the server when a **KV-cache-full retract** fires during high-concurrency long-context decode (`--disable-radix-cache`, `bs128`, `8k` input). `forward_ct` stops advancing → 300 s watchdog → `SIGQUIT`. The TPU is left wedged (only a pod restart recovers it).

## Root cause — bisected to #1234

I bisected this to **#1234** (`fix(swa): SWA budget check throttles decode concurrency for hybrid models`, `48335c97`):

- **Pre-#1234** (`48335c97~1` = `cdb173f1`): the exact same config runs **cleanly** — `#running-req` caps ~24, **0 retracts, no wedge**.
- **With #1234**: hybrid-SWA decode concurrency is unthrottled (~16 → ~70+) → the **SWA KV sub-pool saturates** → triggers a **kv-full retract** → the next forward step deadlocks.

The deadlock itself is **SWA-specific**: I ran a non-SWA model (`Qwen3-30B-A3B`) on the *identical* multi-host config and **force-fired retract hundreds of times** (`SGLANG_TEST_RETRACT=1`), including low-concurrency batches with empty DP groups — it **never wedges**. So the generic retract / overlap / DP-attention / MoE-collective paths are fine; only the **hybrid-SWA retract path** deadlocks.

## Reproduction (deterministic)

Launch MIMO-V2-Flash multi-host (`--tp-size 16 --dp-size 4 --ep-size 16 --moe-backend fused --attention-backend fa --disable-radix-cache`), then from rank-0:
```bash
python -m sgl_jax.bench_serving --backend sgl-jax --host 127.0.0.1 --port 30000 \
  --tokenizer /path/to/MiMo-V2-Flash --dataset-name random \
  --random-input-len 8192 --random-range-ratio 1.0 --flush-cache --seed 12345 \
  --num-prompts 256 --max-concurrency 128 --random-output-len 1024
```
The SWA sub-pool saturates during the prefill ramp → `KV cache pool is full. Retract requests` → a few steps later the server wedges. Fine at `bs64`, and fine for radix-cache-ON / low-concurrency workloads (no retract → no wedge).

## Suggested fix

1. **Mitigation**: cap the hybrid-SWA decode concurrency that #1234 unlocked, so the SWA sub-pool doesn't saturate (no retract → no wedge).
2. **Real fix**: ensure the SWA sub-pool's post-retract allocation state stays cross-host consistent so the next forward step is identical on all hosts. (Suspected divergence: the full pool capacity is `pmin`-synced across hosts, but the SWA sub-pool's per-step alloc/free state is updated locally — needs maintainer confirmation, e.g. via `--disable-hybrid-swa-memory`.)

## Impact

On multi-host v6e, `bs128` long-context decode on hybrid-SWA models is unusable — a single retract deterministically wedges the server and bricks the TPU until the pod is restarted.

zorrofox:
Fixes #1266 — hybrid-SWA models deadlock the TPU on multi-host after a KV-full retract.

cc @pengchengneo (this is a 1-line fix to the `add_chunked_req` early-return path you added in #1234)

## Root cause

#1234 added an early-return in `add_chunked_req` when the hybrid SWA budget is exhausted (`_rem_tokens <= 0 → return req`), but the caller (`scheduler.py:1619`) has already run `init_next_round_input()` which resets `req.fill_ids` to the full input length. The early-return leaves `fill_ids` at full length, so:

| round | what happens | chunked req state |
|---|---|---|
| N | SWA full → `add_chunked_req` early-returns; **`fill_ids` left at 8192** (full input) | prefix=4352, fill_ids=8192 |
| N+1 | `cache_unfinished_req()` sets `prefix = req_to_token[:8192]` (includes 4352–8192 **stale** slots). `init_next_round_input()` → **`extend = 8192 − 8192 = 0`**. SWA now freed (post-retract) → `add_chunked_req` appends with `extend_input_len=0` | seq=8192, **extend=0** in EXTEND batch |

The resulting batch has per-DP `seq_lens=[8192, 4096]`, `extend_seq_lens=[0, 4096]`, `cu_q_lens=[0, 0, 4096, …]`, `distribution=[0, 2, 2]`.

A `q_len==0` seq inside the RPA kernel's `[start_seq_idx, end_seq_idx)` range deadlocks the kernel: the prologue prefetches bq/bkv for seq 0, but `process()` runs **zero** `bq` iterations (so it never advances the double-buffer `sem_ids` or prefetches seq 1's first block), leaving seq 1's `wait_fetch_*` waiting on a mismatched semaphore — all 16 chips wedge in `Execute` ("Slow PjRt TPU operation").

## Fix

Restore `fill_ids` to the actually-committed prefix length before bailing on `_rem_tokens <= 0`.

## Verification (v6e-16, dp=4/tp=16/ep=16, #1266 repro config)

`--disable-radix-cache --swa-full-tokens-ratio 0.15 --max-running-requests 128`, overlap, 256 × (8192 in / 1024 out), conc=128:

| | retract count | result |
|---|---|---|
| main@a98e2931 | 1 | **deadlock** at first post-retract EXTEND (`cu_q=[0,0,4096]`) |
| + this fix | 19 | **PASS** 256/256, 313 tok/s, median ITL 23 ms |

After the fix, the post-retract EXTEND has `cu_q=[0,4096,8192]` (chunked req correctly extends 4096 instead of 0).

## Out of scope (follow-ups)

- RPA kernel robustness to `q_len==0` seqs in the active range — worth hardening so future scheduler bugs degrade to wrong output instead of a TPU wedge.
- `swa-full-tokens-ratio=0.15` + `bs=128` retracts ~19× per 256 reqs (the #1234 throttle doesn't fully cap admission under overlap) — separate tuning issue, not a deadlock.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_policy.py
+++ b/python/sgl_jax/srt/managers/schedule_policy.py
@@ -429,6 +429,15 @@ class PrefillAdder:
             )
         if _rem_tokens <= 0:
             if self.is_hybrid:
+                # Caller already ran init_next_round_input(), which reset
+                # fill_ids to the full input. If we return here without
+                # scheduling, the next round's cache_unfinished_req() will
+                # treat that full length as committed prefix, then compute
+                # extend_input_len==0 and schedule a zero-extend chunked req.
+                # A q_len==0 seq inside the RPA kernel's [start_seq, end_seq)
+                # range deadlocks the DMA pipeline (#1266). Restore fill_ids
+                # to the actually-committed prefix before bailing.
+                req.fill_ids = req.fill_ids[: len(req.prefix_indices)]
                 return req
             _rem_tokens = self.rem_chunk_tokens_list[dp_rank]
         truncated = req.extend_input_len > _rem_tokens


pengchengneo:
well done,  please add accuracy test @zorrofox

pengchengneo:
image
it is ok with mtp test with https://github.com/sgl-project/sglang-jax/pull/1327

pengchengneo (approved):

zorrofox:
GSM8K accuracy (v6e-16 dp=4, nospec, overlap, 200q 5-shot temp=0 max_new_tokens=512):

| config | retract count | accuracy |
|---|---|---|
| swa=0.2 mr=64 (normal, no retract) | 0 | 0.795 |
| swa=0.15 mr=128 (#1266 repro config) | 4 | 0.800 |

Both within historical baseline (0.785–0.875 at this max_new_tokens budget). The second run exercises the fix path (chunked req hits the SWA-full early-return, then retract frees the pool, then the chunked req continues prefill with the restored fill_ids); accuracy is unchanged vs the no-retract run.

Also verified on #1327 + this fix: v6e-16 spec bs=128 swa=0.15 now survives 56× retract (was: deadlock on first retract).

PR #1357, opened 2026-06-12
Add best-fit DP scheduling with stranding-based scoring

Yajingleo:
## Summary
- Add stranding-based best-fit scheduling for DP rank selection using FLOPs (input tokens) and HBM (output tokens) as two dimensions
- Implement configurable weights via `--dp-best-fit-flops-weight` and `--dp-best-fit-hbm-weight` CLI args
- Add comprehensive unit tests for the stranding calculation and DP rank selection logic

## Test plan
- [x] Run unit tests: `python -m pytest python/sgl_jax/test/test_dp_best_fit_scheduling.py`
- [ ] Verify CLI args are properly parsed with `--dp-schedule-policy best_fit`
- [ ] Test with actual DP workloads to validate scheduling behavior

### DIFF
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -834,8 +834,8 @@ class Scheduler(
             flops_util = after_input / flops_cap
             hbm_util = after_output / hbm_cap
 
-        2. Compute inflation factor (driven by bottleneck dimension):
-            inflation = max(flops_cap / after_input, hbm_cap / after_output)
+        2. Compute inflation factor (driven by bottleneck dimension - the one with higher utilization):
+            inflation = min(flops_cap / after_input, hbm_cap / after_output)
 
         3. Compute stranding (wasted capacity) per dimension:
             flops_stranding = 1 - inflation * flops_util
@@ -893,8 +893,8 @@ class Scheduler(
             flops_util = after_input / flops_cap
             hbm_util = after_output / hbm_cap
 
-            # Inflation factor (driven by bottleneck dimension)
-            inflation = max(
+            # Inflation factor (driven by bottleneck dimension - the one with higher utilization)
+            inflation = min(
                 flops_cap / max(after_input, 1),
                 hbm_cap / max(after_output, 1),
             )
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -63,6 +63,7 @@ from sgl_jax.srt.managers.scheduler_output_processor_mixin import (
     SchedulerOutputProcessorMixin,
 )
 from sgl_jax.srt.managers.scheduler_profiler_mixing import SchedulerProfilerMixin
+from sgl_jax.srt.managers.resource_estimator import ResourceEstimator
 from sgl_jax.srt.managers.tp_worker import ModelWorker
 from sgl_jax.srt.managers.tp_worker_overlap_thread import ModelWorkerClient
 from sgl_jax.srt.managers.utils import validate_input_length
@@ -440,6 +441,13 @@ class Scheduler(
         self.chunked_reqs = [None] * self.dp_size  # Per-DP chunked requests
         self.is_mixed_chunk = (
             self.chunked_prefill_size is not None and server_args.enable_mixed_chunk
+        )
+
+        # Init resource estimator for best-fit DP scheduling
+        self.resource_estimator = ResourceEstimator.from_server_args(
+            model_config=self.model_config,
+            server_args=server_args,
+            dp_size=self.dp_size,
         )
 
         # Init pause/continue state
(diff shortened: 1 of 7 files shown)


cjx0709:
@Yajingleo  Hello Yajing, thansks for your contribution. The direction makes sense to us: `best_fit` is trying to make DP rank selection more resource-aware by balancing prefill/FLOPs and decode/HBM pressure separately.

Could you provide more concrete evaluation results to show that this policy is effective in practice? 
The current unit tests cover the scoring logic, but they do not yet demonstrate that the new scheduling policy improves real serving behavior, maybe you can also add new kind of test case for the dp rank load balance.

Yajingleo:
> @Yajingleo Hello Yajing, thansks for your contribution. The direction makes sense to us: `best_fit` is trying to make DP rank selection more resource-aware by balancing prefill/FLOPs and decode/HBM pressure separately.
> 
> Could you provide more concrete evaluation results to show that this policy is effective in practice? The current unit tests cover the scoring logic, but they do not yet demonstrate that the new scheduling policy improves real serving behavior, maybe you can also add new kind of test case for the dp rank load balance.

Sure, let me add some unit tests. Ideally, if someone can help eval in production env, that would be great. The impact is throughput. The stranding on TPU compute and HBM are the bottleneck of utilizations and throughput. When we have diversified shaped requests, the distributed 2-dim bin-packing is significantly more complex than 1-dimensional bin-packing. This bin-packing algorithm can combine the HBM bound requests with TPU bound requests into one DP rank. Main eval metric is TPU duty cycle and HBM utilization. Provided sufficiently high volume of requests, jointly high avg TPU compute utilization + HBM utilization across the fleet or TPU pod (>= 2chips) indicate high throughput.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -774,6 +817,101 @@ def _select_min_running_dp(
...
+            hbm_util = after_output / hbm_cap
+
+            # Inflation factor (driven by bottleneck dimension - the one with higher utilization)
+            inflation = min(
+                flops_cap / max(after_input, 1),
+                hbm_cap / max(after_output, 1),
+            )
+
+            # Stranding per dimension
+            flops_stranding = 1 - inflation * flops_util
```
hlFu: Could you help me understand the formula of flops_stranding, hbm_stranding. If I understand it correctly, their results seem always 0?

Yajingleo: Another formula might be easier to understand. Inputs are `hbm_over_flops(batched requests)` and `hbm_over_flops(dp rank)`.  If `hbm_over_flops(batched requests) > hbm_over_flops(dp rank)`, we strand flops. Vice versa. Thus,
* hbm_stranding = max(0, 1 - hbm_over_flops(batched requests) / hbm_over_flops(dp rank) )
* flops_stranding = max(0, 1 - hbm_over_flops(dp rank) / hbm_over_flops(batched requests) )

hlFu:
Thanks for the PR @Yajingleo .

If I understand it correctly, this algorithm tries to dispatch requests to a DP rank that has balanced compute and HBM utilization. And this commit chooses the total input tokens of running requests to estimate FLOPs and total output tokens to estimate HBM. 

In that case, could you explain more about:
1. The reason behind choosing input token to estimate FLOPs? I think prefill is both memory and compute intensive because it also takes a lot of KV buffer. To get a more practical FLOP number I think we need to collect device signals?
2. Looks like when calculating the utilization ratio, the denominator of FLOP is max chunk size while the denominator of HBM is max token. So we are comparing per-request metrics with overall metrics?
3. For a concrete case like
```
Rank 0: input=900, output=900   (90% on both — nearly full, but perfectly balanced)
Rank 1: input=100, output=500   (10% FLOPs, 50% HBM — lighter but imbalanced)
```
I think there's some chance that the scheduler will dispatch the request to rank 0 using this algorithm? which might not be the situation we want?

Yajingleo:
> Thanks for the PR @Yajingleo .
> 
> If I understand it correctly, this algorithm tries to dispatch requests to a DP rank that has balanced compute and HBM utilization. And this commit chooses the total input tokens of running requests to estimate FLOPs and total output tokens to estimate HBM.
> 
> In that case, could you explain more about:
> 
> 1. The reason behind choosing input token to estimate FLOPs? I think prefill is both memory and compute intensive because it also takes a lot of KV buffer. To get a more practical FLOP number I think we need to collect device signals?

Exactly! Both prefill and decode phases consume both FLOPs and HBM. To be precise, we eventually need to profile real device signals across different execution phases to derive an accurate empirical formula based on input_length and output_length.

The current implementation uses input_token_capacity and output_token_capacity as a first-order approximation. However, since these two dimensions are non-independent and tightly coupled in practice, this approximation has its limitations. I plan to run empirical benchmarks to analyze these dynamics and fine-tune the scoring logic.

> 2. Looks like when calculating the utilization ratio, the denominator of FLOP is max chunk size while the denominator of HBM is max token. So we are comparing per-request metrics with overall metrics?
> 3. For a concrete case like
> 
> ```
> Rank 0: input=900, output=900   (90% on both — nearly full, but perfectly balanced)
> Rank 1: input=100, output=500   (10% FLOPs, 50% HBM — lighter but imbalanced)
> ```
> 
> I think there's some chance that the scheduler will dispatch the request to rank 0 using this algorithm? which might not be the situation we want?

Under this scoring logic, the algorithm selects Rank 0 as the best-fit because it exhibits zero stranding across both dimensions. Conversely, Rank 1 incurs input stranding ($> 0$). Since input capacity maps to our compute dimension, Rank 1 effectively strands FLOPs by leaving compute capacity underutilized relative to memory.

Yajingleo:
With experiments, I found a better solution in https://github.com/sgl-project/sglang-jax/pull/1435. It is also using the input and output separately to do shape-aware io-balance min-running-queue.

PR #1358, opened 2026-06-12
feat(fused-moe-v2): support per-expert SwiGLU activation clamp

Prayer3th:
## Motivation

Some maxtext-trained MoE models (e.g. Ling3-Flash) clamp the SwiGLU activation on
their late layers. The unfused `EPMoE` path already supports this, but the fused
`fused_ep_moe_v2` kernel folds `silu * up` inside its op, so the clamp could not be
applied — models needing it had to fall back to the slower unfused backend. This
adds clamp support to the fused kernel so those layers can stay on the fused path.

## Changes

- **`activation_fn`** gains an optional `swiglu_limit`: `clip(silu(gate), max=L) * clip(up, -L, L)`,
  matching the unfused formula exactly. `None` = disabled (default; existing behavior
  is bit-for-bit unchanged). `swigluoai` keeps its own clamp.
- **`fused_ep_moe_v2` / `_fused_ep_moe_kernel`** thread `swiglu_limit` (routed experts)
  and `shared_swiglu_limit` (shared expert) as static args, applied at the routed and
  shared activation sites respectively. `ref_moe` mirrors both for test parity.
- **`FusedEPMoEV2.__call__`** exposes both limits.
- **`bench_v2`**: `BENCH_SWIGLU_LIMIT` / `BENCH_SHARED_SWIGLU_LIMIT` env knobs.
- **New test** `fused_moe_v2_test.py` (17 cases) — the v2 kernel previously had no unit test.

## Validation (single-host v7x, kernel vs `ref_moe`)

| precision | routed | shared | clamp correct | no regression |
|---|---|---|---|---|
| bf16 | ✓ | ✓ | ✓ | ✓ |
| fp8 block-wise | ✓ | — | ✓ | ✓ |
| fp8 per-channel | ✓ | ✓ | ✓ | ✓ |

`Ran 17 tests ... OK` (bf16 / fp8 block-wise + per-channel / shared expert / grouped
top-k / silu·gelu·swigluoai / SwiGLU clamp routed+shared / block-config `effective_for`).
A small-limit sanity case confirms the clamp genuinely changes the output while still
matching the reference.


JamesBrianD (approved):
LGTM

PR #1363, opened 2026-06-13
test: add MMMU-Pro evaluation support

zhengkezhou1:
## Motivation

This PR adds MMMU-Pro evaluation support to the large accuracy test suite.

We are actively onboarding multimodal models, and MMMU-Pro is an important benchmark for validating vision-language capability. This should also be useful for future multimodal model integrations, so the evaluation path is added as reusable test coverage rather than a one-off script.

The implementation is adapted with reference to both the upstream SGLang evaluation flow and the official MMMU-Pro implementation, while fitting into the existing local `run_eval` / accuracy test structure.

## Checklist

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -162,10 +162,8 @@ jobs:
 
       - name: Setup Python environment
         uses: ./.github/actions/setup-venv
-        with:
-          install-target: python[all,multimodal]
       - name: Evaluate accuracy
-        timeout-minutes: 40
+        timeout-minutes: 20
         shell: bash
         env:
           SGLANG_JAX_IS_IN_CI: true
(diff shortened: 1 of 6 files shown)


JamesBrianD:
Thanks for working on the MMMU-Pro eval path.

After some internal discussion, we are leaning away from adding eval support to `sglang-jax` in this form. The direction we want to take is to keep model evaluation more backend-neutral and run it through [`sgl-project/sgl-eval`](https://github.com/sgl-project/sgl-eval) instead of growing repo-specific `run_eval` coverage here.

So I would not continue expanding this PR as an in-repo eval implementation. Future MMMU-Pro or multimodal eval support should preferably live in `sgl-eval`, with `sglang-jax` acting as one backend/server target rather than owning the benchmark logic directly.

Issue #1240, opened 2026-05-28
[RFC] PD stage 3: multi-host routing and single-entry proxy


cjx0709:
## Summary

Extend PD from a directly addressed `(P, D)` pair to a production-like single-entry deployment shape:

- per-host transfer server identity for multi-host layouts
- single-entry router/proxy for `/generate` and OpenAI `/v1/*`
- request fan-out to one `(P, D)` pair with shared PD request identity and bootstrap fields

This stage is where benchmark and eval drivers can stop talking to P/D separately and start using one entrypoint.

## Position in the roadmap

This issue defines the fourth increment in the PD roadmap:

1. Transfer foundation
2. Host pool + side channel
3. Scheduler + end-to-end
4. Multi-host + routing
5. Production hardening

It builds on the scheduler-integrated single-pair flow and changes the deployment shape from “address P and D directly” to “use a production-like entrypoint”.

## Proposed structure changes

Add or update the following modules:

- `python/sgl_jax/srt/disaggregation/router_args.py`
- `python/sgl_jax/srt/disaggregation/launch_router.py`
- `python/sgl_jax/srt/disaggregation/mini_lb.py`
- `python/sgl_jax/srt/disaggregation/mini_lb_helpers.py`
- tests for launcher args, helper behavior, backend fetch compatibility, and request fan-out

### 1. Per-host transfer identity

In multi-host deployments, each host participating in a prefill role must publish its own `(host_ip, transfer_port, side_channel_port)` through bootstrap so decode can pull from the correct peer.

### 2. Single-entry proxy

Introduce a router/proxy that can expose:

- `/generate`
- `/v1/completions`
- `/v1/chat/completions`
- info/model endpoints

The proxy should fan out one logical request to one selected `(P, D)` pair.

### 3. Shared request identity and bootstrap fields

The proxy should guarantee that the two fan-out legs share the same logical PD identity and routing fields, so P and D observe one coherent request lifecycle.

### 4. Upstream-shaped launcher/MiniLB structure

The launcher and MiniLB should be kept as close as practical to the upstream `sglang_router` shape, while allowing a small local patch layer for `sgl-jax`-specific compatibility.

### Unit tests (CPU)

- launcher / parser tests
- helper tests for bootstrap field and request identity injection
- backend fetch compatibility tests
- request fan-out contract tests for `/generate` and `/v1/*`

### End-to-end tests

- a single-entry proxy can serve native `/generate`
[...]

PR #1365, opened 2026-06-14
fix(disaggregation): support multi-host TP processes in PD KV transfer

### LINKED ISSUE #1364: [Bug] PD disaggregation: KV transfer fails / silently corrupts on multi-host TP processes
## Summary

After #1299, `--disaggregation-mode prefill/decode` works end-to-end on single-host processes (validated on 2×v6e-1 with Qwen3-8B), but **fails when either P or D is a multi-host process** (`jax.process_count() > 1`, required for any model that does not fit on one host's chips).

The decode side observes either a hang or `JaxRuntimeError: INTERNAL: Async transfer object was deleted before transfers completed`, with no indication that the root cause is multi-host sharding. In a same-process probe the transfer completes but with **silently misaligned data**.

cc @cjx0709

## Root cause

`prefill._extract_req_kv` returns a `jax.Array` sharded on `kv_pool.mesh` (the prefill process's full mesh). On a multi-host process this array is **not fully addressable** — each host's HBM holds only `1/process_count` of the shards.

`JaxTransferWrapper.register_pull` → `jax.experimental.transfer.TransferServer.await_pull` only registers **process-local** PJRT buffers. So each prefill host's transfer server exposes only its local slice of the KV.

`decode._build_kv_spec_for_req` builds the pull spec on the **decode** process's full mesh (also multi-host). Each decode host then pulls from a single prefill endpoint (chosen by `bootstrap_room % len(registered_prefills)`), so:

| P shard count registered (1 host) | D spec shard count (1 host) | result |
|---|---|---|
| `local_device_count` | `local_device_count` | shard-index misalignment → silent corruption |
| ≠ | ≠ | `Async transfer object was deleted` / hang |

There is no `process_count()` check anywhere in the PD path.

## Reproduce — same-process (shows the silent path)

On any multi-host slice (tested v6e-64 16-host and v7x 2x2x4 4-host), every host runs:

```python
import socket, jax, numpy as np
from jax.sharding import Mesh, NamedSharding, PartitionSpec as P
from jax.experimental.transfer import start_transfer_server

gdev, ldev = jax.devices(), jax.local_devices()
mesh = Mesh(np.array(gdev).reshape(-1), ("x",))
sh = NamedSharding(mesh, P("x"))
per = 1024 * 1024
arr = jax.make_array_from_callback(
    (len(gdev) * per,), sh,
    lambda s: np.arange(s[0].start, s[0].stop, dtype=np.int32),
)
arr.block_until_ready()

ip = socket.gethostbyname(socket.gethostname())
srv = start_transfer_server(ldev[0].client, f"{ip}:31001", [f"{ip}:0"])
srv.await_pull(100, arr)              # registers ONLY local shards, no error

if jax.process_index() == 0:
    peer_ip = ...                     # any other host in the same process
    spec = jax.ShapeDtypeStruct(arr.shape, arr.dtype, sharding=sh)
    out = srv.connect(f"{peer_ip}:31001").pull(100, [spec])[0]
    out.block_until_ready()
    print(np.asarray(out.addressable_data(0))[:4])
    # -> [8388608 8388609 8388610 8388611]   (peer host's shard data)
    # expected [0 1 2 3]                     (this host's shard)
```

## Reproduce — cross-process (real PD topology)

Two independent v7x slices (`2x2x4` 4-host as P, `2x2x2` 2-host as D). P registers a 32-device-sharded array on every host; D rank-0 connects to one P host and pulls with its own local-mesh spec:

```
[B0] pull(100 sharded(bug)) FAIL: JaxRuntimeError: INTERNAL: Async transfer object was deleted before transfers completed.
```

With P all-gathering to replicated and registering `arr_rep.addressable_data(0)` (single fully-addressable buffer), and D pulling with a single-device spec:

```
[B0] pull(200 allgather+single(fix)): 320MB dt=62.3ms (5139MB/s) [:4]=[0 1 2 3] ... full_ok=True
```

Full 2-slice GKE job spec and logs in the first comment below.

## Impact

| scenario | status |
|---|---|
| P and D each fit on one host (Qwen3-8B, DeepSeek-Distill-1.5B, etc.) | ✅ works (#1299 L4) |
| P or D requires multi-host TP (any model > ~100 GB on v6e, > ~380 GB on v7x) | ❌ fails / silent corruption |

This blocks PD disaggregation for the models that benefit most from it (large-TP MoE where prefill interference is worst).

## Proposed fix

Short-term: make multi-host P/D work via ICI all-gather to a fully-addressable buffer before `register_pull`, and reshard on the decode side after pull. Adds one ICI all-gather on P (sub-ms warm) and one D2H+H2D on D; measured 320 MB end-to-end at 62 ms on v7x. Also add a guard in `JaxTransferWrapper.register_pull` so a non-fully-addressable array raises immediately with an actionable message.

Long-term: #1240 per-host transfer identity + shard-aligned pull (avoids the all-gather, aggregates per-host NIC bandwidth).

I have a ~60-line patch for the short-term fix; will open a PR shortly.

zorrofox:
# fix(disaggregation): support multi-host TP processes in PD KV transfer

Closes #1364

## Summary

`jax.experimental.transfer.await_pull` only registers process-local PJRT buffers. When the prefill process spans multiple hosts (`jax.process_count() > 1`), `_extract_req_kv` currently returns an array sharded on the full mesh, so each host's transfer server exposes only `1/N` of the KV and the decode pull either fails (`Async transfer object was deleted before transfers completed`) or — when shard counts happen to line up — returns misaligned data with no error.

This PR makes the existing single-endpoint transfer path correct on multi-host processes by:

1. **`jax_transfer/wrapper.py`** — `register_pull` raises with an actionable message when given a non-fully-addressable array, instead of letting the transfer layer fail opaquely.
2. **`prefill.py`** — on multi-host, `_extract_req_kv` gathers with `out_sharding=PartitionSpec()` (replicated) and returns `addressable_data(0)`, i.e. one fully-addressable buffer per host. Single-host behaviour is unchanged.
3. **`decode.py`** — on multi-host, `_build_kv_spec_for_req` pulls onto a single local device (matching the 1-buffer registration), and `_write_kv_to_pool` lifts the per-host single-device result onto the pool mesh as replicated before the existing per-layer scatter. Single-host behaviour is unchanged.

This is the minimal-change unblock; #1240 (per-host shard-aligned pull) remains the bandwidth-optimal follow-up.

## Cost

| step | multi-host added cost | measured (v7x, 320 MB bf16, 4-host P → 2-host D) |
|---|---|---|
| P ICI all-gather to replicated | `O(KV_bytes)` over ICI | 121–408 ms cold (jit compile); warm ≪ 1 ms |
| P→D DCN (unchanged, 1 host pair) | — | 62.3 ms = 5.1 GB/s |
| D `np.asarray` + `make_array_from_process_local_data` | one D2H + H2D of `KV_bytes` | not isolated; bounded by PCIe (~tens of ms at this size) |

For a 16K-token request on a 78-layer MLA model (~312 MB KV) the added cost is ≪ prefill time. The D-side D2H can be removed in a follow-up by assembling the global replicated array from device buffers directly; kept simple here.

## Testing

- [x] `pytest python/sgl_jax/test/disaggregation/ -v` (CPU, single-process — guard path not triggered)
- [x] Cross-process probe on 2 independent v7x slices (4-host P, 2-host D): without patch `Async transfer object was deleted`; with all-gather path **byte-equal** at 8 MB int32 and 320 MB bf16
- [ ] Multi-host `--disaggregation-mode prefill` + `--disaggregation-mode decode` end-to-end on a TP>local_device_count model (pending reviewer access to 2× multi-host slices; probe yaml attached in issue)

## Notes

- `is_fully_addressable` is used as the guard rather than `process_count() > 1` in `wrapper.py` so that future callers passing an already-gathered array on a multi-host process are not blocked.
- The prefill/decode branches key on `process_count() > 1` because the all-gather is unnecessary overhead on single-host and the existing behaviour there is already validated by #1299's L4.

### DIFF
--- a/python/sgl_jax/srt/disaggregation/decode.py
+++ b/python/sgl_jax/srt/disaggregation/decode.py
@@ -311,9 +311,17 @@ class SchedulerDisaggregationDecodeMixin:
         padded_pages = _pad_to_page_bucket(num_pages)
         per_layer_tail = kv_pool.kv_buffer[0].shape[1:]
         shape = (kv_pool.layer_num, padded_pages) + per_layer_tail
-        base_spec = kv_pool.kv_sharding.spec
-        stacked_spec = PartitionSpec(None, *base_spec)
-        sharding = NamedSharding(kv_pool.kv_sharding.mesh, stacked_spec)
+        if jax.process_count() > 1:
+            # P registers a single fully-addressable buffer per host (see
+            # prefill._extract_req_kv); pull onto one local device and
+            # let _write_kv_to_pool reshard onto the pool mesh.
+            from jax.sharding import SingleDeviceSharding
+
+            sharding = SingleDeviceSharding(jax.local_devices()[0])
+        else:
+            base_spec = kv_pool.kv_sharding.spec
+            stacked_spec = PartitionSpec(None, *base_spec)
+            sharding = NamedSharding(kv_pool.kv_sharding.mesh, stacked_spec)
         return jax.ShapeDtypeStruct(shape, kv_pool.dtype, sharding=sharding)
 
     def _write_kv_to_pool(self: Scheduler, req: Req, kv_indices, kv: jax.Array) -> None:
@@ -328,6 +336,14 @@ class SchedulerDisaggregationDecodeMixin:
         from jax.sharding import NamedSharding, PartitionSpec
 
         kv_pool = self.token_to_kv_pool_allocator.get_kvcache()
+        if jax.process_count() > 1 and kv.is_fully_addressable:
+            # Each decode host pulled the same single-device buffer; lift
+            # it onto the pool mesh as replicated so the per-layer scatter
+            # below can reshard via out_sharding without cross-host gaps.
+            kv = jax.make_array_from_process_local_data(
+                NamedSharding(kv_pool.mesh, PartitionSpec()),
+                np.asarray(kv),
+            )
         page_size = kv_pool.page_size
         seqlen = len(req.origin_input_ids)
         num_pages = (seqlen + page_size - 1) // page_size
--- a/python/sgl_jax/srt/disaggregation/jax_transfer/wrapper.py
+++ b/python/sgl_jax/srt/disaggregation/jax_transfer/wrapper.py
@@ -154,6 +154,18 @@ class JaxTransferWrapper:
             raise RuntimeError(
                 "JaxTransferWrapper.start() must be called before " "register_pull()"
             )
+        sharding = getattr(data, "sharding", None)
+        if sharding is not None and not data.is_fully_addressable:
+            raise ValueError(
+                f"register_pull: array is sharded across "
+                f"{len(sharding.device_set)} devices but only "
+                f"{jax.local_device_count()} are local to this process. "
+                f"jax.experimental.transfer.await_pull only registers "
+                f"process-local shards, so a remote pull would receive "
+                f"partial/misaligned data. All-gather to a fully-"
+                f"addressable sharding first (see issue #1240 for the "
+                f"per-host coordination plan)."
+            )
         with self._pending_lock:
             if uuid in self._pending:
                 raise RuntimeError(
--- a/python/sgl_jax/srt/disaggregation/prefill.py
+++ b/python/sgl_jax/srt/disaggregation/prefill.py
@@ -252,10 +252,18 @@ class SchedulerDisaggregationPrefillMixin:
             _np.asarray(page_id_source) // page_size,
             idx_sharding,
         )
-        # out_sharding describes the gather output, not the pool.
+        # out_sharding describes the gather output, not the pool. On a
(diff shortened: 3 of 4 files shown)


zorrofox:
Closing this. The all-gather-to-replicated workaround used here for
multi-host KV extraction does not scale: the gather's peak HBM is
×tp_size per in-flight request, so concurrency tops out in the low
single digits and c≥32 OOMs regardless of mem_fraction. Most of the
diff (the per-tick process_allgather sync on both P and D) is patching
divergence that the replicated path itself introduces.

Multi-host PD needs the per-host shard-aligned transfer in #1240; there
is no useful intermediate.

A few independent main-branch bugs were found along the way (chunked
reqs being extracted before the final chunk, padding page-ids read from
uninitialised req_to_token, PD reqs leaking into running_batch). I'll
split those into a separate ~40-line PR since they also affect
single-host.

PR #1366, opened 2026-06-14
feat(disaggregation): per-host KV shard transfer for multi-host PD (#1240 part 1)

zorrofox:
Implements the per-host transfer identity piece of #1240. On a multi-host process (`jax.process_count() > 1`) each P host registers only its own 1/nproc TP-shard of the gathered KV; each D host pulls from the P host with the same `jax.process_index()` and reassembles the global pool sharding zero-copy.

This replaces the all-gather-to-replicated workaround attempted in #1365 (closed): per-request HBM is `×1` instead of `×tp_size`, and the P-side allocator stays SPMD-consistent without a per-tick TPU collective.

## Design

- **bootstrap** (`bootstrap.py`, `runtime.py`): every P host already registers; `PrefillInfo` now carries `jax_process_index` / `jax_process_count` so D can pair by jax rank (k8s pod index ≠ jax process index in general).
- **prefill** (`prefill.py`):
  - `_extract_req_kv` keeps the gather TP-sharded (no all-gather) and returns `_global_to_local_shard(stacked)`: a fully-addressable view of this host's `addressable_shards` on a 1-D local mesh (zero-copy).
  - Padding page-ids are zero-filled instead of read from the uninitialised tail of `req_to_token`, so `device_put`'s cross-process `assert_equal` sees identical indices.
  - Mid-chunk reqs are skipped — extract only on the final chunk.
  - On multi-host, pool pages are released immediately after `register_pull` (the gather output is a fresh buffer), at the same SPMD point on every NP. The per-host ack only triggers `sender.clear()`. This keeps allocator state identical across NPs without `process_allgather`.
- **decode** (`decode.py`):
  - `_pick_prefill_peer_for_this_host()` finds the P host whose `jax_process_index` matches and caches it.
  - `_build_kv_spec_for_req` uses `local_kv_spec_for_pool` on multi-host.
  - `_write_kv_to_pool` lifts the pulled local shard back to the global pool sharding via `make_array_from_single_device_arrays` (each NP contributes its own shards; zero-copy) before the existing scatter.
  - `_drain_transfer_queue_synced` allgathers `(bootstrap_room, state)` in a fixed `(256, 2)` int64 buffer so per-NP entry-count differences cannot desync the SPMD program counter; receiver `poll()` exceptions are caught and treated as FAILED.
- **wrapper** (`wrapper.py`): `register_pull` raises on `not is_fully_addressable` instead of silently registering partial shards.

Single-host behaviour: all multi-host paths are gated on `jax.process_count() > 1`. The chunked-skip and padding-zero changes are not gated and also improve single-host (see testing).

## Testing

- [x] `pytest test/srt/disaggregation/` — 204 passed (7 pre-existing py3.11 `class _Communicator[T]` env failures, unrelated)
- [x] **single-host PD** (1-host P TP=8 ↔ 1-host D TP=8, Qwen3-8B): e2e PASS; GSM8K 50q `acc=0.960` (main `d2a69bc`: `0.840`); bench c=8 32/32 `146.7 tok/s` (main: `91.8`)
- [x] **multi-host PD** (4-host P TP=32 ↔ 4-host D TP=32, two independent v7x slices, Qwen3-8B):
  - e2e cold P 11.06s D 36.20s / warm P 1.01s D 1.59s ×2, text correct
  - GSM8K 200q c=8 `acc=0.955`, invalid=0, errors=0
  - bench c=8 32/32 `164.6 tok/s`, c=32 64/64, **c=64 128/128 `253.7 tok/s` warm / `140-142 tok/s` cold ×2**, p99 32.3s warm
  - 8-host server logs: 0 × `failed to extract / reached failed / RESOURCE_EXHAUSTED / Traceback / Watchdog / launch id`
- [x] synthetic per-host shard probe: 8-device local register + pull byte-correct at 3.7-4.9 GB/s; `make_array_from_single_device_arrays` reassembly 0.4ms zero-copy

## Constraints / known limitations

- **P and D must have the same `jax.process_count()`** (same-TP, same-nnodes). Hetero-TP needs a head-reshard step.
- **Single P process group.** Multi-P routing is the rest of #1240 (router/proxy).
- **`input_len > chunked-prefill-size` is not supported** on either single- or multi-host PD: `init_next_round_input` resets `prefix_indices=[]` for PD reqs each round, so a req longer than one chunk never converges (P loops to KV-pool OOM). Out of scope here; tracked separately. The chunked-skip in this PR handles the multi-req-fills-a-chunk case (each req < chunk size).

### DIFF
--- a/python/sgl_jax/srt/disaggregation/decode.py
+++ b/python/sgl_jax/srt/disaggregation/decode.py
@@ -37,6 +37,9 @@ class DecodeBookkeeping:
     kv_indices: object | None = None
     # Whether the receiver has been initialized + poll started.
     started: bool = False
+    # Set by _drain_transfer_queue_synced on multi-host so downstream
+    # does not re-poll (a poll() that raised would re-raise and desync).
+    synced_state: KVPoll | None = None
 
 
 class DecodePreallocQueue:
@@ -243,7 +246,13 @@ class SchedulerDisaggregationDecodeMixin:
 
         for entry in self._drain_transfer_queue_synced():
             assert entry.receiver is not None
-            state = entry.receiver.poll()
+            state = entry.synced_state
+            if state is None:
+                try:
+                    state = entry.receiver.poll()
+                except Exception:
+                    logger.exception("receiver.poll() raised for req_id=%s", entry.req_id)
+                    state = KVPoll.FAILED
             if state == KVPoll.SUCCESS:
                 try:
                     kv_result = entry.receiver.result
@@ -323,12 +332,13 @@ class SchedulerDisaggregationDecodeMixin:
                 room = getattr(e.req, "bootstrap_room", None)
                 if room in failed:
                     self.disagg_transfer_queue._entries.pop(rid, None)
-                    if e.receiver.poll() != KVPoll.FAILED:
-                        with suppress(Exception):
-                            e.receiver.fail(reason="peer_np_failed")
+                    with suppress(Exception):
+                        e.receiver.fail(reason="peer_np_failed")
+                    e.synced_state = KVPoll.FAILED
                     out.append(e)
                 elif room in success:
                     self.disagg_transfer_queue._entries.pop(rid, None)
+                    e.synced_state = KVPoll.SUCCESS
                     out.append(e)
         return out
 
--- a/python/sgl_jax/srt/disaggregation/decode.py
+++ b/python/sgl_jax/srt/disaggregation/decode.py
@@ -170,7 +170,10 @@ class SchedulerDisaggregationDecodeMixin:
                 from sgl_jax.srt.disaggregation.common.metrics import time_phase
 
                 with time_phase("bootstrap", "decode"):
-                    p_info = self.disagg_bootstrap_client.get_prefill_info(req.bootstrap_room)
+                    if jax.process_count() > 1:
+                        p_info = self._pick_prefill_peer_for_this_host()
+                    else:
+                        p_info = self.disagg_bootstrap_client.get_prefill_info(req.bootstrap_room)
             except Exception:
                 logger.exception(
                     "bootstrap lookup failed for req_id=%s "
@@ -238,7 +241,7 @@ class SchedulerDisaggregationDecodeMixin:
         for entry in self.disagg_prealloc_queue.pop_all():
             self.disagg_transfer_queue.add(entry)
 
-        for entry in self.disagg_transfer_queue.drain_terminal():
+        for entry in self._drain_transfer_queue_synced():
             assert entry.receiver is not None
             state = entry.receiver.poll()
             if state == KVPoll.SUCCESS:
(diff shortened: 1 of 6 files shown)


zorrofox:
cc @cjx0709 — this is the per-host transfer identity piece of #1240; would appreciate a review.

cjx0709:
> cc @cjx0709 — this is the per-host transfer identity piece of #1240; would appreciate a review.

hello, thanks for your commit, now I'm also working for pd disaggreagation stage3 or stage4. Today, I will post a new pr for stage3 and stage4, thus we can test them together.

zorrofox:
@cjx0709 thanks — had a look at #1368. As far as I can tell the two PRs are complementary rather than overlapping in function:

| | #1368 | this PR |
|---|---|---|
| scope | router/mini_lb, path-A D2H staging, decode admission, metrics (stage 3-5) | multi-host path-B per-host shard transfer only |
| HBM strategy | D2H to host pool then free device KV | each host registers/pulls its own 1/nproc TP-shard (device-to-device) |
| tested topology | v6e-1 single-host TP=1 | 2× 4-host TP=32, c=64 |
| `jax.process_count() > 1` transfer path | not touched (single-endpoint pull) | this PR |

i.e. #1368's path-A still registers/pulls one full-KV endpoint, so on a multi-host P (`process_count() > 1`) it would hit the same `await_pull` only-registers-local-shards issue this PR addresses — the per-host pairing here should compose with #1368's D2H staging and router.

The two touch the same five files (`bootstrap.py` / `runtime.py` / `prefill.py` / `decode.py` / `wrapper.py`) so there will be merge conflicts either way. Happy to rebase this on top of #1368 once that lands (probably the cleaner order given the size difference), or the other way round if you prefer — let me know which you'd like.

Separately, neither PR fixes the `IL > chunked-prefill-size` infinite-loop in `init_next_round_input` (the PD `prefix_indices=[]` reset); I have a +13/-4 fix for that verified on single-host and can send it after the merge order is settled.

cjx0709:
> @cjx0709 thanks — had a look at #1368. As far as I can tell the two PRs are complementary rather than overlapping in function:
> 
> #1368	this PR
> scope	router/mini_lb, path-A D2H staging, decode admission, metrics (stage 3-5)	multi-host path-B per-host shard transfer only
> HBM strategy	D2H to host pool then free device KV	each host registers/pulls its own 1/nproc TP-shard (device-to-device)
> tested topology	v6e-1 single-host TP=1	2× 4-host TP=32, c=64
> `jax.process_count() > 1` transfer path	not touched (single-endpoint pull)	this PR
> i.e. #1368's path-A still registers/pulls one full-KV endpoint, so on a multi-host P (`process_count() > 1`) it would hit the same `await_pull` only-registers-local-shards issue this PR addresses — the per-host pairing here should compose with #1368's D2H staging and router.
> 
> The two touch the same five files (`bootstrap.py` / `runtime.py` / `prefill.py` / `decode.py` / `wrapper.py`) so there will be merge conflicts either way. Happy to rebase this on top of #1368 once that lands (probably the cleaner order given the size difference), or the other way round if you prefer — let me know which you'd like.
> 
> Separately, neither PR fixes the `IL > chunked-prefill-size` infinite-loop in `init_next_round_input` (the PD `prefix_indices=[]` reset); I have a +13/-4 fix for that verified on single-host and can send it after the merge order is settled.

that's right, I will review your pr today, it should be merged before #1240. And later when #1240 is ready, can you help  take a look for my one.

python/sgl_jax/srt/disaggregation/decode.py
```diff
@@ -270,6 +273,65 @@ def process_decode_queue(self: Scheduler) -> None:
...
+        )
+        if not success and not failed:
+            return []
+        out: list[DecodeBookkeeping] = []
+        with self.disagg_transfer_queue._lock:
+            for rid, e in list(self.disagg_transfer_queue._entries.items()):
+                room = getattr(e.req, "bootstrap_room", None)
+                if room in failed:
+                    self.disagg_transfer_queue._entries.pop(rid, None)
+                    if e.receiver.poll() != KVPoll.FAILED:
```
cjx0709: `synced_terminal_rooms()` already catches `receiver.poll()` exceptions and treats them as FAILED so all jax processes stay in sync. But this path re-calls `e.receiver.poll()` without a try/except, and `process_decode_queue()` re-polls again after returning the entry. If the original FAILED came from a poll exception, may throw again before cleanup/abort runs.

zorrofox: Good catch — fixed in the latest commit. `_drain_transfer_queue_synced` now records the agreed state on `DecodeBookkeeping.synced_state` and `process_decode_queue` consumes that instead of re-polling, so a `poll()` that raised on one NP can't re-raise after the allgather and diverge the drain set. The redundant pre-`fail()` poll check is gone too (`fail()` is already inside `suppress`). Single-host falls back to a guarded `poll()` as before.

zorrofox:
> And later when #1240 is ready, can you help take a look for my one.

Sure — happy to review #1368 once it's ready, just ping me.

cjx0709 (approved):
LGTM

PR #1368, opened 2026-06-15
[PD disaggregation] mini_lb router, path-A D2H staging, decode admission, PD metrics, and CI tests

### LINKED ISSUE #1240: [RFC] PD stage 3: multi-host routing and single-entry proxy

cjx0709:
Closes #1240

## Motivation

Prefill and decode have opposite resource profiles, so co-locating them forces one scheduling policy onto two workloads and lets a long prefill stall every other request's decode. This PR adds PD (prefill/decode) disaggregation to sglang-jax: a prefill engine produces the KV cache, transfers it to a decode engine, and decode generates the output tokens.

See `docs/design/pd_disaggregation.md` for the design, transfer paths, single/multi-host model, and the full feature support matrix.

## Modifications

**Router**
- `mini_lb` single-entry proxy + launcher (`mini_lb.py`, `mini_lb_helpers.py`, `router.py`, `router_args.py`, `launch_router.py`) that fans each request to a prefill + decode pair and assigns the shared transfer id.

**Prefill KV transfer**
- Path A (`--disaggregation-enable-d2h`): D2H staging into a bounded pinned-host KV pool, then cross-pod pull. Device KV is freed right after staging to bound prefill HBM.
- Path B (default): register the gathered device KV arrays directly for pull (device → device).
- Per-layer KV gather to cap XLA compile-time HBM; host pool buffer sized from `--disaggregation-d2h-max-tokens` (falls back to `max_total_num_tokens`).

**Decode**
- Capacity-gated admission (KV alloc/receiver deferred to an admission step; pending bootstrap reqs hold no slots).
- KV pull moved onto a persistent worker thread (the native TPU pull blocks the event loop).
- `prefill_info` cached and resolved per-room locally (no per-request synchronous GET on the event loop), refreshed on a bounded interval so dead prefills are evicted.

**Observability & hardening**
- PD Prometheus metrics on `/metrics`; `--enable-metrics` wired into `ServerArgs`.
- Event-loop watchdog, per-request time stats, shared-secret auth (bootstrap / side channel), prefill-info validation.

**Tests, scripts & docs**
- Consolidated PD unit tests under `test/srt/disaggregation/`, registered in `test/srt/run_suite.py` (`unit-test-cpu`).
- Single-host eval/benchmark GKE job (`scripts/disaggregation/gke/pd_singlehost_eval_job.yaml` + `deploy.sh`).
- Design doc `docs/design/pd_disaggregation.md`.

### CPU unit tests (mock, no TPU)

```bash
cd test/srt && python3 run_suite.py --suite unit-test-cpu
```

Result: `343 passed, 2 skipped`.

### GKE single-host e2e (TPU v6e-1, `pd-v6e-1`)

DeepSeek-R1-Distill-Qwen-1.5B, TP=1, single chip, through the `mini_lb` router. Reproducible via `scripts/disaggregation/gke/deploy.sh`.

| Check | path A (D2H staging) | path B (direct HBM) |
|---|---|---|
| GSM8K (200q, 5-shot, par 16) | 0.655 | 0.675 |
| Smoke (in 4096 / out 1024, conc 64) | 64/64, 0 OOM | 64/64, 0 OOM |

Baseline (cap 8, conc 48, in 4096 / out 32, 192 prompts): 192/192, 0 OOM. Multi-host + path B validated separately in #1366.

## Checklist
- [x] Format your code according to the Code Formatting with Pre-Commit.
- [x] Add unit tests as outlined in the Running Unit Tests / Adding or Updating Tests.
- [x] Provide the motivation and modifications above.


cjx0709:
@zorrofox hello, I think this pr is ready for review, please help to take a look. Btw, you can click the https://github.com/sgl-project/sglang-jax/invitations to accept for the access.

python/sgl_jax/srt/managers/utils.py
```diff
@@ -43,6 +43,32 @@ def validate_input_length(
...
+    advances past its first chunk and leaks KV until OOM. Guard against it
+    until chunk-prefill-transfer lands.
+
+    Returns an error message if the request must be rejected, else None.
+    """
+    if disaggregation_mode == "null":
+        return None
+    if not chunked_prefill_size or chunked_prefill_size <= 0:
+        return None
+    if len(req.origin_input_ids) > chunked_prefill_size:
```
zorrofox: This only guards `len(origin_input_ids) > chunked_prefill_size`. The other way a PD req gets chunked is when several short reqs share one prefill batch and the adder's remaining budget runs out mid-batch — the last req is truncated regardless of its own length (`schedule_policy.add_one_req`).

That case won't infinite-loop (round 2 sees `extend_input_len = full_IL ≤ chunk_size` and finishes), but because `init_next_round_input` resets `prefix_indices=[]` for PD reqs, round 2 re-allocs a fresh slot+pages and round 1's allocation is never released — so each chunked req leaks one round of pages. Reproduced on this branch: after one c=64 IL=1024 n=128 run on 4-host TP=16, P `token usage` stays at 0.08 (~870 pages) instead of returning to 0; ~12 such runs would fill the pool.

Two options that would close it:
- the `schedule_batch.py:427` change (don't reset `prefix_indices` when `is_chunked > 0`) — that also makes the long-prompt case work instead of being rejected, so this guard could become a soft warning. I have a +13/-4 patch verified on single-host (2167 rounds → 4, GSM8K 0.840 → 0.960) and can send it separately or you can fold it in here.
- or have the prefill adder never truncate a PD req (treat `bootstrap_room is not None` like `add_chunked_req` with no budget cap), so the guard above is actually sufficient.

python/sgl_jax/srt/disaggregation/bootstrap.py
```diff
@@ -431,15 +527,93 @@ def get_prefill_info(self, bootstrap_room: int) -> dict[str, object]:
...
+        self._lock = threading.Lock()
+        self._by_key: dict[str, dict[str, object]] = {}
+        self._sorted_keys: list[str] = []
+        # -inf so the very first lookup always refreshes regardless of clock.
+        self._last_refresh: float = float("-inf")
+        # Transient-failure bookkeeping (throttled logging).
+        self._refresh_failures = 0
+        self._last_fail_log: float = float("-inf")
+
+    def _refresh_locked(self) -> None:
```
zorrofox, line 564: nit: the warm-tick path is zero-I/O as the comment says, but the cold/expired refresh still does a synchronous `httpx.get` on the decode event-loop thread. At `refresh_interval=1.0` that's a few ms of loop stall every second under steady state — probably fine for now, but might be worth moving the actual fetch onto the heartbeat daemon (or a one-shot timer) and having the loop only read the cached snapshot.

python/sgl_jax/srt/disaggregation/jax_transfer/conn.py
```diff
@@ -501,37 +559,18 @@ def poll(self) -> KVPoll:
...
-                        reason="pull_init",
-                    )
-                    with suppress(Exception):
-                        PD_TRANSFER_FAILURES_TOTAL.labels(reason="pull_init", role="decode").inc()
-                    self._mgr._prune_receiver(self._req_id)
-                    return self.state
+            # Hand the blocking pull to the background worker. ``poll()`` stays
+            # non-blocking; a later poll drives ``is_ready()`` -> ack -> SUCCESS
+            # once the worker has stored the results.
+            self._mgr.enqueue_pull(self)
```
zorrofox, line 565: nit: `_transition_to(TRANSFERRING)` happens under the lock but `enqueue_pull(self)` is after it. If the reaper thread fails this receiver in between, `_run_pull` still executes once and its result is discarded. Correctness is fine (the FAILED sticks), just a wasted pull. Not worth changing unless you've seen it in practice.

zorrofox (commented):
Went through the parts that touch / build on #1366 (prefill / decode / wrapper / conn / bootstrap / schedule_batch) plus the path-A staging and admission flow. The single-host path-A + router design looks coherent; left a couple of inline notes, the chunked-prefill guard is the only one I'd call substantive.

Re-ran the `process_count()>1` per-host-shard path from #1366 on this branch (2× 4-host v6e-16, P TP=16 ↔ D TP=16, Qwen3-8B, no D2H) and it still passes:
- e2e cold P 13.29s D 37.16s / warm D 1.19-1.21s, text correct
- bench c=64 IL=1024 n=128: 128/128, 196.4 tok/s, p99 68.5s
- 8-host server logs: 0 fail-lines

so the `_extract_req_kv` list-vs-stacked branching and the new `_jit_write_one_layer` path both work multi-host. 👍

One thing the run did surface (related to the chunked-prefill note below): after the c=64 run completes and all senders ack, P's `token usage` sits at 0.08 instead of returning to 0 — that's the per-chunked-req page leak from the `prefix_indices=[]` reset. Not a blocker at this concurrency, but it accumulates across runs.

python/sgl_jax/srt/managers/utils.py
cjx0709: @zorrofox  Thanks for your comment! I think we can follow the first option, you can just create a pr into #1368, then we merge them into main together.

zorrofox: Sent as #1386 (into this branch). Multi-host verified on 2× v6e-16: IL=5601 now converges in 3 chunks with correct D output, and c=64 P `token usage` returns to 0.00 (was 0.08).

PR #1371, opened 2026-06-17
fix(eplb): explicit out_sharding for topk_ids gather under explicit-axis Mesh

### LINKED ISSUE #1370: [Bug] EPLB --ep-dispatch-algorithm static|dynamic crashes during JIT trace under explicit-axis Mesh
### Describe the bug

When `--ep-dispatch-algorithm static` (or `dynamic`) and `--init-expert-location <map.json>` are set, the server loads weights successfully but crashes at the first precompile/forward with:

```
File ".../sgl_jax/srt/eplb/expert_location.py", line 362, in _topk_ids_logical_to_physical_static
    return info.logical_to_rank_dispatch_physical_map[layer_id, topk_ids]
  ...
  File ".../jax/_src/named_sharding.py", line 520, in _check_mesh_resource_axis
    raise ValueError(
ValueError: Resource axis: data of PartitionSpec('data', None) is not found in mesh: ().
```

This happens on every model that wires `forward_batch.expert_location_metadata` into `gate.py` (gemma4, grok, etc.), so EPLB (#818) is currently unreachable end-to-end since `mesh_utils.create_device_mesh` defaults to `AxisType.Explicit` (#349).

### Minimal repro (CPU, no TPU needed)

```python
import os
os.environ["JAX_PLATFORMS"] = "cpu"
os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=16"
import jax, numpy as np
from jax.sharding import PartitionSpec as P, NamedSharding

mesh = jax.sharding.Mesh(
    np.array(jax.devices()).reshape(1, 16),
    ("data", "tensor"),
    axis_types=(jax.sharding.AxisType.Explicit,) * 2,
)

# what ExpertLocationMetadata.__init__ produces:
# device_array(map, sharding=(None)) -> (None)==None -> device_put(device=None) -> SingleDeviceSharding
mapping = jax.device_put(np.zeros((60, 128), np.int32))
print(mapping.sharding)  # SingleDeviceSharding(...)

# topk_ids inside the jitted forward is sharded on the 'data' axis
topk_ids = jax.device_put(np.zeros((256, 4), np.int32), NamedSharding(mesh, P("data", None)))

@jax.jit
def f(m, idx):
    return m[3, idx]  # _topk_ids_logical_to_physical_static body

with jax.set_mesh(mesh):
    f(mapping, topk_ids)
# -> ValueError: Resource axis: data of PartitionSpec('data', None) is not found in mesh: ().
```

Replacing `mapping` with a fully-replicated `NamedSharding(mesh, P())` array still fails (different error):
```
ShardingTypeError: Use `.at[...].get(out_sharding=)` to provide output PartitionSpec for the gather indexing ...
```

### Root cause

1. `ExpertLocationMetadata.__init__` calls `device_array(x, sharding=(None))`. `(None)` is `None` (not a 1-tuple), so `device_array` falls into the `jax.device_put(x, device=None)` branch, producing a **SingleDeviceSharding** array (abstract mesh = `()`).
2. Inside the jitted forward, `topk_ids` carries `NamedSharding(mesh, P('data', None))`.
3. `arr[layer_id, topk_ids]` is an advanced gather. Under explicit-axis mesh JAX cannot infer the output sharding when the operand and index live on different (abstract) meshes, and raises before lowering.

`_topk_ids_logical_to_physical_dynamic` has the same indexing pattern and fails the same way.

### Environment
- jax 0.8.1 (sglang-jax pinned version)
- sglang-jax `main` (any commit after #818)
- mesh created via `sgl_jax.srt.utils.mesh_utils.create_device_mesh` (default `use_explicit_sharding=True`)

### Expected
Server reaches READY and `topk_ids` are remapped to physical expert ids per `init_expert_location`.

zorrofox:
## Summary

`--ep-dispatch-algorithm static|dynamic` crashes at first JIT trace because `_topk_ids_logical_to_physical_{static,dynamic}` does `replicated_map[layer_id, sharded_topk_ids]`, which JAX cannot resolve under an explicit-axis Mesh (the metadata arrays end up `SingleDeviceSharding` while `topk_ids` is `P('data', None)`). Fixes #1370.

## Changes

- `eplb/expert_location.py`: rewrite both gathers as `map[layer_id].at[topk_ids].get(out_sharding=jax.typeof(topk_ids).sharding.spec)` so the output inherits `topk_ids`' spec and no cross-mesh inference is needed. Works whether the map is `SingleDeviceSharding` or replicated `NamedSharding`.
- `model_executor/model_runner.py`: after `init_expert_location_metadata`, reshard the four metadata arrays to `NamedSharding(self.mesh, P())` so they are addressable on every host (multi-host safety; the maps are ≤30 KB each).

## Test plan

All on jax 0.8.1:
- [x] CPU 16-dev explicit-mesh repro (see linked issue) — crashes before, passes after, output matches numpy reference
- [x] `_topk_ids_logical_to_physical_static` numerical check vs `phy2log[L][topk_ids]` on a real 60×128 mapping
- [x] e2e on TPU v7x 2-host (`tp=16 ep=16 dp=1`, `--ep-dispatch-algorithm static --init-expert-location /tmp/phy2log.json`): server READY, needle@64K 5/5 (weight permute + dispatch remap consistent)
- [ ] CI

## Notes

- `device_array(x, sharding=(None))` in `ExpertLocationMetadata.__init__` is a latent typo (`(None) == None`); left as-is since the reshard in `model_runner.py` supersedes it and changing the call signature would need a `mesh` plumbed in. Happy to fold that in if preferred.
- No existing test covers `ep_dispatch_algorithm`; can add a CPU unit test mirroring the issue repro if wanted.

### DIFF
--- a/python/sgl_jax/srt/eplb/expert_location.py
+++ b/python/sgl_jax/srt/eplb/expert_location.py
@@ -359,7 +359,12 @@ def _topk_ids_logical_to_physical_static(
     info: ExpertLocationMetadata,
     layer_id: int = 0,
 ) -> jax.Array:
-    return info.logical_to_rank_dispatch_physical_map[layer_id, topk_ids]
+    out_spec = jax.typeof(topk_ids).sharding.spec
+    return (
+        info.logical_to_rank_dispatch_physical_map[layer_id]
+        .at[topk_ids]
+        .get(out_sharding=out_spec)
+    )
 
 
 def _topk_ids_logical_to_physical_dynamic(
@@ -367,9 +372,16 @@ def _topk_ids_logical_to_physical_dynamic(
     info: ExpertLocationMetadata,
     layer_id: int = 0,
 ) -> jax.Array:
-    num_valid, selected_expert_replicas = (
-        info.logical_to_all_physical_map_num_valid[layer_id, topk_ids],
-        info.logical_to_all_physical_map[layer_id, topk_ids],
+    out_spec = jax.typeof(topk_ids).sharding.spec
+    num_valid = (
+        info.logical_to_all_physical_map_num_valid[layer_id]
+        .at[topk_ids]
+        .get(out_sharding=out_spec)
+    )
+    selected_expert_replicas = (
+        info.logical_to_all_physical_map[layer_id]
+        .at[topk_ids]
+        .get(out_sharding=out_spec + (None,))
     )
 
     rng_key = jax.random.key(0)
--- a/python/sgl_jax/srt/eplb/expert_location.py
+++ b/python/sgl_jax/srt/eplb/expert_location.py
@@ -361,9 +361,7 @@ def _topk_ids_logical_to_physical_static(
 ) -> jax.Array:
     out_spec = jax.typeof(topk_ids).sharding.spec
     return (
-        info.logical_to_rank_dispatch_physical_map[layer_id]
-        .at[topk_ids]
-        .get(out_sharding=out_spec)
+        info.logical_to_rank_dispatch_physical_map[layer_id].at[topk_ids].get(out_sharding=out_spec)
     )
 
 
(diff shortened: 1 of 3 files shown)


zorrofox:
performance-test-1-tpu flake: `test_input_throughput_default` (Qwen3-8B dense) 28291.4 vs threshold 28299 (-0.027%). This PR only touches `eplb/expert_location.py` + metadata reshard in `model_runner.py`, neither on the dense forward path. Could a maintainer re-run? Happy to dig in if it reproduces.

PR #1372, opened 2026-06-17
feat(qwen3.5): add dense Qwen3.5 (0.8B-27B), validated on 27B

MokusMokun:
## Motivation

Generalize the merged Qwen3.5-35B-A3B MoE path (#1267) to also serve the **dense** `Qwen3_5ForConditionalGeneration` family (0.8B / 2B / 4B / 9B / 27B). Dense and MoE variants share the hybrid backbone (full-attention + Gated DeltaNet, `full_attention_interval=4`), partial M-RoPE, and config layout — only the FFN differs (plain SwiGLU vs 256-expert MoE). Validated end-to-end on 27B.

Follows RFC #1251 / PR #1267.

## Modifications

- **Config** (`srt/configs/qwen3_5.py`): add `intermediate_size`; default all MoE-only fields (`num_experts` / `num_experts_per_tok` / `moe_intermediate_size` / `shared_expert_intermediate_size`) to `None`; add an `is_moe` discriminator (`num_experts is not None`); add `Qwen3_5DenseConfig` (`model_type="qwen3_5"`); widen `get_qwen3_5_hybrid_config` to both root types.
- **Model** (`srt/models/qwen3_5.py`): `Qwen3_5DecoderLayer` builds a plain SwiGLU MLP (`Qwen2MoeMLP`) when `not is_moe` and branches the call site (the two MLPs have different signatures; dense yields `topk_ids=None`, kept out of the EPLB / return-routed-experts list). Add the `Qwen3_5ForConditionalGeneration` entry class. GDN + full-attention paths unchanged.
- **Weight loader** (same file): dense `mlp.{gate,up,down}_proj` mappings; `lm_head.weight` mapping conditioned on `tie_word_embeddings` (tied 0.8B/2B/4B reuse the embedding); skip the pre-fused expert path when dense. 27B bring-up: consumed=851 / skipped=348 (vision+MTP) / missing=0 / unexpected=0.
- **Registration** (`srt/hf_transformers_utils.py`): register `Qwen3_5DenseConfig` for `qwen3_5`.
- **Routed-experts infra** (`srt/layers/routed_experts_capturer.py`, `srt/model_executor/model_runner.py`): alias-aware `get_routed_expert_count` / `get_routed_experts_per_token` helpers resolve the expert count (`num_experts` / `n_routed_experts` / `num_local_experts`) and routing top-k across name aliases and root-vs-text config. `create()` raises a clear error for router-less (dense) models when a capture flag is set; physical-expert count and buffer sizing go through the helpers so non-Qwen MoE configs aren't broken by the `hf_text_config` lookup.
- **Tests**: rework the CPU component tests to be Hub-free (synthetic configs + mapping coverage from per-layer templates snapshotted from the real index), covering MoE + dense (config / MLP dispatch / weight-mapping / tied-variant / capturer guard + build); add a dense class to the TPU smoke.

## Tests

- **Unit (CPU)**: `test_qwen3_5.py` 18/18 — Hub-free, runs in the `unit-test-cpu` CI suite.
- **Smoke (TPU, single-host v6e-4, TP=4)**: `test_qwen3_5_models.py`, MMLU 100-Q thinking mode (asserts > 0.70) — MoE 35B-A3B = **0.920**, dense 27B = **0.910**.
- **Full e2e** (v6e 4×4 16-chip, TP=16/DP=4, thinking: temp=0.6/top-p=0.95/top-k=20/min-p=0/presence=1.5/max-tokens=32768):

  | Benchmark | n | dense 27B | published | delta |
  |---|---|---:|---:|---:|
  | MMLU-Pro | 12,032 | **0.8671** / parsed 0.999 | 0.861 | +0.0061 |
  | GPQA-Diamond | 198 ×5 (median) | **0.8636** (0.848–0.889) | 0.855 | +0.0086 |

  GPQA-Diamond = median of 5 seeds (high-variance in thinking mode + non-deterministic TPU sampling); MMLU-Pro is a single run (large enough to be stable). Dense at TP=16/DP=4 needs `--mem-fraction-static 0.90` (weights replicate across the DP groups) and no `--moe-backend` (no experts).

## Benchmarking and Profiling

Out of scope for this PR (correctness-first); perf follow-up.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/layers/routed_experts_capturer.py
+++ b/python/sgl_jax/srt/layers/routed_experts_capturer.py
@@ -31,6 +31,44 @@ def get_array_size_bytes(t: np.ndarray):
     return np.prod(t.shape) * t.dtype.itemsize
 
 
+# Routed-expert metadata field names vary across HF model families, and may live
+# on the root config or a nested text config. Total experts: ``num_experts``
+# (Qwen), ``n_routed_experts`` (GLM/DeepSeek/MiMo), ``num_local_experts``
+# (MiniMax/Grok). Routing top-k: ``num_experts_per_tok``, ``num_experts_per_token``
+# (Kimi), or ``top_k_experts`` (Gemma4). Plain ``top_k`` is sampling top-k, not
+# routing — never use it here.
+_TOTAL_EXPERT_FIELDS = ("num_experts", "n_routed_experts", "num_local_experts")
+_TOPK_EXPERT_FIELDS = ("num_experts_per_tok", "num_experts_per_token", "top_k_experts")
+
+
+def _first_present_attr(configs, names):
+    for cfg in configs:
+        if cfg is None:
+            continue
+        for name in names:
+            value = getattr(cfg, name, None)
+            if value is not None:
+                return value
+    return None
+
+
+def get_routed_expert_count(model_config):
+    """Total routed experts across name aliases and root/text config; ``None`` if
+    the model is router-less (dense)."""
+    return _first_present_attr(
+        (getattr(model_config, "hf_text_config", None), getattr(model_config, "hf_config", None)),
+        _TOTAL_EXPERT_FIELDS,
+    )
+
+
+def get_routed_experts_per_token(model_config):
+    """Routing top-k across name aliases and root/text config; ``None`` if router-less."""
+    return _first_present_attr(
+        (getattr(model_config, "hf_text_config", None), getattr(model_config, "hf_config", None)),
+        _TOPK_EXPERT_FIELDS,
+    )
+
+
 class RoutedExpertsCapturer(ABC):
     @staticmethod
     def create(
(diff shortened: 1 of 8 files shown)


python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -179,7 +179,7 @@ def init_routed_experts_capturer(self):
                 dist_recorder_buffer_size=self.server_args.expert_distribution_recorder_buffer_size,
                 dist_recorder_output_file=self.server_args.expert_distribution_recorder_output_file,
                 physical_expert_counts=self.server_args.ep_num_redundant_experts
-                + getattr(self.model_config.hf_config, "num_experts", 0),
+                + (getattr(self.model_config.hf_text_config, "num_experts", None) or 0),
```
cjx0709: will this change cause error if the model with `num_experts` in root hf_config but without in hf_text_config?

python/sgl_jax/srt/layers/routed_experts_capturer.py
```diff
@@ -50,6 +50,14 @@ def create(
         physical_expert_counts: int = 256,
     ):
         if enable or enable_balance_debug or enable_dist_recorder:
+            # Dense (router-less) models have num_experts=None; the real capturer
+            # would build buffers from a None top-k and crash. Fail clearly.
+            if getattr(model_config.hf_text_config, "num_experts", None) is None:
```
cjx0709: dose all the models in hf follows this check? Like the comment below.

python/sgl_jax/srt/model_executor/model_runner.py
MokusMokun: covered corner cases with checks of `EXPERT_FIELDS` in `routed_experts_capturer.py`
also added `test_routed_experts_capturer_builds_for_moe` in `test_qwen3_5.py` CI test

python/sgl_jax/srt/layers/routed_experts_capturer.py
MokusMokun: see above https://github.com/sgl-project/sglang-jax/pull/1372#discussion_r3490595642

cjx0709 (approved):
LGTM

PR #1278, opened 2026-06-01
[Model] Support MiniMax-M2

zorrofox:
## Summary

Adds `MiniMaxM2ForCausalLM` (M2/M2.5/M2.7, 230B total / 10B active). Architecture: standard GQA (48h/8kv, head_dim=128) + sigmoid-gated MoE (256 experts, topk=8, `e_score_correction_bias`) + per-layer QK RMSNorm + partial RoPE (`rotary_dim=64`). Official checkpoint is FP8 block-wise 128×128 (same format as DeepSeek-V3).

Closes #909 (roadmap item "Support minimax m2").

## Implementation notes

This is sglang-jax's first **GQA + FP8-static + `tp > num_kv_heads`** combination, which surfaces several edge cases — all handled inside `minimax_m2.py` (no changes to shared code):

| Issue | Fix |
|---|---|
| `QuantizedLinear.from_linear` derives placeholder shapes from `LinearBase.output_size` *before* load-time `kv_head_padding`, so `weight_q` gets padded to `tp` heads but `weight_scale` stays at the unpadded placeholder shape | Size `k_proj`/`v_proj`/`k_norm`/`RadixAttention` for `max(num_kv_heads, tp)` at construction |
| `_apply_kv_head_padding` only matches `"k_proj"/"v_proj"` hf_keys, so the 1D per-layer `k_norm` scale loads unpadded | Post-load per-head tile on host (load replicated; multi-host sharded arrays are non-addressable for `np.asarray`) |
| Per-shard `k/v_proj` out_dim = `head_dim` = 128 = one N-block for any `tp >= num_kv_heads`, tripping the narrow-N guard | `allow_narrow_n_blockwise=True` via `patch_model_config`; GSM8K 5/5 confirms no accuracy loss |
| `qk_norm_type="per_layer"` norms over all heads jointly (`num_heads*head_dim`), variance must reduce across the tensor-sharded axis | `MiniMaxM2QKNorm` with scale sharded `P("tensor")`; GSPMD inserts the cross-shard all-reduce automatically (verified by `test_tp_variance_allreduce`) |

**Topology limitation**: `num_attention_heads=48` is not divisible by `tp=32` or `tp=64`, and sglang-jax has no q-head padding (only kv). M2/M2.5/M2.7 therefore run on **v6e-4/8/16** only among standard topologies.

## Tests (`test_minimax_m2.py`, CPU, <5s)

- Registry registration
- `MiniMaxM2QKNorm` reference + TP variance all-reduce
[...]

PR #1277, opened 2026-06-01
fix(openai): isinstance(dict) -> Mapping for BatchEncoding unwrap

zorrofox:
## Summary
- `tokenizer.apply_chat_template(tokenize=True)` returns `BatchEncoding` (a dict subclass) instead of `list[int]` for some chat templates on transformers>=5 (observed with `MiniMaxAI/MiniMax-M2`, transformers 5.9.0)
- Downstream `GenerateReqInput` validation then rejects with `input_ids should be a list of lists for batch processing`
- Unwrap `prompt_ids["input_ids"]` when a dict is returned

## Repro
```python
from transformers import AutoTokenizer
t = AutoTokenizer.from_pretrained("MiniMaxAI/MiniMax-M2", trust_remote_code=True)
out = t.apply_chat_template([{"role":"user","content":"hi"}], tokenize=True, add_generation_prompt=True)
type(out)  # -> BatchEncoding, not list
```

## Test plan
- [x] `/v1/chat/completions` with MiniMax-M2 returns 200 instead of 400 after fix (verified via hotfix on v6e-16)
- [x] `ruff check` passes

PR #1376, opened 2026-06-18
[Model] MiniMax-M3 with MSA sparse-attention decode

### LINKED ISSUE #1375: [Feature] Support MiniMaxM3SparseForCausalLM (MSA sparse attention)
# [Feature] Support MiniMaxM3SparseForCausalLM (MSA sparse attention)

## Model

[MiniMax-M3](https://huggingface.co/MiniMaxAI/MiniMax-M3) is a 428B-parameter (23B active) MoE model. The key architectural difference from M2 (#1278) is **MSA (MiniMax Sparse Attention)** — a learned block-sparse attention that scales to long context:

- **Index Branch**: each layer has a lightweight (1-head, low-dim) `index_q`/`index_k` projection
- **Prefill**: standard dense FA (chunked); `index_k` is written to a per-page pool alongside KV
- **Decode**: `index_q · index_k_history` scores all KV blocks (page_size=128), top-k=16 blocks are selected, then block-sparse GQA over only those 16×128 tokens

Reference: [MiniMax-AI/MSA](https://github.com/MiniMax-AI/MSA), [arXiv:2606.13392](https://arxiv.org/abs/2606.13392).

## What this needs in sglang-jax

- `models/minimax_m3.py` (mostly reuses M2 MoE + adds Index Branch + `swigluoai` activation)
- `mem_cache/memory_pool.py`: an `IndexKPool` alongside `KVCache` (per-layer `[n_pages, page_size, 1, d_idx]` bf16)
- `layers/attention/flashattention_backend.py`: an `is_msa` decode path that gathers `index_k` history, computes top-k block indices, and rewrites `page_indices`/`kv_lens`/`cu_kv_lens` to the selected blocks before calling RPA v3 (no new kernel needed)
- `layers/moe.py`: `swigluoai` activation (`elif` branch + two default-valued ctor params; existing models unaffected)
- `--page-size 128` is a hard requirement (must equal MSA block size)

## Validation target

- GSM8K (short-ctx, MSA ≡ dense)
- needle@64K (idle + under concurrent load — MSA top-k correctness depends on per-req `page_indices` indexing)
- Throughput vs H200×8 BF16 (sglang.io cookbook reference)

zorrofox:
# [Model] MiniMax-M3 with MSA sparse-attention decode

Closes #1375. Depends on #1277 (chat-template BatchEncoding unwrap) and #1371 (EPLB explicit-mesh gather) — both already open.

## Summary

Adds `MiniMaxM3SparseForCausalLM` (428B / 23B-active MoE, text-only). Architecture is M2 (#1278) + **MSA block-sparse decode**:

- **Prefill**: dense FA via existing chunked-prefill path; additionally writes per-token `index_k` into a paged `IndexKPool`.
- **Decode**: per-req top-k block selection (`index_q · index_k_history → top-16 of 128-token blocks`), then RPA v3 over only the selected 16 pages. No new Pallas kernel — top-k is plain `lax.top_k`, sparse attention reuses RPA with rewritten `page_indices`.

## Shared-code changes (all additive / behind guards)

| File | Change | Effect on other models |
|---|---|---|
| `layers/moe.py` | `swigluoai` activation: new `elif` branch + 2 default-valued ctor params | none (silu/gelu paths unchanged) |
| `layers/attention/flashattention_backend.py` | `_msa_inner` under `if index_k is not None and hasattr(pool, "get_index_k_buffer")` | none (guard is False for non-MSA models) |
| `layers/radix_attention.py` | return `attn_backend(...)` directly (3-tuple passthrough for MSA) | none (existing 2-tuple callers unchanged) |
| `mem_cache/memory_pool.py` | `IndexKPool` class + `get_index_k_buffer` (only allocated when model declares `index_head_dim`) | none |
| `model_executor/model_runner.py` | (a) propagate `num_experts`/`num_hidden_layers` from `text_config` when missing on `hf_config`; (b) `msa_ctx_page_buckets` decode-time page-index truncation | (a) no-op when attrs already exist; (b) `None` → original behavior |
| `model_executor/compilation_manager.py`, `managers/tp_worker.py` | thread `msa_ctx_page_buckets` through | none when `None` |
| `kernels/gmm/megablox_gmm_kernel/gmm_v2.py` | reserve 8MiB headroom in `rhs_vmem_target` heuristic (v7x 64MiB chiplet overflows by ~200KiB at k=6144 n=3072) | safer for all models on v7x; can split into separate PR if preferred |
| `kernels/fused_moe/v1/tuned_block_configs.py` | add v7x ep=16 entries | dict-add only |

`eplb/expert_location.py` and `entrypoints/openai/serving_chat.py` diffs in this branch are **not** part of this PR — they belong to #1371 / #1277 respectively.

## Correctness note: `pi_2d` under continuous batching

`page_indices` from `schedule_batch._merge_cache_loc` is **cumsum-packed ragged** (per-req start = `cumsum(aligned_lens[:r])`), not `[bs, P]` rectangular. The MSA decode path therefore builds `pi_2d` via `cu_kv_lens`-offset gather, not `reshape(bs, P)`. The latter happens to work at bs=1 (offset 0) but at bs>1 misaligns req `k>0` into stale `cache_loc_host_buf` entries (which are intentionally not re-zeroed — see the safety note in `schedule_batch.py`). Symptom: needle@64K passes idle but drops to ~1/5 under any concurrent load. The included `test_minimax_m3_msa.py` covers the bs>1 ragged case.

## Validation (TPU v7x-8, tp=16 ep=16, BF16)

| | result |
|---|---|
| GSM8K-platinum 4-shot, thinking (n=50) | **0.98** |
| needle@64K idle (5 positions) | 5/5 |
| **needle@64K under 2K cc=8 concurrent load** | **5/5** |
| 2K/256 cc=64 output | 843 tok/s (vs H200×8 1120, -25%) |
| 64K/1K cc=32 output | **100.5** tok/s (vs H200×8 101.2, ≈parity) |
| 64K/4K cc=32 output | **201** tok/s (vs H200×8 127, **+58%** — H200 KV-limited to bs=1 at 64K) |

## Test plan

- [x] `pytest test/models/test_minimax_m3.py` (shape/dispatch/weight-mapping checks)
- [x] `pytest test/models/test_minimax_m3_msa.py` (top-k block selection vs reference, bs∈{1,2,9})
- [x] `pytest test/models/test_minimax_m3_msa_trace.py` (jit-trace of `_msa_inner` on CPU mesh)
- [x] needle@64K idle + concurrent on v7x-8
- [x] GSM8K-platinum 4-shot (thinking) 0.98 on v7x-8
- [x] CI lint (pre-commit)

## Launch

```bash
  --model-path MiniMaxAI/MiniMax-M3 --trust-remote-code --dtype bfloat16 \
  --tp-size 16 --ep-size 16 --device tpu \
  --chunked-prefill-size 8192 --context-length 65536 --page-size 128 \
  --attention-backend fa --moe-backend epmoe \
  --precompile-bs-paddings 1 8 12 16 32 64 \
  --precompile-token-paddings 256 1024 2048 8192
```


zorrofox:
Rebased onto main@787c08e5 (only two trivial adjacent-param conflicts in `compilation_manager.py`/`tp_worker.py`; everything else auto-merged). CI should be green again shortly.

@JamesBrianD @MokusMokun — gentle ping, would either of you have time to take a look? This adds MiniMax-M3 with MSA sparse-attention decode; all changes to shared files are behind `if is_msa:` / `isinstance(..., MSATokenToKVPool)` guards. Happy to split further or adjust if that helps review.

Benchmark (v7x-8 vs H200×8 BF16, details in PR body):
- 64K/1K output: matches H200 (100.5 vs 101.2)
- 64K/4K output: **+58%** (201 vs 127) — H200 is bs=1-bound at 64K, TPU KV pool fits bs=11
- GSM8K 4-shot 0.98, needle@64K 5/5 (idle+concurrent)

python/sgl_jax/srt/layers/attention/flashattention_backend.py
```diff
@@ -614,9 +636,178 @@ def _ragged_paged_attention_with_fused_kv(*args):
...
+            else:
+                neg_inf = jnp.finfo(ikp_l.dtype).min
+                ikp_upd = ikp_l.at[q_page, 0].set(neg_inf).at[q_page, 0].max(ik_cast)
+            bs_l = kv_lens.shape[0]
+            assert page_indices.shape[0] % bs_l == 0
+            pages_per_seq = page_indices.shape[0] // bs_l
+            # Skip MSA topk when the (static) page budget cannot exceed topk —
+            # selecting topk from ≤topk blocks is the identity (≡ dense). This
+            # avoids the O(pages_per_seq) ik gather on short-ctx deployments.
+            if is_decode and msa_topk > 0 and pages_per_seq > msa_topk:
```
MokusMokun, line 736: The sparse decode path uses `self.page_size` as the MSA block size throughout (`n_blocks_v`, `q_block`, `bscores`, `msa_kvl`); the model reads `sparse_block_size` (minimax_m3.py:207) but never passes it to the backend, and nothing validates they agree. The repo default is `--page-size 1` (the `<128` hard-fail is PD-mode-only), so launching M3 without `--page-size 128` silently selects top-16 tokens instead of top-16 128-token blocks — accuracy collapses with no error. `--page-size 256` mis-granulates the same way.

Suggestion: hard-fail at model-runner/pool init when `use_sparse_attention` and `page_size != sparse_block_size`, with a unit test that the default config raises.

python/sgl_jax/test/models/test_minimax_m3.py
```diff
@@ -0,0 +1,130 @@
+from types import SimpleNamespace
```
MokusMokun, line 1: None of the three new test files is registered in `test/srt/run_suite.py`, and CI suites don't glob `python/sgl_jax/test/models/` — so CI stays green without running any of them, including `test_pi_2d_ragged_layout`, the regression test for this PR's central bs>1 claim.

Suggestion: add the three files to `unit-test-cpu` with `runner="pytest"`, same as `test_qwen3_5.py`.

python/sgl_jax/test/models/test_minimax_m3_msa_trace.py
```diff
@@ -0,0 +1,317 @@
...
+PAGE_SIZE = 128
+KV_HEADS = 2
+Q_HEADS = 4
+HEAD_DIM = 128
+IDX_HEADS = 2
+IDX_DIM = 64
+POOL_SIZE = 512
+LAYER_NUM = 2
+SPARSE_LAYER = 1
+PAGES_PER_SEQ = 4
```
MokusMokun: With `PAGES_PER_SEQ = 4` and `TOPK = 4`, each data shard sees `pages_per_seq == msa_topk`, so the `pages_per_seq > msa_topk` guard is false and the test only traces the ik-write + dense-RPA path. The top-k selection subgraph (einsum → `top_k` → sort → `take_along_axis` → metadata rewrite) — where the v1 ShardingTypeError lived — is never compiled, in either the direct-backend or the full-integration test.

Suggestion: use `PAGES_PER_SEQ = 8` (or `TOPK = 2`) and assert the sparse branch was taken, e.g. via the metadata shape seen by the RPA stub.

python/sgl_jax/srt/layers/attention/flashattention_backend.py
```diff
@@ -614,9 +636,178 @@ def _ragged_paged_attention_with_fused_kv(*args):
...
+                    inserted_window_dims=(0,),
+                    scatter_dims_to_operand_dims=(0,),
+                )
+                ik_buf_upd = jax.lax.scatter(
+                    ik_buf_l.reshape(-1, d_idx),
+                    flat_idx[order][:, None],
+                    ik_cast[order],
+                    _dn,
+                    indices_are_sorted=True,
+                    unique_indices=True,
```
MokusMokun: Extend batches pad `out_cache_loc` with `-1` (schedule_batch.py:1945) and the clamp above maps them all to flat index 0, so this scatter promises `unique_indices=True` while holding many duplicate zeros — undefined behavior in XLA, on essentially every prefill. Benign today (TPU serializes the writes and page 0 is never read), but a standing miscompile hazard. The prefill trace test has no `-1` entries, so the case the clamp was written for is uncovered.

Suggestion: drop `unique_indices=True` (keep `indices_are_sorted=True`, which the argsort establishes) and re-measure the 367us/layer claim; add padded `-1` entries to the EXTEND trace test.

python/sgl_jax/srt/models/minimax_m3.py
```diff
@@ -0,0 +1,638 @@
...
+        k = k.reshape(-1, self.kv_head_num, self.head_dim, out_sharding=head_sharding)
+        v = v.reshape(-1, self.kv_head_num, self.head_dim, out_sharding=head_sharding)
+        q = self.q_norm(q)
+        k = self.k_norm(k)
+        q, k = self.rotary_emb(positions, q, k)
+
+        ik_upd = None
+        if self.is_sparse and hasattr(token_to_kv_pool, "set_index_k_buffer"):
+            iq, ik_new = self._compute_index_qk(hidden_states, positions)
+            attn_output, kv_fused, ik_upd = self.attn(
```
MokusMokun, line 269: `NativeAttention.__call__` accepts the MSA kwargs via `**kwargs` but returns a 2-tuple, so with `--attention-backend native` this unpack dies with a bare "not enough values to unpack" — after the full 428B weight load.

Suggestion: reject MSA + non-fa backends in the same startup validation as the page-size check.

python/sgl_jax/srt/models/minimax_m3.py
```diff
@@ -0,0 +1,638 @@
+"""MiniMax-M3 (text-only, dense-attention fallback).
```
MokusMokun: The header still describes a dense-attention fallback with `index_*` weights skipped at load, but `_create_weight_mappings` loads them and the MSA decode path is wired — the "dense-attn fallback" log at :486 has the same problem. A maintainer reading either will conclude MSA is not implemented.

Suggestion: rewrite the header to the actual design (dense prefill with per-token `index_k` writes, top-k block selection at decode, text-only VL alias) and fix the log line.

MokusMokun (changes requested):
Reviewed at 10beb85f. No correctness bug found in the core decode path itself (checked: the pi_2d gather against RPA v3's cu_kv_lens page addressing, the msa_kvl rewrite landing the KV write on the same physical slot as the dense path, the -1→page-0 clamp against the allocators' reserved page 0, and per-DP-rank-local cache_loc addressing). Blocking items are in the inline comments.

Non-blocking notes:

- The `serving_chat.py` hunk belongs to #1277 per the PR body — coordinate the rebase so it isn't double-applied.
- The test plan calls `test_minimax_m3.py` "HF-eager parity", but the file has only shape/dispatch checks — commit the parity test or correct the body.
- `MSATokenToKVPool.set_index_k_buffer` is unused at runtime and still has the `-1 // page_size` wraparound — clamp it too, or reduce it to a pure marker.
- Memory estimates count `index_k` but not `index_k_pooled` (1/page_size of ik; worth a comment).
- `SGLANG_MSA_POOLED`: the EXTEND `.set(neg_inf)` wipes a boundary page's pooled max when a chunk starts mid-page — benign only while `chunked_prefill_size % page_size == 0`. The env var is undocumented, and `ik_pooled` is maintained even when it is off.
- The fused MoE kernels hardcode swigluoai `alpha=1.702, limit=7.0` and ignore `config.swiglu_alpha/swiglu_limit` (EPMoE honors them); the fused branch is also unvalidated for M3. A startup assert would prevent silent divergence.
- The `gmm_v2.py` VMEM reserve changes tiling for all untuned shapes on every TPU generation — worth one kernel-performance CI run, or split out as offered.
- `MiniMaxM3Attention.__call__` is annotated as returning a 2-tuple but returns 3.
- The `MHATokenToKVPool.get_index_k_buffer` stub makes the backend's `hasattr` guard always-true for MHA pools; `isinstance(pool, MSATokenToKVPool)` is the honest check.
- Decode precompile now runs bs × 3 ctx buckets with the largest at full context — worth reporting total precompile time on v7x.
- Private iteration codenames leak into shipped comments: "G3" (flashattention_backend/compilation_manager/model_runner), "P0a approximate" / "v2 exact" (flashattention_backend), "R9/R9b" and "v1/v2" (msa_trace test docstrings), and the `minimax_m3.py` header cites `refs/ground-truth.md`, which does not exist in the repo — drop the numbers and the dead link; the surrounding text already carries the meaning.
- `test_pooled_topk_recovers_outlier_block` is missing `@pytest.mark.unit`.

python/sgl_jax/srt/layers/attention/flashattention_backend.py
zorrofox: Added a startup `ValueError` in `model_runner_kv_cache_mixin.py` at `MSATokenToKVPool` creation (before weight load) for both `page_size != sparse_block_size` and `attention_backend != "fa"` — covers the `--page-size 256` mis-granulation case too.

python/sgl_jax/test/models/test_minimax_m3.py
zorrofox: Registered all three in `unit-test-cpu` with `runner="pytest"` (run_suite.py:316-318).

python/sgl_jax/test/models/test_minimax_m3_msa_trace.py
zorrofox: Changed `PAGES_PER_SEQ` to 8. `rpa_stub` now captures the trace-time `page_indices.shape[0]` and `test_msa_decode_shard_map_traces` asserts it equals `bs_per_dp * TOPK` (the sparse branch rewrites `page_indices` to `msa_pi`; the dense/skip branch would leave it at `bs_per_dp * PAGES_PER_SEQ`).

python/sgl_jax/srt/layers/attention/flashattention_backend.py
zorrofox: Dropped `unique_indices=True`, kept `indices_are_sorted=True`. I'll re-run the ik-scatter micro-bench on v7x and update the PR body number — the sorted-only lowering should still be vectorized so I don't expect a large regression, but will confirm.

python/sgl_jax/srt/models/minimax_m3.py
zorrofox: Addressed by the same startup validation as the page-size check (same `ValueError` block in `model_runner_kv_cache_mixin.py`) — fails before weight load.

python/sgl_jax/srt/models/minimax_m3.py
zorrofox: Rewrote the module docstring and the `logger.info` line to describe the actual design (dense prefill + per-token index_k write, top-k block selection at decode, VL alias to text-only).

zorrofox (commented):
Addressed all six inline items in f38a35d6; details in each thread. On the non-blocking notes:

- `serving_chat.py`: our hunk also handles the `list[list[int]]` case (M3's chat template returns a nested list under `apply_chat_template(..., tokenize=True)`) which #1277 doesn't cover. I'll keep it for now and rebase-coordinate if #1277 lands first — happy to move it to a separate PR if preferred.
- Test-plan wording: will update the PR body to say "shape/dispatch/weight-mapping checks" instead of "HF-eager parity" (parity is validated end-to-end via GSM8K/needle in the benchmark section, not by a unit test).
- `set_index_k_buffer`: removed (unused — the actual write is in `_msa_inner`). Switched the model's `hasattr` check to `get_index_k_pooled`, which has no base-class stub so it correctly discriminates `MSATokenToKVPool`.

For inline #4 I'll re-measure the sorted-only ik-scatter micro-bench on v7x-8 and update the 367µs number in the PR body before requesting re-review.

python/sgl_jax/srt/layers/attention/flashattention_backend.py
zorrofox: Re-measured on v7x (single-host 2x2x1, jax 0.8.1): sorted-only is 259µs/layer vs sorted+unique 255µs (+1.6%) at prefill T=8192, so no meaningful regression from dropping the hint. Full comparison:

| variant | prefill 8192 | ×57 |
|---|---|---|
| v0 plain `.at[].set` | 1070µs | 61.0ms |
| v5 sorted+unique (old) | 255µs | 14.5ms |
| **v6 sorted-only (this PR)** | **259µs** | **14.8ms** |

(The 367µs in the PR body was from an earlier libtpu; will update the body to 259µs.)

tsvikas:
> **CORRECTED — please read the next comment.**
> We had the wrong eval invocation (missing `--enable-thinking` and `--platinum`).
> Finding 1 is void: the reported 0.98 reproduces exactly. The accuracy claims in Finding 2 are withdrawn: under the correct protocol `--parallel 64` matches or beats `--parallel 1`. What remains is ordinary non-batch-invariance, not a defect in this PR.
> Original text kept below for the record.

Hi, we've been testing this PR on a v7x-8 slice and found two things we think are worth raising before merge. Both are on pristine `faf93194` with your `## Launch` flags unmodified. Setup details at the bottom. It is possible that we missed something - please tell me if we're holding it wrong.

# Finding 1: we measure much lower GSM8K accuracy than your validation table
<details>
<summary><b>VOID: our eval invocation was wrong</b></summary>
Using the repo's own `benchmark/gsm8k/bench_sglang_jax.py`, 4-shot, `--temperature 0`, n=100, at `--parallel 1`:

| | GSM8K 4-shot |
|---|---|
| our measurement, your `## Launch` flags verbatim | **0.870** (three runs, identical) |
| your validation table (n=50) | 0.98 |

The three runs gave exactly the same score, so this is not run-to-run noise. We cannot get near 0.98.

We used `--parallel 1` deliberately, to get a clean baseline: at higher concurrency the score is worse and varies between runs, which we think is finding 2 below.

We also see the score shift a little with server flags that we would not expect to affect a greedy result (`--context-length`, `--mem-fraction-static`, the padding buckets). We would rather first confirm that our setup is correct, and can go into that separately if useful.
</details>

# Finding 2: identical prompts in the same batch produce different outputs

<details>
<summary><b>VOID: our eval invocation was wrong, and the remaining observation is the expcted non-batch-invariance</b></summary>
Send one request containing the same prompt 64 times with `temperature: 0`. All 64 should be identical. They are not.

```python
import json, urllib.request
prompt = ...  # one 4-shot GSM8K prompt, as bench_sglang_jax.py builds it
req = urllib.request.Request(
    "http://localhost:30000/generate",
    data=json.dumps({
        "text": [prompt] * 64,
        "sampling_params": {"temperature": 0.0, "top_p": 1.0, "max_new_tokens": 512,
                            "stop": ["Question", "Assistant:", "<|separator|>"]},
    }).encode(),
    headers={"Content-Type": "application/json"},
)
outs = [x["text"] for x in json.loads(urllib.request.urlopen(req).read())]
print(len(set(outs)), "distinct outputs from 64 identical prompts")
```

We ran this for 20 different GSM8K questions:

| | |
|---|---|
| questions where the 64 copies gave more than one distinct continuation | **19 / 20** |
| questions where the 64 copies gave more than one distinct final answer | **14 / 20** |
| worst case | 46 distinct continuations from 64 identical prompts |

The divergence includes token-level corruption, not just rephrasing. Same prompt, three different batch positions:

```
pos 0,4,5..  ... then for 20 chickens they would need 3*20=<<3*20=60>>60 cups ...
pos 30,40    ... She needs to give the. chickens an additional 60 - 40 = ...
pos 3        ... She needs to give theircks 60 - 40 = <<60-40=20>>20 cups ...
```

`the. chickens` and `theircks` look like a flipped argmax rather than a different valid phrasing.

**This costs accuracy.** The clearest case is question 15: answered correctly (125) when sent alone, but when sent as 64 identical copies only 7 of 64 were correct, with 57 answering 29 instead.

Aggregated over the 20 questions, the 64-copy batches averaged 0.647 against 0.750 for the same questions sent alone. Note the effect randomizes rather than strictly degrades: two questions that were wrong when sent alone had a substantial fraction of batch positions answer them correctly.

<details>
<summary>Per-question detail (20 questions)</summary>

`alone` is the answer when sent by itself. `batch acc` is the fraction of the 64 identical copies matching the gold answer.

| q | gold | alone | correct alone | batch acc | answers across the 64 copies |
|---|---|---|---|---|---|
| 4 | 20 | 20 | Y | 1.000 | 20x64 |
| 5 | 64 | 64 | Y | 0.703 | 64x45, 60x19 |
| 6 | 260 | 260 | Y | 0.562 | 260x36, 180x20, 80x7, 8x1 |
| 7 | 160 | 160 | Y | 0.531 | 160x34, 120x22, 220x6, 80x2 |
| 8 | 45 | 45 | Y | 0.562 | 45x36, 315x16, 135x7, 2x2 |
| 9 | 460 | 460 | Y | 0.828 | 460x53, 8x5, 520x5, 2x1 |
| 10 | 366 | 366 | Y | 0.953 | 366x61, 474x1, 8x1, 5x1 |
| 11 | 694 | 694 | Y | 0.750 | 694x48, 330x16 |
| 12 | 13 | 12 | N | 0.000 | 12x61, unparseable x3 |
| 13 | 18 | 30 | N | 0.016 | 30x46, 15x10, 22x7, 18x1 |
| 14 | 60 | 60 | Y | 1.000 | 60x64 |
| 15 | 125 | 125 | Y | 0.109 | 29x57, 125x7 |
| 16 | 230 | 460 | N | 0.406 | 460x37, 230x26, 310x1 |
| 17 | 57500 | 57500 | Y | 0.797 | 57500x51, 65000x13 |
| 18 | 7 | 7 | Y | 1.000 | 7x64 |
| 19 | 6 | 6 | Y | 1.000 | 6x64 |
| 20 | 15 | 67 | N | 0.375 | 15x24, 67x14, 3x11, 44x9 |
| 21 | 14 | 2 | N | 0.344 | 2x37, 14x22, 29x5 |
| 22 | 7 | 7 | Y | 1.000 | 7x64 |
| 23 | 8 | 8 | Y | 1.000 | 8x64 |

</details>

The same effect shows up in the full benchmark under concurrency. Same server, same config, `POST /flush_cache` before each run so that runs do not inherit each other's prefix cache:

| `--parallel` | runs | accuracy |
|---|---|---|
| 1 | 3 | 0.870, 0.870, 0.870 |
| 64 | 8 | 0.490, 0.560, 0.560, 0.560, 0.560, 0.550, 0.490, 0.530 |

A useful shortcut if you want to check this quickly: total generated tokens tracks accuracy exactly, so you can detect it without scoring anything. Runs with the same token count score identically, and more tokens always means lower accuracy (14507 tokens gave 0.560 on four separate runs, 17002 gave 0.490 on two, and roughly 8210 at `--parallel 1` gave 0.870).
</details>

# Setup

- 2x `tpu7x-standard-4t` (8 chips, 16 JAX devices), tp=16 ep=16, BF16.
- Weights: `MiniMaxAI/MiniMax-M3` revision `50942730318c7943fe83db7ec8e9f9177ecb1cf8`. All 81 files match by name and size except `README.md` (24-byte docs difference), and we sha256-verified one full shard (`model-00025-of-00059.safetensors`) against the HF LFS oid.
- Source: `faf93194`, verified byte-identical on both hosts, installed into a from-scratch venv.
- jax/jaxlib 0.8.1, libtpu 0.0.30, Python 3.12.
- Your `## Launch` block verbatim, plus the four flags it omits that a 2-host slice needs (`--nnodes 2 --node-rank {0,1} --dist-init-addr <ip>:9000 --host --port`).
- `POST /flush_cache` before every measured run. This matters: a warm prefix cache changes the numbers noticeably, so all figures above are from a flushed cache unless stated.

# Questions

1. Could you share the exact GSM8K invocation behind the 0.98, in particular the concurrency it ran at? Given finding 2, a run at low concurrency would score much higher than one at `--parallel 64`.
2. Had the server already served that 4-shot prefix when you measured? We consistently get better numbers with a warm radix cache than after flushing. One possibility is that a prefix computed once at low batch gets reused by later high-batch requests, effectively carrying the batch-1 result into a batched measurement. That would make a warm-cache number look better than what the model actually produces under load.

Happy to run any diagnostic while we still have the slice.

tsvikas:
Correction to my comment above. We found the config gap, which resolves both findings. Sorry for the noise.

## Finding 1: your number reproduces, we had the wrong eval invocation

We ran `bench_sglang_jax.py` without `--enable-thinking` and without `--platinum`. With both, your number reproduces exactly -- nothing about it needs explaining. Sorry for the mixup (One small suggestion: "GSM8K 4-shot" reads as the stock set, might worth labelling it as "gsm8k-platinum" and "thinking")

## Finding 2: the accuracy claim is withdrawn

Re-running under the correct protocol (`--enable-thinking --platinum`, n=50, flushed between runs):

| `--parallel` | accuracy |
|---|---|
| 1 | 0.980, 0.980 |
| 64 | 0.980, 1.000, 1.000, 1.000, 0.980 |

The accuracy collapse we reported does not exist in this protocol. Please disregard that part.

The remaining observation is that identical prompts in one batch produce different continuations. That is real, and larger in thinking mode, but it is the well-known non-batch-invariance of this class of kernel rather than anything specific to your change.

Thanks for the PR, and sorry again for opening with an accuracy claim that turned out to be our own protocol error.

zorrofox:
Thanks for testing on real hardware and for the quick follow-up — glad it reproduces. Good suggestion on the labelling: updated the validation table to say "GSM8K-platinum 4-shot (thinking)" so the eval protocol is unambiguous. The batch-invariance observation matches what we see too (bf16 accumulation order varies with batch composition); nothing M3-specific.

tsvikas:
Thanks!
By the way, is sparse prefill planned, or is dense prefill the intended end state?

zorrofox:
Dense (chunked) prefill is intentional for this PR, and on TPU it's likely to stay for a while, for two reasons:

1. **Correctness holds up empirically** — needle@64K 5/5 (idle + concurrent) and GSM8K-platinum 0.98 are all measured with dense prefill, and the `index_k` cache written during prefill is exactly what decode's top-k selection consumes.

2. **Prefill on TPU is MoE-bound, not attention-bound** — at 8K chunks, attention is ~65.5ms (~10% of the step) while the MoE stack is ~60%+. Dense FA on TPU actually costs less than the H200's sparse-prefill attention (91.0ms) at the same chunk size, so the sparse-prefill win on TPU is capped to a modest TTFT improvement at very long contexts.

Sparse prefill would need a block-sparse flash-attention Pallas kernel (KV-outer loop like the official SM100 kernel) — a reasonable follow-up if long-context TTFT becomes the bottleneck, but out of scope here.

Issue #458, opened 2025-11-19
[Feature] Introduce mypy in a gradual rollout way


guo-han:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sgl-jax/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Introduce mypy into codebase (in pre-commit maybe) would help catch low-level bugs early, like previously done in vllm, jax, etc. 
Follow a gradual rollout plan to achieve this.

### Related resources

https://docs.jax.dev/en/latest/jep/12049-type-annotations.html
https://github.com/vllm-project/vllm/pull/25313/files#diff-6b980d504c08a48078accc6b5fb392b7b5d4c65dfe293a3956a7109200f184e2

PR #1377, opened 2026-06-18
Add mypy for utils

AmedeoBiolatti:
## Motivation

mypy coverage for utils module, as for https://github.com/sgl-project/sglang-jax/issues/458

## Modifications

- added `"python/sgl_jax/srt/utils/*.py"` mypy to `[tool.mypy]`
- solved minor typing annotation issues 

## Accuracy Tests

no change

## Benchmarking and Profiling

no change

## Checklist

### DIFF
--- a/python/sgl_jax/srt/utils/jax_utils.py
+++ b/python/sgl_jax/srt/utils/jax_utils.py
@@ -195,18 +195,19 @@ def get_available_device_memory(
     return int(free_gpu_memory * (1 << 10))
 
 
-def device_array(*data, sharding=None, **kwargs) -> jax.Array:
+def device_array(data, sharding=None, **kwargs) -> jax.Array:
     if sharding is None:
         return jax.device_put(data, device=sharding, **kwargs)
 
     def _to_device(arr):
+        arr = np.asarray(arr)
+
         def fn(idx, a=arr):
             return a[idx]
 
-        arr = np.asarray(arr)
         return jax.make_array_from_callback(arr.shape, sharding, fn)
 
-    return jax.tree.map(_to_device, *data)
+    return jax.tree.map(_to_device, data)
 
 
 _IS_TPU_RUNTIME_CACHED: bool | None = None
--- a/python/sgl_jax/srt/utils/jax_utils.py
+++ b/python/sgl_jax/srt/utils/jax_utils.py
@@ -47,7 +47,6 @@ def get_device_name(num_devices: int | None = None):
 
 
 def get_device_hbm_limit() -> int:
-
     device_kind = get_device_name()
     if device_kind == "TPU v5p" or device_kind == "TPU v5":
         return 95 * GBYTES
@@ -64,7 +63,7 @@ def get_device_hbm_limit() -> int:
 
 
 def pathways_hbm_usage_gb(live_arrays, devices: Any) -> list[tuple[float, float]]:
-    hbm_used = defaultdict(int)
+    hbm_used: defaultdict[str, int] = defaultdict(int)
     hbm_limit = get_device_hbm_limit()
     for array in live_arrays:
         for buffer in array.addressable_shards:
@@ -113,7 +112,7 @@ def get_original_kv_head_id(tp_rank: int, total_num_kv_heads: int, tp_size: int)
 
 
 def get_available_device_memory(
-    device, distributed=False, empty_cache=True, device_indexes: list[int] = None
+    device, distributed=False, empty_cache=True, device_indexes: list[int] | None = None
 ):
     """
     Get available memory for device:device_id.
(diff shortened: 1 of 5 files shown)


python/sgl_jax/srt/utils/jax_utils.py
```diff
@@ -198,11 +197,14 @@ def _get_available_memory_distributed(a):
 
 def device_array(*data, sharding=None, **kwargs) -> jax.Array:
     if sharding is None:
-        return jax.device_put(*data, device=sharding, **kwargs)
+        return jax.device_put(data, device=sharding, **kwargs)
```
Rodrian7, line 200: jax.device_put(*data, ...) → jax.device_put(data, ...) — this changes runtime behavior, not just type annotations. Was this intentional?

python/sgl_jax/srt/utils/jax_utils.py
```diff
@@ -198,11 +197,14 @@ def _get_available_memory_distributed(a):
...
 
     def _to_device(arr):
+        def fn(idx, a=arr):
+            return a[idx]
+
         arr = np.asarray(arr)
-        return jax.make_array_from_callback(arr.shape, sharding, lambda idx, a=arr: a[idx])
+        return jax.make_array_from_callback(arr.shape, sharding, fn)
```
Rodrian7, line 208: In _to_device, fn is defined before arr = np.asarray(arr), so the callback receives the original JAX array instead of the converted numpy one. Differs from the original behavior.

python/sgl_jax/srt/utils/weight_utils.py
```diff
@@ -1904,7 +1921,7 @@ def load_weights_from_safetensors(
 
                 infos = weight_info[hf_key]
 
-                if isinstance(mapping, str | list):
+                if not isinstance(mapping, WeightMapping):
```
Rodrian7: This changes behavior for tuple mappings — previously they were not wrapped into WeightMapping, now they are. Was this intentional?

python/sgl_jax/srt/utils/jax_utils.py
AmedeoBiolatti: `jax.device_put(*data, device=sharding, **kwargs)` doesn't seem to be compatible in general with `jax.device_put`, since there are only two positional arguments and the second one is already passed with `device=sharding`, so `device_array(x, y)` could raise an error.
 (https://docs.jax.dev/en/latest/_autosummary/jax.device_put.html)

python/sgl_jax/srt/utils/weight_utils.py
```diff
@@ -64,8 +65,8 @@ def _reinterpret_dtype_if_needed(data: np.ndarray, target_dtype: jnp.dtype) -> n
 
 @dataclass
 class WeightMapping:
-    target_path: str | list[str]
-    sharding: tuple | None = None
+    target_path: str | list[str] | tuple[str]
+    sharding: tuple[Any] | tuple[Any, Any] | None = None
```
Rodrian7: sharding: tuple[Any] | tuple[Any, Any] doesn't cover 3-element tuples like ("expert", None, None) — should be tuple[Any, ...].
What's the motivation for adding tuple[str] to target_path?

python/sgl_jax/srt/utils/jax_utils.py
Rodrian7: Good point. But with data, single-arg callers like device_array(np.array(...)) now get a tuple back instead of a single array — does that need handling?

AmedeoBiolatti: We could move the definition to `def device_array(data, sharding=None, **kwargs) -> jax.Array:`, since it is always used with one positional argument and it is more consistent with jax

python/sgl_jax/srt/utils/jax_utils.py
AmedeoBiolatti: I missed the fact that defaults are evaluated at def time in python, good to know. I will fix this

python/sgl_jax/srt/utils/weight_utils.py
AmedeoBiolatti: The code here is already expecting `str | list | WeightMapping` and a `tuple` would raise an error in the old version (there is `mapping.concat_axis` a couple of lines below), but you're right, we should keep the behaviour and make it more explicity. I will check type and raise an error in case of a tuple

python/sgl_jax/srt/utils/weight_utils.py
AmedeoBiolatti: - fixed sharding
- target_path is related to the previous comment about tuple->WeightMapping, I added the tuple type to make it compatible with tuple conversion to WeightMapping, but now there there are more checks I removed it

python/sgl_jax/srt/utils/weight_utils.py
AmedeoBiolatti: done, now I added a stricter check on the mapping type

python/sgl_jax/srt/utils/jax_utils.py
AmedeoBiolatti: fixed

python/sgl_jax/srt/utils/jax_utils.py
AmedeoBiolatti: fixed

Rodrian7 (approved):

cjx0709:
Temporarily closing and reopening to retrigger PR CI on the latest head commit.

PR #1384, opened 2026-06-22
feat(glm5): support GLM-5.2 config (indexer_types + rope_parameters)

### LINKED ISSUE #1383: [Feature] Support GLM-5.2 (`indexer_types` IndexShare config + nested `rope_parameters`)
[GLM-5.2](https://huggingface.co/zai-org/GLM-5.2) reuses `GlmMoeDsaForCausalLM` with the same core architecture as 5.1 (78 layers, 6144 hidden, MLA, 256-expert MoE), but two config additions break loading on the current `glm5_moe.py`:

### 1. `indexer_types` (IndexShare)

5.2 adds a per-layer `indexer_types: ["full","full","full","shared","shared","shared","full",...]` (21 `"full"` / 57 `"shared"`). Layers tagged `"shared"` reuse the previous `"full"` layer's top-k indices and **ship no `self_attn.indexer.*` weights** in the checkpoint (verified via `model.safetensors.index.json`: layer 3/4/5 have 0 indexer keys, layer 2/6 have 5).

The current code builds a `GlmDsaIndexer` per layer unconditionally → weight loading fails on shared layers (model params with no matching checkpoint keys).

### 2. `rope_parameters.rope_theta`

GLM-5.x nests `rope_theta` under `rope_parameters`. `Glm5DecoderLayer` reads `getattr(config, "rope_theta", 1000000)` — 5.1 happens to match the `1e6` fallback so it works by accident; 5.2 uses `rope_theta=8000000` and would silently mis-RoPE.

### Other config diffs (no action needed)

- `max_position_embeddings`: 202752 → 1048576 (already read via getattr)
- `head_dim`: 64 → 192 (`patch_model_config` already overrides to 256 = qk_head_dim)
- `mlp_layer_types`: explicit list, equivalent to existing `first_k_dense_replace=3`
- `index_topk_freq`/`index_skip_topk_offset`/`index_share_for_mtp_iteration`: IndexShare scheduling params, irrelevant on the dense-MLA path (indexer output is discarded)

### Proposed fix

Minimal: skip building/mapping the indexer module when `indexer_types[layer_id] != "full"`, and fall back to `rope_parameters["rope_theta"]` when the flat attr is absent. Both are no-ops for 5.1. This does **not** implement IndexShare semantics — since the dense-MLA path discards indexer output anyway (`_ = self.indexer(...)`), skipping shared layers is lossless and just saves 57 layers of unused forward.

PR incoming.

zorrofox:
Fixes #1383.

## Motivation

GLM-5.2 ([HF](https://huggingface.co/zai-org/GLM-5.2)) reuses `GlmMoeDsaForCausalLM` with identical core architecture (78 layers, 6144 hidden, MLA, 256-expert MoE), but two config additions break loading on the existing 5.1 path:

1. **`indexer_types`** (IndexShare): 57/78 layers are tagged `"shared"` and ship **no indexer weights** in the checkpoint. The current code builds an indexer per layer unconditionally → weight loading fails with unexpected params.
2. **`rope_parameters.rope_theta`**: GLM-5.x nests `rope_theta` under `rope_parameters`. 5.1 happened to match the `1e6` fallback; 5.2 uses `8e6` and would silently mis-RoPE.

## Modifications

`python/sgl_jax/srt/models/glm5_moe.py` (+48/-27):
- `Glm5Attention.__init__`: add `has_indexer: bool = True`, conditionally build `self.indexer`
- `Glm5Attention.__call__`: guard `self.indexer(...)` with `is not None`
- `Glm5DecoderLayer.__init__`: read `rope_theta` from `rope_parameters` fallback; derive `has_indexer` from `config.indexer_types[layer_id] == "full"`
- `_create_glm5_weight_mappings` / `_create_moe_layer_mappings`: skip indexer key mappings on shared layers

Both changes are **no-ops for GLM-5.1** (`indexer_types` absent → all `"full"`; flat `rope_theta` still wins over nested).

Note: this does **not** implement IndexShare semantics (cross-layer top-k reuse). Since the dense MLA path already discards indexer output (`_ = self.indexer(...)`), skipping the module on shared layers is lossless and saves 57 layers of unused indexer forward.

## Verification

TPU v6e-64, W8A8 in-framework (`fp8_w8a8.yaml`), tp64/dp8/ep64, 1K/1K cc=460 ratio=1, ×2, main `0498a2e7`:

| | GLM-5.1 | **GLM-5.2 (this PR)** | Δ |
|---|---|---|---|
| sanity / GSM8K | 3/3 / 5/5 | **3/3 / 5/5** ✅ | — |
| output tok/s | 2744 | **2801** (σ=1.2%) | **+2.1%** |
| Median TPOT | 133.3 ms | **130.4 ms** | -2.2% |
| KV pool tokens (dp8) | 976384 | **1019392** | +4.4% |

The +2.1% throughput / +4.4% KV are consistent with skipping 57 layers of indexer params/forward.

## Checklist

- [x] `pre-commit run --all-files`
- [x] Backward-compat with GLM-5.1 verified (config parsing + e2e on same image)
- [x] e2e sanity + GSM8K on v6e-64 W8A8 dp1/dp8
- [x] Rebased onto `0498a2e7` (post-#1344/#1379)

### DIFF
--- a/python/sgl_jax/srt/models/glm5_moe.py
+++ b/python/sgl_jax/srt/models/glm5_moe.py
@@ -164,6 +164,7 @@ class Glm5Attention(nnx.Module):
         attention_bias: bool = False,
         dtype: jnp.dtype = jnp.bfloat16,
         use_absorbed: bool = True,
+        has_indexer: bool = True,
     ):
         super().__init__()
         self.layer_id = layer_id
@@ -243,15 +244,18 @@ class Glm5Attention(nnx.Module):
             scope_name="o_proj",
         )
 
-        self.indexer = GlmDsaIndexer(
-            hidden_size=hidden_size,
-            q_lora_rank=self.q_lora_rank,
-            index_head_dim=128,
-            index_n_heads=32,
-            mesh=mesh,
-            dtype=dtype,
-            scope_name="indexer",
-        )
+        if has_indexer:
+            self.indexer = GlmDsaIndexer(
+                hidden_size=hidden_size,
+                q_lora_rank=self.q_lora_rank,
+                index_head_dim=128,
+                index_n_heads=32,
+                mesh=mesh,
+                dtype=dtype,
+                scope_name="indexer",
+            )
+        else:
+            self.indexer = None
         self.rotary_emb = RotaryEmbedding(
             head_size=self.qk_rope_head_dim,
             rotary_dim=self.qk_rope_head_dim,
@@ -407,7 +411,8 @@ class Glm5Attention(nnx.Module):
         q, _ = self.q_b_proj(q_compressed)
         q = q.reshape(-1, self.num_heads, self.qk_head_dim)
 
-        _ = self.indexer(hidden_states, q_compressed, positions, self.rotary_emb)
+        if self.indexer is not None:
+            _ = self.indexer(hidden_states, q_compressed, positions, self.rotary_emb)
 
         q_nope = q[:, :, : self.qk_nope_head_dim]
         q_rope = q[:, :, self.qk_nope_head_dim :]
@@ -584,7 +589,8 @@ class Glm5DecoderLayer(nnx.Module):
     ):
         self.layer_id = layer_id
         self.hidden_size = config.hidden_size
-        rope_theta = getattr(config, "rope_theta", 1000000)
+        rope_params = getattr(config, "rope_parameters", None) or {}
+        rope_theta = getattr(config, "rope_theta", None) or rope_params.get("rope_theta", 1000000)
         rope_scaling = getattr(config, "rope_scaling", None)
         max_position_embeddings = getattr(config, "max_position_embeddings", 131072)
         self.head_dim = getattr(config, "head_dim", None) or 128
@@ -592,6 +598,12 @@ class Glm5DecoderLayer(nnx.Module):
 
         partial_rotary_factor = getattr(config, "partial_rotary_factor", 0.5)
         rotary_dim = int(self.head_dim * partial_rotary_factor)
+
+        # GLM-5.2 IndexShare: layers tagged "shared" reuse the previous "full"
+        # layer's top-k and ship no indexer weights. Dense MLA discards indexer
+        # output anyway, so just skip building the module on shared layers.
+        indexer_types = getattr(config, "indexer_types", None)
+        has_indexer = indexer_types is None or indexer_types[layer_id] == "full"
 
         self.self_attn = Glm5Attention(
             hidden_size=config.hidden_size,
@@ -609,6 +621,7 @@ class Glm5DecoderLayer(nnx.Module):
             dtype=dtype,
             mesh=mesh,
             use_absorbed=True,
+            has_indexer=has_indexer,
         )
 
         first_k_dense_replace = getattr(config, "first_k_dense_replace", 0)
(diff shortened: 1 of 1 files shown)


python/sgl_jax/srt/models/glm5_moe.py
```diff
@@ -407,7 +411,8 @@ def __call__(
         q, _ = self.q_b_proj(q_compressed)
         q = q.reshape(-1, self.num_heads, self.qk_head_dim)
 
-        _ = self.indexer(hidden_states, q_compressed, positions, self.rotary_emb)
+        if self.indexer is not None:
```
alexnails, line 414: if has_indexer? (I forget if there are any quirks between these two types of checks)

python/sgl_jax/srt/models/glm5_moe.py
```diff
@@ -924,13 +938,19 @@ def _create_glm5_weight_mappings(self, model_config: ModelConfig) -> dict:
...
         return mappings
 
     def _create_moe_layer_mappings(
-        self, layer_idx: int, target_idx: int, is_mlp_layer: bool, is_static_quant: bool = False
+        self,
+        layer_idx: int,
+        target_idx: int,
+        is_mlp_layer: bool,
+        is_static_quant: bool = False,
+        has_indexer: bool = True,
```
alexnails, line 953: do we want this to be default true?

alexnails (approved):
worth making has_indexer part of model_config? (approving since I can't imagine this goes another round of review, please address feedback)

python/sgl_jax/srt/models/glm5_moe.py
zorrofox: No quirk — `is not None` is an identity check so it never touches `__bool__` (which is the thing nnx Modules can be weird about). It's also the same pattern as `if self.kv_b_proj is None: return` in `post_load_weights` a few lines up. Could also store `self.has_indexer = has_indexer` and branch on that (matching `self.use_absorbed`/`self.use_fused`), but it'd be redundant with `self.indexer` already being `None` on shared layers — let me know if you'd still prefer the explicit bool.

python/sgl_jax/srt/models/glm5_moe.py
zorrofox: Kept it defaulted for the same reason `is_static_quant: bool = False` is — `glm5_moe_nextn.py` borrows this method via a `SimpleNamespace` shim for the MTP layer (layer 78, which does have indexer weights), and any out-of-tree caller that hasn't been updated would get the 5.1 behavior. The only in-tree caller (`_create_glm5_weight_mappings`) always passes it explicitly, so the default is just the safe fallback. Can drop it if you'd rather force explicit.

zorrofox:
`has_indexer` is per-layer (derived from `config.indexer_types[layer_id]`), so it can't be a single `model_config` attribute. The list itself (`indexer_types`) is already on `hf_config` straight from `config.json`, and `Glm5DecoderLayer.__init__` reads it from there — so it's already config-driven, just resolved per-layer at construction time. Happy to move the resolution into `patch_model_config` (write a `_sgl_has_indexer: list[bool]` onto `hf_config`) if you'd prefer the derivation to live there instead of in the layer ctor, but it'd be the same shape.

alexnails:
@zorrofox any concern on merging?

zorrofox:
> @zorrofox any concern on merging?

No. Happy to merge it, thanks!

Issue #1280, opened 2026-06-02
[Roadmap] HiCache Multi-Level KV Cache


cjx0709:
## Motivation

The goal is to add a three-level KV cache hierarchy (L1=HBM / L2=host pinned / L3=persistent storage) to sgl-jax, enabling prefix reuse across requests and across restarts. This targets improved cache hit rate and reduced redundant computation for long-context / multi-tenant workloads.

## Roadmap Semantics

Stages stay unchecked until merged to `main`. Linked sub-issues define scope, design, and acceptance criteria.

### Stage 0: Host pool foundation

- [x] `HostKVPool` core — `alloc` / `free` / `get` / `put` with per-block `jax.device_put`

### Stage 1: UnifiedRadixCache + FullComponent port from sglang

- [x] `kv_cache_builder` registry — support creating `UnifiedRadixCache` via config
- [x] `TreeComponent` ABC definition (3 hooks: `write_through` / `write_back` / `prefetch`)
- [x] `FullComponent` implementation (port from sglang, hooks wired to stub/no-op)
- [x] `UnifiedRadixCache` port (replace existing `RadixCache` — trie, ref counting, eviction)

### Stage 2: HiCacheController + async pipeline

- [ ] `HiCacheController` skeleton — `write` / `load` / `evict_callback`, wired to `HostKVPool` + device `KVCache`
- [ ] D2H/H2D async execution (ThreadPool vs single-thread+future — decided at PR review and the scheduler status)
- [ ] Step-level overlap integration (D2H for step N-1 runs during forward of step N)

### Stage 3: End-to-end validation

- [ ] Wire `FullComponent` hooks to `HiCacheController` (replace stubs)
- [ ] KL equivalence test with `--hicache-storage none` on Qwen3-8B
- [ ] Forward latency regression < 2%
- [ ] Decision point: `HostKVPool` per-block `device_put` vs pre-allocated pool + async ThreadPool vs single-thread future

### Stage 4: L3 persistent storage

- [ ] `HiCacheStorage` ABC (`get` / `set` / `exists` / `delete`)
- [ ] `LocalFileStorage` — page-granular files, parent-hash chain key scheme, atomic rename
- [ ] Controller integration with L3 (`write_back` → `storage.set`; prefetch L3 path → `storage.get`)
- [ ] Restart-and-reload validation (write → restart → `match_prefix` hit)

### Stage 5: Multi-component + tuning

[...]

PR #1396, opened 2026-06-24
feat(mem_cache): HiCache Stage 0 — LRUHostKVPool host pinned KV pool

### LINKED ISSUE #1310: [RFC] HiCache stage 0: HostKVPool — host pinned memory pool
## Summary

Implement `LRUHostKVPool`, the host pinned memory pool for the HiCache L2 layer. It **does not introduce a new abstraction** — it reuses the shared `HostKVPool` ABC from PD disaggregation (`mem_cache/host_kv_pool.py`, #1368), sitting alongside `QueueHostKVPool` (used by PD) as the second concrete implementation.

> **Interface alignment** — This design has been aligned with PD disaggregation (#1368, chosen as the host-pool baseline) and the future tpu-raiden migration. See the decision record `docs/design/hicache_pd_raiden_alignment.md` (D1–D6). The original `put`/`get`/`slot_table` + self-managed `host_sharding` design is **voided** in favor of implementing the existing shared ABC plus a pair of symmetric, `buffer_id`-addressed transfer primitives. #1368 rewrote the ABC to `reserve()/release(buffer_id)/copy_from_device(layers, buffer_id)/array_pytree`; HiCache extends on top of that form.

It provides the upper layer (controller, tree cache) with:

- slot-denominated reserve/release (`reserve()` / `release(buffer_id)`) with LRU + lock_ref retention; each slot's capacity (page count, layer count, per-layer shape) is fixed at construction
- `buffer_id`-addressed symmetric D2H/H2D primitives (`copy_into` / `copy_to_device`)
- capacity queries (`available_size` / `total_size`) for the tree cache to decide host-resident eviction

**Switchability contract (D4):** the upper layer interacts with the pool only via `int buffer_id` + `reserve/release/available_size` + `copy_into/copy_to_device`. **No host `jax.Array` crosses the HiCache upper-layer boundary.** This guarantees that swapping the storage backend from JAX-native to a tpu-raiden engine later requires zero changes to the controller / hooks / tree cache.

## Position in the roadmap

1. **Host pool foundation** *(this issue)*
2. UnifiedRadixCache + FullComponent port
3. HiCacheController + async pipeline
4. End-to-end integration and validation
5. L3 persistent storage
6. Multi-component + tuning

Prerequisite: the shared `HostKVPool` ABC + `QueueHostKVPool` from PD disaggregation (#1368, in progress, chosen as baseline). Can otherwise be developed in parallel with Stage 1.

## Proposed structure changes

- `sgl_jax/srt/mem_cache/host_kv_pool.py` — **modify**: add `copy_into` / `copy_to_device` to the `HostKVPool` ABC (default `raise NotImplementedError`); add `LRUHostKVPool(HostKVPool)`
- `sgl_jax/test/mem_cache/test_host_kv_pool.py` — **new**: `LRUHostKVPool` unit tests (existing `QueueHostKVPool` tests untouched)

### 1. Relationship to the shared ABC

The `HostKVPool` ABC already exists (landed by PD #1368). Current methods:

```python
@dataclass
class StagedData:
    buffer_id: int
    array_pytree: list[jax.Array]   # one host array per layer

class HostKVPool(abc.ABC):
    def reserve(self) -> int | None: ...                          # take a slot id; None when full
    def release(self, buffer_id: int) -> None: ...                # return a slot
    def copy_from_device(self, layers: list[jax.Array],
                         buffer_id: int) -> StagedData: ...        # D2H into an already-reserved slot, returns array_pytree (PD one-shot borrow)
    def available_size(self) -> int: ...
    def total_size(self) -> int: ...
```

`reserve()` is **slot-denominated with entry size fixed at construction** (D6): slot capacity is decided by the constructor params (`max_padded_pages`, `layer_num`, `per_layer_shape`, `dtype`) — PD puts a whole request's padded pages in one slot (few large slots); HiCache puts one page per slot (many small slots). Both are construction-fixed entries — no conflict at the ABC layer, and **no block invariant is forced on PD**.

`copy_from_device(layers, buffer_id)` `device_put`s each layer's device array into pinned host and returns `StagedData.array_pytree` (per-layer host arrays) to the caller — PD hands it to `register_pull` (one-shot borrow, non-retaining). HiCache **does not** use this method (it would hand a `jax.Array` across the boundary); it uses the retaining `copy_into` below.

This issue adds **one pair** of new primitives to the ABC (D2), using "base default + subclass override" (D3), so `QueueHostKVPool` and the PD path get **zero changes**:

```python
class HostKVPool(abc.ABC):
    # ... existing methods unchanged ...

    def copy_into(self, device_indices: list[int], host_buffer_ids: list[int]) -> None:
        """D2H: copy device-pool pages at device_indices into ALREADY-RESERVED
        host slots host_buffer_ids; data is RETAINED in the slot and NO jax.Array
        is returned. Distinct from copy_from_device (which hands array_pytree back
        to the caller, non-retaining); HiCache needs "give a tree node a stable
        slot first, fill it later, keep data in the slot". Default not
        implemented; only LRUHostKVPool overrides."""
        raise NotImplementedError

    def copy_to_device(self, host_buffer_ids: list[int], device_indices: list[int]) -> None:
        """H2D: scatter host slots host_buffer_ids back into device-pool pages at
        device_indices. The existing ABC lacks this primitive (PD needs no local
        H2D). Default not implemented; only LRUHostKVPool overrides."""
        raise NotImplementedError
```

The signatures are **pure integer index addressing** (`device_indices ↔ host_buffer_ids`, no `jax.Array`), deliberately isomorphic to raiden `d2h(src_offsets, dst_offsets)` / `h2d(src_offsets, dst_offsets)`. This is where the switchability contract lands on the transfer surface:

| Direction | HiCache primitive | raiden equivalent |
|---|---|---|
| D2H | `copy_into(device_indices, host_buffer_ids)` | `d2h(src_offsets, dst_offsets)` |
| H2D | `copy_to_device(host_buffer_ids, device_indices)` | `h2d(src_offsets, dst_offsets)` |

### 2. D2H / H2D mechanism (JAX-native implementation)

`LRUHostKVPool` holds a device-pool reference at construction (mirroring raiden's `kv_caches` registration) and uses a `host_sharding` with the same partition spec as the device pool but `memory_kind='pinned_host'` (reusing the ABC module's `_make_host_sharding`):

```python
device_sharding = NamedSharding(mesh, P("data", None, "tensor", None, None))
host_sharding   = NamedSharding(mesh, P("data", None, "tensor", None, None),
                                memory_kind='pinned_host')
```

`copy_into` / `copy_to_device` move data between the two shardings via `jax.device_put`. Since the partition specs match, JAX maps shards chip-to-chip automatically. The D2H transfer format (all-layers-packed vs per-layer direct) is **localized inside this implementation (D5)** and does not enter the public interface.

### 3. Slot content format

Each slot stores all layers of a KV page:

| KV Pool type | Slot shape |
|---|---|
| `MHATokenToKVPool` | `[num_layers, page_size, kv_heads*2//packing, packing, head_dim]` |
| `MLATokenToKVPool` | `[num_layers, page_size//kv_packing, kv_packing, kv_dim]` |

This format is an internal detail of the implementation; it does not affect the upper-layer interface.

### 4. Interface (LRUHostKVPool)

```python
class LRUHostKVPool(HostKVPool):
    """HiCache L2 LRU host pool. Unlike QueueHostKVPool (PD one-shot borrow),
    this carries LRU + lock_ref retention: a slot referenced by a tree node is
    not evicted."""

    def __init__(self, device_pool, pool_size, page_size, layer_num,
                 per_layer_shape, dtype, mesh, partition_spec, *,
                 pool_name="hicache"): ...

    # --- shared ABC ---
    def reserve(self): ...                  # one slot per page; None when full
    def release(self, buffer_id): ...
    def copy_into(self, device_indices, host_buffer_ids): ...
    def copy_to_device(self, host_buffer_ids, device_indices): ...
    def available_size(self): ...
    def total_size(self): ...
    # copy_from_device(layers, buffer_id): thin impl to satisfy the ABC;
    # HiCache upper layer does not use it.

    # --- LRU machinery (private, not on the ABC) ---
    def inc_lock_ref(self, buffer_id): ...
    def dec_lock_ref(self, buffer_id): ...
```

### 5. Why reserve does not auto-evict

Radix tree nodes' `host_value` points to a pool `buffer_id`. If the pool self-evicted, tree nodes would read stale data on the next prefetch. Eviction is therefore driven by the tree cache (`FullComponent.drive_host_eviction`): traverse its own `host_lru_list`, **atomically** call `pool.release(buffer_id)` + set `node.host_value = None`. When full, `reserve()` returns None and the controller asks the tree cache to evict before retrying. `lock_ref` protects slots referenced by in-flight transfers/prefetch from being selected.

### 6. Switchability: raiden migration point

This stage **only guarantees switchability; it writes no raiden adapter code (D4)**. When the switch happens, add `RaidenHostKVPool(HostKVPool)` in one place:

| Now (JAX-native) | After raiden |
|---|---|
| `reserve/release` = pinned-host slot free-list + LRU | = raiden host pool block-offset free-list + LRU |
| `copy_into(device_indices, host_buffer_ids)` | → `raiden.d2h(src_offsets=device_indices, dst_offsets=host_buffer_ids)` |
| `copy_to_device(host_buffer_ids, device_indices)` | → `raiden.h2d(src_offsets=host_buffer_ids, dst_offsets=device_indices)` |
| ctor holds device-pool ref | ctor points to the raiden engine instance |

Cost: ~1 class, 100–150 lines. **Controller / hooks / tree cache: zero changes.** Shared premise: the raiden engine is a process-level resource; PD and HiCache point to the **same engine instance → share one host pool**.

## Acceptance criteria

- ABC extension does not break PD: after adding `copy_into/copy_to_device`, `QueueHostKVPool` instantiates and existing PD tests stay green; both methods `raise NotImplementedError` on `QueueHostKVPool`
- reserve/release: reserve returns non-overlapping buffer_ids; release makes them re-reservable; returns None when full
- lock_ref: slots with lock_ref>0 are not selected by release; eligible again after dec to 0
- copy_into round-trip: `reserve` → `copy_into([d],[b])` → `copy_to_device([b],[d2])` → device pool d2 allclose with original d
- Does not expose jax.Array: upper-layer paths (reserve/release/copy_into/copy_to_device) signatures and returns contain no host `jax.Array`
- available_size: decreases after reserve, recovers after release

## Out of scope

- LRU eviction **policy** (which node to pick) → Stage 1 tree cache `host_lru_lists` (this pool only provides `release` + `lock_ref` mechanism)
- D2H/H2D transfer **scheduling** (threadpool, async) → Stage 2 HiCacheController
- `RaidenHostKVPool` actual implementation → Phase B (when raiden is ready)
- Optionally refactoring PD's `copy_from_device` to reuse the retaining `copy_into` (non-blocking)

cjx0709:
## Summary

HiCache Stage 0 (host pool foundation) from the roadmap #1280. Adds `LRUHostKVPool`, the L2 host-pinned KV pool, as a second concrete implementation of the shared `HostKVPool` ABC (introduced for PD disaggregation in #1368) — no new abstraction.

- **`LRUHostKVPool`** — slot-denominated `reserve`/`release` with LRU + `lock_ref` retention, page-addressed `alloc`/`free`, and capacity queries (`available_size`/`total_size`).
- **Symmetric `buffer_id`-addressed D2H/H2D primitives** — `copy_into`/`copy_to_device` plus a two-phase async transfer path (`stage_backup`/`flush_backup`, `stage_load`/`flush_load`). No host `jax.Array` crosses the upper-layer boundary, so the backend stays swappable (D4 contract).
- **Page-bucket batching** — transfers pad to gather buckets so the Pallas write kernel compiles once per bucket.
- **Shared `write_kv_layer`** — extracted from the PD decode path into `memory_pool.py` and reused by both PD and HiCache (single KV-write kernel wrapper).
- **Page-id boundary guards** — unallocated host ids and out-of-range device pages are rejected at every transfer/lock entry point.

The TPU-only fused-KV Pallas kernel is kept in production; CPU unit tests use a pure-JAX shim via `conftest.py`, so the suite runs on CPU CI.

## Test plan

- [x] `python/sgl_jax/test/mem_cache/test_host_kv_pool.py` — free-list, lock_ref, boundary invariants, D2H/H2D round-trip (bit-exact), page_size>1, bucket padding, two-phase stage/drain errors, precompile. Wired into `unit-test-cpu`.
- [x] Full `mem_cache/` suite green on CPU.

Closes #1310. Part of #1280 (Stage 0).

### DIFF
--- a/python/sgl_jax/srt/mem_cache/host_kv_pool.py
+++ b/python/sgl_jax/srt/mem_cache/host_kv_pool.py
@@ -620,22 +620,48 @@ class LRUHostKVPool(HostKVPool):
             staged_pages.append(packed)
 
         n = len(host_buffer_ids)
-        # Bucket the page count so write_kv_layer compiles once per bucket, not
-        # per distinct n. Pad pages by duplicating page 0; the padding tokens get
-        # loc=-1, which the kernel skips, so the duplicate data is never written.
-        n_b = self._pad_to_page_bucket(n)
-        sel = list(range(n)) + [0] * (n_b - n)
+        dp = self._device_pool
+        dp_size = int(dp.dp_size)
+        num_dev_pages = int(dp.kv_buffer[0].shape[0])
+        pages_per_shard = num_dev_pages // dp_size
+
+        # The KV buffer's page axis is sharded contiguously across the data axis:
+        # shard d owns global pages [d*pages_per_shard, (d+1)*pages_per_shard).
+        # write_kv_layer's shard_map splits loc/data into dp_size equal contiguous
+        # chunks and indexes each shard's LOCAL slice -- so a global page must land
+        # in its owning shard's chunk, addressed as a shard-local offset (passing
+        # global ids writes to the wrong shard under dp>1). Give every shard a
+        # fixed per_shard bucket (pad_to_page_bucket(n) == the worst case where all
+        # n pages fall on one shard) so the compiled shape depends only on
+        # (n bucket, dp_size). For dp_size==1 this collapses to one global==local
+        # chunk -- identical to the single-shard layout.
+        per_shard = self._pad_to_page_bucket(n)
+        n_b = per_shard * dp_size
+        # Bucket each transfer page into its owning shard's slot list; padding
+        # slots keep staged index 0 (data ignored) and loc -1 (kernel skips it).
+        shard_sel = [[0] * per_shard for _ in range(dp_size)]
+        shard_loc = [[-1] * per_shard for _ in range(dp_size)]
+        fill_pos = [0] * dp_size
+        for k, gp in enumerate(int(p) for p in device_indices):
+            d = gp // pages_per_shard
+            pos = fill_pos[d]
+            shard_sel[d][pos] = k
+            shard_loc[d][pos] = gp - d * pages_per_shard
+            fill_pos[d] = pos + 1
+        sel = [k for chunk in shard_sel for k in chunk]
         stack = jnp.stack(
             [staged_pages[i] for i in sel], axis=0
         )  # (n_b, layer_num, *per_layer_shape)
 
-        # Expand global page ids -> absolute device token slots (page*PS + offset),
-        # then pad to n_b pages with -1 (skipped by the kernel).
-        dev_pages = np.asarray(device_indices, dtype=np.int64)
-        loc_np = (dev_pages[:, None] * PS + np.arange(PS, dtype=np.int64)).reshape(-1)
-        if n_b > n:
-            loc_np = np.concatenate([loc_np, -np.ones((n_b - n) * PS, dtype=loc_np.dtype)])
-        dp = self._device_pool
+        # Expand shard-local page ids -> shard-local token slots (page*PS + offset),
+        # -1 for padding. The per-shard chunk order matches the loc/data data-axis
+        # split so each shard receives exactly the pages it owns.
+        loc_pages = np.asarray([lp for chunk in shard_loc for lp in chunk], dtype=np.int64)
+        loc_np = np.where(
+            loc_pages[:, None] < 0,
+            -1,
+            loc_pages[:, None] * PS + np.arange(PS, dtype=np.int64),
+        ).reshape(-1)
         loc = jax.device_put(
             jnp.asarray(loc_np, dtype=jnp.int32),
             NamedSharding(dp.mesh, PartitionSpec(dp.attention_data_partition_axis)),
--- a/python/sgl_jax/srt/mem_cache/host_kv_pool.py
+++ b/python/sgl_jax/srt/mem_cache/host_kv_pool.py
@@ -22,6 +22,7 @@ from dataclasses import dataclass
 from typing import Any
 
 import jax
+import jax.numpy as jnp
 import numpy as np
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 
@@ -108,6 +109,36 @@ class HostKVPool(abc.ABC):
     @abc.abstractmethod
     def total_size(self) -> int:
         """Total entries in the pool (free + in-use)."""
+
+    # ------------------------------------------------------------------
+    # Symmetric buffer_id-addressed transfer primitives (HiCache).
+    #
+    # Defaulted (NOT @abc.abstractmethod) so QueueHostKVPool stays concrete
+    # and the PD path is untouched: only the retaining LRUHostKVPool overrides
+    # these. Signatures are pure-integer index pairs, deliberately isomorphic
+    # to a future raiden ``d2h(src_offsets, dst_offsets)`` / ``h2d(...)`` so
+    # the storage backend can be swapped with zero upper-layer changes.
+    # ------------------------------------------------------------------
+
+    def copy_into(self, device_indices: list[int], host_buffer_ids: list[int]) -> None:
+        """Retaining D2H: copy device-pool page(s) into RESERVED slot(s).
+
+        ``device_indices[i]`` pairs with ``host_buffer_ids[i]``. Unlike
+        :meth:`copy_from_device` (which hands the staged ``array_pytree`` back
+        to the caller and retains nothing), the data STAYS in the slot until
+        :meth:`release`. No ``jax.Array`` crosses the boundary. Only
+        LRUHostKVPool overrides this.
+        """
+        raise NotImplementedError
+
+    def copy_to_device(self, host_buffer_ids: list[int], device_indices: list[int]) -> None:
+        """H2D: scatter slot(s) back into the device pool by index.
+
+        ``host_buffer_ids[i]`` pairs with ``device_indices[i]``. The existing
+        ABC has no local H2D (PD does not need it). Only LRUHostKVPool
+        overrides this.
+        """
+        raise NotImplementedError
 
 
 class QueueHostKVPool(HostKVPool):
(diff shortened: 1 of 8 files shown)


python/sgl_jax/test/mem_cache/test_host_kv_pool.py
```diff
@@ -0,0 +1,477 @@
+"""Unit tests for :class:`LRUHostKVPool` (HiCache Stage 0).
+
+CPU-runnable: ``_make_host_sharding`` falls back to default sharding when
+``pinned_host`` is unsupported, so the round-trip semantics are exercised
+without a TPU. The pinned-host in-place invariant is covered separately by
+``test_single_chip_host_pool`` on a real pod.
```
MokusMokun: It seems `test_single_chip_host_pool` does not exist

python/sgl_jax/srt/mem_cache/host_kv_pool.py
```diff
@@ -270,6 +264,525 @@ def _release(self, buffer_id: int) -> None:
...
+            self._require_allocated(buffer_id)
+            with self._pending_lock:
+                packed = self._pending_gather.pop(buffer_id, None)
+            if packed is None:
+                raise RuntimeError(
+                    f"flush_backup with no staged gather for buffer_id={buffer_id}; "
+                    f"stage_backup() must run (synchronously) first"
+                )
+            host_packed = jax.device_put(packed, self._host_sharding)
+            self._slots[buffer_id] = host_packed
```
MokusMokun: This write is outside `_lock`. If `release` runs between the `_require_allocated` check at `:511` and this line, the id gets freed (`:389-392`) and this write resurrects stale data into a now-reusable slot. Same shape at `:495` and `:736`. Would it be cleaner to have transfer primitives hold `lock_ref` for their duration, so "in-transfer ⇒ not freeable" is a pool invariant?

python/sgl_jax/srt/mem_cache/host_kv_pool.py
```diff
@@ -270,6 +264,525 @@ def _release(self, buffer_id: int) -> None:
...
+        for layer in range(self._layer_num):
+            # One slice for the whole batch's layer (page axis kept), folded to
+            # tokens — no per-page gather/concatenate.
+            layer_kv = stack[:, layer]  # (n_b, page_size, *per_head_tail)
+            layer_kv = jax.lax.reshape(
+                layer_kv,
+                (total_tokens, 1) + tuple(layer_kv.shape[2:]),
+                out_sharding=folded_sharding,
+            )
+            dp.kv_buffer[layer] = write_kv_layer(
```
MokusMokun, line 740: This donates the live KV buffer in-place (`memory_pool.py:858 donate_argnames`), and `stage_backup:487` reshards the page axis out of a `P("data",...)` buffer. 

But the current tests run on a single CPU device at `dp_size=1`, with the real Pallas kernel monkeypatched out (`conftest.py:50`) and `pinned_host` falling back to default sharding — so the real kernel, pinned host, and `dp>1` sharding are all unexercised, which is where a donating in-place write is most likely to bite. 

Could we add a test in `unit-test-tpu-v6e-4`, run at both `tp=4/dp=1` and `tp=2/dp=2`? That covers the real kernel, pinned host, and the dp-sharded page-axis gather/scatter on hardware.

python/sgl_jax/srt/mem_cache/host_kv_pool.py
```diff
@@ -270,6 +264,525 @@ def _release(self, buffer_id: int) -> None:
...
+    def available_size(self) -> int:
+        """Free page slots (HiCache control plane counts in pages)."""
+        with self._lock:
+            return len(self._free_ids)
+
+    def total_size(self) -> int:
+        """Total page slots in the pool (free + in-use)."""
+        return self._pool_size
+
+    def copy_from_device(self, layers: list[jax.Array], buffer_id: int) -> StagedData:
```
MokusMokun: HiCache uses `copy_into` — this override is dead code with no round-trip test coverage. 

Could we make `copy_from_device` default-`raise` on the ABC (same as `copy_into`/`copy_to_device` at `:100-108`) and drop this? 
Removes the `:730→:736` instance of the race above too. `QueueHostKVPool` still overrides it, PD untouched.

python/sgl_jax/srt/mem_cache/host_kv_pool.py
```diff
@@ -270,6 +264,525 @@ def _release(self, buffer_id: int) -> None:
...
+            raise ValueError(f"buffer_id={buffer_id} outside pool range [0, {self._pool_size})")
+        if not self._allocated[buffer_id]:
+            raise RuntimeError(f"buffer_id={buffer_id} is not allocated (in free-list)")
+
+    def _require_device_pages(self, device_indices) -> None:
+        # Reject out-of-range device page ids: a negative id would wrap (JAX
+        # negative indexing on D2H gather) or expand to negative loc on H2D
+        # (only -1 is treated as padding by the kernel; other negatives become
+        # real DMA targets and corrupt pages).
+        n_dev = int(self._device_pool.size) // self._page_size
```
MokusMokun: The allocator hands out 1-based page ids with 0 as sentinel (`allocator.py:177`), and the buffer has an extra page for it (`memory_pool.py:440`). So `size=16, page_size=4` → allocator returns `[1,2,3,4]` but `n_dev=4` rejects page 4. The test at `:164` bakes this in. Could the bound use `kv_buffer[0].shape[0]`?

python/sgl_jax/test/mem_cache/test_host_kv_pool.py
cjx0709: Fixed — removed the stale reference; the docstring now points to `test_host_kv_pool_tpu.py`. (d31681bf9)

python/sgl_jax/srt/mem_cache/host_kv_pool.py
cjx0709: Fixed in d31681bf9. `stage_backup`/`flush_backup`/`stage_load` now publish into `_pending_*`/`_slots` under `_lock` with an allocation re-check (lock order `_lock -> _pending_*`), so a `release`/`free` racing the entry check can no longer resurrect data into a reusable slot.

I went with re-check-under-lock rather than holding `lock_ref` for the transfer duration: the async two-phase split (stage on the KV thread, flush on the worker) means a `lock_ref` leak on any error path would permanently wedge a slot, whereas the re-check is self-contained and has no leak surface.

MokusMokun: Thanks for re-check, but I think there still exists ABA window: `_allocated[buffer_id]` only proves the id is allocated at publish time, not that it is the same allocation instance that passed the earlier transfer check.

Could we track a per-slot generation/epoch and verify `(buffer_id, generation)` before publishing? That should reject stale publishes without needing to hold `lock_ref` across the async transfer.

See this attatched diagram for details
Image

cjx0709: You're right — `_allocated[buffer_id]` only proves the id is allocated at publish time, not that it's the same allocation instance that passed the entry check. Fixed in 046c7656d with the per-slot epoch you suggested.

- `_generation[]` bumped on every `alloc`/`reserve`.
- Each transfer captures the slot's epoch at entry (under `_lock`) and tags its `_pending_*` entry with it.
- Before publishing (`_pending_gather`, `_pending_load`, `_slots`, and the `flush_load` scatter), it re-verifies `_allocated[id] and _generation[id] == captured_epoch` under `_lock`. A `free()+realloc()` that reused the slot mid-transfer bumps the epoch, so the stale publish is dropped — `flush_load` drops a stale page by leaving its `loc=-1` so the kernel skips it. No `lock_ref` held across the async transfer.

Regression tests added in `test_host_kv_pool.py` (`TestLRUHostKVPoolABA`): the real ABA driven by a free()+realloc() injected mid-gather, plus stale-epoch rejection for `flush_backup`/`flush_load`.

MokusMokun: The per-slot generation fix looks good to me — it closes the stale-data ABA corruption. A few follow-ups before Stage 1 wires this into the real controller; none are data-corruption issues:

- **Reclaimed transfer turns into a flush error.** In the real `free()+realloc()` path, `free()` calls `_drop_pending()`, so the later `flush_backup()` / `flush_load()` finds no pending entry and raises `"no staged entry"` — safe, but a normal reclaim can still crash the async worker unless the caller catches it. Could `stage_*` return the still-fresh ids, or could `flush_*` distinguish "this transfer was reclaimed" from "flush called without stage"?

- **`flush_load` relies on scheduler-side serialization.** It snapshots `cur_gen` under `_lock`, but the `write_kv_layer` scatter runs later. This is fine if `flush_load` and host-slot free/alloc are serialized on the KV-owning thread. Can we state that invariant explicitly? If not, the pool would need a final generation check right before scatter, or an in-flight guard.

- **Test fidelity (minor).** The two flush-side ABA tests hand-edit `_pending_*` to an old generation. That exercises the guard logic, but it is not the real reclaim path — real `free()` removes the pending entry first. Once the stale-transfer behavior is decided, it would be worth aligning those tests with the actual path.

cjx0709: Thanks — agreed these are all lifecycle/orchestration concerns rather than data-corruption bugs, and they belong to the controller layer (Stage 1), not this low-level pool.

MokusMokun:
LGTM

MokusMokun (approved):

PR #1398, opened 2026-06-25
feat: Add Stage 1 Qwen-VL multimodal preprocessing path

zhengkezhou1:
## Motivation

This PR adds the Stage 1 multimodal preprocessing path for Qwen-VL models in the regular SGL-JAX serving flow.

The goal is to support tokenizer-time Qwen-VL image preprocessing without requiring the legacy `--multimodal` serving path, while keeping non-multimodal models on the existing text tokenizer path.

The following examples use `Qwen/Qwen2.5-VL-3B-Instruct` with prompt `Describe the image.`. Offsets are inclusive `[start, end]` spans in the expanded `input_ids`.

| Image size | `image_grid_thw` | `item.offsets` | Image token count | `input_ids` length | `pixel_values.shape` | `mrope_positions.shape` |
|---|---:|---:|---:|---:|---:|---:|
| 224x224 | `[(1, 16, 16)]` | `[(15, 78)]` | 64 | 89 | `(256, 1176)` | `(3, 89)` |
| 448x448 | `[(1, 32, 32)]` | `[(15, 270)]` | 256 | 281 | `(1024, 1176)` | `(3, 281)` |

This shows that the image placeholder span is dynamic and derived from `image_grid_thw // spatial_merge_size^2`. The processor writes this span to `MultimodalDataItem.offsets` so scheduler/forward can identify the image token region directly.

## Modifications

- Added a base multimodal processor interface and a Qwen-VL processor implementation.
- Updated tokenizer initialization to load a multimodal processor only when the model is multimodal and a registered processor exists.
- Added fallback behavior so multimodal model configs without a registered processor still use the regular tokenizer.
- Wired tokenizer-time multimodal processing to produce request `input_ids`, image features, image grid metadata, and M-RoPE metadata.
- Passed `MultimodalInputs` through tokenizer and scheduler request plumbing.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/multimodal/processors/qwen_vl.py
+++ b/python/sgl_jax/srt/multimodal/processors/qwen_vl.py
@@ -1,3 +1,5 @@
+import asyncio
+
 import numpy as np
 
 from sgl_jax.srt.multimodal.common.modality_enum import (
@@ -29,9 +31,7 @@ class QwenVLProcessor(BaseMultimodalProcessor):
                 "Please provide text input instead."
             )
 
-        images = [self.load_image(item) for item in self.normalize_data(image_data)]
-        if not images:
-            return None
+        images = await self._load_images_async(image_data)
 
         processor_output = self.processor(
             text=[input_text],
@@ -112,6 +112,11 @@ class QwenVLProcessor(BaseMultimodalProcessor):
             items.append(item)
             offset += count
         return items
+
+    async def _load_images_async(self, image_data):
+        return await asyncio.gather(
+            *(self.load_image_async(item) for item in self.normalize_data(image_data))
+        )
 
     @staticmethod
     def _compute_image_offsets(input_ids, grids, image_token_id, spatial_merge_size):
--- a/python/sgl_jax/srt/multimodal/processors/qwen_vl.py
+++ b/python/sgl_jax/srt/multimodal/processors/qwen_vl.py
@@ -75,7 +75,7 @@ class QwenVLProcessor(BaseMultimodalProcessor):
             mm_items=mm_items,
             input_ids=input_ids,
             im_start_id=self.hf_config.vision_start_token_id,
-            im_end_id=getattr(self.hf_config, "vision_end_token_id", None),
+            im_end_id=self.hf_config.vision_end_token_id,
             im_token_id=self.hf_config.image_token_id,
             mrope_positions=mrope_positions,
             mrope_position_delta=mrope_position_delta,
--- a/python/sgl_jax/srt/multimodal/processors/qwen_vl.py
+++ b/python/sgl_jax/srt/multimodal/processors/qwen_vl.py
@@ -33,7 +33,7 @@ class QwenVLProcessor(BaseMultimodalProcessor):
             text=[input_text],
             images=images,
             padding=True,
-            return_tensors="np",
+            return_tensors="pt",
         )
 
         input_ids_array = self._to_numpy(processor_output.get("input_ids"))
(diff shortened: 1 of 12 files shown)


JamesBrianD:
Can we apply the radix-cache guard to the in-model VLM path as well? Qwen-VL now enters via `model_config.is_multimodal` without `--multimodal`, so radix cache stays enabled unless users remember to pass `--disable-radix-cache`. If Stage 1 assumes radix is off for VLM, we should enforce it or fail fast.

python/sgl_jax/srt/server_args.py
```diff
@@ -336,6 +337,8 @@ def __post_init__(self):
             self.device_indexes = None
         if self.multimodal:
             self.model_path = download_from_hf(self.model_path, allow_patterns=None)
+            if self.limit_mm_data_per_request is None:
+                self.limit_mm_data_per_request = {"image": 16}
```
JamesBrianD, line 352: Can we also set the default `limit_mm_data_per_request` for the in-model VLM path? Right now the default `{ "image": 16 }` is only installed under `server_args.multimodal`, but Qwen-VL enters through `model_config.is_multimodal` without `--multimodal`. In that launch mode `_validate_mm_limits()` becomes a no-op unless users manually pass the flag.

python/sgl_jax/srt/multimodal/processors/qwen_vl.py
```diff
@@ -0,0 +1,152 @@
...
+
+        if isinstance(input_text, list):
+            input_text = self.processor.tokenizer.decode(input_text)
+
+        processor_output = self.processor(
+            text=[input_text],
+            images=images,
+            padding=True,
+            return_tensors="np",
+        )
```
JamesBrianD, line 41: Current issue: if `input_text` is already a token-id list, this decodes it back to text and then lets the HF processor tokenize it again; the returned processor `input_ids` later replace the caller-provided ids. That can drift around non-canonical tokens, special tokens, or placeholders.

Upstream SGLang does decode internally in some loading paths, but its default `SGLANG_MM_AVOID_RETOKENIZE` path preserves the caller-provided token ids and expands only the image placeholders before returning final `input_ids`; it only falls back to decode+retokenize if that reconstruction fails. Can we either add the same avoid-retokenize path here, or explicitly reject multimodal `input_ids` for Stage 1?

python/sgl_jax/srt/multimodal/processors/qwen_vl.py
```diff
@@ -0,0 +1,152 @@
...
+    )
+
+    async def process_mm_data_async(
+        self,
+        image_data,
+        input_text,
+        request_obj,
+        **kwargs,
+    ):
+        images = [self.load_image(item) for item in self.normalize_data(image_data)]
```
JamesBrianD: This runs URL/file/base64 image loading and PIL decode synchronously inside `process_mm_data_async()`, so a slow image fetch can block the TokenizerManager event loop. Upstream SGLang still uses blocking loaders under the hood, but it submits each multimodal load to an IO `ThreadPoolExecutor` and awaits it with `asyncio.wrap_future()`. Can we follow that pattern here, at least for the image loading path?

python/sgl_jax/srt/multimodal/processors/qwen_vl.py
zhengkezhou1: Fixed by explicitly rejecting multimodal input_ids for Qwen-VL for now, instead of silently decoding and retokenizing them.

python/sgl_jax/srt/multimodal/processors/qwen_vl.py
zhengkezhou1: Fixed. Image loading now runs through an IO ThreadPoolExecutor and is awaited from process_mm_data_async().

python/sgl_jax/srt/server_args.py
zhengkezhou1: Fixed by applying VLM defaults after model config detection. This covers the regular Qwen-VL path where server_args.multimodal is false.

PR #1399, opened 2026-06-26
feat(model): Qwen2.5-VL Stage 1 in-model multimodal (owning-rank DP)

lianfang0320:
## Motivation

Stage 1 of migrating Qwen2.5-VL image understanding into the standard `srt/` in-model VLM pipeline, on top of the tokenizer-time multimodal preprocessing from #1398. Adds the in-model `Qwen2_5_VLForConditionalGeneration` that runs vision encode + multimodal merge inside the standard forward, with **owning-rank data-parallel** sharding (zero cross-rank vision collective).

## What's included

- `models/qwen2_5_vl.py` — `Qwen2_5_VLForConditionalGeneration`:
  - `encode_mm`: owning-rank, single `shard_map` — per-image flat ViT (B=1) + local `segment_sum` compaction → `vision_features [dp*per_dp, hidden]` `P("data")`.
  - `merge_mm`: `shard_map` per-rank local cumsum-gather → `inputs_embeds` `P("data")`, no cross-rank collective.
  - reuses the staged Qwen2.5-VL ViT + language backbone.
- `managers/schedule_batch`: host-only `mm_items_by_rank` / `rank_vision_rows` sidecars; dict/dataclass dual-read.
- `model_executor/model_runner`: `prepare_inmodel_multimodal_forward` drives `encode_mm` + `merge_mm` → `forward_batch.input_embedding`.
- `managers/tp_worker`: invoke the in-model multimodal hook before forward.
- `multimodal/.../qwen2_5_vit`: padding-aware vision attention.
- `model_executor/forward_batch_info`: `input_embedding` sharded `P("data", None)`.
- `managers/scheduler`, `io_struct`: mm dual-read + image/video/audio passthrough.
- `test/test_vlm_stage1_framework`: contract / encode / merge / attention coverage.

## Base

Targets `epic/vlm`; builds on #1398 (processor / tokenizer path).

## Notes

- DP=1 verified via `test_vlm_stage1_framework`. The dp>1 owning-rank path is collective-free by construction; TPU dp>1 numerics verification is pending.


python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -2652,6 +2652,7 @@ def get_model_worker_batch(
         mrope_positions = _mm["mrope_positions"]
         apply_for_deepstack = _mm["apply_for_deepstack"]
         deepstack_visual_embedding = _mm["deepstack_visual_embedding"]
```
JamesBrianD, line 2670: Do mrope_positions, apply_for_deepstack and deepstacks require similar DP padding alignment?

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -2956,6 +2959,54 @@ def _extract_mm_value(mm_inputs: Any, key: str):
     return getattr(mm_inputs, key, None)
 
 
+def _count_mm_item_rows(item: Any) -> int:
+    offsets = getattr(item, "offsets", None)
```
JamesBrianD: AI-generated code frequently uses `getattr(item)`; could this be changed? It is an inelegant and unsafe coding practice.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -2956,6 +2959,54 @@ def _extract_mm_value(mm_inputs: Any, key: str):
...
+            mm_items = [item for item in mm_items if _is_image_mm_item(item)]
+            if not mm_items:
+                continue
+            has_mm_items = True
+            mm_items_by_rank[dp_rank].extend(mm_items)
+            rank_vision_rows[dp_rank] += sum(_count_mm_item_rows(item) for item in mm_items)
+
+    if not has_mm_items:
+        return None, None
+    return mm_items_by_rank, rank_vision_rows
```
JamesBrianD: Shouldn't we first check whether it is a VLM model here, and skip this initialization if it isn't?

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -518,6 +522,46 @@ def forward(
...
 
+    def prepare_inmodel_multimodal_forward(
+        self,
+        model_worker_batch: "ModelWorkerBatch",
+        forward_batch: ForwardBatch,
+    ) -> None:
+        if not forward_batch.forward_mode.is_extend():
+            return
+
+        mm_items_by_rank = getattr(model_worker_batch, "mm_items_by_rank", None)
```
JamesBrianD: AI-generated code frequently uses getattr(item); could this be changed? It is an inelegant and unsafe coding practice.

python/sgl_jax/srt/managers/tp_worker.py
JamesBrianD, line 415: `prepare_inmodel_multimodal_forward` should not perform device-side ViT processing operations here.
Please examine the code for `tp_worker` and `tp_worker_overlap_thread` to understand the process—from scheduling through to the final dispatch for device execution—before making any modifications.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -518,6 +522,46 @@ def forward(
...
+        if rank_vision_rows is None:
+            rank_vision_rows = [0 for _ in mm_items_by_rank]
+        if not any(int(rows) > 0 for rows in rank_vision_rows):
+            return
+
+        model = self.model
+        encode_mm = getattr(model, "encode_mm", None)
+        merge_mm = getattr(model, "merge_mm", None)
+        if not callable(encode_mm) or not callable(merge_mm):
+            return
```
JamesBrianD: AI-generated code frequently uses getattr(item); could this be changed? It is an inelegant and unsafe coding practice.

python/sgl_jax/srt/models/qwen2_5_vl.py
```diff
@@ -0,0 +1,738 @@
...
+from jax.sharding import Mesh, NamedSharding, PartitionSpec
+
+from sgl_jax.srt.managers.schedule_batch import _count_mm_item_rows, _extract_mm_value
+from sgl_jax.srt.multimodal.configs.qwen_vl.qwen_2_5_vl_config import (
+    QwenVLModelVitConfig,
+)
+from sgl_jax.srt.multimodal.models.qwen2_5VL.qwen2_5_vit import Qwen2_5_VL_VisionModel
+from sgl_jax.srt.multimodal.models.qwen2_5VL.qwen2_5_vl_generation import (
+    Qwen2_5_VL_Generation,
+)
```
JamesBrianD: This file is entirely "vibe coding"; please reimplement it to ensure a clear architecture and eliminate all unnecessary boilerplate code.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -2956,6 +2985,180 @@ def _extract_mm_value(mm_inputs: Any, key: str):
...
+            if feat is None:
+                raise ValueError(f"IMAGE item in round {k}, dp_rank {r} is missing feature.")
+            feat = np.asarray(feat)
+            if feat.ndim != 2 or feat.shape[0] <= 0:
+                raise ValueError(
+                    "IMAGE item feature must be a non-empty 2D patch array, "
+                    f"got shape={feat.shape} in round {k}, dp_rank {r}."
+                )
+            features[r] = feat
+            patch_rows[r] = int(feat.shape[0])
```
JamesBrianD, line 3121: Could we add a host-side consistency check between the placeholder span length and the number of rows the ViT will produce? `_build_merge_idx()` derives `src_idx/mask` from `item.offsets`, while the merge later assumes those local row ids are valid for the encoded features. For example, if a future processor/dict input provides 8 raw patch rows for Qwen2.5-VL but `offsets=[(0, 3)]` (`spatial_merge_size ** 2 == 4`, so the ViT merger will only produce 2 rows), the merge can over-index `features[src_idx]` or silently leave rows unused. Upstream SGLang checks the multimodal embedding row count against the placeholder mask before scatter; this path should fail fast here before building the plan.

JamesBrianD:
One integration dependency to call out: this forward path assumes the tokenizer/request layer already initializes `model_config`/`mm_processor`, runs the Qwen-VL processor, and sends processor-expanded `input_ids` plus `mm_inputs` to the scheduler. In the current branch, `TokenizerManager` still leaves `self.model_config = None` under `server_args.multimodal`, so `mm_processor` is never created and image requests fail before reaching this scheduler/model-forward path. If this is covered by the companion tokenizer PR, could we make that dependency explicit before merging this stage; otherwise this PR will not be end-to-end runnable by itself.

lianfang0320:
> One integration dependency to call out: this forward path assumes the tokenizer/request layer already initializes `model_config`/`mm_processor`, runs the Qwen-VL processor, and sends processor-expanded `input_ids` plus `mm_inputs` to the scheduler. In the current branch, `TokenizerManager` still leaves `self.model_config = None` under `server_args.multimodal`, so `mm_processor` is never created and image requests fail before reaching this scheduler/model-forward path. If this is covered by the companion tokenizer PR, could we make that dependency explicit before merging this stage; otherwise this PR will not be end-to-end runnable by itself.

This in-model VLM path is intended to run through the normal launch path without --multimodal. The --multimodal flag belongs to the legacy/staged multimodal server path; under that path TokenizerManager.model_config is still None. For this PR, the dependency is that Qwen-VL requests must go through the normal tokenizer path where ModelConfig is initialized and QwenVLProcessor expands input_ids/mm_inputs

python/sgl_jax/srt/managers/schedule_batch.py
lianfang0320: Implemented a host-side Qwen-specific consistency check in the vision metadata builder.

  The check now validates the item contract before the scheduler builds the merge plan:

  - raw patch rows must match `t * h * w` from `image_grid_thw`
  - `h` / `w` must be divisible by `spatial_merge_size`
  - placeholder span length from `item.offsets` must match the ViT merger output rows:
    `t * (h // spatial_merge_size) * (w // spatial_merge_size)`

  I kept this in the Qwen2.5-VL metadata builder rather than in `_build_merge_idx()` because `spatial_merge_size` and `image_grid_thw` are model-specific geometry, while the scheduler should only consume offsets and build the generic merge mapping.

  I also considered a generic forward-side check mirroring upstream’s embedding-vs-placeholder-mask check, but in this JAX path the mask is already device-placed by the time `embed_mm_inputs()` runs. Pulling it back with `device_get()` would add a host/device sync on the hot path. Since this issue is specifically the Qwen `offsets` vs `grid/spatial_merge_size` contract, failing before plan construction in the builder is the cheaper and more targeted check.

Issue #1339, opened 2026-06-09
[RFC] HiCache S5a: Recurrent — make it correct (slot lifecycle + CoW)


MokusMokun:
### Summary

Bring linear-recurrent state (KDA / GDN / GLA) into the prefix tree for cross-request reuse. This sub-item targets correctness only, no kernel dependency: remove the hard guard forcing `--disable-radix-cache` for recurrent, change slots from per-request free to an active/tree/free 3-state lifecycle, **split the release path so the request-finish free does not drop the recurrent snapshot committed to the tree**, and clone state on a prefix fork via slow-path CoW (full-buffer copy).

Depends on **#1311 complete** (Stage 1 fully tested, S1a+S1b+S1c green) — build recurrent on a validated core, per the RFC M1 → M2 critical path. Code-wise it uses the S1a typed-params (`best_match_node`, the CoW hit anchor) and the S1b core.

### Delivery (split per #1380)

Shipped as 3 stacked PRs cut by page regime (tracked as sub-issues below), validated by a nightly suite (#1425, under parent #1338). Split mapping from #1380's review: (1)→PR-1; (2)+(3)+(5)→PR-2; (4)→PR-3.

### Position in the roadmap

partial-Stage-5, sub-item 1/3. Recurrent's value lands at "make it correct"; it does not block on the kernel ("make it fast" = S5b).

### Proposed structure changes

- `sgl_jax/srt/mem_cache/unified_cache_components/recurrent_component.py` — **new**: `RecurrentComponent` (port upstream `MambaComponent`, `MAMBA`→`RECURRENT`)
- `sgl_jax/srt/mem_cache/base_prefix_cache.py` — **modify**: add the `supports_mamba()` capability; `UnifiedRadixCache.supports_mamba()` returns True once it carries the REC component. Upstream keys recurrent dispatch on this capability, not on a concrete cache type.
- `sgl_jax/srt/managers/schedule_batch.py` — **modify**: drive the hit-path clone with `cow_mamba = tree_cache.supports_mamba()`. Today recurrent is keyed off the **pool** type (`isinstance(req_to_token_pool, HybridReqToTokenPool)`), which only gathers `recurrent_indices` — it does not trigger CoW, because recurrent has no tree path yet. Also wire the recurrent indices / ping-pong track buffer for `enable_mamba_extra_buffer`.
[...]

Issue #1338, opened 2026-06-09
[RFC] HiCache stage 5 (partial): device-side Recurrent + SWA


MokusMokun:
## Summary

On top of `UnifiedRadixCache` (Stage 1, #1311), add two device-side `TreeComponent`s: `RecurrentComponent` (KDA / GDN / GLA linear-recurrent models; realizes the #1280 roadmap's `RecurrentStateComponent` checkbox) and `SWAComponent` (sliding-window attention). This lets recurrent state and SWA KV enter the prefix tree and be reused across requests, the same way full KV is.

This is a **partial delivery** of #1280 Stage 5: only the device-side component logic. The HiCache-coupled items (host backup, `--hicache-ratio` benchmark, PD coordination) are **out of scope here** and remain with HiCache (Stage 3/4).

## Position in the roadmap

#1280 Stage 5 ("Multi-component + tuning") mixes two kinds of checkboxes:

| Stage 5 checkbox | True dependency | This issue |
|---|---|---|
| `SWAComponent` | Stage 1 (core) | ✅ device-only |
| `RecurrentStateComponent (Mamba/KDA)` | Stage 1 (core) | ✅ device-only |
| Benchmark across `--hicache-ratio` | Stage 3/4 (HiCache) | ❌ deferred |
| Coordinate with PD on HBM/host split | PD + Stage 3/4 | ❌ deferred |

**Prerequisite = #1311 (Stage 1)** only; no dependency on HiCache (Stage 2/3/4) — HiCache hooks remain no-op placeholders here.

## Sub-issues

| Sub | Scope | Size | Depends on |
|---|---|---|---|
| S5a | Recurrent "make it correct": remove the `--disable-radix-cache` guard + recurrent capability dispatch (`supports_mamba()`) + 3-state slot lifecycle + release-path split (don't free the recurrent snapshot at request finish) + slow-path CoW (full state) + `enable_mamba_extra_buffer` for page≥128 | L | #1311 complete |
| S5b | Recurrent "make it fast": h0-clone Pallas kernel + JIT buffer donation | L | S5a |
| S5c | `SWAComponent` + scheduler SWA capability methods (port upstream `supports_swa()` / `is_chunk_cache()` / `is_tree_cache()`; replace the cache-typed `isinstance` checks) | L | S5a |

Dependency (S1a = typed-params, S1b = core, S1c = M1 tests, all under #1311):

```
[...]

Issue #1311, opened 2026-06-04
[RFC] HiCache stage 1: UnifiedRadixCache + FullComponent port


cjx0709:
## Summary

Port `UnifiedRadixCache` (with the `TreeComponent` ABC and `FullComponent`) from sglang upstream, replacing the current `RadixCache` as the prefix-tree entry point. Core logic (trie, reference counting, eviction) aligns with upstream; adaptation is limited to device KV operations, scheduler integration, and JAX/TPU constraints.

This stage is controlled by `--enable-unified-radix-tree` (default off); when off, behavior matches the pre-port path. HiCache hooks (`build_hicache_transfers` / `commit_hicache_transfer` / `drive_host_eviction`) are no-op placeholders in this stage — real integration lands in Stage 3.

> Note: this stage is not a pure port. `MatchResult` grows from 4 fields to include `best_match_node`, and evict/match/lock signatures converge into typed params (`EvictParams`, etc.). This base-interface change is **not controlled by the flag** — it touches the default `RadixCache` and every caller. It is split out as sub-issue S1a.

## Position in the roadmap

1. Host pool foundation (#1310)
2. **UnifiedRadixCache + FullComponent port** *(this issue)*
3. HiCacheController + async pipeline (#1312)
4. End-to-end integration and validation (#1313)
5. L3 persistent storage
6. Multi-component + tuning (device-side Recurrent + SWA is delivered by the partial-Stage-5 issue)

No prerequisites; can be developed in parallel with Stage 0. It is the core that downstream HiCache / PD / Recurrent all depend on.

## Sub-issues

| Sub | Scope | Size | Depends on |
|---|---|---|---|
| S1a | typed-params migration + `MatchResult.best_match_node` + fix 3 strict 4-tuple unpacks + legacy path compatibility | L | — |
| S1b | `UnifiedRadixCache` core + `UnifiedTreeNode` + `ComponentData`/`ComponentType` + `FullComponent` + toggle flag (incl. `server_args` plumbing) + builder wiring | L | S1a |
| S1c | M1 serving tests: hit/flush determinism + throughput/TTFT instrumentation + MMLU wiring | L | S1a, S1b |

## Proposed structure changes

- `sgl_jax/srt/mem_cache/unified_radix_cache.py` — **new**: `UnifiedTreeNode` / `UnifiedLRUList` / `UnifiedRadixCache`
[...]

PR #1401, opened 2026-06-26
feat(mem_cache): page=1 recurrent state in UnifiedRadixCache (CoW base)

MokusMokun:
## Summary

Refs #1339, #1338

Brings linear-recurrent (KDA / GDN) hybrid models into `UnifiedRadixCache` (#1311) at `page_size=1`: a cross-request continuation reuses the committed recurrent state of its shared prefix via copy-on-write. Behind `--enable-unified-radix-tree`; default off, flag-off paths unchanged.

This is the recurrent base. It is the first of a stacked split — the production `page_size >= 128` extra-buffer path and the cache-aware DP routing policy land in follow-up PRs on top of this one.

Note: earlier revisions of this PR carried a multi-host EXTEND batch-size guard (`SAFE_EXTEND_PER_DP_BS`). It was removed after the failure it worked around could not be reproduced on current code — see the validation comment below for the investigation and evidence.

## What Changed

Paths relative to `python/sgl_jax/srt/` unless noted.

### Recurrent in UnifiedRadixCache (base, `page_size=1`)
| File | Change |
|---|---|
| `mem_cache/unified_cache_components/recurrent_component.py` | **new** — leaf-only recurrent component: finished-request commit, prefix-hit CoW source recording, commit/evict/single-node-lock lifecycle. |
| `mem_cache/unified_radix_cache.py` | modify — wire the recurrent component into match/insert/commit/evict; consume the per-component effective cache length; advance `prefix_indices` on the zero-cache-len skip. |
| `mem_cache/unified_cache_components/tree_component.py` | modify — component interface extended for recurrent (`is_recurrent`, insert result, `prepare_for_caching_req` returns an effective cache length). |
| `mem_cache/unified_cache_components/__init__.py` | modify — export `RecurrentComponent`. |
| `mem_cache/base_prefix_cache.py` | modify — `supports_recurrent()` + recurrent fields on the match/insert/evict/lock parameter structs (CoW source, leaf donation value, per-component lock skip set). |
| `mem_cache/registry.py` | modify — route hybrid-recurrent models to `UnifiedRadixCache` with `tree_components=(FULL, RECURRENT)` (`is_hybrid_recurrent`). |
| `mem_cache/kv_cache_builder.py` | modify — thread `is_hybrid_recurrent` into the build context. |
| `mem_cache/memory_pool.py` | modify — recurrent CoW slot API on `HybridReqToTokenPool` (alloc/free a running slot, slot→tree value, commit-to-tree). |
| `mem_cache/recurrent_state_pool.py` | modify — `copy_slots`: one-shot CoW clone of recurrent + conv state, `optimization_barrier`-wrapped for donated-buffer multi-host SPMD safety. |
| `mem_cache/common.py` | modify — `pool.free()` releases request-owned recurrent slots. |
| `disaggregation/decode.py` | modify — free the recurrent slot on the PD decode path. |
| `model_executor/model_runner.py` | modify — one-shot pre-forward CoW clone before `model(...)`. |
| `model_executor/forward_batch_info.py` | modify — recurrent CoW-src metadata on `ForwardBatch`. |
| `model_executor/model_runner_kv_cache_mixin.py` | modify — recurrent routing + `page_size=1` pool sizing (slot factor 2 = 1 running + 1 headroom), admission cap kept dp-aligned; enforce the recurrent server constraints. |
| `managers/scheduler.py` | modify — construct `ScheduleBatch` with `is_hybrid_recurrent`. |
| `managers/schedule_batch.py` | modify — recurrent request fields + CoW-src metadata build, and the `cache_protected_len` fix below. |
| `kernels/gdn/gated_delta.py` | modify — EXTEND gather-site `optimization_barrier` (donated-buffer aliasing guard). |
| `server_args.py` | modify — `--enable-unified-radix-tree` help text (now also routes hybrid recurrent models). |

Two correctness fixes (general hardening, no-op for non-recurrent requests; net-identical for EAGLE):
- `schedule_batch.py` `prepare_for_extend` — `cache_protected_len = last_matched_prefix_len`, not `len(prefix_indices)` (the latter can include a request-owned, un-published tail, which on finish/retract frees only part of the committed range and orphans KV pages → pool leak → crash).
- `unified_radix_cache.py` — on an insert that caps to zero newly-cacheable tokens, advance `prefix_indices` to the committed range (mirroring `ChunkCache`) so the next chunked round extends from the right offset instead of re-allocating over committed KV.

### Tests
| File | Change |
|---|---|
| `test/srt/test_recurrent_cow_metadata.py` | **new** — recurrent CoW src-index metadata + all-zero skip. |
| `test/srt/test_prepare_for_extend_protected_len.py` | **new** — `cache_protected_len` regression (sentinel-guarded). |
| `python/sgl_jax/test/test_model_runner_kv_cache_mixin.py` | modify — recurrent server-constraint enforcement + dp-aligned admission cap. |
| `python/sgl_jax/test/mem_cache/test_unified_radix_cache.py` | modify — recurrent CoW ledger, copy-equivalence, eviction, lock lifecycle, effective-cache-length, abort/retract, full-only re-match. |
| `python/sgl_jax/test/mem_cache/test_hybrid_req_to_token_pool.py` | modify — recurrent slot ledger. |
| `python/sgl_jax/test/test_gdn_attention.py`, `test_kda_attention.py` | modify — recurrent CoW `copy_slots` equivalence (GDN + KDA). |
| `test/srt/run_suite.py` | modify — register the new recurrent CPU tests (`unit-test-cpu`). |

## Live Validation

CPU unit tests run per-PR in `unit-test-cpu`; the recurrent backend equivalence tests run in `unit-test-tpu-v6e-4`. Validated on a v6e (single-host 2×2), a CPU runner, and 16-chip multi-host (4 hosts).

| Area | Result |
|---|---|
| Recurrent cache correctness — `unit-test-cpu` | **pass** — 89 tests: CoW ledger/lifecycle, eviction, effective-cache-length, zero-cache-len skip, `cache_protected_len`, dp-aligned admission cap |
| Recurrent CoW `copy_slots` equivalence — v6e | **pass** — GDN (11) + KDA (10) on TPU, including bit-exact recurrent copy-on-write on both backends |
| `page_size=1` serving determinism | **bit-identical** — cold == warm == no-cache (`--disable-radix-cache` control) |
| Ownership leak fix | **clean** — 12 serial + 300 concurrent >512-token requests, no slot leak / crash |
| Multi-host serving — GKE v6e-16 (4 hosts) | **clean** — cross-request continuation workloads with genuine cached-prefix EXTEND rows (`#cached-token` up to ~6000) at per-DP padded batch 16 (tp16/dp4) and 32 (tp16/dp1): 0 NaN across 96 + 48 generations |
| Flag-off | **byte-identical** — `--enable-unified-radix-tree` off leaves all paths unchanged |

## Deferred / Out of scope

Land in follow-up PRs stacked on this base:
- Production `page_size >= 128` extra-buffer path (page-aligned ping-pong track slots + scheduler boundary splitting) and the recurrent pool sizing / admission-backpressure robustness.
- `--dp-schedule-policy cache_aware` DP routing.
- HiCache host tier; speculative / mixed-chunk + extra-buffer.

## Known Limitations

- `page_size=1` only; the production `page_size >= 128` path is the follow-up extra-buffer PR.

## Next Work

1. `page_size >= 128` extra-buffer recurrent path + recurrent pool sizing / admission backpressure.
2. `cache_aware` DP routing to lift the cross-request reuse ceiling under data parallelism.


cjx0709:
the code in this pr seems being too over-commented, maybe can refine and delete some.

python/sgl_jax/srt/utils/common_utils.py
```diff
@@ -39,6 +39,12 @@
 PRECOMPILE_DEFAULT_TOKEN_PADDINGS = [1 << i for i in range(6, 14)]
 PRECOMPILE_DEFAULT_BS_PADDINGS = [1 << i for i in range(0, 9)]
 
+# Largest EXTEND per-DP batch size that compiles correctly for the recurrent
```
cjx0709: this guard looks strange, if it is a real bug, maybe we should at least leave a issue to describe about it. And it looks not really related to this pr, we use gdn model use chunk cache can also cause this nan problem(but maybe we don't test page_size=1 before, so it happends at first time). 
I think it's ok to add this guard in this pr, but maybe leave a issue to describe it.

MokusMokun:
**Follow-up on the multi-host EXTEND batch-size guard (`SAFE_EXTEND_PER_DP_BS`) — removed in b7f38aeb after re-validation.**

Investigation summary:

- The guard originated from a multi-host NaN observed during development: on a 16-chip page=1 recurrent run, a cached-prefix continuation row inside the per-DP=16 EXTEND executable produced NaN logits (`!!!` output), while per-DP=8 was clean. In-situ debugging at the time ruled out donated-buffer aliasing (an `optimization_barrier` on the KV buffer immediately before the attention kernel did not help, and q/k/v at the first-NaN layer were verified finite), so a shape-keyed serving-layer guard was added.
- That observation predates the donated-buffer entry barriers later added on the GDN EXTEND gather sites (included in this PR), and no unguarded run had been re-tested since.
- Re-validation on current code (16 chips / 4 hosts, guard disabled): the originally-failing executable shape — cached-prefix continuation rows inside the per-DP=16 EXTEND bucket (tp16/dp4) — and the per-DP=32 variant (tp16/dp1) both produce coherent output: 0 NaN across 96 + 48 generations, with scheduler logs confirming genuine cached-prefix rows (`#cached-token` up to ~6000) in every burst. Plain full-query EXTEND at per-DP=16 (radix off, page 1 and 64) was also clean (0/288 each).
- With the trigger no longer reproducible, the guard (admission deferral + EXTEND bucket capping + dispatch backstop, ~400 lines of scheduling machinery) is removed; EXTEND returns to the legacy largest-bucket padding on all paths. Post-removal validation: the recurrent CPU unit gate (89 tests) plus a fresh multi-host smoke on the removed code (same cached-prefix workload, multi-bucket `[32, 64]`, 0 NaN).

If the NaN ever resurfaces, the removal commit message documents the failing shape, and the reproduction workload is: seed N distinct conversations serially (each finishes and commits its recurrent state), then burst their continuations concurrently so cached-prefix EXTEND rows form inside a per-DP>8 padded batch.

python/sgl_jax/srt/utils/common_utils.py
MokusMokun: removed the guard - it's been verified unnecessary in https://github.com/sgl-project/sglang-jax/pull/1401#issuecomment-4863507373

cjx0709 (commented):
Overall the quality and validation look solid. One small ask before merging: the comment density is a bit high — a lot of them explain what the code does rather than why. Would suggest trimming them down, keeping the load-bearing "why" ones (e.g. the aliasing-barrier and leak-fix rationale).

cjx0709 (approved):
LGTM

PR #1402, opened 2026-06-26
feat(scheduler): cache-aware DP routing policy

MokusMokun:
## Summary

Adds an opt-in `--dp-schedule-policy cache_aware`. Under data parallelism,
load-only DP routing caps cross-request prefix reuse at ~`1/dp_size`: a prefix
cached on one rank is invisible to the others, so a shared system prompt is
re-cached on every rank it lands on. `cache_aware` routes each request to the
rank that already holds its cached prefix, with soft load balancing, lifting
reuse without overloading a hot rank.

This is a general full-KV `cache_aware` semantics change, **not**
recurrent-specific — it works for any radix-backed dp-keyed cache. Default stays
`min_running_queue`; `round_robin` and all flag-off paths are unchanged.

## What Changed

Paths under `python/sgl_jax/srt/` unless noted.

| File | Change |
|---|---|
| `managers/dp_schedule_policy.py` *(new)* | Dependency-free policy module (no JAX / Scheduler imports, so it unit-tests standalone): `pick_cache_aware_dp` (soft affinity-vs-load) and `req_prefix_match_key` (effective probe key). |
| `managers/scheduler.py` | `_select_cache_aware_dp` probes each eligible rank's dp-keyed radix tree via `match_prefix` and routes by the policy; `_dp_load_and_eligible` factored out of `_select_min_running_dp`; `cache_aware` branch added to `select_dp_for_request`. |
| `server_args.py` | Add `cache_aware` to `--dp-schedule-policy` choices + help. |
| `test/srt/test_dp_schedule_policy.py` *(new)* | 18 unit cases over the pure policy + effective-key extraction; registered in `test/srt/run_suite.py` (`unit-test-cpu`). |
| `benchmark/hicache/bench_unified_radix_ab.py` | Offline `--compare` gate: merge the `aggregates` from separate single-config result JSONs and run the soft-target A/B with no serving. |

**Policy.** Balance on large load skew (`max-min > 32` and `max > 1.1*min`,
upstream sgl-router defaults); else route to the least-loaded rank among the
*holders* — ranks whose cached prefix covers > 50% of the probe length — so a hot
prefix spreads across its holders instead of piling on the single longest match;
else least-loaded. Load order `(running, tokens, rank)`. The per-rank admission
cap stays the eligibility bound, so a hot prefix whose rank is full spills to the
next-best rank.

**Why the effective probe key.** The probe derives the same key the request's
real radix lookup will use, so it routes on a match the request can actually
reuse: `lora_id` is folded into `extra_key` (as `Req.__init__` does), and the
length is the reusable prefix `input_len - 1` (clamped by `logprob_start_len`
under `return_logprob`, as `Req.adjust_max_prefix_ids` does). Using `input_len-1`
also stops page-aligned `match_prefix` from reporting a phantom final-page hit
the request drops, and it is the holder denominator.

## Live Validation

| Area | Result |
|---|---|
| `test_dp_schedule_policy.py` (18 cases) | **pass** |
| `unit-test-cpu` full suite (39 files) | **pass** (pinned env, jax 0.8.1 / transformers 4.57) |
| pinned pre-commit (isort / ruff / black / codespell / mypy) | **pass** |
| Full-KV A/B (Qwen3-30B-A3B, tp4/dp2/ep4) | cache_aware **> min_running** |

### Full-KV A/B (Qwen3-30B-A3B, tp4/dp2/ep4, unified radix on, gsp shared-prefix)

Pure full-attention MoE, one v6e-4. Two server launches differing **only** in
`--dp-schedule-policy`; gsp = 16 groups × 8 prompts, 2048-token shared prefix;
mean of 2 post-warmup reps:

| metric | min_running_queue | cache_aware | delta |
|---|---|---|---|
| cache hit rate | 0.676 | 0.698 | +0.022 |
| p50 TTFT | 1177 ms | 874 ms | **−25.8%** |
| throughput | 32635 tok/s | 34780 tok/s | **+6.6%** |

The hit-rate headroom is bounded at dp2 (the ~`1/dp_size` floor is already ~0.5);
the user-visible win is the **−25.8% p50 TTFT** from better prefill cache
locality.

Reproduce (run the launch + client once per policy, identical otherwise):

```bash
# POLICY in {min_running_queue, cache_aware}
python -m sgl_jax.launch_server --model-path <model> --port 30000 \
  --tp-size 4 --dp-size 2 --ep-size 4 --attention-backend fa \
  --enable-unified-radix-tree --enable-cache-report --page-size 128 \
  --context-length 8192 --chunked-prefill-size 512 --mem-fraction-static 0.8 \
  --max-running-requests 64 --dtype bfloat16 --random-seed 3 \
  --skip-server-warmup --trust-remote-code \
  --precompile-bs-paddings 8 16 32 64 --precompile-token-paddings 2048 4096 \
  --dp-schedule-policy $POLICY

python -m sgl_jax.bench_serving --backend sgl-jax --base-url http://127.0.0.1:30000 \
  --model <model> --dataset-name generated-shared-prefix \
  --gsp-num-groups 16 --gsp-prompts-per-group 8 --gsp-system-prompt-len 2048 \
  --gsp-question-len 128 --gsp-output-len 128 --num-prompts 128 --max-concurrency 64
```

The committed `bench_unified_radix_ab.py` benchmarks **cache configs**
(no-cache / radix / unified), not DP policies; the policy A/B above is the two
launch commands, and the new offline `--compare` gate merges the two result
JSONs into one A/B report.

### Downstream (validated once the recurrent path lands) — related context only

- Recurrent shared-prefix A/B at dp4: hit 0.51 → 0.76, p50 TTFT −59%, tput +36%.
- Real eval (token-weighted cache-hit): 0.289 → 0.444.

## Known Limitations

- The probe uses `match_prefix`, which is **not read-only**: it refreshes the
  matched path's LRU and may split a node on a partial match. In the common case
  a prefix lives on one rank, so only the rank the probe routes to is touched. A
  non-mutating `peek_prefix_len` would remove the cross-rank touch (Next Work).
- Recurrent / hybrid models still require `--disable-radix-cache` on this branch,
  so `cache_aware` applies to full-KV radix paths only here; recurrent reuse is
  gated on the recurrent-radix path.

## Next Work

1. Non-mutating `peek_prefix_len` probe to drop the LRU-refresh / node-split side
   effect.

### DIFF
--- a/benchmark/hicache/bench_unified_radix_ab.py
+++ b/benchmark/hicache/bench_unified_radix_ab.py
@@ -80,7 +80,14 @@ def parse_args():
         choices=["random", "gsp"],
         default=["random", "gsp"],
     )
-    p.add_argument("--max-concurrency", type=int, default=64)
+    p.add_argument(
+        "--max-concurrency",
+        type=int,
+        default=64,
+        help="Client-side in-flight request load (NOT a server capacity knob; the "
+        "server cap is --max-running-requests). With cache_aware DP routing, keep "
+        ">= dp_size so all ranks receive load.",
+    )
     p.add_argument("--num-prompts", type=int, default=256)
     p.add_argument("--random-input-len", type=int, default=1024)
     p.add_argument("--random-output-len", type=int, default=128)
@@ -125,7 +132,24 @@ def parse_args():
             "ids) when the host has no ShareGPT access."
         ),
     )
+    p.add_argument(
+        "--compare",
+        nargs="+",
+        default=None,
+        metavar="RESULT_JSON",
+        help=(
+            "Offline A/B gate: merge the 'aggregates' from two or more "
+            "--output-json files (each a single-config run) and run the soft-target "
+            "gate across them, no serving. This lets runs that produce one config "
+            "at a time form an A/B gate. Combine with --strict to fail on a missed "
+            "target."
+        ),
+    )
     args = p.parse_args()
+    if args.compare:
+        # Offline merge of prior results: no server is launched, so the
+        # server-shape validations below do not apply.
+        return args
     assert args.page_size >= 128, f"--page-size must be >= 128, got {args.page_size}"
     if args.tp_size > 1:
         bad = [b for b in args.precompile_bs_paddings if b % args.tp_size != 0]
@@ -371,9 +395,16 @@ def check_sweep_complete(args, raw):
 
 
 def soft_targets(args, agg):
-    """Print PASS/WARN per check. Returns True if all checks passed."""
+    """Print PASS/WARN per check. Returns ``(all_pass, n_gates_fired)``.
+
+    ``n_gates_fired`` counts the A/B comparison checks that actually ran (both
+    sides present). A run where no gate fired (e.g. only one config supplied)
+    must not be mistaken for a pass; callers using ``--strict`` should fail when
+    it is zero.
+    """
     print("\n--- soft-target report ---")
     all_pass = True
+    fired = 0
 
     def have(config, workload):
         return agg.get(config, {}).get(workload) is not None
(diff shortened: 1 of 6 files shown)


cjx0709 (approved):
LGTM, and I think the "next work" can be left with a issue so that we will not forget it.

MokusMokun:
> LGTM, and I think the "next work" can be left with a issue so that we will not forget it.

issue established
https://github.com/sgl-project/sglang-jax/issues/1414

PR #1057, opened 2026-05-11
fix: upgrade JAX from 0.8.1 to 0.9.2

Aharrypotter:
## Summary

- Upgrade JAX/JAXlib from 0.8.1 to 0.9.2 and libtpu from 0.0.30 to 0.0.38.
- Fix JAX 0.9.1+ strict `shard_map` `in_specs` enforcement across FlashAttention, MLA, KV cache, linear attention, and fused MoE paths.
- Fix JAX 0.9.2 compatibility issues around `reshape` `out_sharding`, `pallas_call` Manual mesh axes, and the `jax.reshard` API namespace.
- Add sharding-aware helper paths that derive shard specs from actual `NamedSharding` / tracer sharding instead of relying on divisibility heuristics.
- Fix FlashAttention GQA and `attention_sink` sharding behavior under local head-ratio constraints.
- Harden LoRA batching with explicit batch metadata, stable base-model sentinel handling, no-LoRA row preservation, clearer slot accounting, and stronger runtime validation.
- Remove confirmed dead LoRA helper code and tighten backend contracts.
- Stabilize LoRA debug dump request ordering and padded-batch `lora_ids` handling for DP-aware batches.

Closes #1025

## Motivation

JAX 0.9.1+ tightened several sharding contracts that were previously permissive. In particular, `shard_map` now requires `in_specs` to match the actual input sharding instead of silently resharding mismatched inputs. JAX 0.9.2 also exposes stricter behavior around `reshape(..., out_sharding=...)`, Pallas mesh axes, and reshard API placement.

This PR brings SGLang-JAX in line with the newer JAX/libtpu stack used by TPU inference work, while preserving model accuracy and making LoRA batch handling explicit enough to catch invalid runtime state early.

### Core JAX 0.9.2 sharding compatibility

**FlashAttention / MLA / linear attention**
- Add effective-axis helper logic that reads the actual input sharding, including tracer `aval.sharding` inside `jax.jit`.
- Derive `shard_map` `in_specs` / `out_specs` from actual tensor, DP, cache, sequence-length, and metadata sharding.
- Fix FlashAttention GQA behavior when local Q heads are not divisible by local KV heads by resharding Q to replicated where required.
- Fix `attention_sink` sharding by deriving from the actual Q sharding and resharding to replicated when Q is replicated on the tensor axis.

**KV cache / Pallas**
- Add local/test-safe update paths and wrap only multi-device paths with `shard_map`.
[...]

PR #1407, opened 2026-06-28
fix(sharding): derive FlashAttention shard_map in_specs from actual sharding

Aharrypotter:
## Summary

- Derive the FlashAttention `shard_map` `in_specs` / `out_specs` from the **actual** input
  sharding instead of hard-coded `P("data", "tensor", ...)`, so they satisfy the stricter
  `shard_map` `in_specs` enforcement introduced in JAX 0.9.1+.
- Add a small `effective_axis(x, dim, expected)` helper that reads an array's real
  `NamedSharding` (including tracer `aval.sharding` inside `jax.jit`).
- Keep the dependency pinned at `jax[tpu]==0.8.1`; this change is **dual-version compatible**
  (no behavior change on 0.8.1, satisfies the strict check on 0.9.2).

This is the first minimal, real call-site fix from the JAX 0.8.1 → 0.9.2 upgrade
(splitting #1057 into small, reviewable PRs). It is intentionally scoped to the single
concrete 0.9.2 `in_specs` incompatibility — multi-chip `num_kv_heads < tp_size` robustness
is deliberately **not** included (see Scope).

## Motivation

JAX 0.9.1+ makes `shard_map` enforce that `in_specs` match the actual sharding of each
input; the previous silent resharding of mismatched inputs is gone. FlashAttention hard-codes
`P("data", "tensor", ...)` for every argument, but the replicated 1-D auxiliary arrays —
`attention_sink` (`float32[num_heads]`) and `custom_mask` (`bool[...]`) — are actually
`P(None)`. On JAX 0.9.2 this raises:

```
ValueError: in_specs passed to shard_map: P('tensor',) does not match the specs of the
input: P(None,) for arg: float32[32]. `in_specs` is an optional argument ... If you want
to reshard your inputs, you can use `jax.reshard` on the arguments and then pass those args
to shard_map.
```

This is **not** a multi-device-only problem: it reproduces on a single chip, because even
with a tensor-axis size of 1 JAX 0.9.2 treats `P('tensor') != P(None)`. The fix follows the
direction agreed in the #1286 review thread: land one concrete, dual-version-compatible
call-site fix while keeping the runtime JAX pinned at 0.8.1.

## Modifications

**FlashAttention (`srt/layers/attention/flashattention_backend.py`)**
- In `FlashAttention._paged_attention`, derive `in_specs` / `out_specs` for `q/k/v`,
  `kv_cache_fused`, the metadata arrays (`kv_lens`, `page_indices`, `cu_q_lens`,
  `cu_kv_lens`, `distribution`, `custom_mask`), and `attention_sink` from each input's
  actual sharding via `effective_axis`, instead of hard-coded axis names.
- Replicated aux arrays now correctly get `P(None)`; tensor/DP-sharded arrays keep their
  axes. No production tensor is resharded.

**Sharding helper (`srt/utils/jax_utils.py`)**
- Add `effective_axis(x, dim, expected)`: returns `expected` only when `x` is exactly
  sharded that way on `dim` (reading `x.sharding` or, inside `jit`, `x.aval.sharding`);
  otherwise `None`. It does not infer or simplify composite axes.

**Tests (`test/test_jax_utils.py`)**
- Unit tests for `effective_axis`: concrete `NamedSharding` match / mismatch, replicated
  fallback to `None`, dim out of range.

## Accuracy Tests

Representative end-to-end model, before/after the change, on the same hardware and JAX
version (the dependency is pinned at 0.8.1). Dense MMLU via the maintained server path
(`fa` backend), `Qwen/Qwen3-8B`, `num_examples=200`, single chip, JAX 0.8.1:

```
cd test/srt && python -m unittest test_eval_accuracy_large.TestEvalAccuracyLarge.test_mmlu
```

| Run | MMLU score | stem / humanities / social / other | Test time |
| --- | --- | --- | --- |
| `main` (before) | **0.725** | 0.786 / 0.629 / 0.714 / 0.809 | 256.16 s |
| this PR (after) | **0.725** | 0.786 / 0.629 / 0.714 / 0.809 | 245.92 s |

The MMLU score is **identical** overall and per-category, confirming the in_specs derivation
is a semantic no-op on JAX 0.8.1 (no correctness regression on a real 8B model). End-to-end
latency is within run-to-run noise (no throughput regression / no added communication); no
production tensor is resharded on the common path.

JAX 0.9.2 forward-compatibility for this path is covered by the single-chip unit-test matrix
below (this PR is green on 0.9.2 where `main` fails) and by the full upgrade in #1057.

### FlashAttention unit tests — single-chip dual-version matrix

Run `python python/sgl_jax/test/test_flashattention_{mha,gqa,misc}.py` on TPU v6e restricted
to a single chip (`TPU_CHIPS_PER_HOST_BOUNDS=1,1,1 TPU_HOST_BOUNDS=1,1,1 TPU_VISIBLE_CHIPS=0`),
across three JAX versions in separate venvs on the same hardware:

| Code | JAX 0.8.1 | JAX 0.9.2 | JAX 0.10.2 |
| --- | --- | --- | --- |
| `main` (hard-coded specs) | PASS | **FAIL** — `in_specs` on attention_sink / custom_mask | PASS |
| **this PR** (derived specs) | PASS | **PASS** | PASS |

(PASS = mha 10/10, gqa 10/10, misc 6/6.) `main` fails only inside the strict `in_specs`
window: JAX 0.9.1 introduced the Explicit-mode assertion, and JAX 0.10.x no longer raises it
for a size-1 mesh axis. Deriving the specs from the actual sharding is **version-agnostic** —
green on 0.8.1, on the strict 0.9.x window, and on 0.10.x — so the fix is the robust choice
regardless of which JAX version the gray period and the eventual bump land on. The full
upgrade (dependency bump + end-to-end model runs) is exercised in #1057.

### Helper unit tests

```bash
python python/sgl_jax/test/test_jax_utils.py   # effective_axis: 4 tests, PASS
```

## Benchmarking and Profiling

Throughput/latency are reported via the Accuracy Tests table above (MMLU total latency on
`main` vs this branch). No separate serving benchmark is included: on JAX 0.8.1 the derived
specs are identical to the previous hard-coded specs on the common path, and no production
tensor is resharded, so no hot path changes behavior.

## Scope (explicitly deferred to a follow-up PR)

These are **multi-chip-only** and unrelated to the 0.9.2 `in_specs` strictness; they are
no-ops on a single chip and are validated separately on multi-chip hardware:

- `num_kv_heads < tp_size` divisibility fallback in `MHATokenToKVPool`
  (`effective_kv_partition_axis`).
- GQA head-ratio / v-axis / attention_sink resharding in the kernel call path.
- Reference-path host-side gather metadata in `ref_ragged_paged_attention` (test-only).

## References

- Part of [#1025](https://github.com/sgl-project/sglang-jax/issues/1025): the JAX upgrade
  tracking issue. This PR does not close it — it lands one dual-version-compatible call-site
  fix during the gray period while the dependency stays pinned at 0.8.1. (Per the discussion
  on that issue, the eventual runtime target may advance to 0.10.2; the `in_specs` strictness
  this PR addresses was introduced in 0.9.1 and still applies on 0.10.x, so the fix remains
  forward-compatible.)
- Relates to [#1057](https://github.com/sgl-project/sglang-jax/pull/1057): the full JAX
  0.8.1 → 0.9.2 upgrade; this PR is the first call-site fix split out of it, and #1057 also
  provides the 0.9.2 runtime + end-to-end model validation.
- Supersedes [#1286](https://github.com/sgl-project/sglang-jax/pull/1286): the helper-only
  foundation PR; per that review thread, this PR lands the helper together with a real,
  dual-version-compatible call-site fix instead of in isolation.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/layers/attention/flashattention_backend.py
+++ b/python/sgl_jax/srt/layers/attention/flashattention_backend.py
@@ -534,9 +534,10 @@ class FlashAttention(AttentionBackend):
         elif hasattr(token_to_kv_pool, "remap_cache_loc") and self.page_size == 1:
             page_indices_arg = token_to_kv_pool.remap_cache_loc(page_indices_arg, layer.layer_id)
 
-        q_arg = q.reshape(q.shape[0], -1, self.head_dim)
-        k_arg = k.reshape(k.shape[0], -1, self.head_dim)
-        v_arg = v.reshape(v.shape[0], -1, self.head_dim)
+        head_dim = getattr(layer, "head_dim", self.head_dim)
+        q_arg = q.reshape(q.shape[0], -1, head_dim)
+        k_arg = k.reshape(k.shape[0], -1, head_dim)
+        v_arg = v.reshape(v.shape[0], -1, head_dim)
 
         # JAX 0.9.1+ enforces shard_map in_specs match actual input sharding.
         # Derive effective axes from the actual array sharding rather than
--- a/python/sgl_jax/srt/layers/attention/flashattention_backend.py
+++ b/python/sgl_jax/srt/layers/attention/flashattention_backend.py
@@ -19,7 +19,7 @@ from sgl_jax.srt.mem_cache.memory_pool import KVCache
 from sgl_jax.srt.model_executor.forward_batch_info import ForwardBatch, ForwardMode
 from sgl_jax.srt.speculative.eagle_util import EagleDraftInput
 from sgl_jax.srt.utils import cdiv
-from sgl_jax.srt.utils.jax_utils import device_array
+from sgl_jax.srt.utils.jax_utils import device_array, effective_axis
 from sgl_jax.srt.utils.profiling_utils import named_scope
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


python/sgl_jax/test/test_jax_utils.py
```diff
@@ -0,0 +1,42 @@
+import unittest
+
+import jax
+import jax.numpy as jnp
+import numpy as np
```
cjx0709: this import of `numpy` seems not be used

python/sgl_jax/srt/layers/attention/flashattention_backend.py
```diff
@@ -534,43 +534,78 @@ def __call__(
         elif hasattr(token_to_kv_pool, "remap_cache_loc") and self.page_size == 1:
             page_indices_arg = token_to_kv_pool.remap_cache_loc(page_indices_arg, layer.layer_id)
 
+        q_arg = q.reshape(q.shape[0], -1, self.head_dim)
```
cjx0709: maybe can't just do this change, we do 

Image because of the hybird model, like mimo-v2-flash or gemma4. In gemma4, different layer has different head dim.

Aharrypotter: make sense

python/sgl_jax/test/test_jax_utils.py
Aharrypotter: fixed

cjx0709:
Thanks for the work here. After looking at the latest CI and the JAX `shard_map` behavior, I think this PR is not ready to merge in its current form.

Besides the lint failure, `accuracy-test-4-tpu (0)` is now failing with a FlashAttention shape mismatch:

```text
ValueError: Expected k.shape=(16, 2, 80) to be equal to v.shape=(16, 8, 80)
```
This looks related to deriving q/k/v/kv_cache specs from actual sharding, which can change the local-shape contract expected by the FA kernel on multi-chip paths.
Given that our final target is JAX 0.10.1, I think we should avoid adding a broad intermediate compatibility layer for the 0.9.x strict shard_map window. The better plan is to upgrade directly to JAX 0.10.1 and fix any remaining shard_map mismatches in a targeted way, keeping production tensor specs explicit/intended and only handling replicated aux inputs where needed.
So I suggest we hold off on merging this PR and instead fold the necessary pieces into the direct 0.10.1 upgrade path.

Aharrypotter:
Agreed. I'll close this PR and push the JAX upgrade directly against 0.10.1.

PR #1408, opened 2026-06-29
feat(mem_cache): page>=128 recurrent extra-buffer + pool sizing/admission

MokusMokun:
## Summary

Adds the production **page>=128 recurrent radix cache** path behind
`--enable-recurrent-extra-buffer`, plus root-cause recurrent pool sizing and admission
backpressure. Builds on the page=1 recurrent base (#1401, merged). Off by default; flag-off and
non-recurrent paths are unchanged.

Tracked by #1424 (under #1339 / #1338). Builds on #1401 (merged).

### Extra-buffer path (page>=128)
| Area | Change |
|---|---|
| `memory_pool.py` | Page-aligned ping-pong track slots (2 with overlap scheduling, 1 without) + request-owned-slot ledger and track-slot API. |
| `managers/schedule_batch.py`, `managers/schedule_policy.py` | Scheduler splits prefill at track-interval boundaries so the published snapshot is always the forward's final state (no mid-forward extraction). |
| `model_executor/forward_batch_info.py` | `recurrent_track_indices` / `recurrent_track_mask` metadata. |
| `kernels/gdn/gated_delta.py`, `attention/linear/{kda,gdn}_backend.py`, `attention/hybrid_linear_attn_backend.py`, `attention/linear/short_convolution.py` | Masked track scatter into the keep slot, wrapped in `optimization_barrier` (donated-buffer aliasing guard under multi-host SPMD). |
| `unified_cache_components/recurrent_component.py`, `unified_radix_cache.py` | Extra-buffer commit/donation + shortfall eviction. |
| `server_args.py`, `cache_init_params.py`, `kv_cache_builder.py`, `registry.py`, `base_prefix_cache.py` | `--enable-recurrent-extra-buffer` + `--recurrent-track-interval` (defaults to `chunked_prefill_size` to avoid the chunked-prefill chunk-size accuracy penalty); reject unsupported combinations at init (speculative decoding, mixed chunked prefill, GLA/Lightning, page<=1, disable-radix-cache). |

### Pool sizing / admission (over-subscription throttles instead of crashing)
| Area | Change |
|---|---|
| `model_executor/model_runner_kv_cache_mixin.py` | Split the back-cap reservation factor from actual per-request consumption; overlap-aware `request_owned_slots`; explicit snapshot-headroom contract (the auto/memory-ratio path reserves headroom and caps running by consumption; the explicit `--max-recurrent-state-size` path keeps the operator's max_running and warns on near-zero headroom). Fail-fast at startup when a rank has fewer recurrent slots than one request consumes. |
| `managers/scheduler.py` | Recurrent admission backpressure: defer a request when a rank's free + evictable recurrent slots cannot cover demand, instead of an `alloc_req_slots` crash. The check is match-aware — a cross-request prefix hit keeps its matched snapshot locked for the request's lifetime, so that slot is discounted from the evictable pool. Recurrent-aware diagnostic on the residual allocation error. |

### Benchmark
`benchmark/hicache/bench_recurrent_reuse_sweep.py` (reuse K-sweep + capacity model) and a
`unified-recurrent` A/B mode + external-server mode in `bench_unified_radix_ab.py`.

> The dense offline `--compare` gate is already on main via #1402; net-new in this PR is the recurrent A/B (`unified-recurrent`) and external-server modes of `bench_unified_radix_ab.py`.

## Live validation

| Area | Result |
|---|---|
| CPU `unit-test-cpu` (full suite) | **pass** — CI green on this head; 221 tests across the PR-touched files locally |
| Recurrent serving determinism (GDN single-host tp4/dp2; KDA multi-host tp16/dp4) | **bit-identical** — cache-ON == no-cache, byte-for-byte, under real boundary reuse (GDN 5120-tok / KDA 3584-tok warm hits) → cache reuse is logits-transparent; multi-host ep>1 caveat below |
| Reuse K-sweep + pool robustness (multi-host, size 192) | **no crash** — through K=256 (over-capacity throttles); predict_knee K\*=128 matches `size − request_owned_slots·parallel` |
| Admission backpressure / sizing | **throttles, never crashes** — over-subscription defers (no SIGQUIT); match-aware backpressure + fail-fast undersized-pool guard, unit-tested |
| Accuracy — gsm8k (GDN Qwen3.5-35B-A3B, 200q 5-shot greedy) | **0.865 == matched no-cache control 0.865, exactly equal** (no published reference — the model card reports HMMT/AIME, not gsm8k) |
| Accuracy — GPQA-Diamond (GDN Qwen3.5-35B-A3B, robust `simple_eval`, 5 seeds/arm) | **cache-ON median 0.8333 ≥ matched no-radix median 0.8283** (same seeds; per-seed spread 0.818–0.859; published 84.2) |
| Accuracy — MMLU-Pro (GDN Qwen3.5-35B-A3B, robust `simple_eval`, 4 shuffled seeds/arm) | **cache-ON 0.846 ± 0.009 == matched no-cache control 0.847 ± 0.007** (published 85.3; the ~0.7pp offset is harness-side and arm-symmetric), ~0% parse-failures |
| Accuracy — MMLU-Pro (KDA Kimi-Linear-48B-A3B, cache-ON full-set 12,032 q) | **0.6720 ≥ 0.6574 bring-up full-set baseline** (0.8% parse-failures; no protocol-comparable published number — the model card's 51.0 uses a 4k-ctx base-model protocol) |

The recurrent cache reuse is output-transparent by construction (greedy cache-ON == no-cache is byte-identical single-host and at ep=1 multi-host, and the CPU split-equivalence test proves the recurrent scan is chunk-invariant), so cache reuse does not change the model's logits. Matched no-radix controls (same seeds, same `simple_eval` harness) close the loop empirically: gsm8k is exactly equal, the GPQA cache-ON median sits above the control median, and the MMLU-Pro means are statistically identical — residual offsets vs the model-card numbers (85.3 MMLU-Pro / 84.2 GPQA for Qwen3.5-35B-A3B) are harness-side and affect both arms equally. The low readings seen during bring-up were eval-harness artifacts (max_tokens truncation of thinking-mode CoT + a strict `ANSWER:`-only extractor), not the cache; the extractor fragility is tracked in #1409. The `recurrent_track_interval = chunked_prefill_size` default removes the chunked-prefill chunk-size accuracy penalty (a smaller interval force-splits prefill and lowers GPQA ~3.5pp).

## Performance (A/B via this branch's `bench_unified_radix_ab.py`, single-host v6e-4, GDN Qwen3.5-35B-A3B tp4)

`--configs no-cache unified-recurrent --workloads gsp random --repeats 3` (gsp: 64 groups × 16 prompts sharing a 2048-token system prompt, max concurrency 64; random: 256 × 1024/128, no shared prefix; cache flushed between reps):

| workload | config | p50 TTFT (ms) | out tok/s | total tok/s | hit rate |
|---|---|---|---|---|---|
| gsp (shared prefix) | no-cache | 36811 | 111.7 | 2062 | — |
| gsp (shared prefix) | unified-recurrent | **8095 (0.22×)** | **409.5** | **7559 (+267%)** | 0.80 |
| random (no sharing) | no-cache | 17270 | 222.6 | 2004 | — |
| random (no sharing) | unified-recurrent | 17190 | 224.3 | 2019 (+0.7%) | 0 |

Shared-prefix traffic gets 4.5× lower p50 TTFT and +267% throughput at a 0.80 hit rate; traffic with no shared prefix is overhead-neutral. A lighter-load run through the nightly A/B suite (parallel 16) corroborates: p50 TTFT ratio 0.100, hit rate 0.8020, random throughput +14.4%.

## Known limitations / next work
- Multi-host `ep>1` (GDN tp16/dp4/ep4): cache resume is within-bf16-numerics of full recompute but not byte-identical (shape-dependent argmax flip; pre-existing — reproduced identically on the pre-PR tree via a flush cold-vs-warm test). Single-host and `ep=1` multi-host are byte-identical.
- Production recurrent cache path is page>=128 with `--enable-recurrent-extra-buffer`; the page=1
  base path (#1401) remains for single-tiling KL==0.
- An explicit `--max-recurrent-state-size` still needs headroom for reuse; oversubscription
  throttles rather than crashing.
- Cache-aware DP routing (the reuse-rate uplift) is a separate PR (#1402); with the default
  `min_running` policy the per-rank reuse plateau is the ~1/dp floor.
- The bring-up accuracy dip traced to MCQ answer-extraction, not the cache. The repo's `simple_eval` MCQ extractors for `gsm8k`/`mgsm`/`math` share a strict-`ANSWER:` fragility for reasoning models (no `strip_reasoning`); tracked separately in #1409.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -23,7 +23,7 @@ import logging
 import os
 import threading
 from http import HTTPStatus
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar, NamedTuple
 
 import jax
 import numpy as np
@@ -232,6 +232,13 @@ class Req:
         # One-shot recurrent CoW src slot: set on a fresh match, consumed at the
         # next extend forward, cleared everywhere else (re-match, merge, retract).
         self.recurrent_cow_src_index: int | None = None
+        # Extra-buffer ping-pong track slots holding materialized page-boundary
+        # snapshots (dormant unless enable_recurrent_extra_buffer).
+        self.recurrent_ping_pong_track_buffer: list[int] | None = None
+        self.recurrent_next_track_idx: int | None = None
+        # Watermark seqlen of the last materialized track slot; set ONLY by a real
+        # track scatter, never from a prefix match.
+        self.recurrent_last_track_seqlen: int | None = None
 
         # Check finish
         self.tokenizer = None
@@ -460,10 +467,19 @@ class Req:
                 self.last_host_node = tree_cache.root_node
                 self.host_hit_length = 0
             else:
+                # A running recurrent req already OWNS its state (running slot):
+                # no re-clone (cow_recurrent=False), FULL-only match -- its
+                # unfinished prefix is not gated on recurrent-tree validation.
+                is_running_recurrent = (
+                    tree_cache.supports_recurrent() and self.recurrent_pool_idx is not None
+                )
                 match_result = tree_cache.match_prefix(
                     MatchPrefixParams(
                         key=RadixKey(self.adjust_max_prefix_ids(), self.extra_key, self.dp_rank),
-                        cow_recurrent=tree_cache.supports_recurrent(),
+                        cow_recurrent=(
+                            tree_cache.supports_recurrent() and not is_running_recurrent
+                        ),
+                        full_only=is_running_recurrent,
                         req=self,
                     )
                 )
@@ -647,6 +663,9 @@ class Req:
         self.cache_protected_len = 0
         self.recurrent_pool_idx = None
         self.recurrent_cow_src_index = None
+        self.recurrent_ping_pong_track_buffer = None
+        self.recurrent_next_track_idx = None
+        self.recurrent_last_track_seqlen = None
 
     def set_finish_with_abort(self, error_msg: str):
         # set it to one token to skip the long prefill
@@ -723,6 +742,10 @@ class ScheduleReqsInfo:
     # Recurrent CoW src slot per req (0 = no clone), per DP
     recurrent_cow_src_indices: np.ndarray | None = None
 
+    # Recurrent track metadata per DP (extra-buffer; set only on a track boundary).
+    recurrent_track_indices: np.ndarray | None = None
+    recurrent_track_mask: np.ndarray | None = None
+
 
 def _build_recurrent_cow_src_indices(reqs: list[Req]) -> np.ndarray | None:
     """None when no clone is pending, so cold extends skip the donated-buffer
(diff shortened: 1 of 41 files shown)


python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -2842,6 +2970,38 @@ def get_model_worker_batch(
             ):
                 recurrent_cow_src_indices_cpu = None
 
+        # Step 5.5c: Merge recurrent track metadata (extra-buffer snapshot at
```
cjx0709: maybe we should not let the step notes in our pr code

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -2842,6 +2970,38 @@ def get_model_worker_batch(
...
+            for info in self.reqs_info:
+                info.recurrent_track_indices = None
+                info.recurrent_track_mask = None
+            # Backstop: no boundary hit this batch -> drop both to None so
+            # the backend skips the snapshot path entirely.
+            if not recurrent_track_mask_cpu.any():
+                recurrent_track_indices_cpu = None
+                recurrent_track_mask_cpu = None
+
         # Step 5.6: has_initial_state[i] = True iff slot i already holds
```
cjx0709: also like comment before

python/sgl_jax/srt/managers/schedule_batch.py
MokusMokun: Done in 808bbd93 — dropped the step numbering from both comments, and swept the rest of the diff for the same class: no other flow-numbering or internal labels remain (also replaced two PR-body "Next Work" comment references with the tracking issue #1425). The remaining `Step 1-7` notes in this function are pre-existing main code.

python/sgl_jax/srt/managers/schedule_batch.py
MokusMokun: Done in 808bbd93.

cjx0709:
I took another pass through the logic and the public/shared surfaces touched by this PR. I do not see any remaining blocking issues from my side: the extra-buffer path is guarded behind the new flag, the default/off path stays well isolated, and the recurrent pool/admission/accounting changes look consistent with the added tests and green CI.

One non-blocking cleanup suggestion: the code comments are useful, especially where they explain ownership, donation, admission backpressure, and boundary-splitting invariants, but the PR is a bit comment-heavy now. I would consider trimming some repeated explanations and keeping only the comments that protect non-obvious invariants or dangerous future edits. In particular:

- Keep the “why” comments around slot ownership, replacement-before-donation, locked snapshot admission, and `optimization_barrier` usage.
- Consolidate repeated recurrent track metadata explanations into one field/function-level comment, and keep later call sites shorter.
- Move benchmark/accuracy rationale from inline comments into the PR description or a design note where possible; in code, a shorter statement like “default to chunked_prefill_size to avoid adding extra prefill split boundaries” should be enough.
- Drop comments that restate immediately obvious control flow, such as simple decode/boundary checks or one-line slot consumption statements.

So overall: logic LGTM from me; comment density could be reduced as a follow-up polish item, but I would not block the PR on it.

MokusMokun:
Done in 0c10e194 (plus a second stricter pass) — track-metadata explanations consolidated to the field/function declarations, accuracy rationale moved out of `server_args`, obvious control-flow / one-liner comments deleted, new test module docstrings cut to 1-4 lines, and the ownership / donation / admission / `optimization_barrier` why-comments kept. Net: added comment lines in `srt/` down ~30%. Also rebased onto latest main.

CI is green on 0c10e194 — @cjx0709

cjx0709 (approved):
LGTM

PR #1411, opened 2026-06-29
feat(hicache): integrate offload into UnifiedRadixCache

### LINKED ISSUE #1312: [RFC] HiCache stage 2: HiCacheController — L1↔L2 async pipeline
## Summary

Implement `HiCacheController`, the control plane for L1 (HBM) <-> L2 (host pinned) KV page transfers:

- **D2H write** (async): when a tree cache node is reused (hit), asynchronously copy its KV from L1 to L2
- **H2D load** (sync): when prefill discovers an L1 miss but L2 hit, synchronously copy from host pool back to device pool
- **Eviction coordination**: when L2 is full, the tree cache layer initiates eviction, and the controller executes slot release (this stage does not connect L3 — eviction discards data)

> **Interface alignment** — Transfers go exclusively through the `HostKVPool` `buffer_id`-addressed primitives (`copy_into` / `copy_to_device`); the controller does **not** call `jax.device_put` itself. The A/B transfer format is now an implementation detail of the pool, not the controller. See the decision record `docs/design/hicache_pd_raiden_alignment.md` (D4/D5). The shared ABC (#1368) is `reserve()/release(buffer_id)/copy_from_device(layers, buffer_id)`; HiCache reserves a slot and fills it via `copy_into`. This keeps the controller raiden-switchable with zero changes (see §8).

## Position in the roadmap

1. Host pool foundation
2. UnifiedRadixCache + FullComponent port
3. **HiCacheController + async pipeline** *(this issue)*
4. End-to-end integration and validation
5. L3 persistent storage
6. Multi-component + tuning

Depends on Stage 0 (`LRUHostKVPool` + ABC transfer primitives) and Stage 1 (UnifiedRadixCache + FullComponent).

## Proposed structure changes

- `sgl_jax/srt/mem_cache/hicache_controller.py` — **new**: `HiCacheController`
- `sgl_jax/test/mem_cache/test_hicache_controller.py` — **new**: unit tests
- `sgl_jax/test/disagg_feasibility/t_hicache_d2h_format.py` — **new**: D2H format benchmark (spike; decides the `LRUHostKVPool` internal format, not the controller interface)
- `sgl_jax/test/disagg_feasibility/t_hicache_host_lifecycle.py` — **new**: host memory lifecycle validation (spike)

### 1. Data organization

**Device side (L1)**

The device pool (`MHATokenToKVPool`) stores per-layer. `kv_buffer` is `list[jax.Array]` of length `num_layers`, each with shape:

```text
kv_buffer[layer_id]: (num_pages, page_size, kv_heads*2//packing, packing, head_dim)
                      ^^^^^^^^                ^^^^^^^^^^^^^^^^^^^
                      DP shard ("data")       TP shard ("tensor")
```

**Host side (L2)**

`LRUHostKVPool` (Stage 0) manages host pinned slots. The controller **never touches** a slot's internal `jax.Array` — it holds only `int buffer_id` and moves data via `host_pool.copy_into` / `copy_to_device`. Host layout (per-slot single array vs list, all-layers-packed or not) is a pool implementation detail invisible to the controller.

**D2H / H2D transfer format (pool-internal detail)**

The transfer format (all-layers-packed vs per-layer direct) is **fully localized inside `LRUHostKVPool.copy_into` / `copy_to_device` (D5)** and does not enter the controller interface:

- **A (all-layers-packed):** pool-internal `jnp.stack` then bulk move; fewer calls, temp HBM overhead.
- **B (per-layer direct):** pool-internal per-layer move; no temp HBM, more calls.

Final choice is decided by the Stage 0 `LRUHostKVPool` benchmark. **Regardless of choice, the controller interface is unchanged**: it always calls `copy_into([device_idx], [buffer_id])` / `copy_to_device([buffer_id], [device_idx])`. Under raiden the host layout is engine-decided, making this moot for the controller.

### 2. Interface

```python
class HiCacheController:
    def __init__(self, host_pool: HostKVPool, device_pool: KVCache,
                 device_allocator: BaseTokenToKVPoolAllocator):
        self._executor = ThreadPoolExecutor(max_workers=1)
        self._pending: list[Future] = []

    def write(self, device_indices, host_buffer_ids) -> None:
        """Async D2H: copy device page(s) into host pool slot(s).
        device_indices[i] pairs with host_buffer_ids[i] (int buffer_id from reserve())."""

    def load(self, host_buffer_ids, device_indices) -> None:
        """Sync H2D: copy host pool slot(s) back to device pool page(s)."""

    def evict_callback(self, host_buffer_ids) -> None:
        """L2 eviction callback. This stage frees slots; Stage 4 connects L3."""

    def check_write_status(self) -> None:
        """Non-blocking poll: clean up completed D2H futures, release node locks."""

    def drain_pending(self) -> None:
        """Blocking wait for all pending D2H (e.g. before L1 eviction, shutdown)."""

    def pending_count(self) -> int: ...
```

`write`/`load` **do not call `jax.device_put` themselves** — they delegate to the `HostKVPool` `buffer_id`-addressed primitives:

```python
def _do_d2h(self, device_indices, host_buffer_ids):
    self._host_pool.copy_into(list(device_indices), list(host_buffer_ids))

def _do_h2d(self, host_buffer_ids, device_indices):
    self._host_pool.copy_to_device(list(host_buffer_ids), list(device_indices))
```

This is where the switchability contract (D4) lands in the controller: **the controller only ever sees `int buffer_id`, never a host `jax.Array`.** The actual movement (`jax.device_put`, host layout, format A/B) is sealed inside the pool. It does not reuse `KVCache.get_cpu_copy` / `load_cpu_copy` (which go `jax.device_get` -> numpy, non-pinned, non-async).

### 3. Async execution model

D2H uses `ThreadPoolExecutor` (1 worker). JAX's `device_put` releases the GIL at the C++ level, so D2H can run in parallel with the main thread's forward.

```python
def write(self, device_indices, host_buffer_ids):
    future = self._executor.submit(self._do_d2h, device_indices, host_buffer_ids)
    self._pending.append(future)

def check_write_status(self):
    done = [f for f in self._pending if f.done()]
    for f in done:
        f.result()  # raise exceptions if any
        self._pending.remove(f)

def drain_pending(self):
    for f in self._pending:
        f.result()
    self._pending.clear()
```

H2D (`load`) runs synchronously on the main thread — ~45.7 GB/s, microsecond-range per page. It also delegates to `host_pool.copy_to_device`; the controller does no direct `device_put`.

### 4. D2H write strategy

D2H is not triggered on every insert. `UnifiedRadixCache` maintains `hit_count` per node; when a node is **reused** (`match_prefix` hit) and `hit_count >= write_through_threshold`, `write_backup` (D2H) triggers.

```text
insert creates new node -> hit_count = 0, no D2H
subsequent match_prefix hit -> hit_count += 1
hit_count >= threshold -> _inc_hit_count triggers write_backup -> controller.write()
```

`write_through_threshold` defaults to 1. This means: decode does not D2H every step; only reused pages are worth D2H; partially-filled pages do not D2H.

After submission nodes are locked; `check_write_status()` polls completion, then releases locks and marks `node.backuped = True` (called before `get_new_batch_prefill`).

### 5. H2D load timing

H2D (prefetch) occurs during prefill batch construction. When `match_prefix` returns `last_host_node` (L1 miss but L2 hit):

```text
get_new_batch_prefill
  +- check_write_status()  -> non-blocking poll D2H completion
  +- req.init_next_round_input -> match_prefix
       +- if host_hit_length > 0:
            +- device_indices = allocator.alloc(host_hit_length)
            +- controller.load(host_buffer_ids, device_indices)  -> sync H2D
            +- update tree node: value = device_indices, host_value = None
```

### 6. Step-level timing overview

```text
Step N-1 forward completes
  +- process_batch_result -> insert -> _inc_hit_count -> threshold
       +- write_backup -> controller.write(device_idx, [buffer_id]) -> async D2H
Step N: get_next_batch_to_run
  +- check_write_status()  -> non-blocking poll
  +- get_new_batch_prefill -> match_prefix -> [if L2 hit] controller.load() -> sync H2D
  +- run_batch -> forward (D2H runs async in background)
```

**D2H latency estimates** (Qwen3-8B, page_size=1, bf16):

| Scenario | Pages | Size | D2H (@2.8 GB/s) | Forward |
|---|---|---|---|---|
| decode step (single page hit) | 1 | ~128 KB | ~0.045 ms | ~10-20 ms |
| prefill 64 tokens hit | 64 | ~8 MB | ~2.9 ms | ~15-30 ms |
| prefill 2048 tokens hit | 2048 | ~256 MB | ~91 ms | ~200-500 ms |

`drain_pending` is only called at consistency-critical points.

### 7. Lifecycle

```text
init_hicache():
  host_pool = LRUHostKVPool(device_pool, pool_size, page_size, layer_num, per_layer_shape, dtype, mesh, partition_spec)
  controller = HiCacheController(host_pool, device_pool, allocator)

Running:
  insert -> _inc_hit_count -> threshold:
    buffer_id = host_pool.reserve()                -> buffer_id
    controller.write(device_indices, [buffer_id])  -> async D2H
    tree_node: lock, host_value = buffer_id
  check_write_status(): D2H done -> unlock, node.backuped = True
  match_prefix -> L1 miss, L2 hit:
    controller.load(host_buffer_ids, device_indices)  -> sync H2D
  L2 full (reserve returns None):
    FullComponent.drive_host_eviction -> controller.evict_callback([buffer_id])
      -> host_pool.release(buffer_id); tree_node.host_value = None
Shutdown: controller.drain_pending()
```

Throughout, the upper layer sees only `int buffer_id`; no host `jax.Array` is exposed.

### 8. Switchability: raiden migration

The controller's contract (D4): **only `int buffer_id` + device index, moving via `host_pool.copy_into` / `copy_to_device`, never touching a host `jax.Array` or calling `jax.device_put` directly.** So at raiden time:

| Change | Location | Controller impact |
|---|---|---|
| Add `RaidenHostKVPool(HostKVPool)`, `copy_into→raiden.d2h`, `copy_to_device→raiden.h2d` | Stage 0 (host pool backend) | **zero** |
| `init_hicache` swaps `LRUHostKVPool` → `RaidenHostKVPool` | wiring/config | **zero** |

`HiCacheController`, `FullComponent` hooks, `UnifiedRadixCache` are all untouched.

### Spike 1: D2H transfer format benchmark

Compare approach A (all-layers-packed) vs B (per-layer direct) on v6e-4 (Qwen3-8B: 32 layers, 8 kv_heads, 128 head_dim, bf16). **This decides the `LRUHostKVPool` internal format (Stage 0), not the controller interface** — the controller always uses the `buffer_id` primitives.

1. Single page D2H latency (median of 100); 2. Batch D2H (64/256 pages); 3. Device HBM peak increment; 4. D2H + forward overlap.

| Metric | Choose A if | Choose B if |
|---|---|---|
| Single page D2H latency | A >30% lower | A no advantage / slower |
| HBM peak increment | stack temp < 1% total HBM | unacceptable |
| Overlap | both overlap | — |

### Spike 2: Host pinned memory lifecycle validation

Verify managing host pinned memory via JAX (create via `device_put`, reclaim via GC) works: reference hold/release, batch release, repeated alloc/free (1000x, RSS bounded), multi-slot coexistence, long-term data integrity. Key concerns: delayed reclamation/fragmentation for `pinned_host`, TPU runtime alloc/free frequency limits, many small vs few large array GC behavior. Env: v6e-4. Code in `sgl_jax/test/disagg_feasibility/`.

## Acceptance criteria

- **Spike: D2H format** — A/B benchmark complete with clear data
- **Spike: host memory lifecycle** — all items pass
- write correctness: D2H then `copy_to_device` round-trip data allclose with device original
- load correctness: H2D then device KV buffer allclose with host original
- Does not expose jax.Array: controller public method signatures contain only `int buffer_id` / device index
- Async: `write` returns immediately without blocking
- check_write_status / drain_pending: correct poll / drain behavior
- Overlap: forward + D2H total time < forward + D2H
- Round-trip: D2H -> H2D data allclose with original
- evict_callback: after eviction, slot freed, `available_size()` increases

### LINKED ISSUE #1313: [RFC] HiCache stage 3: end-to-end integration and validation
## Summary

Wire the modules from Stage 0-2 (`HostKVPool`, `UnifiedRadixCache` + `FullComponent`, `HiCacheController`) end-to-end. This stage adds the complete **control plane** — the decision layer that determines when to transfer, which nodes to transfer, and how to handle resource exhaustion — on top of the Stage 2 controller (which only handles the "how to transfer" execution).

Specifically:
1. Replace `FullComponent` hook stubs with real implementations
2. Add control plane methods to `UnifiedRadixCache` (write_backup, load_back, eviction paths, lock management, async event management)
3. Integrate with the scheduler at three call sites
4. Validate correctness (KL equivalence) and performance (forward latency regression)

## Position in the roadmap

1. Host pool foundation
2. UnifiedRadixCache + FullComponent port
3. HiCacheController + async pipeline
4. **End-to-end integration and validation** *(this issue)*
5. L3 persistent storage
6. Multi-component + tuning

This issue depends on all three prior stages. Stage 4 (L3 storage ABC) can start in parallel.

## Proposed structure changes

- `sgl_jax/srt/mem_cache/unified_radix_cache.py` — **modify**: add `write_backup`, `load_back`, `writing_check`, `loading_check`, `check_hicache_events`, `flush_write_through_acks`, `init_load_back`, eviction paths (`_evict_device_leaf`, `_evict_to_host`, `_evict_host_leaf`, `evict_host`), leaf set management, lock mechanism
- `sgl_jax/srt/mem_cache/unified_cache_components/full_component.py` — **modify**: stub -> real hook implementation (`build_hicache_transfers`, `commit_hicache_transfer`, `evict_component`, `drive_eviction`, `drive_host_eviction`, `acquire_component_lock`)
- `sgl_jax/srt/mem_cache/kv_cache_builder.py` — **modify**: call `init_hicache()` in builder
- `sgl_jax/srt/managers/scheduler.py` — **modify**: add `check_hicache_events` + `init_load_back` in `get_new_batch_prefill`; add `flush_write_through_acks` in `update_running_batch`
- `sgl_jax/test/mem_cache/test_hicache_e2e.py` — **new**: end-to-end tests

### 1. Control plane overview

The control plane is distributed across three modules:

| Module | Responsibility |
|---|---|
| `UnifiedRadixCache` | Node-level D2H/H2D decisions, eviction paths, async event management |
| `FullComponent` | Resource release, heap-sorted eviction, lock management, transfer construction |
| `Scheduler` | Step-level scheduling entry points |

Stage 2 already provides the controller layer (`write`, `load`, `check_write_status`, `drain_pending`, `evict_callback`). Stage 3 adds the following control plane on top.

### 2. write_backup: D2H write decision

`write_backup` is the D2H **decision layer** (controller.write is the execution layer):

1. **Parent invariant** (Full attention only): under write_through, parent must be backed up before child, enforced recursively. This ensures prefetch can find complete host data along the parent chain. This invariant relies on Full attention's prefix semantics — SWA's window-external parent KV has no meaning, and linear attention has no traditional KV cache. Backup strategy must be re-evaluated when extending to these architectures.
2. **Host space pre-eviction**: when space is insufficient, call `evict_host()` to free host leaf nodes; abandon if still insufficient.
3. **Node locking**: `inc_lock_ref` after submission to prevent eviction during in-flight D2H. Lock released by `writing_check` upon completion.
4. **Trigger condition**: called by `_inc_hit_count` when `hit_count >= write_through_threshold` (Stage 2 §5.1).

### 3. load_back: H2D load decision

`load_back` is the H2D decision layer:

1. **Transfer construction**: walk up from `best_match_node` collecting host_indices of all evicted+backuped nodes.
2. **Device space pre-eviction**: evict device leaf nodes when device pool is insufficient; abandon if still insufficient. HiCache does not reserve device pool space — reservation relies on the scheduler's existing `new_token_ratio` mechanism (decode headroom).
3. **mem_quota check**: scheduler-provided memory quota limits single load size to prevent device pool exhaustion.
4. **Path locking**: `inc_lock_ref` locks the device path from root to best_match_node; `loading_check` releases after completion.
5. **Commit**: write device_indices back to node after H2D, clear host_value, mark as non-evicted.

### 4. Device eviction paths

`_evict_device_leaf` selects a path based on backup state:

| Condition | Path | Description |
|---|---|---|
| `backuped == True` | `_evict_to_host` | Release device resources, demote node to host-only (stays in tree) |
| `not backuped` + `write_through` | cascade delete | No backup and no remedy — completely remove from tree |

`_evict_to_host`: call `evict_component(target=DEVICE)` to free device slot -> mark `evicted = True` -> update leaf set.

`drive_eviction` (FullComponent): heap-sort from `evictable_device_leaves` by priority, evicting one by one until enough tokens are freed.

### 5. Host eviction path

Triggered when `host_pool.alloc()` returns None:

1. `drive_host_eviction` (FullComponent): heap-sort traverse `evictable_host_leaves`
2. `_evict_host_leaf`: atomically release host resources, remove node from tree, cascade-delete tombstone parents

**Leaf set management**: maintain `evictable_device_leaves` and `evictable_host_leaves` sets, updated after every eviction/load/backup.

| Type | Criteria |
|---|---|
| Device leaf | non-root, non-evicted, `lock_ref == 0`, no children holding device value |
| Host leaf | non-root, evicted, backuped, `host_lock_ref == 0`, no children |

### 6. Lock mechanism

Two-layer locking prevents node eviction during async transfers:

| Lock | Protects | Lock time | Release time |
|---|---|---|---|
| `lock_ref` | device KV data | write_backup submission, load_back path locking | `writing_check` / `loading_check` completion |
| `host_lock_ref` | host KV data | load_back host node locking | `loading_check` completion |

`acquire_component_lock`: walk from node up to root, skip evicted segments, `lock_ref += 1` for device-on nodes and remove from leaf set. First version H2D is synchronous, so `host_lock_ref` is not needed yet.

### 7. Async event management

On top of Stage 2's `check_write_status` (future poll), add tree state updates:

- **`writing_check`**: mark `node.backuped = True` on D2H completion, release `lock_ref`, update leaf set
- **`loading_check`**: release `lock_ref` on H2D completion

Poll uses `Future.done()` non-blocking check.

### 8. FullComponent hook implementation

Replace Stage 1 stub hooks with real implementations:

- **`build_hicache_transfers`**: BACKUP_HOST phase returns device_indices; LOAD_BACK phase walks parent chain collecting evicted nodes' host_indices
- **`commit_hicache_transfer`**: BACKUP_HOST writes host_value; LOAD_BACK restores device value and clears host_value
- **`evict_component`**: release device slot and/or host slot by target

### 9. Scheduler integration

```text
Scheduler.event_loop_normal
  +- get_next_batch_to_run
       +- get_new_batch_prefill
       |    +- [entry] check_hicache_events()  -> non-blocking poll D2H/H2D completion
       |    +- init_next_round_input -> match_prefix
       |    |    +- [if L2 hit] init_load_back -> load_back -> controller.load()
       |    +- [tail] ready_to_load_host_cache()  -> first version H2D is sync, no-op here
       +- update_running_batch
       |    +- flush_write_through_acks()  -> early release D2H locks, increase evictable space
       +- run_batch -> forward (D2H runs async in background)
```

`_inc_hit_count` -> `write_backup` -> `controller.write()` is triggered automatically by insert, no additional scheduler call needed.

### 10. Initialization assembly

`init_hicache()` is called in `kv_cache_builder.py` after creating `UnifiedRadixCache` (when `hicache_storage != "disable"`). It creates `HostKVPool` and `HiCacheController`, and wires them to the cache and FullComponent.

### 11. Stage 3 scope summary

| Component | Stage 2 provides | Stage 3 adds |
|---|---|---|
| D2H execution | `controller.write()` | — |
| D2H decision | — | `write_backup` (parent invariant, host pre-eviction, locking) |
| H2D execution | `controller.load()` | — |
| H2D decision | — | `load_back` (device pre-eviction, mem_quota, path locking, commit) |
| D2H poll | `check_write_status()` | `writing_check` (+ tree state update + lock release) |
| H2D poll | — | `loading_check` (lock release) |
| Device eviction | — | `_evict_device_leaf`, `_evict_to_host`, `drive_eviction` |
| Host eviction | `evict_callback` (slot free) | `_evict_host_leaf`, `drive_host_eviction`, leaf set management |
| Lock | — | `inc_lock_ref` / `dec_lock_ref`, `acquire_component_lock` |
| FullComponent hooks | stub | `build_hicache_transfers`, `commit_hicache_transfer`, `evict_component` |
| Scheduler entry | — | `check_hicache_events`, `flush_write_through_acks`, `init_load_back` |
| Init | — | `init_hicache` + builder integration |

## Testing

Stage 1 covers basic radix cache correctness. Stage 3 focuses on HiCache-specific scenarios; the full test matrix will be refined during implementation.

### KL equivalence test

| Config | Value |
|---|---|
| Model | Qwen3-8B |
| Hardware | v6e-1 |
| Control | Existing `RadixCache` (no HiCache) |
| Experiment | `UnifiedRadixCache` + `--hicache-storage none` |
| Metric | All output token logits KL divergence < 1e-6 |
| Test set | Multi-turn conversation (triggers prefix reuse + eviction + prefetch) |

### HiCache-specific test scenarios

| Scenario | Validation point |
|---|---|
| L1 eviction -> L2 hit -> prefetch | H2D correctly restores KV, skips prefix forward |
| L1 + L2 both evicted -> full miss | Falls back to complete forward, results correct |
| write_through trigger condition | Only nodes with `hit_count >= threshold` trigger D2H |

### Performance validation

| Metric | Threshold |
|---|---|
| Single step forward latency increase | < 2% |
| Measurement | 100 step average (excluding first XLA compilation) |
| Hardware/model | Qwen3-8B @ v6e-1 |

## Acceptance criteria

- KL equivalence: Qwen3-8B multi-turn conversation, logits KL < 1e-6
- Latency regression: forward latency increase < 2%
- HiCache-specific test scenarios pass
- No memory leak: host pool `available_size()` stable during sustained operation
- write_through correctness: only nodes with `hit_count >= threshold` trigger D2H

cjx0709:
## Summary

Integrate HiCache offload control plane into `UnifiedRadixCache`: page-level H2D/D2H scheduling over `LRUHostKVPool` (Stage 0, #1396), `HiCacheController` single-worker async dispatch, write policies (write-through / write-back), tombstone demotion & revival, donation-safe eviction barrier, and full scheduler/server_args wiring.

Closes #1312
Closes #1313

## What changed

| Area | Change |
|---|---|
| `unified_radix_cache.py` | `_inc_hit_count` / `_evict_device_leaf` / `write_backup` / `_donation_barrier` |
| `hicache_controller.py` | Single-worker `ThreadPoolExecutor`, async write/load, inflight guards, evict callback |
| `kv_cache_builder.py` | `init_hicache()` assembles `LRUHostKVPool` + `HiCacheController` |
| `scheduler.py` | `_wait_donation_safe` barrier for overlap + HiCache |
| `schedule_policy.py` | HiCache overlap scheduling + chunked-budget guard before H2D |
| `server_args.py` | `--hicache-storage`, `--hicache-ratio`, `--hicache-write-policy`, `--hicache-write-through-threshold` |

## Review fixes

All 10 review comments from @MokusMokun addressed. 8 fixed, 2 deferred to #1429.

| # | Issue | Status |
|---|-------|--------|
| 1 | inflight registration after `submit()` race | Fixed |
| 2 | `_evict_host_leaf` raises on inflight pages | Fixed |
| 3 | `init_load_back` before chunked rejection | Fixed |
| 4 | `reset()` doesn't drain controller | Fixed |
| 5 | `write_backup` stale `host_value` on failure | Fixed |
| 6 | cascade delete detaches backed-up subtree | Fixed |
| 7 | dead `_full_kv_pool_host` attribute | Fixed |
| 8 | missing equivalence/overlap/dp-rank tests | Fixed (+10 tests) |
| 9 | unbounded `_pending` | → #1429 |
| 10 | sync `block_until_ready` in write-through | → #1429 |

## New tests added

| Class | Tests | Coverage |
|-------|-------|----------|
| `TestHiCacheEquivalence` | 2 | Dual-run HiCache ON vs OFF match agreement |
| `TestOverlapE2E` | 3 | Donation barrier fires, full overlap cycle bit-exact, None barrier safe |
| `TestDpRankMapping` + `Page4` | 6 | `_to_global_device_pages` dp_rank=0/1 mapping, page_size=1+4 |

All tests in CI via `unit-test-cpu` suite.

### L1 — CPU unit tests (271 passed, 0 failed)

```bash
PYTHONPATH=python python -m pytest python/sgl_jax/test/mem_cache/ -v
```

### L2 — TPU (v6e-4, 15/15 passed)

```bash
PYTHONPATH=python python -m pytest python/sgl_jax/test/mem_cache/test_host_kv_pool_tpu.py -v
PYTHONPATH=python python -m pytest python/sgl_jax/test/mem_cache/test_host_kv_pool_tpu_dp.py -v
PYTHONPATH=python python -m pytest python/sgl_jax/test/mem_cache/test_hicache_e2e_tpu.py -v
```

### L3 — E2E benchmark (Qwen3-8B, v6e-4)

| Configuration | Duration | Throughput | Success | Errors |
|---|---|---|---|---|
| HiCache ON | 89.92 s | 1.07 req/s | 96/96 (100%) | 0 |
| HiCache OFF | 91.46 s | 1.05 req/s | 96/96 (100%) | 0 |


python/sgl_jax/srt/mem_cache/unified_radix_cache.py
```diff
@@ -660,11 +857,144 @@ def _is_device_leaf(self, node: UnifiedTreeNode) -> bool:
...
+            if parent is not None and self._is_host_leaf(parent):
+                heapq.heappush(heap, parent)
+        return num_freed
+
+    def _evict_host_leaf(self, node: UnifiedTreeNode) -> int:
+        """Release a host leaf's page ids and delete it from the tree."""
+        cd = node.component_data[BASE_COMPONENT_TYPE]
+        buffer_ids = cd.host_value
+        num = len(buffer_ids)
+        self.hicache_controller.evict_callback([int(b) for b in buffer_ids])
```
MokusMokun: `_evict_host_leaf()` always calls `evict_callback()`, and `evict_callback()` raises if the page is still in `_inflight` / `_inflight_load`. Since `write_backup()` publishes `host_value` before the D2H future completes — and the single-worker backlog stretches that in-flight window arbitrarily, not just per-transfer milliseconds — ordinary host-pool pressure can select that page as an LRU host victim and turn cache pressure into an uncaught scheduler exception. Make `evict_host()` skip or wait on in-flight victims instead of using the raise as the pressure signal.

python/sgl_jax/srt/managers/schedule_policy.py
```diff
@@ -628,6 +633,23 @@ def add_one_req(self, req: Req):
                 swa_needed = self._swa_budget_for_req(req.extend_input_len, dp_rank)
                 if swa_needed >= self.rem_swa_tokens_for_dp(dp_rank):
                     return AddReqResult.NO_TOKEN
+
+            # HiCache: after budget gate, pull host-only prefix back to device.
+            # Must happen after NO_TOKEN check so rejected reqs never trigger H2D.
+            if getattr(self.tree_cache, "hicache_enabled", False) and req.host_hit_length > 0:
+                mem_quota = self.token_to_kv_pool_allocator.available_size(dp_rank)
+                new_indices, last_node, flush_plan = self.tree_cache.init_load_back(
```
MokusMokun, line 652: `init_load_back()` allocates device pages, un-tombstones nodes, and queues H2D work before the chunked branch can still reject with `trunc_len <= 0`. The scheduler then still harvests `adder.pending_h2d`, so a rejected request can scatter into live, evictable, unlocked pages. Rather than patching individual gates, either size admission with a side-effect-free estimate (`extend_input_len - host_hit_length`) before any load-back, or make load-back transactional and roll it back on rejection; holding the reload-chain lock until `finish_load_back()` would also close the window.

python/sgl_jax/srt/mem_cache/hicache_controller.py
```diff
@@ -0,0 +1,188 @@
...
+        """Async D2H: copy device pages into reserved host page slots.
+
+        The device-side gather (stage_backup) runs synchronously on the caller's
+        thread because the forward path donates kv_buffer every step — reading it
+        from the worker would race that deletion. Only the slow host transfer
+        (flush_backup) is offloaded.
+        """
+        host_buffer_ids = list(host_buffer_ids)
+        self._host_pool.stage_backup(list(device_indices), host_buffer_ids)
+        future = self._executor.submit(self._do_d2h, host_buffer_ids)
```
MokusMokun: Both `write()` and `stage_load()` submit work before adding ids to `_inflight` / `_inflight_load`. This is mostly latent in the current single scheduler-thread model, but if a worker ever finishes before registration, its cleanup removes nothing and the ids are later marked busy forever. Register before `submit()` and roll back that registration if submission fails.

python/sgl_jax/srt/mem_cache/unified_radix_cache.py
```diff
@@ -135,6 +138,16 @@ def __init__(
...
+        self.host_pool = None
+        self.write_through_threshold: int = 1
+        self.write_policy: str = "write_through"
+        # Injected by the overlap scheduler so write_back's eviction-time D2H
+        # gather waits for the prior forward's replace_all (donation-safe).
+        self._donation_barrier = None
+
         self.reset()
 
     def reset(self):
```
MokusMokun, line 158: `reset()` clears `evictable_host_leaves` and `ongoing_write` but does not drain the controller or free pages still allocated in `host_pool`. After `flush_cache()`, the tree forgets those pages while the host pool still considers them allocated. Drain pending transfers and free all old `host_value` pages, or recreate the host pool/controller on reset.

python/sgl_jax/srt/mem_cache/hicache_controller.py
```diff
@@ -0,0 +1,188 @@
...
+        self,
+        host_pool: HostKVPool,
+        device_pool: Any,
+        device_allocator: Any = None,
+    ) -> None:
+        self._host_pool = host_pool
+        self._device_pool = device_pool
+        self._device_allocator = device_allocator
+        self._executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="hicache-d2h")
+        self._pending: list[Future] = []
```
MokusMokun, line 35: `_pending` is unbounded, and each `write()` leaves materialized staging arrays in `LRUHostKVPool._pending_gather` until the single worker finishes `flush_backup()`. A write-through hit burst can grow HBM staging pressure without limit and also stretches the in-flight window behind the host-eviction crash. Cap pending pages or bytes, and skip/defer write-through backups once the cap is reached.

python/sgl_jax/srt/mem_cache/unified_radix_cache.py
```diff
@@ -633,17 +814,33 @@ def _remove_leaf_from_parent(self, node: UnifiedTreeNode) -> None:
...
+
+        if self.hicache_enabled and node.backuped:
+            # Demote to host tier: keep the tombstone in-tree.
+            self._update_evictable_leaf_sets(node)
+            self._update_evictable_leaf_sets(node.parent)
+            return
+
         parent = node.parent
         self._remove_leaf_from_parent(node)
         self._update_evictable_leaf_sets(parent)
```
MokusMokun, line 951: If eviction-time backup fails because `_reserve_host_slots()` returns `None`, the un-backed-up node is cascade-deleted. `_is_device_leaf()` still allows tombstoned children, so deleting that parent can detach host-backed descendants and make their host pages unreachable to matching. Free the backed-up subtree before deleting, or refuse the cascade delete when backed-up children exist.

python/sgl_jax/test/mem_cache/test_hicache_e2e.py
MokusMokun, line 1: The Stage 3 KL-equivalence and normal-load latency claims are not backed by tests or matching evidence, and the committed HiCache e2e tests do not exercise scheduler overlap or `dp_size > 1` page mapping. Add a KL dual-run against `RadixCache`, an overlap e2e that forces demote/reload, and a `dp_rank=1` mapping test.

python/sgl_jax/srt/mem_cache/unified_radix_cache.py
```diff
@@ -616,6 +687,116 @@ def _insert_helper(
...
+        self.ongoing_write[future] = (node, host_pages)
+        return num_pages
+
+    def precompile_hicache_transfers(self) -> None:
+        """Precompile host<->device transfer kernels at startup (no-op if disabled)."""
+        if not self.hicache_enabled or self.host_pool is None:
+            return
+        self.host_pool.precompile_transfers()
+
+    def writing_check(self) -> None:
```
MokusMokun, line 858: `write_backup()` publishes `host_value` before the worker flush completes, but `writing_check()` does not roll that state back if `flush_backup()` fails. If the exception is ever handled non-fatally, the node can still point at an empty or stale host slot. On worker failure, clear `host_value`, free the host pages, and remove the node from host eviction state.

python/sgl_jax/srt/mem_cache/unified_radix_cache.py
```diff
@@ -616,6 +687,116 @@ def _insert_helper(
...
+        device_pages = device_tokens[::PS] // PS
+        num_pages = len(device_pages)
+        host_pages = self._reserve_host_slots(num_pages)
+        if host_pages is None:
+            return 0
+
+        self.inc_lock_ref(node)
+        node_dp_rank = node.key.dp_rank if node.key and node.key.dp_rank is not None else 0
+        global_pages = self._to_global_device_pages(device_pages, node_dp_rank)
+        future = self.hicache_controller.write(global_pages, host_pages)
```
MokusMokun, line 838: With overlap scheduling, the next batch can be launched before the previous batch's result processing finishes. When that processing inserts through an existing HiCache node, `_inc_hit_count()` may call `write_backup()`, and `write_backup()` still calls `stage_backup()` synchronously on the scheduler/KV-owning thread; `stage_backup()` then `block_until_ready()`s the gathered KV pages before the async host put can be queued. So hit-time write-through backups can put a device wait back into the overlapped CPU path, especially with the default `write_through_threshold=1`, and this likely explains part of the reported 16.6% overhead despite #1312's "write returns immediately" goal.

That wait is not obviously removable: the synchronous gather is the donation-safety mechanism, because an off-thread read of `kv_buffer` could race the next donated forward.

I suggest to do an optimization (maybe to be handled in a follow-up PR): defer gathers to a donation-safe pre-launch window such as the `_flush_pending_h2d` slot, with a cross-round pending-backup queue, locks/epochs for nodes that may be evicted in between, and a higher default threshold or cap. For this PR, document the tradeoff and avoid presenting write-through as fully async.

python/sgl_jax/srt/mem_cache/kv_cache_builder.py
```diff
@@ -48,3 +51,62 @@ def build_kv_cache(
...
+    )
+    controller = HiCacheController(host_pool, device_pool)
+
+    cache.host_pool = host_pool
+    cache.hicache_controller = controller
+    cache.hicache_enabled = True
+    cache.write_through_threshold = server_args.hicache_write_through_threshold
+    cache.write_policy = server_args.hicache_write_policy
+    for component in cache._components_tuple:
+        component._full_kv_pool_host = host_pool
```
MokusMokun: The PR wires `_full_kv_pool_host`, but the attribute has no production readers and the HiCache control plane still hardcodes `BASE_COMPONENT_TYPE` in `UnifiedRadixCache` — a deviation from #1313 §8, which specifies FullComponent hooks as the seam for Stage 5 multi-component reuse. Either implement the planned component hooks or document the centralized-control-plane design change on the issue and remove the dead attribute.

cjx0709 (commented):
All 10 review comments addressed:

**Fixed (8/10):**
1. ✅ inflight registration before submit (hicache_controller.py)
2. ✅ host-evict skip inflight pages (unified_radix_cache.py)
3. ✅ chunked-budget check before init_load_back (schedule_policy.py)
4. ✅ reset() drain controller (unified_radix_cache.py)
5. ✅ writeback rollback on failure (unified_radix_cache.py)
6. ✅ cascade delete: keep tombstone when backed-up children exist (unified_radix_cache.py)
7. ✅ dead _full_kv_pool_host removed (kv_cache_builder, full_component, tests)
8. ✅ test coverage: +10 tests (equivalence, overlap, dp-rank mapping)

**Deferred to #1429 (2/10):**
9. 🔜 unbounded _pending cap/backpressure
10. 🔜 async write-through optimization

L2+L3 TPU tests all pass (15/15 + 2 benchmarks).

python/sgl_jax/srt/mem_cache/unified_radix_cache.py
MokusMokun: `reset()` now drains pending transfers, but it still drops the old tree without returning allocated `host_value` pages to `host_pool`. After backing up pages and then calling `cache.reset()`, those host pages remain allocated with no tree owner left to free them, so `flush_cache()` can still leave the L2 pool artificially full.

__Suggestion__: Walk the old tree and free every node's `host_value` before replacing the root, or recreate/reset the host pool/controller. Please also add a reset/`flush_cache()` test that asserts full host capacity is restored.

MokusMokun (approved):
LGTM

PR #1413, opened 2026-06-29
fix(qwen3): reasoning_content split + epmoe decode bs alignment

### LINKED ISSUE #1412: [Bug] Qwen3.5-35B-A3B: empty reasoning_content + epmoe DECODE precompile crash on default bs_paddings
Two issues hit when launching `Qwen/Qwen3.5-35B-A3B` (model_type=`qwen3_5_moe`) on TPU v7x, tp=16, default `--moe-backend=epmoe`:

### 1. DECODE precompile crash: `Expected num_tokens=1 to be aligned to ep_size=16`

Without an explicit `--precompile-bs-paddings`, default `PRECOMPILE_DEFAULT_BS_PADDINGS=[1,2,4,...]` is used. `fused_ep_moe` requires `num_tokens % (ep_size * t_packing) == 0` where `ep_size = get_ep_size(mesh) = dp*tp = 16` and `t_packing=2` for bf16 (kernel.py:255/277, tuned_block_configs.py:268). Decode `num_tokens == bs`, so bs=1 fails.

`_compute_bs_buckets` already filters `bs < tp_size*2` for `moe_backend in ("fused", "fused_v2")` (compilation_manager.py:86), but not for `"epmoe"` (the default).

```
ValueError: Expected num_tokens=1 to be aligned to ep_size=16.
  File ".../kernels/fused_moe/v1/tuned_block_configs.py", line 268, in get_simplified_key
```

Workaround: `--precompile-bs-paddings 32 64`.

### 2. `reasoning_content` always `None` for Qwen3

Qwen3/Qwen3.5 chat_template emits the `<think>\n` opener **as part of the prompt** (`add_generation_prompt=True` produces `...<|im_start|>assistant\n<think>\n`), so the completion text starts mid-reasoning **without** the start tag. `Qwen3Detector` has `force_reasoning=False` and therefore requires `<think>` in the completion to enter reasoning mode — which never happens. Result: `reasoning_content=None`, and `content` contains the full `...thinking...</think>\n\nanswer`.

```python
>>> ReasoningParser(model_type="qwen3").parse_non_stream("step1\n</think>\n\nans")
('', 'step1\n</think>\n\nans')   # expected: ('step1\n', 'ans')
```

The `enable_thinking=False` case (where completion has no reasoning at all) is already gated upstream in `serving_chat._get_reasoning_from_request`, so `force_reasoning=True` is safe here (matches `DeepSeekR1Detector`).

### Env
- sglang-jax `main` (image built 2026-06-27)
- transformers 4.57.1
- TPU v7x 2x2x2, tp=16 dp=1

### Fix
See linked PR — both are one-line fixes with CPU unit tests.

zorrofox:
Fixes #1412

## Summary

Two independent one-line fixes hit while bringing up Qwen3.5-35B-A3B on TPU v7x:

1. **`Qwen3Detector` force_reasoning=True** — Qwen3/Qwen3.5 chat_template emits `<think>\n` as part of the *prompt*, so the completion starts mid-reasoning without the start tag. With `force_reasoning=False` the detector required the start tag and returned `reasoning_content=None`, leaving the entire `...thinking...</think>\n\nanswer` in `content`. The `enable_thinking=False` case is already gated upstream in `serving_chat._get_reasoning_from_request`, so the detector is only invoked when reasoning is active.

2. **`_compute_bs_buckets` filter `epmoe`** — `fused_ep_moe` requires `num_tokens % (ep_size * t_packing) == 0` (kernel.py:255/277). Decode `num_tokens == bs`, so `bs_buckets` must be `>= tp_size*2`. The filter already existed for `fused`/`fused_v2` but missed `epmoe` (the default), so MoE models on the default backend crashed in DECODE precompile with `Expected num_tokens=1 to be aligned to ep_size=16`.

## Repro (before fix)

Qwen3.5-35B-A3B, tp=16, default `--moe-backend=epmoe`, no `--precompile-bs-paddings`:
- DECODE precompile → `ValueError: Expected num_tokens=1 to be aligned to ep_size=16`
- After working around with `--precompile-bs-paddings 32 64`: `/v1/chat/completions` returns `reasoning_content: null` with `</think>` left in `content`

## Test plan

- [x] `pytest test/srt/test_reasoning_parser.py` — 6/6 (4 new Qwen3 cases)
- [x] `pytest test/srt/test_compilation_manager.py` — 4/4 (new file)
- [x] pre-commit (isort/ruff/black/mypy)
- [x] e2e: Qwen3.5-35B-A3B tp=16 v7x without `--precompile-bs-paddings` → `[DECODE] bs_paddings=[32, 64]` auto-filtered, fired_up, `reasoning_content` populated (len=526, content='2')


test/srt/test_compilation_manager.py
```diff
@@ -0,0 +1,53 @@
+import unittest
```
MokusMokun: there is already a `python/sgl_jax/test/test_compilation_manager.py`
I suggest to move the epmoe test case onto that file

python/sgl_jax/srt/model_executor/compilation_manager.py
```diff
@@ -83,7 +83,9 @@ def _compute_bs_buckets(self, user_paddings: list[int] | None) -> list[int]:
         for bs in bs_list:
             if (
                 bs <= self.max_padded_batch_size
-                and (self.moe_backend not in ("fused", "fused_v2") or bs >= self.tp_size * 2)
+                and (
+                    self.moe_backend not in ("epmoe", "fused", "fused_v2") or bs >= self.tp_size * 2
```
MokusMokun: see issue comment https://github.com/sgl-project/sglang-jax/issues/1412#issuecomment-4850158176

test/srt/test_compilation_manager.py
zorrofox: Moved to `python/sgl_jax/test/test_compilation_manager.py` in e1faa31 (uses the existing `_make_server_args` helper).

python/sgl_jax/srt/model_executor/compilation_manager.py
zorrofox: Reverted the `"epmoe"` addition and reworked per the issue comment — `CompilationManager` now takes the resolved `moe_backend` from `ModelConfig` (which detects hard-coded-FusedEPMoE archs). See e1faa31.

python/sgl_jax/srt/configs/model_config.py
```diff
@@ -134,6 +137,24 @@ def __init__(
             )
         _assert_fused_moe_v2_supported(self.moe_backend, self.hf_config.architectures)
 
+        # Models whose MoE block hard-codes FusedEPMoE (fused_ep_moe v1 kernel)
+        # instead of dispatching on --moe-backend. Resolve the effective backend
+        # to FUSED so downstream guards keyed on the backend string
+        # (CompilationManager bs-bucket filter, tp_worker align_bs, scheduler
+        # ep-size warning) see the actual kernel constraints.
```
MokusMokun: Much closer — the effective-backend resolution fixes the root cause without over-filtering GMM-EPMoE models, and the reasoning side looks good.

### Still Problem
the resolved backend only reaches `CompilationManager`; the `ModelConfig` comment also names `tp_worker.align_bs` and the scheduler warning, but both still read raw `server_args`. The crash-relevant one is `tp_worker.py:196` — gated on `server_args.moe_backend == "fused"` **and** `server_args.ep_size > 1`, then `align_bs_for_fused_ep(max_running_requests, server_args.ep_size)`. 

For Qwen3.5 with default `--ep-size` that's `ep_size=1`, so the block is skipped (align also passes through at ep≤1). But `fused_ep_moe` derives EP from the mesh (`data*tensor`) — exactly why `bs=1` crashed at tp=16. So `max_running_requests` is never down-aligned, and `_compute_bs_buckets` appends it as the top decode bucket → it can still crash if that bucket is not launchable by the mesh EP size. Small bs is fixed; the top bucket is not fully protected.

### Fix
To keep this PR scoped, scheduler can be left as a diagnostic follow-up, but the tp_worker max-running-requests alignment should be wired to the effective fused backend because it controls the appended top decode bucket.

zorrofox: Wired in 9abd850: `tp_worker` now reads `effective_moe_backend = model_config.moe_backend.value` and aligns against `mesh_ep_size = tp_size * dp_size` (matching `get_ep_size(mesh)` in the kernel), so the appended top decode bucket is protected. Kept the guard at `== "fused"` (not extending to `fused_v2`) to avoid changing Bailing/MiMo behaviour. For existing explicit `--moe-backend fused` users `mesh_ep_size == server_args.ep_size`, so no regression there.

e2e counterexample on 35B-A3B tp=16 with `--max-running-requests 100`:
```
Qwen3_5MoeForConditionalGeneration hard-codes FusedEPMoE; resolving effective moe_backend epmoe -> fused
Adjusted max_running_requests from 100 to 64 for fused MoE (mesh_ep_size=16, bt must be in {2,4,8k})
[DECODE] Begin to precompile bs_paddings=[32, 64]
```
fired_up + GSM8K sanity 5/5.

`scheduler.py:308` diagnostic still reads raw `server_args`; left as follow-up as suggested.

MokusMokun: Thanks for the update. The original dp=1 Qwen3.5 repro looks covered now, but I think there are still two follow-ups needed before this is structurally consistent:

- `mesh_ep_size` in `tp_worker` should be computed from the actual mesh, not as `self.tp_size * self.dp_size`.

  The fused kernel uses `get_ep_size(mesh) = mesh.shape["data"] * mesh.shape["tensor"]`. Scheduler creates the mesh as `[dp_size, tp_size // dp_size]`, so the kernel sees `dp_size * (tp_size // dp_size) == tp_size`.

  With `tp_size=16, dp_size=4`, the current code computes `64`, while the kernel EP size is `16`. That can over-align `max_running_requests` or reject values that are valid for the real fused_ep mesh.

  Suggested fix:

  ```python
  mesh_ep_size = self.mesh.shape.get("data", 1) * self.mesh.shape.get("tensor", 1)
  ```

- The `ModelConfig` comment still says the resolved backend reaches the scheduler warning, but `scheduler.py` intentionally still reads raw `server_args.moe_backend`. Either wire scheduler to the effective backend too, or narrow the comment to only mention the paths that are actually fixed in this PR.

The `fused_v2` exclusion and the updated `CompilationManager` direction look reasonable to me.

zorrofox: Both addressed in ab8144c + 147de3d:

- `mesh_ep_size` now reads `self.mesh.shape.get("data",1) * self.mesh.shape.get("tensor",1)` — matches `get_ep_size(mesh)` exactly. At tp=16 dp=4 that's 4*4=16 (was wrongly 64 with tp*dp). dp=1 case unchanged so the earlier e2e (`mr=100 -> 64`, `mesh_ep_size=16`) still holds.
- ModelConfig comment narrowed to only name CompilationManager + tp_worker (dropped scheduler).

Also ran an internal adversarial review pass and added a `mimo` test case (shares `Qwen3Detector`, same `force_reasoning=True` behaviour, gated by `serving_chat._get_reasoning_from_request`). Checked the other `server_args.ep_size` reader in `model_runner.py` → only feeds `routed_experts_capturer` device-balance debug metrics (default off), not on the crash path — left out of scope.

MokusMokun (approved):
LGTM

Issue #1427, opened 2026-07-02
[Design] Pathways single-controller PD: alternative TransferBackend for #1196


zorrofox:
# [Design] Pathways single-controller PD: alternative transfer backend for #1196

> Sub-issue of #1196 (PD Disaggregation Roadmap). Design doc for review per the contribution guide. PR: `#<PR-NUM>`.

## 1. Motivation

The current `JaxTransferBackend` (#1199) moves KV across pods via `jax.experimental.transfer`, which stages `device → host → NIC → host → device` because TPU lacks device-to-device RDMA across pods. On v6e-1 the staged D2H + p2p pull segments dominate the transfer timeline (~2.5 GB/s/host effective).

**Pathways IFRT proxy already solves the per-host fan-out inside the runtime.** A single Python process sees all devices across N slices; `jax.device_put(array, sharding_on_other_slice)` is lowered to per-host DCN sends without surfacing host buffers to Python. Same hardware (2× v7x-8 slices, DCN not ICI), measured:

| transport | aggregate BW | per-host | payload |
|---|---|---|---|
| `jax.experimental.transfer` (`JaxTransferBackend` primitive) | 9.0 GB/s | 4.5 GB/s | 2048 MiB |
| Pathways IFRT cross-slice `device_put` | **28.9 GB/s** | 14.4 GB/s | 2048 MiB |
| — | — | — | 576 MiB (≈16K-tok MLA KV): **23 ms** |

3.2× is pure software-stack difference; the runtime is the same DCN NIC.

This proposal adds Pathways as an **alternative `TransferBackend`** under the #1199 abstraction (and #1196 roadmap), plus two roadmap items it unblocked along the way: a measured "CO-interference" dimension for PD model selection, and an SWA-aware migrate design (Stage 5).

## 2. Positioning

**Not a parallel architecture.** This slots under the existing PD stack:

- **Reuse** (#1241/#1240/#1368, future Stage 4/6): mini_lb router, bootstrap room, decode admission, PD metrics, P/D ratio policy.
- **Replace** only the #1199 `TransferBackend` layer: instead of `register_pull → pull → release` over `jax.experimental.transfer`, do `gather → stack → device_put → scatter` in one address space.
- **Adds** a single-controller scheduler variant (`_pd_prefill_loop`) that drives N prefill workers + 1 decode worker from one process — #1196 Stage 6 "multi-P single-D" falls out naturally (no inter-process bootstrap).

[...]

Issue #1196, opened 2026-05-25
[Roadmap] Prefill-Decode Disaggregation


cjx0709:
## Motivation

Add Prefill-Decode Disaggregation (PD) to sgl-jax so prefill and decode can run on different pods, improving throughput and resource utilization on multi-host TP deployments (the typical production layout).

Cross-pod transfer throughput baseline: 8.86 GB/s single-channel on a v6e-16 slice (JAX 0.8.1).

## Roadmap semantics

This roadmap reflects the capability of the **`main` branch**.

That means:

- stages remain **unchecked** until the corresponding functionality is merged into `main`
- linked sub-issues define the intended stage scope, design direction, and acceptance criteria
- the roadmap should be read as the merge target for upstream, not as a record of in-progress branch state

### Stage 0: Transfer foundation
- [x] JAX transfer wrapper + connection ABC + single backend + byte-correct transfer validation (#1199)

### Stage 1: Host pool + side channel
- [x] `QueueHostKVPool` foundation + out-of-band ZMQ pull-done side channel + event-driven sender/receiver lifecycle (#1242)

### Stage 2: Scheduler + end-to-end
- [x] Bootstrap server + scheduler mixins + tokenizer passthrough + prefill-only response contract (#1241)

### Stage 3: Multi-host + routing
- [x] Per-host transfer server registration + single-entry router/proxy for `/generate` and OpenAI `/v1/*` (#1240)
- [x] Production-like benchmark/eval entrypoint on the PD single-entry path

### Stage 4a: Minimum viable PD (pre-production capability)

Scope rule: an item belongs here only if it can be validated **without** a real production environment, and skipping it would make PD not even a usable toy. Capability/correctness, not hardening.

- [x] Fully plumb path-A (`P HBM -> P host -> ... -> D HBM`) end-to-end
- [x] Functional capacity at the target shape (`64` concurrency / `4k` input / `1k` output, no chunked prefill): run completes 100% with zero aborts. Throughput is **not** a goal at this stage.
- [x] Baseline observability: Prometheus PD metrics emitted on `/metrics` (state transitions, transfer bytes/duration/inflight, host-pool usage, failures, registry size)
- [x] Minimal admission cap at the single-entry proxy (mini_lb): bound the number of concurrently-running requests; excess requests are held **pending** (never returned as client errors, never aborted)

### Stage 4b: Production hardening (deferred — requires a real production environment to validate)
[...]

PR #1428, opened 2026-07-02
feat(pd): Pathways single-controller PD as alternative TransferBackend

zorrofox:
**Design issue**: #1427 (sub-issue of #1196)
**Depends on**: #1404 (glm5 static-fp8 fused-mlp fix — included here so e2e passes; will rebase once #1404 merges)

---

## Summary

Implements the Pathways single-controller PD backend described in `#1427`: one Python process on a CPU head sees all TPU devices across ≥2 slices via IFRT proxy, runs N prefill `ModelWorker`s + 1 decode `ModelWorkerClient`, and moves KV pages P→D with cross-mesh `jax.device_put` (29 GB/s vs `jax.experimental.transfer`'s 4.5 GB/s on the same v7x DCN).

Addresses #1196 Stage 5 (SWA model support — MiMo-V2.5-Pro validated) and Stage 6 (multi-P single-D — falls out of the single-controller loop).

## Changes by area

| area | files | notes |
|---|---|---|
| **transfer core** | `disaggregation/pathways_pd.py` (new, 281 L) | `make_slice_meshes` / `PathwaysPDKVTransfer` / `migrate_reqs_p_to_d` / SWA dual-pool |
| **scheduler** | `managers/scheduler.py` (+551) | `_pd_prefill_loop` async thread, dual-worker init, admission accounting |
| | `managers/tp_worker_overlap_thread.py` | `gc.freeze()` + gen2 threshold (Max-ITL 2247→269 ms) |
| | `managers/scheduler_{profiler_mixing,output_processor_mixin}.py` | Pathways `gs://` profiler out, PD-aware output |
| **weight load** | `utils/weight_utils.py` | `_PD_WEIGHT_CACHE` cross-slice reuse (D load 80 min→24 s), `BULK_READ` (cold 3.05×), `BULK_BLOCK` (bound head RSS for >768 G ckpt) |
| **compat** | `kernels/_pathways_compat.py` (new) | `_slice_memref` monkeypatch (jax 0.8.1 lowering vs Pathways verify) |
| | `kernels/fused_moe/v1/kernel.py` | `memory_stats()` fallback on IFRT-proxy device |
| | `kernels/mla/v2/kernel.py`, `blockwise_kernel.py` | small-batch BlockSpec compat |
| | `utils/jax_utils.py` | `live_arrays()` dedup by buffer ptr (proxy wrapper double-count) |
| | `configs/model_config.py` | deepcopy lru_cached `hf_config` (P/D cross-pollution) |
| **misc** | `server_args.py` | `--pd-disaggregation=pathways` + `--pd-{num-prefill,prefill/decode-mem-fraction,*-max-tokens}` |
| | `pyproject.toml` | pin `pathwaysutils<0.1.9` (0.1.9 needs jax≥0.8.3) |
| | `models/glm5_moe.py` | #1404 (static-fp8 fused-mlp guard) — will drop on rebase after #1404 merges |
| **tests** | `test/srt/disaggregation/test_pathways_pd.py` (new, 9 tests) | mesh split / byte-exact transfer / bucket pad / SWA dual-pool / SWA page mapping / `migrate_reqs_p_to_d` e2e |

`git diff origin/main --stat`: 17 files, +1521/−159.

All new behavior is behind `--pd-disaggregation=pathways` or `JAX_PLATFORMS=proxy`; the native multi-process path and existing `--disaggregation-mode` are untouched.

## Benchmarks (v7x, 2× 2x2x2 slices, same total HW, IL=4K OL=1K, ×2 σ<1%)

| model | metric | colocated tp32 | PD p1d1 | Δ |
|---|---|---|---|---|
| GLM-5.2 W8A8 | MedITL @0.19 | 40.61 | 35.35 | −13 % |
| GLM-5.2 W8A8 | P99-ITL @0.28 | 2162 | 82 | **−96 %** |
| GLM-5.2 FP8 | MedITL @0.19 | — | 36.50 | — |
| MiMo-V2.5-Pro (SWA) | MedITL @0.19 | 20.05 | 57.82 | +188 % (see design §4: SWA CO doesn't interfere → PD is net-negative, expected) |

Regression check on this branch (`0fcb99c`, post-rebase + ici_pd removal) vs `cb92c6f` baseline, GLM-5.2-FP8 @rate=0.19, 2 runs σ<3%: MedITL **36.74** (baseline 36.50, Δ +0.6%), P99-ITL **72.66** (baseline 72.7, Δ −0.1%), TTFT **1883** (baseline 1893, Δ −0.5%) → no regression.

GSM8K exposed a latent bug: with cps=4096 and 32 concurrent short reqs, `PrefillAdder` truncates the last req of a batch → `chunked_reqs` non-empty → PD assert. Fixed by raising `chunked_prefill_size` to `pd_prefill_max_tokens` at PD init (commit `0fcb99c`), so the adder stops on P-pool budget instead. The assert now only fires on the actual limitation (single-req IL > P pool).

## Test plan

- [x] `pre-commit run --all-files`
- [x] CPU unit: `XLA_FLAGS=--xla_force_host_platform_device_count=8 JAX_PLATFORMS=cpu pytest test/srt/disaggregation/test_pathways_pd.py -v` → 9 passed
- [x] v7x e2e correctness: GLM-5.2-FP8 78L greedy `Paris` ✓
- [x] v7x perf regression: GLM-5.2-FP8 PD MedITL 36.74 vs baseline 36.50 (Δ +0.6%, 2 runs σ 0.4%)
- [x] GSM8K 200q on GLM-5.2-FP8 PD: acc=0.955 (5-shot, temp=0, parallel=32)
- [x] v7x SWA e2e: MiMo-V2.5-Pro PD correctness ✓, MedITL 57.78 vs baseline 57.82 (Δ −0.07%, 2 runs σ 0.2%)

## Known limitations (tracked in design issue)

- No chunked prefill (`IL ≤ chunked_prefill_size`); dp=1 only
- Pathways runtime image is GCP-only (client Apache-2.0; same closed-surface constraint as `libtpu`)
- Cold load ~18 min for 78L GLM

## Deployment (reviewer reproduction)

<details><summary>PathwaysJob CRD + launch command</summary>

```yaml
# PathwaysJob: 2x tpu7x 2x2x2 slices + c4 head
spec:
  workers:
    - type: tpu7x-standard-4t
      topology: 2x2x2
      numSlices: 2
```
```bash
JAX_PLATFORMS=proxy python -m sgl_jax.launch_server \
  --model-path <GLM-5.2-FP8> --trust-remote-code --dtype=bfloat16 \
  --tp-size 16 --ep-size 16 --dp-size 1 --page-size 128 \
  --pd-disaggregation pathways --pd-num-prefill 1 \
  --pd-prefill-max-tokens 32768 --pd-decode-max-tokens 262144 \
  --pd-prefill-mem-fraction 0.88 --pd-decode-mem-fraction 0.88 \
  --chunked-prefill-size 4096 --disable-radix-cache --attention-backend fa \
  --enable-single-process --host 0.0.0.0 --port 30000
```
</details>


python/sgl_jax/srt/server_args.py
```diff
@@ -772,6 +778,47 @@ def add_cli_args(parser: argparse.ArgumentParser):
             nargs="+",
             help="The device indexes to use build mesh. Defaults is all if not specified.",
         )
+        parser.add_argument(
+            "--pd-disaggregation",
+            dest="pd_disaggregation",
```
cjx0709, line 788: Could we make `--pd-disaggregation pathways` mutually exclusive with the existing `--disaggregation-mode prefill/decode` path? #1427 says the existing multi-process path is untouched, but today both knobs can be set together: the scheduler can initialize Pathways P/D workers while `dispatch_scheduler_event_loop()` still selects the legacy multi-controller PD loop from `disaggregation_mode`. A small validation error here would keep the two PD modes much easier to reason about.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -295,9 +295,35 @@ def __init__(
...
         if mesh is not None:
             self.mesh = mesh
+        elif self.pd == "pathways":
+            from sgl_jax.srt.disaggregation.pathways_pd import make_slice_meshes
+
+            server_args.disable_radix_cache = True
+            # PD does not support chunked prefill; raise cps to the P-pool
+            # capacity so PrefillAdder never truncates a req mid-batch when
+            # packing many short reqs (e.g. 32-way GSM8K -> 6x~700 > 4096).
+            server_args.chunked_prefill_size = max(
```
cjx0709: `chunked_prefill_size` defaults to `None`, so enabling `--pd-disaggregation pathways` without explicitly passing `--chunked-prefill-size` looks like it will fail at startup with `TypeError` from `max(None, int)`. #1427 documents that Pathways does not support chunked prefill, but it would be nice for the default Pathways startup path to initialize cleanly.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -959,8 +1113,21 @@ def event_loop_normal(self):
     def event_loop_overlap(self):
         """A scheduler loop that overlaps the CPU processing and Accelerator computation."""
         self.result_queue = deque()
+        _pd_iter_trace = self.pd == "pathways"
+
+        import gc as _gc
+
+        _gc.collect()
+        _gc.freeze()
```
cjx0709: This GC tuning seems Pathways-specific in #1427, but it is currently applied to every overlap scheduler. Since it changes global Python runtime behavior for ordinary overlap serving too, could we either gate it on `self.pd == "pathways"` or add a note/test result showing it is also intended for the existing pd transfer path?

cjx0709 (commented):
Reviewed with #1427 in mind; I focused on Pathways vs existing PD boundaries and a few lifecycle edges worth tracking.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -2094,6 +2321,294 @@ def run_batch(self, batch: ScheduleBatch) -> GenerationBatchResult:
...
+        _ann = jax.profiler.TraceAnnotation("pd_drain_ready")
+        _ann.__enter__()
+        for r, p_slots, p_pages in p_pages_per_req:
+            seq_len = len(r.fill_ids)
+            n_pages = len(p_pages)
+            d_slots = self.token_to_kv_pool_allocator.alloc(
+                n_pages * self.page_size, dp_rank=r.dp_rank or 0
+            )
+            if d_slots is None:
+                raise RuntimeError("[pd_async] D pool OOM during insert")
```
cjx0709: One lifecycle edge to track: the async Pathways path allocates D KV only after prefill completes, and D-pool exhaustion currently raises here. That may be OK for the first experimental version, but it would be good to capture a follow-up for D-side KV reservation/backpressure or a pending-ready path.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -320,12 +346,140 @@ def __init__(
...
+                        p_alloc=self.p_allocs[i],
+                        d_alloc=d_alloc,
+                        page_size=server_args.page_size,
+                    )
+                    for i, (pm, w) in enumerate(zip(p_meshes, self.tp_workers_p))
+                ]
+                self.kv_transfer = self.kv_transfers[0]
+                self._pd_prefill_qs = [_Queue(maxsize=2) for _ in range(n_p)]
+                self._pd_prefill_q = self._pd_prefill_qs[0]
+                self._pd_ready_q: _Queue = _Queue()
```
cjx0709: Small lifecycle note: `_can_flush_cache()` only accounts for the legacy disagg queues, not Pathways `_pd_prefill_qs`, `_pd_ready_q`, or `_pd_inflight`. This does not need to be solved immediately if the backend is still experimental, but it would be worth tracking so flush/abort/pause eventually cover the async Pathways queues too.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -2094,6 +2321,294 @@ def run_batch(self, batch: ScheduleBatch) -> GenerationBatchResult:
...
+                mwb = batch.get_model_worker_batch(
+                    *paddings, self.page_size, self.server_args.enable_static_lora
+                )
+                _, ntok_dev, _ = worker.forward_batch_generation(mwb, sampling_metadata=None)
+                t_disp = time.perf_counter()
+                ntok = np.asarray(jax.device_get(ntok_dev))
+                t_fwd = time.perf_counter()
+                self._extract_dp_output_ids(ntok, mwb, batch)
+                result = GenerationBatchResult(
+                    logits_output=None,
```
cjx0709: If Pathways initially targets plain generation only, this is probably fine as a first step. Since the async prefill result sets `logits_output=None` while reusing the normal prefill output processor, could we document or explicitly reject logprob/hidden-state requests for now, or track carrying those fields through the async path later?

python/sgl_jax/srt/kernels/mla/v2/kernel.py
```diff
@@ -1276,6 +1276,11 @@ def prepare_q_inputs(
         q_packing,
         head_dim,
     )
+    # When actual_num_q_heads % q_packing == 0 (e.g. 2 heads on bf16), the
+    # head-axis pad is (0,0) and XLA may fuse pad+reshape into a lazy view
+    # whose physical layout != [mnt, h//p, p, D]. pallas_call then gets a
+    # non-contiguous HBM ref and the DMA BoundsCheck trips. Force a copy.
+    q = jax.lax.optimization_barrier(q)
```
cjx0709, line 1283: Could you add a small note in the PR or a focused test for the shape/runtime case that needs this barrier? This changes the global MLA kernel path, not just Pathways PD, and `optimization_barrier` can affect fusion/materialization. The comment explains the intent well; I mainly want to make sure we have coverage or regression data for non-Pathways MLA too.

python/sgl_jax/srt/kernels/quantized_matmul/quantized_matmul_kernels/blockwise_kernel.py
```diff
@@ -69,6 +69,10 @@ def quantized_matmul_kernel(
             w_q_dtype=jnp.dtype(w_q.dtype).name,
         )
     batch_block_size = tuned_value.batch_block_size
+    # Mosaic requires last-2 block dim %8==0 or ==array dim. tuned batch_block_size
+    # can be 1; for n_batch in (2..7) that fails the check. Pad block to 8.
+    if orig_n_batch > 1 and batch_block_size % 8 != 0:
+        batch_block_size = next_multiple(batch_block_size, 8)
```
cjx0709, line 75: This looks like a general Mosaic small-batch kernel fix rather than something specific to the PD transfer backend. Could we either split it into a small prerequisite PR, or add a focused test/shape note here? It would help reviewers understand why a transfer-engine PR needs to change quantized matmul behavior globally.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -2094,6 +2321,294 @@ def run_batch(self, batch: ScheduleBatch) -> GenerationBatchResult:
             ret.accept_lens = batch_output.accept_lens
         return ret
 
+    # ---- pathways_pd helpers ----
```
cjx0709: The Pathways-specific scheduler code is getting fairly large inside `scheduler.py`. Could we consider moving these helpers into a small mixin or `disaggregation/pathways_scheduler.py`, leaving the main scheduler with just the wiring/hooks? That would make the boundary with the existing scheduler easier to maintain.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -1596,14 +1778,26 @@ def _get_swa_token_info(self):
         )
 
     def get_next_batch_to_run(self) -> ScheduleBatch | None:
+        if self.pd == "pathways":
+            return self._pd_get_next_batch_async()
+        # PD: 先尝试迁移之前 D 满时暂存的 batch（D running req 完成后 free 出空间）
```
cjx0709: Tiny style nit: could we keep new comments/docstrings in English here? A few new Pathways helper comments are currently Chinese, while the surrounding scheduler code is mostly English. Keeping one language makes future maintenance a bit easier for the broader contributor base.

python/sgl_jax/srt/managers/scheduler_profiler_mixing.py
```diff
@@ -220,10 +220,23 @@ def start_profile(
...
-            profiler_options=profiler_options,
-        )
+        if self.profiler_output_dir.startswith("gs://"):
+            # Pathways: worker device trace only. Client-side python tracer
+            # OOMs head pod on stop (millions of events); patch it out.
+            from pathwaysutils import profiling as _pwp
+
+            _pwp._original_start_trace = lambda *a, **k: None
+            _pwp._original_stop_trace = lambda *a, **k: None
+            jax.profiler.start_trace(
```
cjx0709, line 232: Could you clarify whether the local profiling path remains equivalent after this change? The non-`gs://` branch now calls `pathwaysutils.profiling._original_start_trace/_original_stop_trace` instead of `jax.profiler.start_trace/stop_trace`, and the `gs://` branch patches pathwaysutils internals globally. It would be helpful to gate this by Pathways/proxy mode or document why it is safe for non-Pathways profiling.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -501,18 +502,27 @@ def _forward(
...
-        if self.tp_size == 1 and isinstance(pool_updates, list):
-            target_sharding = self.token_to_kv_pool.kv_sharding
-            pool_updates = [jax.device_put(kv, target_sharding) for kv in pool_updates]
-        self.memory_pools.replace_all(pool_updates)
+            # tp_size==1: sharding constraint is lost after JIT; re-place explicitly.
+            # See https://github.com/sgl-project/sglang-jax/issues/233
+            if self.tp_size == 1 and isinstance(pool_updates, list):
+                target_sharding = self.token_to_kv_pool.kv_sharding
+                pool_updates = [jax.device_put(kv, target_sharding) for kv in pool_updates]
+            self.memory_pools.replace_all(pool_updates)
```
cjx0709, line 523: The `_donate_lock` itself is nicely optional, but this also moves `memory_pools.replace_all(pool_updates)` inside the `count_pjit_cpp_cache_miss()` context. Is that intentional? If the goal is only to serialize the donate/replace window for Pathways, it may be cleaner to keep the cache-miss measurement scope unchanged and only guard the minimal shared KV-buffer section.

python/sgl_jax/srt/utils/weight_utils.py
```diff
@@ -2324,38 +2384,13 @@ def _load_dummy_weights(
...
-                else:
-                    return jnp.zeros(shard_shape, dtype=dtype)
-
-            dummy_weight = jax.make_array_from_callback(shape, sharding, make_shard)
-            model_param.value = dummy_weight
+            # jit(zeros) compiles to a device-side XLA constant -> 0 H2D. The
+            # original make_array_from_callback path serializes ndev host
+            # callbacks per param under Pathways IFRT (ocean_remote_python.py
+            # iterates devices synchronously) -> ~27K RPCs for a 78L MoE.
+            model_param.value = jax.jit(
```
cjx0709, line 2391: This changes dummy weights from deterministic small random values to all zeros for every backend, not just Pathways. That is probably fine for shape/compile-only dummy loading, but it is a semantic change for any smoke tests or routing behavior that depend on nonzero dummy weights. Could you add a short note or test coverage confirming this is OK?

python/sgl_jax/srt/server_args.py
cjx0709, line 788: For the Pathways tuning knobs below, would it be cleaner to group them under a Pathways-specific config object/file or at least use a `pathways-pd-*` prefix? These are quite backend-specific, and keeping the public CLI surface clearly separated would make the new mode feel less mixed into the generic server args.

cjx0709 (commented):
A second pass focused on PR shape: keeping Pathways-specific support fixes scoped, preserving existing behavior where possible, and making the scheduler changes easier to maintain.

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Tracked as follow-up in #1427 §8. Admission already caps `_pd_inflight` against D r2t slots so this only fires when a burst of long-IL reqs lands after admission but before D alloc; will add proper reservation/backpressure in a follow-up.

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Fixed in 8f869a46: gated under `if self.pd == "pathways"`.

python/sgl_jax/srt/server_args.py
zorrofox: Kept `--pd-*`: these knobs (`num-prefill` / `{prefill,decode}-max-tokens` / `*-mem-fraction`) describe the single-controller P/D split, not Pathways specifically — a future single-controller `TransferBackend` (e.g. tpu-raiden in-process) would reuse them. `--pd-disaggregation=pathways` is the only backend-naming flag. Open to a config-object grouping if that reads cleaner.

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Fixed in 8f869a46: `or 0` guard on `chunked_prefill_size`.

python/sgl_jax/srt/server_args.py
zorrofox: Fixed in 8f869a46: `__post_init__` now raises `ValueError` when both `--pd-disaggregation` and `--disaggregation-mode` are set.

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Done in a30a36e9: `_pd_*` helpers + P/D worker init moved to `disaggregation/pathways_scheduler.py::PathwaysPDSchedulerMixin`. scheduler.py 3033→2598 lines; the remaining hooks are mesh dispatch, tp_worker init dispatch, `get_next_batch_to_run` dispatch, and the (now gated) gc block.

python/sgl_jax/srt/kernels/quantized_matmul/quantized_matmul_kernels/blockwise_kernel.py
zorrofox: The trigger is `n_batch ∈ [2,7]` with `tuned batch_block_size=1` — v6e-64 tp32 GLM-5.1-FP8 hits it on the native backend too, so it's a general prerequisite for FP8 on that shape, not PD-specific. The inline comment documents the Mosaic constraint. Bundled here because that's the config the PD path was validated on; can split into a 4-line prerequisite PR if you prefer.

python/sgl_jax/srt/kernels/mla/v2/kernel.py
zorrofox: The barrier only fires when the head-axis pad is `(0,0)`, which is the exact case XLA otherwise fuses into a non-contiguous view. GLM-5.2-FP8 (MLA, native backend) MedITL on this branch vs the pre-barrier baseline is +0.6% (36.74 vs 36.50, 2 runs σ 0.4%), so the fusion loss is within noise. Gating on backend would need runtime device introspection inside the kernel path; happy to add if you have a shape you're worried about.

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Fixed in 8f869a46: `_pd_get_next_batch_async` now asserts `not r.return_logprob` before dispatch with an explicit "not yet supported" message. Carrying logits through the async path is a follow-up.

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Tracked as follow-up. `--pd-disaggregation` currently forces `disable_radix_cache=True` so `flush_cache` is a no-op on this path; will wire `_pd_prefill_qs`/`_pd_ready_q`/`_pd_inflight` into the flush/abort surface once radix support lands.

python/sgl_jax/srt/utils/weight_utils.py
zorrofox: `--load-format dummy` is documented as shape/compile-only; the previous `rng.uniform(-1e-3,1e-3)` values were already meaningless for MoE routing (random gate → random expert). The proxy backend can't use the callback path (~27K sync RPCs for a 78L MoE). If there's a specific test that depends on nonzero dummy values I'll add an env-gated fallback — could you point me at it?

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Fixed in 8f869a46 (all remaining Chinese comments translated).

python/sgl_jax/srt/managers/scheduler_profiler_mixing.py
zorrofox: Fixed in 8f869a46: both start/stop now take the original `jax.profiler.*` path when `JAX_PLATFORMS != "proxy"`; the `pathwaysutils` imports only happen under proxy.

python/sgl_jax/srt/model_executor/model_runner.py
zorrofox: Fixed in 8f869a46: `_donate_lock` is now the outer context (covers forward-donate through `replace_all`), `count_pjit_cpp_cache_miss()` is nested inside and scopes only the `jitted_run_model` call — matches main's original measurement window.

zorrofox:
Pushed 8f869a46 (7 gate/validate/docs fixes) + a30a36e9 (PathwaysPDSchedulerMixin extraction). CPU pytest 9/9, pre-commit clean. Re-ran GLM-5.2-FP8 PD e2e on the mixin commit (image a30a36e): MedITL **36.61** (2 runs 36.71/36.50, σ 0.6%) vs pre-review d2b20c3 36.74 vs baseline 36.50 → Δ +0.3%, no regression.

Follow-ups tracked in #1427: D-side KV reservation/backpressure, `_can_flush_cache` coverage of async queues, carrying logits through async prefill.

cjx0709 (approved):
LGTM

Issue #1419, opened 2026-06-30
[Feature] Bring up larger Qwen3.5 MoE sizes (122B-A10B / 397B-A17B)


MokusMokun:
Part of the Qwen3.5 model series roadmap (#1421). Discharges the size-coverage deferral in #1251 §1.2.

## Motivation

The merged MoE (#1267) and dense (#1372) paths validate the Qwen3.5 backbone on 35B-A3B and up to 27B dense. The same architecture extends to the larger MoE checkpoints — **122B-A10B** and **397B-A17B** — the production-scale targets. M1 was written to avoid 35B-A3B-only hardcoding so these land incrementally; this issue closes that out and validates them at scale.

**Start with 122B-A10B** (a near-drop-in scale-up of the merged path); treat **397B-A17B** as the capacity-gated scale follow-up.

## Per-checkpoint config (to verify against each checkpoint's index)

| Size | model_type | Total experts | Top-k | moe_intermediate_size |
|---|---|---:|---:|---:|
| 35B-A3B (done) | `qwen3_5_moe` | 256 | 8 | 512 |
| 122B-A10B | `qwen3_5_moe` | 256 | 8 | 1024 |
| 397B-A17B | `qwen3_5_moe` | 512 | 10 | 1024 |

Expert fields live under `text_config` for all three. 397B-A17B's larger expert count (512) and top-10 routing are the points most likely to expose hardcoded assumptions.

## Scope

**In**
- **Checkpoint-format verification as an explicit blocker.** The merged loader assumes the pre-fused `gate_up_proj` / `down_proj` expert layout; the `Qwen3_5HybridConfig.moe_pre_fused` field exists but **only the pre-fused path is implemented** (no consumer of `moe_pre_fused=False` exists today). Confirm each larger checkpoint's `model.safetensors.index.json` matches the pre-fused layout before bring-up; if a variant ships per-expert tensors, the non-pre-fused loader path must be **implemented** here, not assumed.
- Multi-host serving + sharding at scale (TP/EP/DP mesh, `mem-fraction-static`, page/prefill sizing) per checkpoint.
- Accuracy validation (MMLU-Pro / GPQA-Diamond) per the #1251 §4.3 protocol.

**Out**
- New backbone architecture work — none expected; any divergence becomes its own follow-up.
- FP8/INT8 quantization (roadmap backlog), which the 397B-A17B footprint may eventually motivate.

## Depends on

[...]

PR #1434, opened 2026-07-03
fix(qwen3.5): use make_array_from_callback in _put for multi-host loading

zorrofox:
Toward #1419 — unblocks multi-host loading for the larger Qwen3.5 MoE sizes. Scoped to the loader fix only; not claiming the full #1419 accuracy acceptance.

## Summary

`models/qwen3_5.py:_put` (the special-path loader for GDN fused projections and MoE `gate_up_proj`) used `jax.device_put(jnp.asarray(arr), NamedSharding)`. On multi-host that triggers `multihost_utils.assert_equal → process_allgather`, which materializes N_hosts copies per device. On 122B-A10B / v6e 8×8 (16 hosts) the MoE `gate_up` w1 tensor (~1.5G) becomes ~24G/device and OOMs after regular weights are already resident (~28G, 3.74G free). 35B-A3B on 2 hosts (2×0.5G) fit under the bar so this wasn't visible in M1.

Switch `_put` to `jax.make_array_from_callback` so each host uploads only its addressable shard — same pattern the regular `WeightLoader` path in `weight_utils.py` already uses (see comment at :678).

## Checkpoint-format note (both #1419 sizes)

Both 122B-A10B and 397B-A17B ship the pre-fused `experts.gate_up_proj`/`down_proj` layout under `model.language_model.*`, matching 35B (the per-expert `experts.N.{gate,up,down}_proj` keys in the index are all under `mtp.*`, which the loader skips). No non-pre-fused loader path needed.

## Test plan

- [x] `pytest python/sgl_jax/test/models/test_qwen3_5.py` — 18/18
- [x] pre-commit
- [x] 122B-A10B v6e 8×8 `--tp-size 64 --dp-size 4`: fired_up (was OOM before), loader-assert 831 loaded / 1118 skipped (333 visual + 785 mtp), no `[missing]`/`[unexpected]`
- [x] 122B-A10B v7x 2×2×2 `--tp-size 16 --dp-size 1`: fired_up, same loader-assert, GSM8K sanity 5/5
- [x] GPQA-Diamond full (198, §4.3 protocol, `max_tokens=30720`): **0.864** on v6e (target ≥85.1; official 86.6) — correctness holds after the loader change

### DIFF
--- a/python/sgl_jax/srt/models/qwen3_5.py
+++ b/python/sgl_jax/srt/models/qwen3_5.py
@@ -619,9 +619,18 @@ class Qwen3_5MoeForConditionalGeneration(nnx.Module):
     # Weight loading
     # ------------------------------------------------------------------
     def _put(self, arr, spec):
-        """Host array -> sharded bf16 device array under the model mesh."""
-        return jax.device_put(
-            jnp.asarray(arr, dtype=self.dtype), NamedSharding(self.mesh, P(*spec))
+        """Host array -> sharded device array under the model mesh.
+
+        Uses make_array_from_callback so each host uploads only its local
+        shard; jax.device_put on a host-replicated array triggers a
+        multihost assert_equal -> process_allgather (16 hosts * 1.5G MoE
+        gate_up = 24G per device -> OOM on 122B/v6e-64). Matches the
+        pattern in weight_utils.py (see comment at :678).
+        """
+        sharding = NamedSharding(self.mesh, P(*spec))
+        arr = np.asarray(arr)
+        return jax.make_array_from_callback(arr.shape, sharding, lambda idx: arr[idx]).astype(
+            self.dtype
         )
 
     @staticmethod


MokusMokun (approved):
Reviewed and validated on v6e 4×4 (4-host, tp16/dp4) at 9c1ddf72.

Code: confirmed the claimed mechanism in jax `dispatch.py` (numpy/uncommitted input + multi-host sharding with all processes participating → `assert_equal` → `process_allgather`, i.e. process_count copies per device), and the new `_put` is bit-identical to the old path on CPU for bf16/fp32 sources including the non-contiguous transpose views at the w1/qkvz call sites; `make_array_from_callback(...).astype(...)` matches the existing idiom at weight_utils.py:1142/:1236, and the no-op astype is zero-copy when ckpt dtype == model dtype.

Validation: 35B-A3B GPQA-Diamond 5-seed median 0.838 (0.833–0.874, gate ≥0.827 passes); 122B-A10B fires up on 16 chips with the exact loader summary reported here (831 consumed / 1118 skipped / 0 missing / 0 unexpected), clean multi-host decode (no NaN/`!`), GPQA 5-seed 0.833–0.899 (median 0.843, mean 0.857) — consistent with the single-run 0.864 in the PR given the known ~5pp seed spread.

MokusMokun:
Follow-up investigation on the TPU/XLA side of the loader OOM mechanism, using the NCCL collectives doc as the comparison point.

NCCL's useful mental model here is rank-level data movement: AllGather gives every rank the concatenation of all ranks' inputs, ReduceScatter reduces then leaves each rank with one slice, and AlltoAll exchanges rank-specific chunks. StableHLO/XLA has the same semantic level, but expressed as compiler IR collectives (`all_gather`, `all_reduce`, `all_to_all`, `reduce_scatter`, `collective_broadcast`, `collective_permute`) over process groups / replica groups. On TPU this is not a user-visible NCCL call; the XLA backend/runtime chooses the device communication implementation from the StableHLO/HLO collective semantics and sharding metadata.

For this PR, the important finding is that the problematic allgather is not a model HLO collective from Qwen3.5 execution. It is a host-side JAX placement safety path:

- `jax.device_put(host_np_array, NamedSharding(global_mesh, pspec))` with a non-fully-addressable multi-host sharding enters JAX's sharded `device_put` implementation. When the source is a NumPy / host array and every host participates in the target sharding, JAX checks that all hosts are passing the same replicated value before it builds the deferred shard argument. Current JAX source: https://github.com/jax-ml/jax/blob/7d2cb09952/jax/_src/dispatch.py#L504-L520
- That equality check is `multihost_utils.assert_equal`, which calls `process_allgather(..., tiled=True)` and then compares the gathered values. Source: https://github.com/jax-ml/jax/blob/7d2cb09952/jax/experimental/multihost_utils.py#L168-L180
- `process_allgather` documents that NumPy / fully-addressable inputs are gathered across processes, with the tiled case concatenating outputs. Source: https://github.com/jax-ml/jax/blob/7d2cb09952/jax/experimental/multihost_utils.py#L138-L155

So the observed scaling matches the source-level mechanism: for a full host tensor, the equality check can materialize `process_count` copies before the intended sharding happens. On the 16-host 122B-A10B case, the ~1.5G `gate_up` tensor therefore becomes roughly ~24G of gathered payload, before accounting for the already-loaded model state.

The new `_put` path is the right primitive for this loader shape. `jax.make_array_from_callback` asks the callback for each addressable shard index and then builds the global array from those per-device values. Public JAX docs say the callback fetches data for each addressable shard, and the implementation constructs `device_to_index_map` for non-replicated shardings before `batched_device_put`: https://docs.jax.dev/en/latest/_autosummary/jax.make_array_from_callback.html and https://github.com/jax-ml/jax/blob/7d2cb09952/jax/_src/array.py#L783-L835

Net conclusion: this PR is not relying on a TPU/XLA collective to redistribute a full tensor after upload. It changes the host-to-device contract from "replicated full host value; verify by process allgather; then shard" to "materialize only the local addressable shard values requested by the global sharding." That is exactly the distinction we want for large fused MoE weights, and it matches the existing lazy loader pattern in `weight_utils.py`.

PR #1435, opened 2026-07-04
feat(scheduler): add shape_aware DP scheduling policy

Yajingleo:
## Summary

Adds a shape-aware data-parallel routing policy, `shape_aware`, alongside `round_robin`, `min_running_queue`, and `cache_aware`. It balances prefill (input) and decode (output) token load **jointly** across DP replicas, routing each request to the replica whose *bottleneck* dimension stays smallest after admission:

```
score(rank) = max( Σinput + input_r,  Σoutput + output_r )   →  argmin
```

The `max` keeps whichever resource is binding on each rank (prefill/FLOPs vs. decode/KV-HBM) lowest, so a prefill-heavy request is drawn toward a decode-heavy rank and vice versa, co-locating complementary shapes rather than piling like-shaped work onto one rank.

Per-rank load includes the input/output split of requests assigned earlier in the **same intake tick** (pending-io tracking). This is load-bearing at low concurrency: without it, a burst routes against a stale snapshot and mis-balances (see the ablation below). It plugs into `assign_dp_ranks` the same way `cache_aware` does; the decision math is a pure, dependency-free function in `dp_schedule_policy.py`. Routing is `O(replicas)` plain arithmetic on the CPU scheduler thread, off the accelerator critical path.

## Why this score

Currently, we don't distinguish FLOP-bound and HBM-bound requests when scheduling DP. This is inspired by a cluster (k8s) scheduling score to balance CPU and RAM pressure per device

```
max(FLOPs Util, HBM Util)
```

Here, we simply use input token to represent FLOPs pressure, and use output token to represent HBM pressure. Ideally, we need normalization or weight to apply to input and output. 

```
max(input token / prefill capacity, output token / decode capacity)
```

Here we simply assume prefill and decode capacity are the same. We can fine tune the weights in the future, using cost factor or real FLOPs / HBM GB estimations.

## When it helps

Shape-aware routing matters only when compute is **stranded** — long outputs + low concurrency + a saturated KV pool, where a small decode batch is memory-bandwidth-bound and FLOPs sit idle (the operating point for long-generation / reasoning serving). On ordinary high-concurrency traffic there is no imbalance to exploit and it is within noise of `min_running_queue` (no regression).

## Experiment results

Single-node **TPU v6e-4**, **Qwen3-8B** bf16, `dp=4`. Bimodal workload: compute-heavy stream A (`in=2000/out=32 ×60`) + memory-heavy stream B (`in=64/out=20000 ×16`), stranded config (`CONC=8`, `RATE=2`, `mem-fraction-static=0.50`), averaged over 2 seeds. Metrics are for stream B (the memory-heavy stream that dominates makespan). Δ is vs `min_running_queue` (the stronger baseline):

| Metric (stream B) | round_robin | min_running_queue | **shape_aware** | Δ vs mrq |
|---|---|---|---|---|
| Throughput (tok/s) ↑ | 228 | 227 | **280** | **+23%** |
| P90 E2E latency ↓ | 104.7 s | 96.7 s | **71.6 s** | **−26%** |
| P99 E2E latency ↓ | 105.5 s | 107.1 s | **84.0 s** | **−22%** |

The gain is consistent across both seeds (throughput 277.6 / 281.6 tok/s). On compute-bound / unimodal workloads all policies are within noise (no regression).

### Ablation: pending-io tracking is required

Same config, comparing the policy **with vs. without** the per-tick pending-io split:

| Variant | B throughput | B P90 | Notes |
|---|---|---|---|
| **with** pending-io (this PR) | **280 tok/s (+23%)** | **71.6 s** | stable across seeds |
| without pending-io | 225 tok/s (~tie) | 121.5 s | unstable: 164 vs 287 tok/s by seed |

Without pending-io, a low-concurrency burst routes against a stale snapshot and mis-balances — throughput collapses to a wash and swings wildly by arrival order. This is why the per-tick split is included.

## Reproduce

On a TPU v6e-4 host. Start the server (swap `--dp-schedule-policy` between `min_running_queue` and `shape_aware`):

```bash
python -m sgl_jax.launch_server --model-path Qwen/Qwen3-8B --trust-remote-code \
  --device tpu --dtype bfloat16 --tp-size 4 --dp-size 4 \
  --dp-schedule-policy shape_aware \
  --mem-fraction-static 0.50 --chunked-prefill-size 2048 --page-size 128 \
  --host 127.0.0.1 --port 30000
```

Drive the two streams **concurrently** (compute-heavy A + memory-heavy B), then compare stream B's `output_throughput` / P90 / P99 across policies:

```bash
# Stream A (compute-heavy: long input, short output)
python -m sgl_jax.bench_serving --backend sglang --model Qwen/Qwen3-8B \
  --base-url http://127.0.0.1:30000 --dataset-name random \
  --num-prompts 60 --request-rate 2 --max-concurrency 8 \
  --random-input-len 2000 --random-output-len 32 --random-range-ratio 0.6 --seed 1 &

# Stream B (memory-heavy: short input, long output) -- the reported metrics
python -m sgl_jax.bench_serving --backend sglang --model Qwen/Qwen3-8B \
  --base-url http://127.0.0.1:30000 --dataset-name random \
  --num-prompts 16 --request-rate 2 --max-concurrency 8 \
  --random-input-len 64 --random-output-len 20000 --random-range-ratio 0.6 --seed 1 &
wait
```

The effect appears only in this stranded regime (small saturated KV pool + low concurrency + long outputs); on compute-bound traffic the policies tie.

## Testing

- `test/srt/test_dp_schedule_shape_aware.py` — 8 pure-function unit tests for `pick_shape_aware_dp` (no JAX dependency; run with `PYTHONPATH=python pytest`).
- `test/srt/test_dp_rank_assignment.py` — dispatcher characterization for the new policy path.
- Full CPU unit suite green (`python test/srt/run_suite.py --suite unit-test-cpu`): **33 passed** on the v6e-4 CI image. Server boots and routes cleanly under `--dp-schedule-policy shape_aware`.

### DIFF
--- a/test/srt/test_dp_rank_assignment.py
+++ b/test/srt/test_dp_rank_assignment.py
@@ -38,6 +38,9 @@ class _DpAssignmentHarness:
         self.per_dp_max_running_requests = per_dp_max_running_requests
         self.round_robin_counter = 0
         self.cache_aware_calls: list[tuple[str | list[str] | None, list[int], list[int]]] = []
+        # Records (item_input, item_output, extra_input_counts, extra_output_counts)
+        # per shape_aware dispatch, so tests can assert the same-tick pending-IO split.
+        self.shape_aware_calls: list[tuple[int, int, list[int], list[int]]] = []
 
     def select_round_robin_dp(self) -> int:
         dp_rank = self.round_robin_counter % self.dp_size
@@ -80,8 +83,12 @@ class _DpAssignmentHarness:
         extra_input_counts: list[int],
         extra_output_counts: list[int],
     ) -> int | None:
-        # These characterization tests exercise the dispatcher, not the shape-aware
-        # math (covered by test_dp_schedule_shape_aware.py); reuse min-running here.
+        # Record the IO-specific args so the dispatcher's same-tick pending-IO
+        # tracking is under test; the routing decision itself (shape-aware math)
+        # is covered by test_dp_schedule_shape_aware.py, so reuse min-running here.
+        self.shape_aware_calls.append(
+            (item_input, item_output, list(extra_input_counts), list(extra_output_counts))
+        )
         return self.select_min_running_dp(extra_counts, extra_token_counts)
 
     @staticmethod
@@ -201,6 +208,25 @@ def test_cache_aware_receives_pending_load_from_prior_requests():
     assert result.pending_reqs == []
     assert new.dp_rank == 1
     assert harness.cache_aware_calls == [("new", [1, 0], [8, 0])]
+
+
+def test_shape_aware_receives_pending_io_split_from_prior_requests():
+    # Guards the load-bearing same-tick pending-IO tracking: a request assigned
+    # earlier in this intake tick (the sticky request on rank 0) must be reflected
+    # in the input/output split passed to select_shape_aware_dp for the next
+    # request -- distinct input/output values here catch a swapped or dropped axis.
+    harness = _DpAssignmentHarness()
+    sticky = _req("sticky", dp_rank=0, input_len=10, max_new_tokens=3)  # io = (10, 3)
+    new = _req("new", input_len=5, max_new_tokens=7)  # item io = (5, 7)
+
+    result = _assign(recv_reqs=[sticky, new], policy="shape_aware", harness=harness)
+
+    assert result.ready_reqs == [sticky, new]
+    assert result.pending_reqs == []
+    # select_shape_aware_dp was called once (for `new`), seeing the incoming item's
+    # (input, output) = (5, 7) and the sticky request's IO accumulated on rank 0:
+    # per-rank pending input = [10, 0], pending output = [3, 0].
+    assert harness.shape_aware_calls == [(5, 7, [10, 0], [3, 0])]
 
 
 def test_dp_size_one_assigns_generate_reqs_and_preserves_control_reqs():
--- a/test/srt/test_dp_rank_assignment.py
+++ b/test/srt/test_dp_rank_assignment.py
@@ -71,10 +71,27 @@ class _DpAssignmentHarness:
         self.cache_aware_calls.append((req.rid, list(extra_counts), list(extra_token_counts)))
         return self.select_min_running_dp(extra_counts, extra_token_counts)
 
+    def select_shape_aware_dp(
+        self,
+        item_input: int,
+        item_output: int,
+        extra_counts: list[int],
+        extra_token_counts: list[int],
+        extra_input_counts: list[int],
+        extra_output_counts: list[int],
+    ) -> int | None:
+        # These characterization tests exercise the dispatcher, not the shape-aware
+        # math (covered by test_dp_schedule_shape_aware.py); reuse min-running here.
+        return self.select_min_running_dp(extra_counts, extra_token_counts)
+
     @staticmethod
     def estimate_req_tokens(req: TokenizedGenerateReqInput) -> int:
         max_new_tokens = req.sampling_params.get("max_new_tokens", 0)
         return len(req.input_ids or []) + max_new_tokens
+
+    @staticmethod
+    def estimate_req_io_tokens(req: TokenizedGenerateReqInput) -> tuple[int, int]:
+        return len(req.input_ids or []), req.sampling_params.get("max_new_tokens", 0)
 
     def _eligible(self, extra_counts: list[int]) -> list[int]:
         return [
@@ -103,7 +120,9 @@ def _assign(
         select_round_robin_dp=harness.select_round_robin_dp,
         select_cache_aware_dp=harness.select_cache_aware_dp,
         select_min_running_dp=harness.select_min_running_dp,
+        select_shape_aware_dp=harness.select_shape_aware_dp,
         estimate_req_tokens=harness.estimate_req_tokens,
+        estimate_req_io_tokens=harness.estimate_req_io_tokens,
     )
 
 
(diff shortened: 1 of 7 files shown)


test/srt/test_dp_schedule_shape_aware.py
```diff
@@ -0,0 +1,84 @@
...
+
+    score(rank) = max(input_counts[rank] + item_input,
+                      output_counts[rank] + item_output)
+
+Route to the eligible rank minimizing that bottleneck dimension. Like the
+other policies this lives in the dependency-free helper module, so the test
+imports neither Scheduler nor JAX. The end-to-end throughput/latency
+improvement is validated separately by a bimodal A/B (see the PR description).
+"""
+
```
JamesBrianD, line 14: Please add this test into the run_suite.

JamesBrianD:
@cjx0709 I reviewed this PR and the implementation looks reasonable to me. The shape-aware policy has a clear purpose: compared with balancing only by request count or queue length, it uses the input/prefill and output/decode token shapes as separate load signals, which should help avoid assigning requests with similar bottlenecks to the same DP rank.

My main concern is more about the long-term policy surface. As we add more DP scheduling policies, it may become harder for users to understand which one to choose for a given workload. This is not a blocker for this PR, but I think we should eventually consider having a recommended auto/default path, and keep the individual policies as explicit tuning or debugging options with documented workload assumptions.

test/srt/test_dp_schedule_shape_aware.py
Yajingleo: Done.

cjx0709:
> @cjx0709 I reviewed this PR and the implementation looks reasonable to me. The shape-aware policy has a clear purpose: compared with balancing only by request count or queue length, it uses the input/prefill and output/decode token shapes as separate load signals, which should help avoid assigning requests with similar bottlenecks to the same DP rank.
> 
> My main concern is more about the long-term policy surface. As we add more DP scheduling policies, it may become harder for users to understand which one to choose for a given workload. This is not a blocker for this PR, but I think we should eventually consider having a recommended auto/default path, and keep the individual policies as explicit tuning or debugging options with documented workload assumptions.

Now we have the `["round_robin", "min_running_queue", "cache_aware", "shape_aware"]`(including this pr), `round_robin` is the basic, `cache_aware` is working for the radix cache situation, `shape_aware` seems performances better than `min_running_queue`.

I think we can do like this:
```
if args.dp_schedule_policy is not None:
  dp_schedule_policy = args.dp_schedule_policy
else:
  if args.enable_radix_cache:
    dp_schedule_policy = cache_aware
  else:
    dp_schedule_policy = shape_aware
```
And we update the args notes in the dp structure doc. What do you think?

JamesBrianD:
> > @cjx0709 I reviewed this PR and the implementation looks reasonable to me. The shape-aware policy has a clear purpose: compared with balancing only by request count or queue length, it uses the input/prefill and output/decode token shapes as separate load signals, which should help avoid assigning requests with similar bottlenecks to the same DP rank.
> > My main concern is more about the long-term policy surface. As we add more DP scheduling policies, it may become harder for users to understand which one to choose for a given workload. This is not a blocker for this PR, but I think we should eventually consider having a recommended auto/default path, and keep the individual policies as explicit tuning or debugging options with documented workload assumptions.
> 
> Now we have the `["round_robin", "min_running_queue", "cache_aware", "shape_aware"]`(including this pr), `round_robin` is the basic, `cache_aware` is working for the radix cache situation, `shape_aware` seems performances better than `min_running_queue`.
> 
> I think we can do like this:
> 
> ```
> if args.dp_schedule_policy is not None:
>   dp_schedule_policy = args.dp_schedule_policy
> else:
>   if args.enable_radix_cache:
>     dp_schedule_policy = cache_aware
>   else:
>     dp_schedule_policy = shape_aware
> ```
> 
> And we update the args notes in the dp structure doc. What do you think?

I agree with your suggestion.

test/srt/test_dp_rank_assignment.py
```diff
@@ -71,11 +71,28 @@ def select_cache_aware_dp(
...
+        item_input: int,
+        item_output: int,
+        extra_counts: list[int],
+        extra_token_counts: list[int],
+        extra_input_counts: list[int],
+        extra_output_counts: list[int],
+    ) -> int | None:
+        # These characterization tests exercise the dispatcher, not the shape-aware
+        # math (covered by test_dp_schedule_shape_aware.py); reuse min-running here.
+        return self.select_min_running_dp(extra_counts, extra_token_counts)
```
MokusMokun, line 92: The pure `pick_shape_aware_dp` tests cover the scoring logic, but the dispatcher path for `shape_aware` is still untested. In `test_dp_rank_assignment.py`, no test calls `_assign(policy="shape_aware")`, and the harness implementation of `select_shape_aware_dp` ignores the IO-specific arguments and delegates to min-running.

That means the load-bearing same-tick pending IO tracking can regress without any test failing: swapping `pending_input_counts`/`pending_output_counts`, failing to update them in `track()`, or wiring the wrong `(item_input, item_output)` into `select_shape_aware_dp` would still pass the current suite.

__Suggestion__: Add a dispatcher-level `shape_aware` characterization test, similar to `test_cache_aware_receives_pending_load_from_prior_requests`, that records `item_input`, `item_output`, `extra_input_counts`, and `extra_output_counts`, then asserts that prior same-tick requests are reflected in the IO split passed to `select_shape_aware_dp`.

cjx0709 (approved):

Yajingleo:
Thanks so much for the careful and quick reviews with insights @JamesBrianD @cjx0709 @MokusMokun!

PR #1443, opened 2026-07-07
[Feat] Add skills for model-eval-benchmark

lingebeng:
I have tested them in claude-opus-4.8 and gpt-5.5.

## Example1

> In 30000 port of ssh tpu-v6e，use gsm8k to eval --limit 20.

### Output

# Benchmark

## Status

- State: success
- Target resource: `tpu-v6e` (ssh host `tpu-v6e`, GCP `ct6e-standard-1t`)
- Endpoint: `http://127.0.0.1:30000/v1/chat/completions`
- Model path: `/models/mistralai/Mistral-7B-Instruct-v0.3`
- Server: healthy
- Server id: PID 343391
- Hardware:
  - TPU type: tpu-v6e
  - Topology: 1x1 (v6e-1)
  - Chips: 1
- Versions:
  - engine (sglang-jax): `0.0.0.dev558+ga64070690` @ 6f6d2844 (haifeng/feat/mistral-7B)
  - `evalscope`: 1.8.1
  - `jax`: 0.8.1
  - `libtpu`: 0.0.30
- Server launch args:

```bash
.venv/bin/python -u -m sgl_jax.launch_server --model-path /models/mistralai/Mistral-7B-Instruct-v0.3 \
  --trust-remote-code --tp-size 1 --device tpu --random-seed 3 --mem-fraction-static 0.85 \
  --max-prefill-tokens 8192 --chunked-prefill-size 2048 --download-dir /dev/shm --dtype bfloat16 \
  --skip-server-warmup --host 0.0.0.0 --port 30000 --attention-backend fa --page-size 256 \
  --max-running-requests 16
```

### gsm8k

```bash
evalscope eval \
  --model /models/mistralai/Mistral-7B-Instruct-v0.3 \
  --api-url http://127.0.0.1:30000/v1/chat/completions \
  --eval-type openai_api \
  --datasets gsm8k \
  --limit 20 \
  --eval-batch-size 16 \
  --generation-config '{"max_tokens":4096,"temperature":0.0}' \
  --work-dir /tmp/evalscope_gsm8k_20_20260707_123026 \
  --no-timestamp
```

| Dataset | Metric   | Sample count | Correct | Score |
| ------- | -------- | ------------ | ------- | ----- |
| gsm8k   | mean_acc | 20           | 9/20    | 0.45  |

| Average latency (s) | Average throughput (tok/s) | Average input (tok) | Average output (tok) |
| ------------------- | -------------------------- | ------------------- | -------------------- |
| 3.66                | 65.72                      | 729.25              | 240.5                |

- Artifacts:
  - `/tmp/evalscope_gsm8k_20_20260707_123026/reports/Mistral-7B-Instruct-v0.3/gsm8k.json`
  - `/tmp/evalscope_gsm8k_20_20260707_123026/reports/report.html`

## Notes

- None

## Example2

> In 30000 port of ssh tpu-v6e，use sharegpt to bench output=128， and concurrency = "1 2 4 8"

### Output

# Benchmark

## Status

- State: success
- Target resource: `tpu-v6e` (ssh host `tpu-v6e`, GCP `ct6e-standard-1t`)
- Endpoint: `http://127.0.0.1:30000`
- Model path: `/models/mistralai/Mistral-7B-Instruct-v0.3`
- Server: healthy
- Server id: PID 343391
- Radix cache: enabled
- Hardware:
  - TPU type: tpu-v6e
  - Topology: 1x1 (v6e-1)
  - Chips: 1
- Versions:
  - engine: sglang-jax `0.0.0.dev558+ga64070690` @ 6f6d2844 (haifeng/feat/mistral-7B)
  - `jax`: 0.8.1
  - `libtpu`: 0.0.30
- Server launch args:

```bash
.venv/bin/python -u -m sgl_jax.launch_server --model-path /models/mistralai/Mistral-7B-Instruct-v0.3 \
  --trust-remote-code --tp-size 1 --device tpu --random-seed 3 --mem-fraction-static 0.85 \
  --max-prefill-tokens 8192 --chunked-prefill-size 2048 --download-dir /dev/shm --dtype bfloat16 \
  --skip-server-warmup --host 0.0.0.0 --port 30000 --attention-backend fa --page-size 256 \
  --max-running-requests 16
```

## Speed Benchmark

```bash
DS=/home/gcpuser/.cache/huggingface/hub/datasets--anon8231489123--ShareGPT_Vicuna_unfiltered/snapshots/192ab2185289094fc556ec8ce5ce1e8e587154ca/ShareGPT_V3_unfiltered_cleaned_split.json
for bs in 1 2 4 8; do
  .venv/bin/python -m sgl_jax.bench_serving --backend sgl-jax --host 127.0.0.1 --port 30000 \
    --dataset-name sharegpt --dataset-path "$DS" --sharegpt-output-len 128 \
    --max-concurrency $bs --num-prompts $((bs * 3)) \
    --output-file /tmp/sgl_speed_sharegpt_20260707_123938/bs_$bs/result.jsonl
done
```

- Status: completed
- Output dir: `/tmp/sgl_speed_sharegpt_20260707_123938`
- Summary CSV: `/tmp/sgl_speed_sharegpt_20260707_123938/summary.csv`
- Backend: sgl-jax
- Dataset: sharegpt
- Workload shape: `sharegpt_output_len=128` (num_prompts = bs × 3)
- Cache hit rate: bs_1 0.0, bs_2 0.47, bs_4 0.15, bs_8 0.45
- Metrics:

| Max Concurrency | Input tok/s | Output tok/s | Peak output tok/s | Total tok/s | Mean E2E ms | Mean TTFT ms | Mean TPOT ms | Mean ITL ms | P99 ITL ms |
| --------------- | ----------: | -----------: | ----------------: | ----------: | ----------: | -----------: | -----------: | ----------: | ---------: |
| 1               | 141.36      | 76.78        | 78.0              | 218.14      | 1661.79     | 32.15        | 12.83        | 12.83       | 14.13      |
| 2               | 216.40      | 151.23       | 154.0             | 367.63      | 1687.29     | 30.15        | 13.05        | 13.05       | 14.22      |
| 4               | 979.76      | 287.53       | 308.0             | 1267.28     | 1775.24     | 68.50        | 13.44        | 13.45       | 14.45      |
| 8               | 1755.65     | 557.57       | 600.0             | 2313.22     | 1830.67     | 81.61        | 13.77        | 13.79       | 15.23      |

- Artifacts:
  - `/tmp/sgl_speed_sharegpt_20260707_123938/bs_1/result.jsonl`
  - `/tmp/sgl_speed_sharegpt_20260707_123938/bs_2/result.jsonl`
  - `/tmp/sgl_speed_sharegpt_20260707_123938/bs_4/result.jsonl`
  - `/tmp/sgl_speed_sharegpt_20260707_123938/bs_8/result.jsonl`

## Notes

- Ran via the `scripts/speed_benchmark.sh` wrapper; the command block above shows the expanded raw `bench_serving` loop it executed (per-point `--max-concurrency`/`--num-prompts` set by the wrapper, `num_prompts = bs × 3`).
- Radix cache is **enabled** on the server. ShareGPT prompts share some prefixes, so `cache_hit_rate` was non-zero (0.15–0.47) at bs 2/4/8, which inflates `Input tok/s` and lowers `Mean TTFT ms` at those points. Decode-side metrics (`Output tok/s`, `Mean TPOT/ITL ms`) are unaffected. For clean prefill/input-throughput numbers, relaunch the server with `--disable-radix-cache` — not changed here since it is a server-launch setting and you did not request a restart.


.claude/skills/model-eval-benchmark/scripts/render_benchmark_report.py
```diff
@@ -0,0 +1,182 @@
...
+    if speed_header:
+        speed_lines.append("")
+        speed_lines.append(_markdown_table(speed_header, speed_rows))
+    if args.speed_artifact:
+        speed_lines.append("")
+        speed_lines.append("- Raw result files:")
+        speed_lines.append(_format_paths(args.speed_artifact))
+    speed_lines.extend(f"- {note}" for note in args.speed_note)
+
+    report = f"""# Benchmark
```
cjx0709: [P1] This renderer still does not emit the report shape required by `SKILL.md`. The skill contract requires a top-level `## Status` section with state/server/version fields, and the included accuracy/speed sections need fields such as sampling params, summary, `Summary CSV`, etc. This generated markdown starts directly at `## Accuracy Benchmark` and uses `#### Results`, so a third-party user following the helper produces a report that violates the skill. Please either update the renderer to match the required template exactly, or remove/relax the helper from the workflow.

.claude/skills/model-eval-benchmark/scripts/render_benchmark_report.py
```diff
@@ -0,0 +1,182 @@
...
+    for row in rows:
+        padded = row + [""] * (len(header) - len(row))
+        lines.append("| " + " | ".join(padded[: len(header)]) + " |")
+    return "\n".join(lines)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--out", required=True, help="Output markdown path.")
+    parser.add_argument("--dataset", required=True, help="Accuracy dataset name.")
```
cjx0709: [P2] These required accuracy/speed args make the helper incompatible with the skill's accuracy-only and speed-only rules. `SKILL.md` says to omit an entire benchmark section when it was not run, but `--dataset`, `--accuracy-command`, `--speed-command`, and `--sample-count` are always required here. Please make each benchmark block optional, or split this into helpers that match the supported report shapes.

.claude/skills/model-eval-benchmark/SKILL.md
```diff
@@ -0,0 +1,399 @@
...
+Preferred rendering flow:
+
+1. Run the benchmarks and collect the concrete commands that were executed.
+2. Save raw artifacts such as the evalscope result json/work directory, benchmark log, `summary.csv`, and `bs_<N>/result.jsonl`.
+3. Generate the final markdown document directly, or with `scripts/render_benchmark_report.py` if the helper still matches the executed commands and produced artifacts.
+
+Remote sync step when wrappers are missing:
+
+```bash
+scp model-eval-benchmark/scripts/accuracy_benchmark.sh model-eval-benchmark/scripts/speed_benchmark.sh <target_host>:<sglang-jax root>/scripts/
```
cjx0709: [P2] These paths do not exist from the repo root. The files added by this PR live under `.claude/skills/model-eval-benchmark/scripts/...`, so `scp model-eval-benchmark/scripts/...` and the renderer example below will fail for a third-party developer copy-pasting from a checkout. Please use repo-root paths such as `.claude/skills/model-eval-benchmark/scripts/...`, or explicitly tell the user to `cd .claude/skills` before running these commands.

.claude/skills/model-eval-benchmark/scripts/speed_benchmark.sh
```diff
@@ -0,0 +1,184 @@
...
+  --request-rate VALUE
+  --python-bin VALUE
+  --append-output                  Append to existing result.jsonl instead of replacing it
+  -h, --help
+  --                               Pass remaining args directly to bench_serving
+
+Examples:
+  scripts/speed_benchmark.sh --batch-sizes "32 64 128"
+  scripts/speed_benchmark.sh --model mistralai/Mistral-7B-Instruct-v0.3 --out /tmp/mistral_speed
+  scripts/speed_benchmark.sh --dataset-name sharegpt --output-len 1024
```
cjx0709: [P2] This ShareGPT example is misleading: `--output-len` is only translated to `--random-output-len` when `DATASET_NAME == random`, so with `--dataset-name sharegpt` it is silently ignored. Please either show the pass-through form, e.g. `-- --sharegpt-output-len 1024`, or add a dedicated wrapper flag that maps to `bench_serving`'s ShareGPT option.

cjx0709 (changes requested):
Thanks for the lint fix; CI is green now. I still think this needs changes before merge from a third-party developer usability perspective.

I also compared this with the closest skill in ../sglang, `.claude/skills/llm-serving-auto-benchmark`. There is no exact same skill there: the SGLang one is a broader config/schema/validator/results-normalizer workflow for cross-framework serving benchmarks, while this PR is a narrower SGL-JAX/TPU/evalscope runbook. That narrower scope is fine. The key thing to borrow from the SGLang skill is the discipline around a single output contract: docs, helper scripts, examples, and generated reports should all agree. Right now the renderer, path examples, and ShareGPT example still diverge from the written contract, which makes the skill harder for external contributors to use reliably.

JamesBrianD:
I have a higher-level concern about the skill boundary here, separate from the existing inline comments about renderer/path/example details.

Accuracy eval and serving benchmark feel like two different workflows and should not be merged into one mandatory skill path. Eval is about dataset/scoring correctness (`evalscope eval` or `test/srt/run_eval.py` fallback), with artifacts like per-dataset reports and scores. Serving benchmark is about throughput/latency sweeps (`sgl_jax.bench_serving`), with artifacts like `summary.csv` and `bs_<N>/result.jsonl`. They have different triggers, required inputs, validation checks, and output fields.

Combining them makes the skill over-eager: a user asking only for eval may inherit benchmark-only requirements, and a user asking only for serving throughput may inherit eval-only fields. That is also why the report contract becomes hard to keep consistent.

Could we split this into a standalone eval skill plus a standalone serving benchmark skill? If we still want a one-shot "run both" experience, I would make that a thin orchestrator that explicitly invokes the two narrower flows, rather than one large skill owning both contracts.

lingebeng:
> I have a higher-level concern about the skill boundary here, separate from the existing inline comments about renderer/path/example details.
> 
> Accuracy eval and serving benchmark feel like two different workflows and should not be merged into one mandatory skill path. Eval is about dataset/scoring correctness (`evalscope eval` or `test/srt/run_eval.py` fallback), with artifacts like per-dataset reports and scores. Serving benchmark is about throughput/latency sweeps (`sgl_jax.bench_serving`), with artifacts like `summary.csv` and `bs_<N>/result.jsonl`. They have different triggers, required inputs, validation checks, and output fields.
> 
> Combining them makes the skill over-eager: a user asking only for eval may inherit benchmark-only requirements, and a user asking only for serving throughput may inherit eval-only fields. That is also why the report contract becomes hard to keep consistent.
> 
> Could we split this into a standalone eval skill plus a standalone serving benchmark skill? If we still want a one-shot "run both" experience, I would make that a thin orchestrator that explicitly invokes the two narrower flows, rather than one large skill owning both contracts.

Thanks a lot. You're right, I'll split it into two parts.

.claude/skills/model-eval-benchmark/scripts/render_benchmark_report.py
lingebeng: Thanks a lot, I'll redesign it.

cjx0709 (approved):

PR #1444, opened 2026-07-07
Qwen2.5-VL per-request encode plan

lianfang0320:
## Summary

This PR changes Qwen2.5-VL vision encode scheduling from per-image rounds to per-request rounds.

For a single request with multiple images, the scheduler now packs all vision patch rows for that request into one ViT encode round on the owning DP rank, then merges the encoded rows back into the request's placeholder positions. This reduces multi-image requests from N vision encode calls to 1 while preserving the existing owning-rank DP merge path.

## Implementation Details

- `ScheduleBatch.get_model_worker_batch` keeps the existing multimodal merge fields (`input_embedding`, mRoPE positions, deepstack inputs) and only builds an embed plan for normal prefill batches with multimodal inputs.
- `mm_utils.build_mm_embed_plan` collects request-level encode units per DP rank. Each unit owns all Qwen vision items for one request, along with the request's rank-local token base.
- Each embed round derives patch concat order, request-level metadata, `src_idx`, and `mask` from the same `unit.images` traversal. This is the core invariant: the encoded feature row order must match the placeholder scatter order.
- Qwen2.5-VL metadata is packed at request granularity. Windowed ViT blocks continue to use `cu_window_seqlens`; full-attention blocks use `cu_image_seqlens` so packed images remain attention-isolated.
- `_device_put_embed_plan` keeps embed-plan leaves data-leading for DP execution. Forward then loops over embed rounds, runs `get_image_feature`, applies `merge_jit`, and passes the fused embedding into the existing Qwen2 backbone path.
- `merge_jit`, `embed_mm_inputs`, and the language backbone JIT remain structurally unchanged.

## End-to-End Flow

```mermaid
flowchart TD
    A["ScheduleBatch.get_model_worker_batch"] --> B["_merge_multimodal<br/>input_embedding / mRoPE / deepstack"]
    A --> C{"contains_mm_inputs<br/>and ForwardMode.EXTEND"}
    C -- no --> Z["mm_embed_plan = None"]
    C -- yes --> D["mm_utils.build_mm_embed_plan<br/>reqs_info, dp_size, model_config, per_dp_token"]

    D --> E["Collect request-level encode units per DP rank<br/>ReqEncodeUnit(images, req_base)"]
    E --> F["n_rounds = max request units across ranks<br/>missing ranks use dummy lanes"]
    F --> G["_build_embed_round"]

    G --> H["Single traversal of unit.images"]
    H --> H1["concat patch rows<br/>pixels"]
    H --> H2["builder.get_metadata(images)<br/>request-level Qwen metadata"]
    H --> H3["item.offsets -> src_idx / mask<br/>merge_row increments across images"]

    H1 --> I["pad and stack pixels<br/>[dp, patch_k, dim]"]
    H2 --> J["builder.stack_metadata<br/>request metadata padded by patch_k"]
    H3 --> K["data-leading src_idx / mask"]
    I --> L["MultimodalEmbedPlan<br/>IMAGE: EmbedRound[]"]
    J --> L
    K --> L

    L --> M["ForwardBatch.init_new<br/>_device_put_embed_plan"]
    M --> N["model_runner.forward"]
    N --> O["general_mm_embed_routine<br/>embed_mm_inputs loop"]
    O --> P["Qwen2.5-VL get_image_feature<br/>encode_jit"]
    P --> Q["ViT attention<br/>windowed: cu_window_seqlens<br/>full-attention: cu_image_seqlens"]
    Q --> R["merge_jit<br/>scatter encoded rows into placeholders"]
    R --> S["Qwen2 backbone JIT<br/>consumes fused input embedding"]
```

## Validation

Validated feature branch: `feat/qwen25vl-per-request-encode@db64b6b7`  
Baseline: `epic/vlm@3e0b4a54`

### CPU Unit Tests

`test_vlm_stage1_framework.py`: `33 passed, 2 skipped`.

This suite directly tests the local plan/metadata/merge invariants:

- request-level metadata packing for multi-image inputs
- single-image degeneration to the old behavior
- block-diagonal attention semantics on a small CPU ViT
- per-request embed plan construction
- placeholder row count vs encoded feature row count checks
- DP dummy lane padding
- uneven multi-image requests across DP ranks
- encode/merge sharding paths

### TPU Functional Tests

All TPU validation phases passed.

| Area | What was tested | Result |
| --- | --- | --- |
| Single-image E2E regression | Sent the existing Qwen-VL request matrix through the server: text-only, real image QA, OCR, counting, chart understanding, multi-image prompts, and interleaved text/image prompts. This checks that the per-request encode change does not regress existing single-image and mixed multimodal behavior. | 9/9 passed. |
| Multi-image E2E routing | Sent concurrent real-image requests where each request contains two images and expects an answer tied to its own images. This detects cross-request contamination and cross-image routing mistakes after packing. | N=16, CONC=8, `correct_routed=16/16`. |
| Multi-image ViT parity vs HF | Compared packed sglang-jax ViT output against HF Qwen2.5-VL multi-image ViT output on the same multi-image input. Outputs were split back per image and compared numerically. | gcos small/mid/large = `0.99993 / 0.99954 / 0.99986`. |
| Cross-image leakage | Re-encoded the same target image while changing or clearing neighboring images in the same packed request. If full-attention boundaries leaked across images, the target image output would change. | Target image output unchanged, `max_abs=0`. |
| Multi-image logprobs parity | Sent 2-image and 3-image requests with `return_logprob`, then compared tokenization alignment, image placeholder counts, greedy first token, argmax agreement, and input-token logprobs against sglang/HF references. | Text input ids matched 22/22; image pad counts matched 256/688; first greedy token matched across sglang-jax / sglang / HF; sglang-jax vs HF itlp Pearson `0.99983`. |
| DP1 concurrent serving | Ran concurrent mixed single-image and multi-image requests on dp1. This checks batching, per-request packing, and merge routing under concurrent load. | conc 24/24; real-image concurrent routing `correct_routed=32/32`. |
| DP2 concurrent serving | Ran the same concurrent mixed workload on dp2. This checks owning-rank assignment, dummy lanes, per-DP request grouping, and merge isolation across DP ranks. | conc 24/24; real-image concurrent routing `correct_routed=32/32`; per-DP distribution `[4,4]/[4,3]`. |
| High-concurrency soak | Ran a larger concurrent multi-image workload to check stability under longer packed ViT sequences and repeated batching. This targets OOM, request retraction, and server crashes. | N=64, CONC=32, `correct_routed=64/64`; no OOM/retract/crash; server stayed alive; 5.45 req/s. |
| Multi-host tp16×dp4 | Ran multi-image smoke and concurrent real-image routing on 4-host TPU with tp16×dp4. This checks the same packing/merge behavior across host boundaries and DP replicas. | real_conc 32/16, `correct_routed=32/32`; per-DP distribution `[2,2,2,2]/[5,4,3,3]`; no OOM/crash. |
| Performance A/B | Compared `epic/vlm` per-image encode vs this PR's per-request encode on N=1/2/3 image requests. Checked ViT call count, latency, and memory. | ViT calls confirmed N→1. Small-N E2E latency was noise-level comparable: baseline `33 / 71 / 166 ms`, feature `33 / 74 / 165 ms`; no single-image or memory regression. |

## Performance Notes

The structural optimization is confirmed: a multi-image request now performs one ViT encode instead of one encode per image.

In the measured N=1/2/3 setup, E2E latency is roughly flat versus `epic/vlm` because LLM prefill dominates this configuration. The main expected win is reduced vision scheduling/encode overhead for larger multi-image or higher-concurrency workloads.

### DIFF
--- a/python/sgl_jax/srt/managers/mm_utils.py
+++ b/python/sgl_jax/srt/managers/mm_utils.py
@@ -6,11 +6,26 @@ the token merge and host loop that call model-specific encoders before the
 language backbone forward.
 """
 
+import dataclasses
 import functools
+from typing import Any
 
 import jax
 import jax.numpy as jnp
+import numpy as np
 from jax.sharding import PartitionSpec
+
+from sgl_jax.srt.multimodal.common.mm_plan import (
+    EmbedRound,
+    MultimodalEmbedPlan,
+    VisionEncodeInputs,
+)
+from sgl_jax.srt.multimodal.common.modality_enum import (
+    Modality,
+    MultimodalDataItem,
+    MultimodalInputs,
+)
+from sgl_jax.srt.multimodal.common.vision_metadata import VisionMetadataBuilderProtocol
 
 _MERGE_IN_SPECS = (
     PartitionSpec("data", None),  # running   [total_tok, H]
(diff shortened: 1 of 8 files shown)


python/sgl_jax/srt/managers/mm_utils.py
```diff
@@ -20,6 +35,184 @@
...
+                    f"got {type(req.mm_inputs).__name__}."
+                )
+            image_items = []
+            for item in mm_items:
+                if not isinstance(item, MultimodalDataItem):
+                    raise TypeError(
+                        "build_mm_embed_plan expects mm_items to contain "
+                        f"MultimodalDataItem, got {type(item).__name__}."
+                    )
+                if item.is_image() or item.is_video():
```
JamesBrianD, line 78: Non-blocking for this PR: Qwen2.5-VL image/video can share the same ViT encode after preprocessing, so reusing the encode implementation is fine. Since video is not fully supported in the current in-model path yet, I think this PR can land as image-focused. When adding video later, we should keep the control-plane contract separate from image: build `Modality.VIDEO` rounds, add `get_video_feature()` (it can call the same `visual.encode`), pass `video_data` through tokenizer/processor, and cover `video_grid_thw` / offsets / M-RoPE cases in tests.

python/sgl_jax/srt/managers/mm_utils.py
```diff
@@ -20,6 +35,184 @@
...
+
+def build_mm_embed_plan(
+    reqs_info: list[Any] | None,
+    dp_size: int,
+    model_config: Any,
+    per_dp_token: int,
+) -> MultimodalEmbedPlan | None:
+    """Build the owning-rank DP multimodal embed plan (host-side, numpy).
+
+    Round-loop: process one request's images per rank per round. ``n_rounds`` =
```
JamesBrianD, line 185: Non-blocking note: this PR moves from per-image encode rounds to per-request encode rounds, which is a good step for multi-image requests. Upstream SGLang is a bit more general here: it groups multimodal items by modality and calls the modality embedder with the whole item list, e.g. `get_image_feature(items)` / `get_video_feature(items)`. So per-request packing is fine for this stage, but for later performance work we may want to keep the plan extensible toward per-modality batching, especially for reducing ViT launches across multiple requests.

PR #1448, opened 2026-07-09
Improve DP scheduling defaults for radix cache

### LINKED ISSUE #1442: Follow up: improve DP scheduling defaults and cache_aware miss fallback
## Background

PR #1435 adds an opt-in `shape_aware` DP scheduling policy. It balances prefill/input and decode/output token load separately and uses the post-admission bottleneck dimension as the score:

```python
score(rank) = max(input_load + item_input, output_load + item_output)
```

This is a good first step, but we should follow up on how `shape_aware` composes with the existing policies, especially `cache_aware`, and on what the recommended/default policy should be.

## Problem

Today the DP policy surface is becoming harder to reason about:

- `round_robin`: simple baseline.
- `min_running_queue`: balances by current request count / total scheduled token load.
- `cache_aware`: prefers DP ranks with reusable radix-cache prefix, then falls back to least-loaded behavior.
- `shape_aware`: balances input/prefill and output/decode pressure separately.

The main gap is in `cache_aware`: when there is no usable cached prefix holder, or every eligible rank is effectively a cache miss, it currently falls back to least-loaded / min-running-style selection. In that miss path, it does not use the newly introduced `shape_aware` signal, so cold-cache or low-reuse workloads may lose the input/output balancing benefit even if `cache_aware` is selected as the recommended policy when radix cache is enabled.

### 1. Add shape-aware fallback for cache_aware miss path

Keep cache affinity as the first-order behavior, but improve the cache-miss fallback:

- If there are substantial cache holders, keep routing within the holder set as today.
- If there is no holder, no usable prompt prefix, or all eligible ranks are cache misses, choose the DP rank using `shape_aware` instead of plain least-loaded.
- Preserve the current large-load-skew protection (`BALANCE_ABS` / `BALANCE_REL`) so cache affinity cannot force traffic onto an overloaded rank.
- Consider whether a later step should use shape-aware tie-breaking within the holder set as well; for the first change, the lower-risk target is only the no-holder / miss fallback.

Expected behavior:

```python
if has_substantial_cache_holder:
    pick among holders by existing cache-aware load logic
else:
    pick_shape_aware_dp(eligible, input_counts, output_counts, item_input, item_output)
```

### 2. Add lightweight tests

Suggested coverage:

- No cached match: `cache_aware` chooses the same rank as `shape_aware`, not just the least request-count rank.
- Cache holder exists: holder preference still wins over a non-holder that would otherwise be shape-aware optimal.
- Large load skew: existing load-balancing override still applies.
- Pending input/output counts are included when `cache_aware` falls back to shape-aware for multiple requests in the same intake tick.

These should remain unit/characterization tests; no heavyweight serving e2e test should be required for this follow-up.

### 3. Define recommended/default DP policy behavior

After #1435 and the cache-aware fallback cleanup, consider making the default behavior easier to understand. One possible direction:

```python
if args.dp_schedule_policy is not None:
    dp_schedule_policy = args.dp_schedule_policy
else:
    if args.enable_radix_cache:
        dp_schedule_policy = "cache_aware"
    else:
        dp_schedule_policy = "shape_aware"
```

This should be done carefully because changing defaults has a wider blast radius than adding an opt-in policy.

Suggested approach:

- Keep this as a separate PR from #1435.
- Document workload assumptions for each policy in the DP structure / serving args docs.
- Confirm no regression on ordinary workloads using existing test/benchmark coverage and targeted serving results where practical.
- Keep explicit policy values available for tuning/debugging.

## Acceptance criteria

- `cache_aware` preserves cache-hit behavior.
- `cache_aware` uses shape-aware selection when there is no meaningful cache holder.
- Unit tests cover cache-hit, cache-miss, load-skew, and pending-count behavior.
- Recommended/default policy behavior is documented before any default change is merged.
- Any default-policy change is reviewed separately from the opt-in `shape_aware` implementation in #1435.

cjx0709:
## Summary

Closes #1442.

- Make `--dp-schedule-policy` unset by default so startup can distinguish an omitted policy from an explicit user override.
- Default DP scheduling to `cache_aware` when radix cache is enabled and `min_running_queue` when radix cache is disabled or Pathways PD is enabled.
- Keep `shape_aware` and `round_robin` as explicit tuning policies, and make cache-aware cache misses fall back to shape-aware load balancing.
- Document the new DP policy behavior and refresh the top-level server arguments page with several missing common serve flags.

## Expected behavior

```python
# ServerArgs normalization, after radix/hicache args are normalized.
if dp_schedule_policy is None:
    if disable_radix_cache or pd_disaggregation:
        dp_schedule_policy = "min_running_queue"
    else:
        dp_schedule_policy = "cache_aware"

# Cache-aware DP assignment.
if policy == "cache_aware":
    if eligible_ranks is empty:
        return None
    if eligible load skew is large:
        return least_loaded(eligible_ranks)
    if radix prefix holders exist:
        return least_loaded(prefix_holders)
    return shape_aware_pick(eligible_ranks, pending_input, pending_output, req_shape)
```

Explicit `--dp-schedule-policy` values are preserved, so users can still force `min_running_queue`, `cache_aware`, `shape_aware`, or `round_robin` for workload-specific tuning.

## v6e-4 validation

Environment:

- GKE cluster: `tpu-service-473302/asia-northeast1-b/niu-v6e-4x4-asia`
- TPU: v6e-4 from node pool `tpu-v6e-2x2`
- Model: `Qwen/Qwen3-8B`
- Server: `tp=4`, `dp=2/4`, TPU, bf16, FA

### Radix enabled

Deltas are relative to `min_running_queue`; the proposed default is `cache_aware`.

| dp | scenario | cache-aware result |
|---:|---|---|
| 2 | gsp | hit +4.2pp, total +6.6%, p90 -21.2% |
| 2 | gsp_no_ignore_eos | hit +1.4pp, total -5.2%, output +3.2%, p90 +0.3% |
| 2 | mixed_gsp | hit +5.6pp, total +20.8%, p90 -4.9% |
| 2 | mixed_random | hit +0.0pp, total +9.5%, p90 +1.8% |
| 2 | random_no_share | hit +0.0pp, total +2.4%, p90 -1.3% |
| 4 | gsp | hit +16.9pp, total +12.5%, p90 -34.7% |
| 4 | gsp_no_ignore_eos | hit +17.0pp, total +26.4%, p90 -11.0% |
| 4 | mixed_gsp | hit +20.6pp, total +4.9%, p90 +74.0% |
| 4 | mixed_random | hit +0.0pp, total +1.6%, p90 -37.5% |
| 4 | random_no_share | hit +0.0pp, total +33.7%, p90 -9.7% |

Takeaways:

- `cache_aware` is positive on most radix/cache-reuse cases, especially dp=4 GSP cases.
- Cache-miss fallback to shape-aware was neutral to positive for no-share/random cases in this run.
- Caveat: dp=2 `gsp_no_ignore_eos` regressed total throughput by 5.2%, although output throughput improved by 3.2%.
- `shape_aware` is not a safe radix default in this matrix: dp=4 `mixed_gsp` regressed total throughput by 14.3% and p90 by 168.7%.
- `round_robin` can be strong on homogeneous random/mixed throughput but does not preserve the cache-hit objective, so it remains an explicit tuning policy.

### Radix disabled

The proposed default remains `min_running_queue` for no-radix serving.

| dp | scenario | shape-aware result vs min_running_queue |
|---:|---|---|
| 2 | random_no_share | total +1.4%, p90 -1.8%, TTFT -6.3% |
| 2 | shape_prefill | total +0.0%, p90 +0.0%, TTFT -0.4% |
| 2 | shape_decode | total +2.0%, p90 -0.2%, TTFT +1.5% |
| 4 | random_no_share | total -0.2%, p90 +0.2%, TTFT +5.0% |
| 4 | shape_prefill | total +0.1%, p90 +0.7%, TTFT -0.8% |
| 4 | shape_decode | total -5.0%, p90 -0.6%, TTFT +2.9% |

Takeaway: no-radix should not default to `shape_aware` yet; it is mostly neutral but has a dp=4 shape-decode total throughput regression. Keeping `min_running_queue` avoids making this path riskier by default.

## Test plan

- `UV_PROJECT_ENVIRONMENT=/tmp/sgl-jax-pr-venv uv run --project python --extra cpu --with pytest python -m pytest -q test/srt/test_dp_schedule_policy.py test/srt/test_dp_schedule_shape_aware.py test/srt/test_dp_rank_assignment.py python/sgl_jax/test/mem_cache/test_unified_radix_tree_flag.py`
  - `46 passed in 13.65s`
- `git diff --check`

### DIFF
--- a/python/sgl_jax/srt/server_args.py
+++ b/python/sgl_jax/srt/server_args.py
@@ -123,7 +123,7 @@ class ServerArgs:
 
     # Data parallel
     dp_size: int = 1
-    dp_schedule_policy: str = "min_running_queue"
+    dp_schedule_policy: str | None = None
 
     # Logging
     log_level: str = "info"
@@ -478,6 +478,10 @@ class ServerArgs:
                     "hicache_write_policy must be one of write_through/"
                     f"write_through_selective/write_back, got {self.hicache_write_policy}"
                 )
+
+        if self.dp_schedule_policy is None:
+            use_no_radix_default = self.disable_radix_cache or bool(self.pd_disaggregation)
+            self.dp_schedule_policy = "min_running_queue" if use_no_radix_default else "cache_aware"
 
         if self.nnodes > 1 and self.device_indexes is not None:
             logger.warning("In a multi-machine scenario, device_indexes will be set to None.")
@@ -1175,6 +1179,8 @@ class ServerArgs:
             default=ServerArgs.dp_schedule_policy,
             help=(
                 "DP scheduling policy for assigning dp_rank to new requests. "
+                "When unset, defaults to 'cache_aware' with radix cache enabled "
+                "and 'min_running_queue' with radix cache disabled or Pathways PD. "
                 "'cache_aware' routes by cache affinity with soft load balancing: "
                 "it balances on large load skew, else picks the least-loaded rank "
                 "among those holding a substantial cached prefix, so a hot prefix "
--- a/docs/architecture/03-scheduler.md
+++ b/docs/architecture/03-scheduler.md
@@ -80,7 +80,7 @@ The `Scheduler` class inherits from three mixins:
 | `grammar_queue` | `list[Req]` | Asynchronous grammar compilation queue |
 | `aborted_reqs` | `dict[str, Req]` | Aborted requests (Request ID → Req mapping) |
 | `dp_size` | `int` | Number of Data Parallel ranks |
-| `dp_schedule_policy` | `str` | Policy for assigning new requests to DP ranks |
+| `dp_schedule_policy` | `str` | Policy for assigning new requests to DP ranks; auto-derived at startup when unset |
 
 ---
 
@@ -236,13 +236,14 @@ The `DFS_WEIGHT` policy traverses the Radix Tree, computing the number of waitin
 
 ### Data Parallel Request Assignment
 
-Data Parallel does not change the number of Scheduler processes. After receiving a new request, the Scheduler assigns `Req.dp_rank`, after which all batch construction, KV Cache allocation, and Attention Metadata are partitioned by DP rank. Three DP assignment policies are currently supported:
+Data Parallel does not change the number of Scheduler processes. After receiving a new request, the Scheduler assigns `Req.dp_rank`, after which all batch construction, KV Cache allocation, and Attention Metadata are partitioned by DP rank. If `--dp-schedule-policy` is unset, startup chooses `cache_aware` when radix cache is enabled and `min_running_queue` when `--disable-radix-cache` is set or Pathways PD is enabled. Explicit policy values remain available for tuning and debugging:
 
 | `dp_schedule_policy` | Description |
 |---|---|
-| `min_running_queue` | Picks the DP rank with the fewest currently running requests |
+| `cache_aware` | Prefers eligible DP ranks that hold a substantial reusable radix-cache prefix, preserves large-load-skew balancing, and uses shape-aware selection on cache misses |
+| `shape_aware` | Balances input/prefill and output/decode token load separately by minimizing the post-admission bottleneck dimension |
+| `min_running_queue` | Picks the DP rank with the fewest currently running requests, then scheduled tokens |
 | `round_robin` | Rotates new requests across ranks |
-| `cache_aware` | Prefer ranks with a substantial cached prefix while falling back to load balancing under skew |
 
 `waiting_queue` is still a Scheduler-level queue, but each `Req` already carries a target `dp_rank`. When constructing batches, `ScheduleBatch.reqs_info` keeps a `ScheduleReqsInfo` per rank, containing that rank's `reqs`, `chunked_req`, `input_ids`, `req_pool_indices`, `seq_lens`, and `out_cache_loc`; `per_dp_bs_size` records each rank's padded batch size, used by the execution side to construct DP-uniform inputs.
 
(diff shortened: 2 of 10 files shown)


python/sgl_jax/srt/server_args.py
```diff
@@ -479,6 +479,11 @@ def __post_init__(self):
                     f"write_through_selective/write_back, got {self.hicache_write_policy}"
                 )
 
+        if self.dp_schedule_policy is None:
```
MokusMokun, line 482: Pathways PD only sets `disable_radix_cache = True` later, in `_pd_make_meshes` (pathways_scheduler.py:48; the deepcopied P/D args at :72/:75 never re-run `__post_init__`), after this auto default has already resolved to `cache_aware`. So `--pd-disaggregation pathways` with no explicit policy runs `cache_aware` over ChunkCache, where `match_prefix` always returns an empty match: no request ever finds a holder, and every request takes the new shape-aware fallback instead of the documented `min_running_queue` no-radix default — the configuration the no-radix matrix above rejects as a default (dp=4 shape_decode total -5.0%). Reachable whenever pathways runs with `dp_size > 1`.

**Suggestion**: `pd_disaggregation` is already a `ServerArgs` field known at this point, so fold it into the resolution condition (treat `bool(self.pd_disaggregation)` as no-radix when picking the default) rather than keeping the sentinel alive past scheduler init. A regression test for the pathways unset-policy case next to `test_unset_dp_schedule_policy_follows_normalized_radix_state` would cover this.

MokusMokun (dismissed):
One blocking issue: the auto DP policy is resolved before Pathways PD disables radix.

JamesBrianD (approved):

PR #1451, opened 2026-07-10
perf(moe): O(n) scatter inverse permutation in EPMoE unpermute

ChengFu0118:
# perf(moe): O(n) scatter inverse permutation in EPMoE unpermute

## Motivation

`EPMoE._unpermute` rebuilds the inverse of the routing permutation so expert
outputs can be scattered back to their original token order. It did this with a
second full **stable argsort**:

```python
argsort_indices = jnp.argsort(sorted_selected_experts, stable=True)
```

But `sorted_selected_experts` is itself an `argsort` output produced in
`_permute` (`sorted_selected_experts = jnp.argsort(flatten_selected_experts, stable=True)`),
i.e. a **permutation of `arange(n)`**. The inverse of a permutation is obtainable
in **O(n)** with a single scatter (`inv[p] = arange(n)`) instead of an
**O(n log n)** sort. This removes one of the two per-MoE-layer sorts on the
decode hot path, and it is **model-agnostic** — every model using `EPMoE`
(Qwen3-MoE, DeepSeek, Grok, GLM, Bailing, etc.) benefits.

## Modifications

- `python/sgl_jax/srt/layers/moe.py` — in `EPMoE._unpermute`, replace the second
  `jnp.argsort` with a scatter-based inverse permutation
  (`jnp.zeros(n).at[sorted_selected_experts].set(jnp.arange(n))`). The scattered
  indices are **bit-identical** to the argsort result (a permutation has a unique
  inverse), so the unpermute output is **fp-identical**; only the algorithm that
  builds the index array changes.

Single-file change, +11/-1 lines.

## Accuracy Tests

The change is **fp-identical by construction**: `argsort` of a permutation and
the scatter inverse produce the same index array (a permutation has a unique
inverse), and `jnp.take` with those identical indices yields identical values.

Verified offline on CPU (jax 0.8.1) by calling the real `EPMoE._unpermute` and
comparing against the previous argsort-based path, bit-for-bit, across shapes
including a heavily imbalanced case (many tokens, few experts):

```
tokens=  20 experts=8: real _unpermute == old argsort path -> True | finite=True
tokens= 128 experts=8: real _unpermute == old argsort path -> True | finite=True
tokens= 300 experts=3: real _unpermute == old argsort path -> True | finite=True
tokens=   1 experts=1: real _unpermute == old argsort path -> True | finite=True
```

`unit-test-cpu` suite: passing (no new test added — the change is fp-identical
and existing MoE model paths exercise `EPMoE`).

## Benchmarking and Profiling

Because the output is fp-identical, this is a pure device-time reduction (removes
one sort per MoE layer per decode step). Measured on **v7x-8, gpt-oss-120b bf16,
TP8** (`bench_one_batch`, bs64, decode window):

- decode router/sort device-time: **88.7ms → 57.9ms (-34.7%)**
- sort op count: **8601 → 4360** (one sort/layer-step removed)
- end-to-end at c256 unchanged (router/sort is ~3% of decode there; that regime
  is collective/compute-bound): 1k/1k c256 ≈ 4611 tok/s, ITL ≈ 45.7ms (== base)

> Note for reviewers: the perf numbers above were measured on TPU v7x; the
> correctness/fp-identity claim is hardware-independent and verified above.

## Checklist

- [ ] (Optional) The necessary documentation update. — n/a (internal perf change)

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -726,7 +726,11 @@ class EPMoE(nnx.Module):
                 padding = jnp.zeros((padding_size, intermediate.shape[1]), dtype=intermediate.dtype)
                 intermediate = jnp.concatenate([intermediate, padding], axis=0)
 
-        argsort_indices = jnp.argsort(sorted_selected_experts, stable=True)
+        argsort_indices = (
+            jnp.zeros(expected_tokens, dtype=jnp.int32)
+            .at[sorted_selected_experts]
+            .set(jnp.arange(expected_tokens, dtype=jnp.int32))
+        )
         unsort_intermediate = jnp.take(intermediate, indices=argsort_indices, axis=0)
 
         total_tokens = weights.shape[0] * weights.shape[1] // self.num_experts_per_tok


cjx0709:
@ChengFu0118  hello, thanks for your contribution. The changes made here to improve performance truely make sense. But the pr body "Measured on v7x-8, gpt-oss-120b bf16, TP8 (bench_one_batch, bs64, decode window)" looks very strange, because in our sglang-jax main branch, we are not supporting gpt-oss model. Can you provide the commands to reproduce the performance improvement corresponding to your PR?

ChengFu0118:
Yeah, I added the support for `gpt-oss-120b` locally. 

I will update the performance with models supported on main branch.

ChengFu0118:
I re-benchmarked on two **supported** EPMoE models (Qwen3-30B-A3B, DeepSeek-V2-Lite), TPU v7x, TP8/EP8, `--moe-backend epmoe`.

**Correctness:** output is byte-identical to `main` (fp-identical by construction — the inverse permutation is unique; verified on both models incl. Qwen3 with real weights).

**Compiled graph:** one fewer `sort` op per MoE layer (the `_unpermute` argsort is removed; TopK + `_permute` unchanged):

| Model | main | this PR |
|---|---:|---:|
| Qwen3-30B-A3B | 144 | **96** |
| DeepSeek-V2-Lite | 78 | **52** |

**Throughput:** on par with `main` (within run-to-run noise), no regression:

| output tok/s | main | this PR |
|---|---:|---:|
| Qwen3-30B-A3B | 1492.8 | 1546.3 |
| DeepSeek-V2-Lite | 1796.4 | 1823.5 |

Repro (swap in DeepSeek-V2-Lite / `--page-size 256` for the second):

```bash
python -m sgl_jax.bench_offline_throughput \
  --model-path Qwen/Qwen3-30B-A3B --load-format dummy --trust-remote-code \
  --tp-size 8 --ep-size 8 --moe-backend epmoe --dtype bfloat16 \
  --attention-backend fa --page-size 128 \
  --dataset-name random --random-input 128 --random-output 128 --num-prompts 64
```

cjx0709 (approved):
LGTM

Issue #1452, opened 2026-07-12
[perf] jitted_run_model cpp fastpath miss every decode step under dp=1 + Pathways proxy backend


zorrofox:
# [perf] `jitted_run_model` cpp fastpath miss every decode step under dp=1 + Pathways proxy backend

## Symptom

On GLM-5.2-FP8 (epmoe/MLA) with `--dp-size=1` under Pathways single-controller (`JAX_PLATFORMS=proxy`), every warm decode step reports `#cache_miss: 1` from `count_pjit_cpp_cache_miss()` around `jitted_run_model`. Observed on both **colocated** and PD-disaggregated configs — this is not PD-specific.

Not observed with `--dp-size=2` (MiMo-V2 tp8 dp2: `#cache_miss: 0` in steady state).

Overhead: ~<1ms/step (<1.1% of 89ms ITL on GLM tp16), so not a throughput blocker, but it defeats the pjit cpp fastpath and forces the Python slow path on every call.

## Root cause

`ForwardBatch.from_model_worker_batch` (via `device_array`) constructs a **fresh `NamedSharding` object per call** for each metadata array (`input_ids`, `positions`, `seq_lens`, `cache_loc`, ...). Instrumented `id(sharding)` differs every step; `id(sharding.mesh)` is stable.

jaxlib `PjitFunctionCache` key (`CallSignature`, `jaxlib/jax_jit.{h,cc}` @ jax-v0.8.1):
- `HashShardingForJit(NamedSharding)` = `absl::Hash<void*>()(mesh().ptr())` — stable ✓
- `EqualShardingsForJit(NamedSharding a, b)`:
  - fast-path `a.ptr() == b.ptr()` — **fails** (fresh object each call)
  - slow-path compares `mesh().ptr()` (stable) + `spec` (stable) + `memory_kind.equal()` + `logical_device_ids().equal()`

One of `memory_kind` / `logical_device_ids` / `layout` (also in `CallSignature`) differs across fresh `NamedSharding` objects under dp=1 (size-1 `data` axis). Under dp=2 the same fields are stable, so the slow-path `==` returns true and the cache hits.

Verified: `_get_fastpath_data` returns non-None (so it's not an output-type or effects issue); `JAX_EXPLAIN_CACHE_MISSES` shows tracing-cache hits after the first per-bs call — the miss is purely the cpp `PjitFunctionCache` lookup.

## Repro

```bash
python -m sgl_jax.launch_server \
[...]

PR #1453, opened 2026-07-13
feat(pd): Pathways single-controller PD Stage 2 — perf + multi-D + hetero-TP

zorrofox:
# Pathways single-controller PD Stage 2: perf + multi-D + hetero-TP

## Summary

Follow-up to #1428 (Stage 1 base). Adds:

**Performance (single-controller dispatch overhead)**
- Fuse `run_model + sampler + resolve/set_future` into one jit → 3 Executes/tick to 1 on the D path
- Fold dp>1 gather/scatter into single jit (per-layer 48 Executes → 1 gather + ~13 batched scatter)
- SWA models: gather/scatter only the sliding-window tail pages instead of full seq_len (16K/256 → ~2.6GB/done cross-slice down to ~40MB)
- P-thread depth-1 overlap pipeline: eager-dispatch chunk N+1 under N's device wait
- Vectorize `_merge_cache_loc` (O(bs) loop → single gather); cap `cache_loc` bucket at pool size
- Pin scatter jit in/out shardings so donate aliases in-place (avoid full-pool copy)

**`--pd-num-decode` (1 prefill → N decode fan-out)**
- Batch-level round-robin routing across N decode meshes; each done batch goes to one D
- Per-D `running_batch`/`tp_worker`/allocator state, switched via `_pd_set_d`
- Dedicated `event_loop_overlap_pd_nd` when N>1

**Startup precompile (fixes first-request cold-compile timeout)**
- Base scheduler skips `run_precompile()` under PD; this PR precompiles P workers (extend-only) + all D workers (decode-only) + KV-transfer gather/scatter for every `_NPG_BUCKETS` shape at init
- GLM-5.2 78L: correctness first-request goes from 180s+ timeout (8-16 failed) to immediate PASS (336/336); startup +~16min (P extend 220s + D decode 680s + KV 50s)

**`--pd-prefill-tp-size` (heterogeneous P/D tensor parallelism)**
- Build P mesh with `tp_prefill` devices, D mesh with `--tp-size`; slices assigned by device count
- Currently supports MLA models (KV pool tensor-replicated so per-page shape matches across tp); MHA/GQA requires `kv_heads >= max(attention_tp_p, attention_tp_d)`
- Hetero-tp gather: stack L layers → 1 device_put → unstack (L ReshardOps → 1; ~350ms → ~3ms on 78L 32→16)
- Requires a hand-written JobSet to expose slices of both topologies (PathwaysJob CRD supports one `workers[]` entry only) — sample YAML below

**Correctness fixes** (found during Stage 2 development, all covered by the test matrix)
- `copy()` must not share `reqs` list with `running_batch` (tp16 CONC≥3 overlap corruption)
- `prefix_indices[:pre_len]` snapshot for P-thread race in `prepare_for_extend`
- SWA-tail slice from real pages (not bucket-padded); per-req default; `_swa_tail_pages=0` for non-SWA pools
- dp>1 admission per-dp/total unit consistency; per-rank D-pool capacity gate
- 1P-ND per-dp RR to break dp↔D phase lock

## Test plan

All runs on TPU via Pathways single-controller. Debug logs (`[pd-*]`) gated behind `SGLANG_PD_DBG` / `SGLANG_PD_DBG_TIMING`.

| model | KV pool | tp | dp | nP-nD | IL/OL | conc | correctness | tok/s (r2/r3) | gsm8k |
|---|---|---|---|---|---|---|---|---|---|
| MiMo-V2-Flash | SWA | 8 | 2 | 2P1D | 16K/4K | 128 | 336/336 | 15439 / 15341 | 0.885 |
| MiMo-V2-Flash | SWA | 16 | 2 | 2P1D | 16K/4K | 128 | 336/336 | 16102 / 16070 | 0.865 |
| MiMo-V2-Flash | SWA | 8 | 2 | 1P2D | 4K/16K | 128 | 336/336 | 3866 / — | 0.895 |
| MiMo-V2.5-Pro | SWA | 16 | 2 | 1P1D | 16K/4K | 64 | 336/336 | 7293 / 7282 | 0.980 |
| MiMo-V2.5-Pro | SWA | 16 | 2 | 2P1D | 16K/4K | 64 | 336/336 | 7391 / 7388 | 0.980 |
| GLM-5.2-FP8 | MLA | 16 | 1 | 1P1D | 16K/4K | 64 | 336/336 | 1078³ | 0.955 |
| GLM-5.2-FP8 | MLA | 32→16 | 1 | 1P1D-hetero | 16K/4K | 64 | 336/336 | 936³ | 0.965 |

³ GLM tp16 D-pool is HBM-capped at ~397K tokens (MLA KV ~88KB/token × 78L ≈ 35GB/device tensor-replicated); at 16K IL that bounds `#running` to ~20 (vs conc 64), so throughput is D-pool-bound. The MiMo rows above (SWA pool, `#running`=128) are the P-bound throughput measurements. MiMo-V2.5-Pro rows use `pd_decode_mem_fraction=0.75` / `max_running=64` (963G weights leave ~36G/device HBM headroom).

- Correctness: 336 requests × {SHORT/MID/LONG input} × {CONC 4/8/16}, greedy `max_new_tokens=8`, byte-equal per CONC group
- Throughput: `bench_serving --num-prompts 512 --request-rate inf`, 3 rounds, r2/r3 warm
- gsm8k: 200q, `max_new_tokens=1024`
- CPU unit tests: `pytest test/srt/disaggregation/test_pathways_pd.py` (11 pass, 10 skip on CPU)

### vs Stage 1 (#1428)

Stage 1 established functional parity (rate-limited MedITL Δ+0.6% vs colocated on GLM-5.2 tp16). Stage 2 adds the throughput-oriented optimizations above; the MiMo tp8/tp16 2P1D rows are the first `request-rate inf` measurements on this path (Stage 1 did not include a saturated-throughput test plan).

GLM-5.2 gsm8k is unchanged Stage 1 → Stage 2 (0.955), confirming the perf commits do not touch model forward/sampling.

MiMo-V2.5-Pro at the Stage-1 rate-limited operating point (4K/1K, rate=0.19/0.28, same hardware):

| rate | Stage 1 PD MedITL | Stage 2 1P1D | Stage 2 2P1D |
|---|---|---|---|
| 0.19 | 57.80ms | **16.19ms** (-72%) | 16.33ms |
| 0.28 | 75.07ms | **17.28ms** (-77%) | 17.36ms |

Primary contributors are the fused D-tick (3 Executes → 1) and SWA-tail-only KV transfer (Stage 1 sent full 4K seq to D swa pool). 2P1D ≈ 1P1D at these rates (P is not the bottleneck). Config note: Stage 1 was dp=1 `fused`; Stage 2 uses dp=2 `fused_v2` — not a strict apples-to-apples on the moe backend axis.

### Shared-code perf notes

The two shared-file changes that touch the compiled path:

- **`compilation_manager` cache_loc bucket cap** (`tp_worker.py:277`): capping the per-bs bucket at pool size shrinks the largest H2D array (e.g. 134MB → ~15MB at bs=128 ctx=262K), saving ~25ms/tick on Pathways gRPC H2D. On native TPU the smaller/odd bucket shape yields a slower Pallas ragged-attention kernel (v6e-1 CI `test_bench_serving_dense` input_throughput 28285 → 26557, -6.1%), so the cap is gated on `JAX_PLATFORMS=="proxy"` (e72e96ca). Applies to Pathways colocated too (same gRPC H2D).
- **`FusedEPMoEV2` explicit reshard** (`fused_moe.py:603`): under jax 0.8.1 the `shard_map` `in_specs` already declares `P((dp,tp))`, but a same-node A/B (v7x tp8 dp2 2P1D fused_v2 16K/4K C128) shows the explicit reshard still helps the compiled path — with reshard r2/r3 = 15493/15426 tok/s vs without 15313/15245 (-1.16%/-1.17%, per-run variance 0.4–0.7%). XLA appears to place the collective differently when the sharding is asserted before the shard_map boundary. The eager-mode overhead is Python-level and does not survive jit. The analogous `logits_processor` reshard was reverted (0bbf6e4e) since it showed no e2e delta. Only reachable via `moe_backend=fused_v2`.

### Reproducing the tp8 dp2 2P1D run

<details>
<summary>PathwaysJob + launch + benchmark commands (v7x, MiMo tp8 dp2 2P1D 16K/4K C128)</summary>

**1. PathwaysJob** (3×2x2x1 slices + CPU head; requires the PathwaysJob CRD and pre-created v7x nodepools):

```yaml
apiVersion: pathways-job.pathways.domain/v1
kind: PathwaysJob
metadata:
  name: pathways-pd-b5
spec:
  maxRestarts: 20
  pathwaysVersion: jax-0.8.1
  pathwaysDir: gs://<YOUR_BUCKET>/pathways-staging
  customComponents:
    - componentType: proxy_server
      customFlags:
        - --temporary_flags_for_debugging=temporary_flag_for_debugging_disable_concurrent_dispatch=false
    - componentType: pathways_server
      customFlags:
        - --temporary_flags_for_debugging=temporary_flag_for_debugging_persistent_compilation_cache=false
  workers:
    - type: tpu7x-standard-4t
      topology: 2x2x1
      numSlices: 3
  controller:
    deploymentMode: default
    mainContainerName: main
    template:
      spec:
        nodeSelector:
          cloud.google.com/gke-nodepool: <CPU_NODEPOOL>
        containers:
          - name: main
            image: <YOUR_SGLANG_JAX_IMAGE>
            ports: [{ containerPort: 30000 }]
            env: [{ name: HF_HOME, value: /tmp/hf }]
            command: ["/bin/bash","-c","sleep infinity"]
```

For models >700G, raise the `pathways-proxy` initContainer memory limit after the head pod is running (see Known issues #6):

```bash
kubectl patch pod <HEAD_POD> --subresource resize --type=strategic \
  -p '{"spec":{"initContainers":[{"name":"pathways-proxy","resources":{"limits":{"memory":"150Gi"}}}]}}'
```

**2. Launch server** (exec into the head pod's `main` container):

```bash
env SGLANG_MOE_BULK_READ=1 SGLANG_MOE_BULK_BLOCK=1 SGLANG_PD_WEIGHT_CACHE=1 \
  --model-path <MODEL_PATH> --trust-remote-code --dtype=bfloat16 \
  --tp-size 8 --ep-size 8 --dp-size 2 --dp-schedule-policy round_robin --moe-backend fused_v2 \
  --page-size 256 --swa-full-tokens-ratio 0.2 --context-length 262144 \
  --chunked-prefill-size 2048 --max-prefill-tokens 16384 --max-running-requests 256 \
  --precompile-bs-paddings 1 4 8 16 32 64 128 256 --precompile-token-paddings 4096 \
  --disable-radix-cache --attention-backend fa --skip-server-warmup --random-seed 3 \
  --pd-disaggregation pathways --pd-num-prefill 2 --pd-num-decode 1 \
  --pd-prefill-max-tokens 32768 --pd-decode-max-tokens 800000 \
  --pd-prefill-mem-fraction 0.84 --pd-decode-mem-fraction 0.84 \
  --disaggregation-max-inflight-transfers 32 \
  --host 0.0.0.0 --port 30000
```

**3. Benchmark** (after `fired up`, from the same container):

```bash
# throughput (r1/r2/r3 rows in the test-plan table)
python3 -m sgl_jax.bench_serving --backend sgl-jax --host 127.0.0.1 --port 30000 \
  --dataset-name random --num-prompts 384 --request-rate inf --max-concurrency 128 \
  --random-input 16384 --random-output 4096 --random-range-ratio 1.0 --seed 12345 \
  --extra-request-body '{"sampling_params":{"temperature":0.1,"top_p":0.95,"max_new_tokens":4096,"ignore_eos":true}}'

# gsm8k
python3 benchmark/gsm8k/bench_sglang_jax.py --base-url http://localhost:30000 \
  --num-questions 200 --num-shots 5 --max-new-tokens 1024 --temperature 0 --parallel 32
```

`<MODEL_PATH>` in the reported runs is a local checkpoint of MiMo-V2-Flash mounted via PVC; a HuggingFace path with `--download-dir` also works if the head pod has enough ephemeral storage.

</details>

## Known issues

1. `schedule_policy.py:453/458/600/667` potential `prefix_indices` race under radix-cache (PD forces `--disable-radix-cache` so not triggered on this path)
2. 1P-ND N≥2 tp16: main-thread bound (~50ms/tick vs device 25ms) — throughput does not scale linearly with N
3. dp=1 + Pathways proxy backend: `jitted_run_model` cpp fastpath miss every step (~<1ms/step overhead; dp>1 unaffected). Also present on colocated — pre-existing on main, tracked in #1452.
4. Hetero-TP requires a hand-written JobSet (PathwaysJob CRD does not support multiple `workers[]` entries)
5. `kv_cache_dtype=fp8` not supported (`_init_kv_cache_dtype` handles auto/bf16 only)
6. Large models (>700G) need `pathways-proxy` initContainer memory limit raised from the CRD default 100G (peak proxy rss ≈10% of model bytes during load; e.g. `kubectl patch pod <head> --subresource resize -p '{"spec":{"initContainers":[{"name":"pathways-proxy","resources":{"limits":{"memory":"400Gi"}}}]}}'`)

## Hetero-TP JobSet sample

```yaml
apiVersion: jobset.x-k8s.io/v1alpha2
kind: JobSet
metadata: { name: <NAME> }
spec:
  replicatedJobs:
    - name: pathways-head
      template:
        spec:
          template:
            spec:
              containers:
                - name: pathways-rm
                  image: <PATHWAYS_SERVER_IMAGE>
                  args:
                    - --server_port=29001
                    - --expected_instances=<TPU_TYPE>:<P_TOPOLOGY>,<TPU_TYPE>:<D_TOPOLOGY>
                    - --instance_count=2
                # ... proxy + main containers (see PathwaysJob-generated head pod for full spec)
    - name: worker-0  # P slice, larger topology
      template:
        spec:
          parallelism: <P_HOSTS>
          template:
            spec:
              nodeSelector: { cloud.google.com/gke-tpu-topology: <P_TOPOLOGY> }
              containers:
                - name: worker
                  env:
                    - { name: MEGASCALE_SLICE_ID, value: "0" }
    - name: worker-1  # D slice, smaller topology
      # ... same, MEGASCALE_SLICE_ID=1, <D_TOPOLOGY>
```

Server args: `--pd-disaggregation pathways --tp-size <D_TP> --pd-prefill-tp-size <P_TP> --pd-prefill-ep-size <P_EP>`

### DIFF
--- a/python/sgl_jax/srt/disaggregation/pathways_pd.py
+++ b/python/sgl_jax/srt/disaggregation/pathways_pd.py
@@ -3,23 +3,25 @@
 Architecture: one Python process (Pathways head) sees all devices across
 N slices via IFRT proxy. Build prefill_mesh on slice 0, decode_mesh on
 slice 1, run two ModelRunners. KV transfer = jax.device_put across meshes
-(IFRT runtime handles per-host DCN fan-out, ~14 GB/s/host on v7x vs ~4.5
+(IFRT runtime handles per-host DCN fan-out, ~14 GB/s/host measured vs ~4.5
 for jax.experimental.transfer on the same hardware).
 """
 
 from __future__ import annotations
 
 import logging
+import os
+import time
 from collections import defaultdict
 
 import jax
-import jax.numpy as jnp
 import numpy as np
 from jax.sharding import Mesh, NamedSharding
 from jax.sharding import PartitionSpec as P
 
 __all__ = [
     "group_by_slice",
+    "group_pages_by_dp_rank",
     "make_slice_meshes",
     "PathwaysPDKVTransfer",
     "migrate_reqs_p_to_d",
@@ -27,12 +29,13 @@ __all__ = [
 ]
 
 logger = logging.getLogger(__name__)
+_PD_DBG = bool(os.environ.get("SGLANG_PD_DBG"))
 
-_NPG_BUCKETS = (8, 16, 32, 64)
-# Upper bound on pages per stacked transfer. 64pg x 78L x 80KB ~= 0.39 GB
-# transient HBM (x2 with the recv-side donate buffer) — fits the smallest
-# headroom we measured (v6e fp8: ~1.3 GB free after weights + KV + jit exec).
-_NPG_BATCH_MAX = 64
+_NPG_BUCKETS = (8, 16, 32, 64, 128, 256)
+# Upper bound on pages per stacked transfer. 256pg x 78L x 80KB ~= 1.6 GB
+# transient on each side; mf=0.84 on 96GB devices leaves >8 GB headroom.
+# 16K/ps256 = 64pg fits one shot, 16K/ps128 = 128pg also one shot.
+_NPG_BATCH_MAX = 256
 
 
 def _bucket_npg(n: int) -> int:
@@ -52,6 +55,51 @@ def slots_to_ordered_pages(slots: np.ndarray, page_size: int) -> np.ndarray:
     pages = np.asarray(slots, np.int64) // page_size
     _, first_idx = np.unique(pages, return_index=True)
     return pages[np.sort(first_idx)].astype(np.int32)
+
+
+def group_pages_by_dp_rank(
+    pages_by_req: list[tuple[int, np.ndarray]], dp_size: int
+) -> tuple[np.ndarray, np.ndarray]:
+    """Group per-req (dp_rank, local_pages) pairs into the [dp_size, max_n]
+    layout the dp>1 shard_map kernels (_make_pool_jits' gather_jit /
+    scatter_one_jit) require -- each row must hold ONLY that rank's own
+    LOCAL page indices (allocator.py: indices aren't comparable across
+    ranks), never a flat cross-rank concat.
+
+    Returns (pages, valid) both shaped [dp_size, max_n]: `pages[i, j]` for
+    j >= real-count is padded by repeating rank i's own last real page
+    (harmless self-overwrite/re-gather, same scheme _bucket_npg already
+    uses within a single rank). `valid[i, j]` is False for those padding
+    slots -- callers MUST check it before trusting/using the corresponding
+    output.
+
+    KNOWN LIMITATION (not yet handled by any caller): a rank with ZERO real
+    pages this round has no "own last page" to repeat, so its row is left
+    as all-zero-index/all-invalid. Naively feeding that row into
+    scatter_one_jit would overwrite page 0 of THAT rank's D pool with
+    whatever stale P-side data rides along -- a real corruption, not a
+    harmless no-op (unlike the same-rank padding case). Callers must NOT
+    invoke the scatter kernel across a batch where any rank's row is
+    all-invalid; either defer that rank's migration to a round where it
+    has >=1 real req, or split into per-rank-populated sub-batches.
+    """
+    per_rank: list[list[int]] = [[] for _ in range(dp_size)]
+    for dp_rank, pages in pages_by_req:
+        per_rank[dp_rank].extend(int(p) for p in np.asarray(pages, np.int64))
+    max_n = max((len(p) for p in per_rank), default=0)
+    if max_n == 0:
+        return np.zeros((dp_size, 0), np.int32), np.zeros((dp_size, 0), bool)
+    pages_out = np.zeros((dp_size, max_n), np.int32)
+    valid_out = np.zeros((dp_size, max_n), bool)
+    for i, p in enumerate(per_rank):
+        if not p:
+            continue  # no real pages for this rank -- see KNOWN LIMITATION above
+        arr = np.asarray(p, np.int32)
+        valid_out[i, : len(arr)] = True
+        if len(arr) < max_n:
+            arr = np.concatenate([arr, np.full(max_n - len(arr), arr[-1], np.int32)])
+        pages_out[i] = arr
+    return pages_out, valid_out
 
 
 def migrate_reqs_p_to_d(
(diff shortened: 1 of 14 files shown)


python/sgl_jax/srt/disaggregation/pathways_scheduler.py
```diff
@@ -46,37 +48,64 @@ class PathwaysPDSchedulerMixin:
...
         p_args.disable_radix_cache = True
+        if self._pd_tp_p != self.tp_size:
+            # ModelWorker/ModelRunner read tp/ep from server_args (tp_worker.py
+            # :75/:130, moe.py:79 world_size from mesh); mesh alone isn't
+            # enough -- attention_tp = tp_size//dp_size drives kv_heads
+            # replication + validate_tensor_parallel_config. ep must divide
+            # both p_mesh world_size and num_experts; default = keep D's
+            # ep/tp ratio (pure-EP stays pure-EP on the bigger P slice).
+            p_args.tp_size = self._pd_tp_p
+            p_args.ep_size = server_args.pd_prefill_ep_size or (
```
cjx0709: I think there may be a corner case here. With `D TP=16`, `P TP=8`, and the default `EP=1`, this derives `p_args.ep_size=0`, which would then reach the MoE initialization path with an invalid EP size. Would it make sense to validate the derived P-side TP/EP combination up front, or define a safer default for this case?

python/sgl_jax/srt/layers/fused_moe.py
```diff
@@ -600,6 +600,10 @@ def __call__(
             )
 
         direct_scaled_dot = w1_scale is not None
+        kernel_sharding = jax.sharding.NamedSharding(self.mesh, P(("data", "tensor"), None))
+        hidden_states = jax.sharding.reshard(hidden_states, kernel_sharding)
+        topk_weights = jax.sharding.reshard(topk_weights, kernel_sharding)
+        topk_ids = jax.sharding.reshard(topk_ids, kernel_sharding)
```
cjx0709, line 606: Could you share a bit more context on why these input reshard calls are needed? I was wondering whether this is related to the explicit-sharding behavior in newer JAX versions (e.g. JAX >= 0.9), since the repo currently pins JAX 0.8.1. This also affects every `FusedEPMoEV2` caller, not just Pathways PD. I saw matching outputs in a small v7x check, although the eager-layer timing was higher; a compiled or end-to-end measurement would help clarify the overall impact.

python/sgl_jax/srt/managers/tp_worker.py
```diff
@@ -71,6 +73,7 @@ def __init__(
         # Parse args
         self.tp_size = server_args.tp_size
         self.dp_size = server_args.dp_size
+        self._pd_fuse_sample = os.getenv("SGLANG_PD_FUSE_SAMPLE") == "1"
```
cjx0709: Just checking the intended scope here: should `SGLANG_PD_FUSE_SAMPLE` also be usable for normal non-PD workers? Since this is read by the shared `ModelWorker`, an environment setting can change the normal serving path as well. If this is only intended for Pathways PD, would it be clearer to gate it on the Pathways PD configuration or set it from the Pathways setup?

python/sgl_jax/srt/managers/tp_worker.py
```diff
@@ -255,6 +268,14 @@ def __init__(
             page_size=self.page_size,
             max_req_len=self.max_req_len,
             vocab_size=self.model_config.vocab_size,
+            # cache_loc bucket cap only under Pathways proxy: it saves ~25ms/tick
+            # gRPC H2D there, but on native TPU the smaller/odd bucket shape
+            # (page_indices ~554 vs 81920) yields a slower Pallas ragged-attention
+            # kernel (-6% input_throughput on v6e-1 CI). Native H2D is fast enough
+            # that the uncapped bucket is not on the critical path.
+            max_total_num_tokens=(
+                self.max_total_num_tokens if os.getenv("JAX_PLATFORMS") == "proxy" else 0
```
cjx0709, line 278: Small question about the scope: the comment describes this as Pathways-proxy-only, while the condition currently checks `JAX_PLATFORMS == "proxy"` without checking the PD mode. Does proxy imply Pathways PD in all supported setups? If not, would it be clearer to scope this to Pathways PD explicitly? The detailed benchmark numbers might also fit better in the PR or design doc, while keeping the correctness rationale here.

cjx0709:
One small cleanup suggestion: there are quite a few new comments that walk through the implementation or include benchmark/debug history. It might be nice to keep the concurrency and sharding invariants in the code, and move the more detailed measurements or implementation narrative to the PR or design doc. That would make the shared paths a little easier to read and maintain.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
zorrofox: Good catch — with P tp < D tp and D ep=1 the integer division yields 0. Fixed in 1e2a6dbf: floor at `max(1, ...)` and assert the derived ep_size divides p_tp so a bad combination fails loudly at init instead of inside MoE. The current KV-reshard path is only exercised with P tp >= D tp; P < D would also need the reverse gather/scatter direction, so for now the validate just makes the failure explicit.

python/sgl_jax/srt/layers/fused_moe.py
zorrofox: Ran a same-node A/B on the compiled path (v7x tp8 dp2 2P1D fused_v2 16K/4K C128, identical hosts/hour):

```
with reshard    r2/r3 = 15493/15426 tok/s  MedITL 32.3-32.5ms
without reshard r2/r3 = 15313/15245 tok/s  MedITL 33.0-33.1ms
delta -1.16%/-1.17% (per-run variance 0.4-0.7%)
```

So under jit the explicit reshard does help — XLA appears to place the collective differently when the sharding is asserted before the shard_map boundary vs. left to in_specs, even though the specs match. The eager-layer overhead you saw is Python-level and does not survive compilation. Keeping this one; I did drop the analogous logits_processor reshard in 0bbf6e4e since that one showed no e2e delta (hidden_states there presumably already carries the target spec). Added the numbers to the PR body.

python/sgl_jax/srt/managers/tp_worker.py
zorrofox: Agreed, the env alone leaks into non-PD workers. Gated on `server_args.pd_disaggregation == "pathways"` in 1e2a6dbf so a stray env var cannot flip the fused path outside Pathways PD.

python/sgl_jax/srt/managers/tp_worker.py
zorrofox: proxy != PD-only — Pathways colocated also runs under `JAX_PLATFORMS=proxy` and does H2D over the same gRPC path, so the cap helps there too. Kept the condition and reworded the comment to say "Pathways proxy backend (PD or colocated)" and dropped the benchmark numbers (moved to PR body).

For the cache_loc `.copy()` gate added in 0bbf6e4e I had used the same proxy check, but that race is PD-eager-stash-specific (colocated has a single scheduler thread), so switched it to `global_server_args_dict["pd_disaggregation"] == "pathways"` in 1e2a6dbf.

zorrofox:
Agreed — trimmed the shared-path comments in f326646e to keep only the concurrency/sharding invariants (page-contiguous allocator guarantee, PD eager-stash race, cpp-fastpath sharding stability, donate-input window) and dropped the profiled numbers / implementation-choice narrative. Shared-file added-comment lines vs main went from 94 to 43. Left the disaggregation/ module comments as-is since that's the PD-specific design surface.

python/sgl_jax/srt/disaggregation/pathways_pd.py
```diff
@@ -127,68 +175,210 @@ def group_by_slice(devs: list) -> dict[int, list]:
...
+    scatter_one_jit below do the same: `idx`/`page` carry an explicit
+    dp_size-sized leading axis sharded on 'data', so each rank's shard_map
+    instance only ever sees and applies its OWN local indices.
+
+    dp_size==1 fast path: shard_map's manual-sharding lowering is a
+    functional no-op here but was NOT compile/dispatch-free on real TPU --
+    a live 78-layer model showed gather (all layers in one XLA program,
+    unlike scatter's per-layer calls) climbing 18s->37s and NOT plateauing
+    across repeat calls, vs <1s before shard_map existed.
+    Since dp_size==1 is 100% of production traffic today (dp>1 scheduler-
```
cjx0709: Small doc cleanup: this says `dp_size==1` is 100% of production traffic and that the `dp>1` scheduler wiring is not done yet, but this PR now includes the dp>1 scheduling path. Could we update or remove this sentence so the comment does not become misleading?

cjx0709:
One small follow-up on reproducibility: the test matrix is very helpful, but reproducing the end-to-end Pathways runs still requires reconstructing the JobSet/Falcon configuration, server launch command, and benchmark invocation. Since this is still experimental, would it be possible to add the exact config and commands for one representative setup (for example, the MiMo v7x tp8 dp2 2P1D run) directly to the PR description? That would make the throughput and correctness numbers much easier to reproduce without adding permanent docs yet.

python/sgl_jax/srt/disaggregation/pathways_pd.py
zorrofox: Removed in 8f1c9d5a — the dp>1 scheduling path is in this PR (via `_pd_group()` + the shard_map branch below), so kept only the perf rationale for the dp==1 fast path.

zorrofox:
Added a "Reproducing the tp8 dp2 2P1D run" section to the PR body with the full PathwaysJob yaml, launch_server flags, and bench_serving/gsm8k invocations for the MiMo v7x tp8 dp2 2P1D config (placeholders for image/bucket/model-path/nodepool).

cjx0709 (approved):
LGTM

PR #1455, opened 2026-07-13
chore(skills): add profiling-capture skill

Yanko-7:
## Motivation

Add a **capture-only** XProf profiling skill for sgl-jax (JAX/TPU) under `.claude/skills/profiling-capture/`. It documents how to reliably *produce* trace artifacts (`*.xplane.pb` + `trace.json.gz`) for later analysis, avoiding common silent failures (cold runs, prefix-cache hits faking prefill, truncated/empty windows, LLO dump flags set after `import jax`). Analysis is out of scope.

## Modifications

Docs only — no runtime code changed. Adds three files:

- `SKILL.md` — path A/B/C selection (live server / offline batch / kernel microbench),
  capture-quality musts, handoff contract.
- `references/capture-recipes.md` — step definition, load shaping.
- `references/pallas-kernel-profiling.md` — Pallas/Mosaic LLO/HLO/XProf recipe + pitfalls.

## Accuracy Tests

N/A — documentation only, no model output changes.

## Benchmarking and Profiling

N/A — documentation only, no inference-speed impact.

## Checklist

### DIFF
--- a/.claude/skills/profiling-capture/SKILL.md
+++ b/.claude/skills/profiling-capture/SKILL.md
@@ -1,0 +1,146 @@
+---
+name: profiling-capture
+description: "Use when you need to CAPTURE / PRODUCE XProf (jax.profiler) profiling artifacts from sgl-jax (SGLang on JAX/TPU) for later performance analysis: driving a live server (/start_profile, stage-separated prefill/decode), offline bench_one_batch --profile, or kernel microbench. This skill produces the trace files; it does not analyze them."
+---
+
+# XProf Profiling Capture
+
+Produce XProf profiling files from sgl-jax for later analysis. sgl-jax is **JAX/TPU**, so the
+profiler is `jax.profiler` and it emits **XProf/TensorBoard** artifacts (`*.xplane.pb` +
+`trace.json.gz` under `plugins/profile/`), not `torch.profiler`. Scope is **capture only** —
+producing the trace, not reading it. Analysis belongs to a separate skill.
+
+## Workflow
+
+1. Pick a capture path (A/B/C) from the table below by your situation.
+2. Warm up before tracing so JIT/autotune is excluded — **never trace a cold run**. Path A
+   uses `--warmup-requests`; Path B auto-warms.
+3. Default to **one non-stage trace**; prefill/decode are separated at analysis time (out of
+   scope here). Add `--profile-by-stage` only when you need pre-split dirs.
+4. For prefill, force cache-missing prompts (`--dataset-name random` / `--disable-radix-cache`)
+   plus `--flush-cache` — unless the change under test is cache-related.
+5. Run the capture. The trace dir is **server-side** — confirm it is writable on the host.
+6. Wait for flush: confirm `*.xplane.pb` exists and `profile_status == idle`. Whether the
+   needed device compute actually landed is confirmed at analysis time, not here.
+7. Prefer a **single-rank** trace when TP emits one per rank.
+8. Hand off (see Handoff) — do **not** analyze; analysis is a separate skill.
+
+## Pick a capture path
+
+| Situation | Path |
+| --- | --- |
+| running HTTP server | **A** — live capture (non-stage default) |
+| one batch shape, no server | **B** — `bench_one_batch --profile` |
+| one kernel | **C** — kernel microbench |
+
+## A — live server
+
+`bench_serving --profile` is the recommended default: it warms up, arms the profiler, drives
+a real dataset workload, then stops — **one trace over the whole run**. A single window
+reliably captures device ops for **both** stages; separating prefill from decode is an
+analysis-time step (out of scope here).
+
+Default command:
+
+```bash
+python -m sgl_jax.bench_serving \
+  --backend sgl-jax --base-url http://127.0.0.1:30000 \
+  --dataset-name random --num-prompts 8 \
+  --random-input-len 4096 --random-output-len 128 --warmup-requests 10 \
+  --flush-cache --profile --profile-num-steps 5
+```
+
+Key flags:
+
+| Flag | Effect |
+| --- | --- |
+| `--profile` | arm the profiler for this run |
+| `--profile-num-steps` | number of steps to trace after warmup |
+| `--warmup-requests` | untraced requests first, so JIT/autotune is excluded |
+| `--flush-cache` | clear the radix cache once before the run (see Capture quality) |
+| `--profile-by-stage` | write pre-split `prefill/` + `decode/` dirs (prefill unreliable) |
+| `--profile-stages prefill decode` | with `--profile-by-stage`, which stages to profile |
+
+**Need pre-split `prefill/` + `decode/` dirs** to compare the stages directly? Add
+`--profile-by-stage`, optionally with `--profile-stages prefill decode` — pass one stage to
+profile only that stage, omit for both.
+
+**Manual path:** `POST /start_profile` (body = `ProfileReqInput`, may include `output_dir`) →
+send requests → `POST /stop_profile` → `GET /profile_status` until `message == "idle"` (else
+`"in_progress"`).
+
+**Traces are written server-side**, into `SGLANG_JAX_PROFILER_DIR` (default `/tmp`), read from
+the server process's environment. To steer `bench_serving` output, export it **at server
+launch** — a running process cannot be changed. The manual `/start_profile` path can pass
+`output_dir` per request instead. Either way the dir must be writable on the host.
+
+## B — offline batch
+
+`bench_one_batch` takes the **same server args as `launch_server`**, so fill the model and
+parallelism from your own deployment context, then add the batch shape and `--profile`:
+
+```bash
+python -m sgl_jax.bench_one_batch \
+  --model-path <model> --tp-size <n> --device tpu ... \  # your launch_server config
+  --batch-size 1 --input-len 4096 --output-len 128 --profile   # -> *.tb/
+```
+
+- **Match your server config.** Pass the same parallelism/model flags you launch with
+  (`--tp-size`, `--dp-size`, `--trust-remote-code`, `--dtype`, …); mismatched config profiles a
+  different program than you run in production.
+- **Auto-warms up** first (a full untraced run), so JIT/autotune is already excluded — do not
+  add warmup here.
+- The trace covers prefill **plus all decode steps mixed in one `*.tb/` dir**, with no stage
+  separation.
+- Only **tp_rank 0** writes the trace.
+- Dir name pattern: `<prefix>_batch<bs>_input<il>_output<ol>.tb` (`--profile-filename-prefix`).
+
+## C — kernel microbench
+
+`jax.profiler.trace(dir)` around a warmed-up `run_kernel()`. Produces Pallas/Mosaic LLO
+detail. For LLO / custom-call / perf-counter detail, set the dump flags **before `import jax`**
+and pin libtpu ≥ 0.0.39. Full recipe — region vs counter mode, flags, verification, delivery,
+existing harnesses — in [references/pallas-kernel-profiling.md](references/pallas-kernel-profiling.md).
+
+## Capture quality
+
+Details in [references/capture-recipes.md](references/capture-recipes.md). Musts:
+
+- **Warm up** (`--warmup-requests 5`) so JIT/autotune is excluded (a *step* = one forward over
+  the batch; decode step = one token, prefill step ≈ one prompt batch).
+- **Capture both stages in one non-stage trace**; splitting prefill from decode is done at
+  analysis time. Reach for `--profile-by-stage` only when you need pre-split dirs, knowing its
+  prefill trace is unreliable.
+- **Avoid the prefix-cache shortcut for prefill.** `--dataset-name random` gives unique
+  (cache-missing) prompts, or run with `--disable-radix-cache`. Add `--flush-cache` too —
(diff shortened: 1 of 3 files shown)


.claude/skills/profiling-capture/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+`output_dir` per request instead. Either way the dir must be writable on the host.
+
+## B — offline batch
+
+`bench_one_batch` takes the **same server args as `launch_server`**, so fill the model and
+parallelism from your own deployment context, then add the batch shape and `--profile`:
+
+```bash
+python -m sgl_jax.bench_one_batch \
+  --model-path <model> --tp-size <n> --device tpu ... \  # your launch_server config
```
cjx0709: This copy-paste command does not continue onto the next line. In POSIX shells, the backslash escapes the following space and `#` starts a comment, so line 85 is parsed as a separate command and fails with `--batch-size: command not found`. Please move the comment outside the command block, or make `\` the final character on the line.

.claude/skills/profiling-capture/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+| `--flush-cache` | clear the radix cache once before the run (see Capture quality) |
+| `--profile-by-stage` | write pre-split `prefill/` + `decode/` dirs (prefill unreliable) |
+| `--profile-stages prefill decode` | with `--profile-by-stage`, which stages to profile |
+
+**Need pre-split `prefill/` + `decode/` dirs** to compare the stages directly? Add
+`--profile-by-stage`, optionally with `--profile-stages prefill decode` — pass one stage to
+profile only that stage, omit for both.
+
+**Manual path:** `POST /start_profile` (body = `ProfileReqInput`, may include `output_dir`) →
+send requests → `POST /stop_profile` → `GET /profile_status` until `message == "idle"` (else
```
cjx0709: The manual flow polls the wrong response field. `/profile_status` returns `{"status":"idle"}` or `{"status":"in_progress"}`, not a `message` field, so a client following this instruction cannot observe completion. Please change this to `status == "idle"` (ideally with a copy-pasteable `curl | jq -r .status` example).

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
```diff
@@ -0,0 +1,104 @@
...
+    for _ in range(100):                      # a ns-scale kernel needs many iters or the window is ~empty
+        r = run_kernel()
+    jax.block_until_ready(r)                  # block_until_ready MUST be inside the block
+```
+
+Loop count: one call of a fast (single-digit-ns) kernel leaves too little in the window — a
+valid-but-empty trap. Loop until the region is milliseconds of device work (hundreds–thousands
+of iters for a tiny kernel); block **once at the end, still inside the block**.
+
+Name the phases with `jax.named_scope("w1_load" / "gemm" / "store_output")` so
```
cjx0709: This is not valid Python: it evaluates `"w1_load" / "gemm" / "store_output"` before calling `jax.named_scope`, which raises `TypeError`. Please show the actual context-manager form (for example, `with jax.named_scope("gemm"):` around the relevant operations) rather than slash-separated illustrative names inside a call.

.claude/skills/profiling-capture/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+$SGLANG_JAX_PROFILER_DIR/            # server-side; default /tmp
+    plugins/profile/<run>/<host>.xplane.pb           # default: single tree (+ .trace.json.gz)
+# with --profile-by-stage: split into prefill/ and decode/ subtrees instead
+    prefill/plugins/profile/<run>/<host>.xplane.pb
+    decode/plugins/profile/<run>/<host>.xplane.pb
+```
+
+## Handoff
+
+Do not analyze — hand the trace to the analysis skill. The dir containing `plugins/profile` is
```
cjx0709: The handoff is not currently actionable: this repository does not contain the separate analysis skill referenced here. Please name and link the expected skill if it exists elsewhere, or explicitly state that analysis is not bundled and provide the manual XProf/TensorBoard next step.

.claude/skills/profiling-capture/SKILL.md
Yanko-7: Fixed.

.claude/skills/profiling-capture/SKILL.md
Yanko-7: changed to status and added curl example.

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
Yanko-7: Fixed — now uses the real context-manager form: one with jax.named_scope("..."): block per phase inside run_kernel(). Also added @jax.jit.

.claude/skills/profiling-capture/SKILL.md
Yanko-7: Fixed — Handoff now states no analysis skill is bundled, and adds the manual step: TensorBoard profile plugin (or XProf) on the dir containing plugins/profile.

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
```diff
@@ -0,0 +1,113 @@
...
+    with jax.named_scope("gemm"):
+        acc = matmul(x, w1)
+    with jax.named_scope("store_output"):
+        store(acc)
+    return acc
+
+opts = jax.profiler.ProfileOptions()          # add counter config only for counters mode
+for _ in range(3):                            # warm up so compile/autotune is outside the trace
+    jax.block_until_ready(run_kernel())
+with jax.profiler.trace(f"{OUT}/xprof"):      # trace lands beside the LLO/HLO dumps under $OUT
```
cjx0709: Looks like `OUT` got dropped when this example was rewritten. `OUT=/tmp/pallas-profile` in the shell block above is only a shell variable, while this snippet needs a Python variable here, so this hits `NameError` even after the kernel placeholders are filled in. Could we define `OUT` in the Python block or read it from `os.environ`?

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
```diff
@@ -0,0 +1,113 @@
...
+    # wrap each phase in its own named_scope context manager
+    with jax.named_scope("w1_load"):
+        w1 = load_weights()
+    with jax.named_scope("gemm"):
+        acc = matmul(x, w1)
+    with jax.named_scope("store_output"):
+        store(acc)
+    return acc
+
+opts = jax.profiler.ProfileOptions()          # add counter config only for counters mode
```
cjx0709: Small follow-up: `opts` is currently unused. If someone adds the suggested counter configuration there, it never reaches the profiler. Could we either remove this line for the plain-trace example, or pass it through as `jax.profiler.trace(..., profiler_options=opts)`? `bench_v2.py` already uses that pattern.

Yanko-7: replaced with jax.profiler.trace(..., profiler_options=opts).

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
Yanko-7: we define OUT in Python code.

.claude/skills/profiling-capture/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+| running HTTP server | **A** — live capture (non-stage default) |
+| one batch shape, no server | **B** — `bench_one_batch --profile` |
+| one kernel | **C** — kernel microbench |
+
+## A — live server
+
+`bench_serving --profile` is the recommended default: it warms up, arms the profiler, drives
+a real dataset workload, then stops — **one trace over the whole run**. A single window
+reliably captures device ops for **both** stages; separating prefill from decode is an
+analysis-time step (out of scope here).
```
JamesBrianD, line 41: For the live server, need to configure sglang-jax server so that it does not disable precompilation.

.claude/skills/profiling-capture/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+  `.xplane.pb` file itself):
+
+  ```bash
+  pip install tensorboard tensorboard-plugin-profile
+  tensorboard --logdir /tmp/myprofile   # open http://localhost:6006 → PROFILE tab
+  ```
+
+- Or load the `*.xplane.pb` directly in the standalone [XProf](https://github.com/openxla/xprof) viewer.
+
+Either way, the dir containing `plugins/profile` is what the tool consumes. Always report:
```
JamesBrianD: https://openxla.org/xprof/jax_profiling
It is now officially recommended to use xprof by passing the full logdir to it; please prioritize this approach.

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
```diff
@@ -0,0 +1,113 @@
...
+--xla_enable_custom_call_region_trace=true \
+--xla_xprof_register_llo_debug_info=true \
+--xla_mosaic_dump_to=$OUT/compiler/llo"
+export XLA_FLAGS="${XLA_FLAGS:+$XLA_FLAGS }\
+--xla_dump_to=$OUT/compiler/hlo --xla_dump_hlo_as_text --xla_dump_hlo_as_proto"
+```
+
+- `--xla_mosaic_dump_to` → Pallas/Mosaic LLO
+- `--xla_enable_custom_call_region_trace=true` → custom-call regions enter the trace
+- `--xla_xprof_register_llo_debug_info=true` → XProf can align custom-call ↔ LLO
```
JamesBrianD: https://openxla.org/xprof/custom_call_profiling
Note that these two options should be enabled only as needed, as they increase profile size and collection overhead.

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
```diff
@@ -0,0 +1,115 @@
...
+## Three artifact types
+
+| Artifact | When | Path |
+| --- | --- | --- |
+| Mosaic **LLO** dump | compile time | `compiler/llo/` |
+| HLO / MLIR dump | compile time | `compiler/hlo/` |
+| **XProf** trace / xplane | run time | `plugins/profile/<ts>/` |
+
+Having an LLO dump does **not** mean you captured a runtime trace — a failed
+compile can still leave LLO/HLO but no valid `.xplane.pb`.
```
JamesBrianD: The four capture modes provide different kinds of evidence and have different costs. They should not all be enabled by default. They can be combined, but each additional mode should be explicitly selected based on the profiling question.</p><div><div>
Mode | What it provides | How to enable it | Best suited for
-- | -- | -- | --
Plain trace | JAX/TPU runtime timeline, kernel duration, and named scopes | jax.profiler.trace(...) | “Which kernel is slow, and how long does it take?”
LLO utilization | LLO bundle utilization inside a Pallas custom call | Add the two custom-call profiling flags to LIBTPU_INIT_ARGS | “How are MXU, DMA, and wait time distributed inside the kernel?”
Compiler dump | Compile-time HLO, MLIR, and Mosaic LLO files | --xla_mosaic_dump_to and --xla_dump_to | “What instructions and layouts did the compiler generate?”
Periodic counters | Fine-grained, time-series hardware counters on Ironwood/TPU v7x | ProfileOptions.advanced_configuration | “Why is the MXU idle? Are spills or other hardware bottlenecks occurring?”

</div></div><p>I suggest structuring the recipe as four independent, opt-in layers:</p><ol start="1"><li>Use a <strong>plain trace</strong> as the low-overhead default.</li><li>Add <strong>LLO utilization</strong> when custom-call bundle visibility is required.</li><li>Enable <strong>compiler dumps</strong> only when investigating lowering, layouts, or generated instructions.</li><li>Enable <strong>periodic counters</strong> only for focused Ironwood/v7x bottleneck analysis.</li></ol><p>These modes are composable, but enabling all of them unconditionally can substantially increase compilation time, profile size, and runtime overhead.</p></body></html>

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
JamesBrianD: “--xla_enable_custom_call_region_trace=true” This flag incurs a performance penalty; this needs to be noted in the `skills` section as well.

.claude/skills/profiling-capture/SKILL.md
Yanko-7: It has been clarified that the live server retain the default startup precompilation.

.claude/skills/profiling-capture/SKILL.md
Yanko-7: The XProf has been set as the preferred option.

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
Yanko-7: Changed the two custom-call profiling flags to be enabled on demand, with notes on their profile size and collection overhead.

.claude/skills/profiling-capture/references/pallas-kernel-profiling.md
Yanko-7: Refactor Pallas profiling into a four-layer structure with plain trace as the default and the other three modes independently opt-in.

JamesBrianD (approved):

PR #1461, opened 2026-07-15
feat(jax): upgrade jax[tpu] to 0.10.2

Aharrypotter:
## Motivation

Upgrade SGLang-JAX directly from JAX/JAXlib 0.8.1 to 0.10.2 and libtpu 0.0.42.1.

JAX 0.9.1+ enforces `shard_map` input contracts that older versions could satisfy with an
implicit reshard. The stricter checks exposed a real EAGLE3 producer bug and several unit-test
fixtures that created replicated arrays even though their production producers return sharded
arrays. JAX 0.10.2 also removes or deprecates several APIs used by the current TPU kernels.

This PR keeps production `shard_map` specs explicit. It does not add a generic consumer-side
reshard wrapper: production layout errors are fixed at their producer, and direct kernel tests
place their fixtures on the layouts produced in serving.

Supersedes #1407 and relates to #1025.

### Dependency and API migration

- Upgrade `jax[tpu]`, `jax[cuda12]`, and `jax[cpu]` from 0.8.1 to 0.10.2.
- Upgrade libtpu from 0.0.30 to 0.0.42.1.
- Replace removed or deprecated JAX/Pallas APIs:
  - `with mesh:` with `jax.set_mesh(mesh)`;
  - `jax.experimental.shard_map` with `jax.shard_map`;
  - `pltpu.MemorySpace.ANY` with `pl.ANY`;
  - TPU semaphore and device-id APIs with their current `jax.experimental.pallas` forms;
  - deprecated `jnp.clip(a_min=/a_max=)` argument names.

### Sharding contract fixes

- Fix the EAGLE3 `v_proj` producer layout by changing `kernel_axes` from
  `("tensor", None)` to `(None, "tensor")`, matching `q_proj`, `k_proj`, its weight mapping,
  and the attention K/V contract.
- Place dynamically expanded quantized weight scales on their kernel layout when the
  `QuantizedLinear` parameter is constructed. This mirrors the existing abstract/static path
  instead of repairing the scale at the matmul consumer.
- Update direct backend fixtures for FlashAttention, MLA, Lightning, KV cache, quantized linear,
  and fused MoE to use the layouts their production producers provide.
- Keep reference inputs replicated where the reference implementation is intentionally serial;
  only the arrays passed to the backend are placed on backend layouts.
- Do not change the production Lightning, KDA, MLA, or FlashAttention `shard_map` contracts and
  do not introduce a blanket `reshard` helper.

### JAX 0.10.x runtime fixes

- Bound EAGLE tree back-traversal by `draft_token_num` so device execution cannot stall in an
  unbounded data-dependent `while_loop`.
- Cap the v5/v6 custom-mask query tile at 16 rows to stay within scoped VMEM; add a focused
  regression test for the block-size decision.
- Migrate the remaining MiMo Audio `jnp.clip(a_max=...)` call and add direct target-version
  coverage for the ISTFT path.
- Materialize EAGLE kernel outputs in tests so asynchronous device failures cannot be hidden by
  shape-only assertions.
- Run the FlashAttention files through pytest because it tears down their shared mesh cleanly on
  JAX 0.10.x.
- Make the chunked-prefill accuracy test deterministic and rebaseline the bf16 logprob output on
  JAX 0.10.2 with a one-ULP-rejecting tolerance while preserving exact token id and token text checks.

### Direct-backend test topology

Four FlashAttention sink tests intentionally use logical Q/KV head counts that are not divisible
by TP=4. Production rejects non-divisible Q heads and replicates KV heads before constructing the
backend; these lower-level tests bypass that preparation. They now run on a TP=1 submesh that is
compatible with their logical shapes, while the regular GQA/MHA cases continue to exercise the
full four-device mesh.

## Accuracy Tests

MMLU on `Qwen/Qwen3-8B` remained identical to the JAX 0.8.1 baseline:

| Stack | Overall | STEM | Humanities | Social sciences | Other |
| --- | ---: | ---: | ---: | ---: | ---: |
| JAX 0.8.1 baseline | 0.725 | 0.786 | 0.629 | 0.714 | 0.809 |
| JAX 0.10.2 | 0.725 | 0.786 | 0.629 | 0.714 | 0.809 |

## Benchmarking and Profiling

The JAX upgrade invalidated the old v6e RPA mixed-prefill tuning point for the Qwen3-8B serving
shape. Retuning the exact 1024/2048-token cells on the target JAX 0.10.2 runtime selected
`(32, 256, 32, 256)`. The final isolated v6e single-chip gate reached 30,454.80 input tok/s and
45.74 ms median TTFT without changing the performance thresholds.

## Test Plan

Validated with JAX/JAXlib 0.10.2 and libtpu 0.0.42.1 on TPU v6e-4. The four non-TP-divisible
direct-backend cases and final serving performance gate were run with the process physically
restricted to one chip.

| Area | Result |
| --- | --- |
| Lightning backend | `22 passed` |
| KDA | `5 passed` |
| FlashAttention MHA | `10 passed` |
| FlashAttention GQA + MISC, full four-chip run | `16 passed` |
| Non-TP-divisible sink cases, `TPU_VISIBLE_CHIPS=0` | `4 passed` |
| GQA custom mask | `1 passed` |
| MHA custom mask | `1 passed` |
| New v6 custom-mask block-size regression | `1 passed` |
| MiMo ISTFT target-version regression | `1 passed` |
| Logprob end-to-end test | `1 passed` |
| Tuned-table CPU tests | `6 passed` |
| v6e single-chip serving performance gate | input `30454.80`, output `2813.62` tok/s; TTFT `45.74` ms; ITL `13.73` ms |
| Pre-commit on all PR files | passed |

Representative commands:

```bash
python -m pytest -q python/sgl_jax/test/layers/test_lightning_backend.py
python -m pytest -q python/sgl_jax/test/kernels/kda_test.py
python -m pytest -q \
  python/sgl_jax/test/test_flashattention_gqa.py \
  python/sgl_jax/test/test_flashattention_misc.py

TPU_VISIBLE_CHIPS=0 python -m pytest -q \
  python/sgl_jax/test/test_flashattention_gqa.py::TestFlashAttentionGQA::test_gqa_4q1kv_attention_sink_decode_accuracy \
  python/sgl_jax/test/test_flashattention_gqa.py::TestFlashAttentionGQA::test_gqa_4q1kv_attention_sink_prefill_accuracy \
  python/sgl_jax/test/test_flashattention_misc.py::TestFlashAttentionMisc::test_single_head_attention_sink_decode_accuracy \
  python/sgl_jax/test/test_flashattention_misc.py::TestFlashAttentionMisc::test_single_head_attention_sink_prefill_accuracy
```

## Checklist

- [x] The test plan, such as providing test commands.


Aharrypotter:
@cjx0709 cc

cjx0709:
Thanks for your contribution, I will test and review for this. I think maybe you can rebase origin/main at first
image

python/sgl_jax/srt/layers/attention/linear/lightning_backend.py
```diff
@@ -214,24 +221,27 @@ def _decode_fn(q_l, k_l, v_l, gamma, buf_l, idx_l, has_l):
...
-                P("data", "tensor", None),
-                P("data", "tensor", None),
-                P("data", "tensor", None),
-                P("tensor"),
-                P("data", "tensor", None, None),
-                P("data"),
-                P("data"),
-            ),
+            in_specs=in_specs,
             out_specs=(
```
JamesBrianD, line 229: This is problematic. The JAX 0.9.1 changelog states that users must explicitly call `jax.reshard` when resharding is intended. However, the inputs here are expected to already have shardings matching `in_specs`.

If `shard_map` rejects them because of a mismatch, that indicates a bug in data preparation or an upstream computation. We should fix the sharding at its source rather than mask the contract violation with `reshard` here.

python/sgl_jax/srt/layers/attention/linear/kda_backend.py
```diff
@@ -12,6 +12,7 @@
 )
 from sgl_jax.srt.layers.attention.linear.short_convolution import short_convolution
 from sgl_jax.srt.model_executor.forward_batch_info import ForwardMode
+from sgl_jax.srt.utils.jax_utils import shard_map_reshard
 from sgl_jax.srt.utils.profiling_utils import named_scope
```
JamesBrianD: As with the Lightning comment, we should avoid overusing reshard.

Aharrypotter: you are right

python/sgl_jax/srt/layers/attention/linear/lightning_backend.py
Aharrypotter: Agreed. I removed the consumer-side reshard instead of masking the contract violation. In the final tree, `lightning_backend.py` has no diff from `main` and there is no `shard_map_reshard` helper. The unit fixture now places Q/K/V and recurrent-state buffers on the layouts that production already provides via `LinearBase` and the state-pool producer. The focused v6e-4 Lightning suite passes 22/22. I also rewrote the commit history so the rejected blanket reshard no longer appears in an earlier commit.

python/sgl_jax/srt/layers/attention/linear/kda_backend.py
Aharrypotter: Agreed. The final KDA consumer is unchanged from `main`: no reshard import, wrapper, or call site remains. Replicated ad-hoc inputs were fixed at their test producer/fixture boundaries, while production `in_specs` stay explicit. The v6e-4 KDA suite passes 5/5, and the rewritten history removes the earlier blanket-reshard implementation entirely.

Aharrypotter:
Updated the PR to JAX/JAXLIB 0.10.2 (libtpu 0.0.42.1). This follow-up fixes the remaining MiMo `jnp.clip` API break, rebaselines and tightens the bf16 logprob check, and retunes the two v6e RPA mixed-prefill cells; the final single-chip gate reached 30,454.80 input tok/s and 45.74 ms TTFT without changing thresholds.

JamesBrianD:
@Aharrypotter I try to fix the tpu v6e-1 performance test failed, by tuning new rpa blocks

Aharrypotter:
> @Aharrypotter I try to fix the tpu v6e-1 performance test failed, by tuning new rpa blocks

Thanks, that matches my findings.

cjx0709 (approved):
LGTM

PR #1469, opened 2026-07-20
feat(pd): Pathways PD follow-ups — D-side KV admission reservation + logprob on the async prefill path

zorrofox:
Implements two follow-ups tracked in #1427 (promised in the #1428 review), plus one general tokenizer fix found while validating them.

Related but orthogonal to #1410 (multi-host DP logprob host transfer): that PR makes `_materialize_logprobs_to_host` sharding-aware for data-sharded tables across hosts; this one carries the (already host-materialized) logprobs through the Pathways single-controller async prefill path, where all arrays are locally addressable. The only shared file is `tp_worker.py` in adjacent, non-overlapping hunks — either PR rebases cleanly over the other.

## 1. D-side KV reservation / backpressure

The drain-time capacity gate + defer (pending-ready path) landed in #1453; this adds the admission-side reservation half:

- **Token ledger** (`_pd_inflight_tokens`): every admitted-but-not-yet-migrated req reserves its page-padded full-input footprint; released at migrate or on prefill failure. Idempotent per rid, so chunked re-admissions don't double-count.
- **Admission gate** (`_pd_token_gate_closed`): NEW admissions pause when `inflight footprint + decode-growth reserve (disaggregation_num_reserved_decode_tokens × (running + inflight)) >= total D free space` — instead of burning P compute on prefills that pile up in `_pd_defer` holding P KV. Chunked continuations still build. On SWA allocators the gate reads the FULL pool (the swa pool's transient full-seq occupancy during migrate recovers via `free_swa`; min()-gating would wedge admission for window-sized swa pools).
- **Starvation diagnostics**: a periodic warning with need/reserved/avail when a deferred ready item stays wedged, and an admission-paused counter log. `SGLANG_PD_NO_TOKEN_GATE=1` reverts to pre-gate behavior.

No new CLI flags; reuses `disaggregation_num_reserved_decode_tokens`.

## 2. logprob on the async prefill path

The async prefill result previously dropped `logits_output` and the per-req extend lens, so `return_logprob` was rejected at admission. The P-thread forward already computes and host-materializes every logprob field for logprob batches (they bypass the fused-sample jit), so:

- Keep `logits_output` in the async `GenerationBatchResult`; the shared prefill output processor consumes it unchanged, including chunked-prefill input-logprob accumulation.
- Snapshot `extend_input_len` / `extend_logprob_start_len` on the main thread at batch-build time: the depth-1 pipeline builds chunk N+1 (mutating those fields) while the P thread is still between forward and result construction — same reason `run_batch` copies them before overlap dispatch.
- Gate the Pathways fused-sample jit on `return_logprob` too (it only checked `return_output_logprob_only`), so a decode batch carrying a full-logprob req falls back to the unfused path instead of silently dropping `next_token_logprobs`. Unreachable outside Pathways PD.
- `return_hidden_states` remains rejected at admission.

## 3. fix(tokenizer): per-token batch_decode for logprob text under transformers>=5

Independent of PD (found while validating the logprob change in section 2 above, needed to observe it end-to-end). `detokenize_logprob_tokens` passes a flat list of ints to `batch_decode`. Under the pinned `transformers~=4.57.1` that already decodes per-token, so **this change is a byte-identical no-op on the pinned env**. Under transformers >= 5, `batch_decode` treats a flat int list as ONE sequence and returns a single string — the zip then silently truncates every `meta_info` output/input/top logprob list to one entry. This is reachable today: GLM-5 tokenizers don't load under transformers 4.x (`TokenizersBackend does not exist`), so any GLM-5 deployment already runs transformers 5 and gets truncated logprobs. Wrapping each id as its own sequence decodes per-token on both. Verified on 4.57.6 / 5.3.0 / 5.14.1 with two tokenizers: v4 flat == v4 wrapped == v5 wrapped == per-token; v5 flat == one concatenated string. The existing `test_logprobs.py` asserts values, not list length vs token count, so it doesn't catch the truncation on a v5 env. Happy to split this into a small prerequisite PR if preferred.

## Test plan

Environment: TPU v7x via Pathways single-controller, MiMo-V2-Flash, tp8 ep8 dp2, 2P1D (3 slices), 16K/4K, same config as the #1453 test plan.

**Standard regression** (`--pd-decode-max-tokens 800000`, pool not saturated — gate never fires):

| item | this branch | #1453 baseline | delta |
|---|---|---|---|
| correctness SHORT/MID/LONG × C4/8/16 | 336/336 | 336/336 | — |
| C128 r2 tok/s (Med ITL) | 15411 (32.52ms) | 15493 (32.3-32.5ms) | −0.53% (within node-to-node variance) |
| C128 r3 tok/s (Med ITL) | 15378 (32.69ms) | 15426 | −0.31% |
| gsm8k 200q | 0.890, Invalid 0 | 0.865–0.885 | — |
| `[pd-backpressure]` lines during bench | 0 | — | zero perturbation at standard load |

**Backpressure stress** (`--pd-decode-max-tokens 300000` → full pool 2,120,192 tok; C128 × (16K+4K) ≈ 2.6M demand → genuinely saturated; 256 prompts, single round each arm):

| metric | gate ON | gate OFF (`SGLANG_PD_NO_TOKEN_GATE=1`, = pre-PR behavior) |
|---|---|---|
| completed | 256/256 | 256/256 |
| crashes | 0 | 0 |
| total tok/s | 14117 | 14301 (−1.3%, single round; per-run variance 0.4–0.7%) |
| Median / P99 ITL | 30.54 / 123.4ms | 30.72 / 121.8ms |
| admission-paused episodes | 32 (opens/closes with pool availability) | 0 |
| ledger leak check | single request served immediately after the run | n/a |

The small gate-ON deficit at saturation is the expected trade: reqs wait in the waiting queue instead of pre-prefilled in the defer list; at standard load there is no cost.

**GLM-5.2-FP8 (MLA, non-SWA KV pool) — the gate's real-world engagement case** (tp16 ep16 dp1, 1P1D, 2 slices, 16K/4K C64, 192 prompts, 2 warm rounds each arm; this config's D pool is HBM-capped so it saturates at standard load and the gate genuinely engages, unlike the MiMo rows where it never fires):

| item | gate ON | gate OFF | #1453 baseline |
|---|---|---|---|
| correctness 336 | 336/336 | 336/336 | 336/336 |
| gsm8k 200q | 0.950 | 0.965 | 0.955 (±1-2 questions run-to-run) |
| C64 tok/s, 2-round mean | 1067.6 (r1/r2 1068.5/1066.6) | 1076.6 (1076.5/1076.7) | 1078 |
| Median / P99 ITL | 88.9–89.2 / 94.1–94.2ms | 89.3 / 93.6ms | — |
| admission-paused | 154+ episodes | 0 | — |
| crashes | 0 | 0 | — |

Gate ON costs a reproducible **−0.83%** at this saturated operating point (ITL unchanged): with the gate off, deferred pre-prefilled requests refill a freed D slot one admission→prefill cycle sooner. We think the trade is worth it (bounded defer, P KV not held hostage, requests remain abortable in the waiting queue), and `SGLANG_PD_NO_TOKEN_GATE=1` restores the old behavior — but happy to discuss adding a pre-staging slack or flipping the default if preferred.

**logprob e2e** (async prefill path, both models):

| test | result |
|---|---|
| short prompt, `return_logprob` + `top_logprobs_num=5`, temperature=0 (MiMo + GLM) | per-token logprobs == completion length; 5 candidates per position; sampled token == top-1 with equal logprob |
| MiMo: 8001-token prompt (4 × 2048 chunks), `logprob_start_len=0` | 8001 input logprob entries (first `None` by convention), all finite ≤ 0, no crash |
| GLM: 5001-token prompt (2 × 4096 chunks), `logprob_start_len=0` | 5001 input logprob entries, all finite ≤ 0; output logprobs == completion length under `ignore_eos` |
| mixed concurrency: 2 logprob + 2 plain | all 4 outputs byte-equal to a plain reference run |

- CPU unit tests: 7 new (ledger idempotence, gate projection math, multi-D/rank summing, SWA full-pool semantics, env escape); `test_pathways_pd.py` 28 pass
- Known not-covered: the `return_output_logprob_only`-only path (bare `return_logprob` requests are converted to it by the tokenizer manager) shares the same `next_token_logprobs` production but has no dedicated assertion in the e2e above.

### DIFF
--- a/python/sgl_jax/srt/disaggregation/pathways_scheduler.py
+++ b/python/sgl_jax/srt/disaggregation/pathways_scheduler.py
@@ -227,6 +227,11 @@ class PathwaysPDSchedulerMixin:
         # reqs pushed to prefill_q but not yet drained into running_batch;
         # tracked by rid so a chunked req counts once, not once-per-chunk.
         self._pd_inflight_rids: set[str] = set()
+        # D-pool tokens the inflight rids will need at migrate time (page-
+        # padded full IL per req) -- the admission-side reservation ledger.
+        self._pd_inflight_tokens = 0
+        self._pd_gate_ticks = 0
+        self._pd_defer_ticks = 0
         self._pd_reserved_per = server_args.disaggregation_num_reserved_decode_tokens
         self._pd_sliding_window = self.tp_workers_p[0].sliding_window_size or 0
         self._pd_next_p = 0
@@ -647,7 +652,7 @@ class PathwaysPDSchedulerMixin:
                     if r.req_pool_idx is not None:
                         p_tree.cache_finished_req(r)
                         p_r2t.free_slots.append(r.req_pool_idx)
-                    self._pd_inflight_rids.discard(r.rid)
+                    self._pd_untrack_inflight(r)
                     all_reqs.append(r)
             self.stream_output(all_reqs, False, False)
             for dp in chunked_excl:
@@ -671,7 +676,22 @@ class PathwaysPDSchedulerMixin:
                 for rk, nd in need_by_rank.items()
             ):
                 self._pd_defer.append(item)
+                # Starvation diagnostic: defer is silent by design (retried
+                # every drain tick); if it stays wedged (e.g. a single item
+                # whose need can never fit next to the reserve) surface the
+                # numbers instead of hanging mutely. ~30s at a ~29ms tick.
+                self._pd_defer_ticks += 1
+                if self._pd_defer_ticks % 1024 == 0:
+                    logger.warning(
+                        "[pd-backpressure] ready item deferred x%d: need=%s "
+                        "reserved/rank=%d avail=%s",
+                        self._pd_defer_ticks,
+                        dict(need_by_rank),
+                        reserved_per_rank,
+                        {rk: allocator.available_size(dp_rank=rk) for rk in need_by_rank},
+                    )
                 return
+            self._pd_defer_ticks = 0
         d_pages_all = []
         for r, p_slots, p_pages in done_per_req:
             seq_len = len(r.fill_ids)
@@ -773,7 +793,7 @@ class PathwaysPDSchedulerMixin:
                 (time.perf_counter() - t_dequeue_ready) * 1e3,
             )
         for r, _, _ in done_per_req:
-            self._pd_inflight_rids.discard(r.rid)
+            self._pd_untrack_inflight(r)
         if _PD_DBG and done_per_req and hasattr(p_alloc, "swa_available_size"):
             logger.info(
                 "[pd-diag] after free %d done: P full_avail=%d swa_avail=%d",
@@ -827,6 +847,65 @@ class PathwaysPDSchedulerMixin:
                 break
         return n_calls
 
+    def _pd_req_kv_tokens(self, r) -> int:
+        """Page-padded D-pool token footprint a req needs at migrate time
+        (migrate allocs len(p_pages) * page_size for the full input)."""
+        n = len(r.origin_input_ids)
+        return -(-n // self.page_size) * self.page_size
+
+    def _pd_track_inflight(self, r) -> None:
+        """Reserve r's D-pool footprint at admission. Idempotent per rid so
+        a chunked req's chunk-2..N batches don't double-count."""
+        if r.rid not in self._pd_inflight_rids:
+            self._pd_inflight_rids.add(r.rid)
+            self._pd_inflight_tokens += self._pd_req_kv_tokens(r)
+
+    def _pd_untrack_inflight(self, r) -> None:
+        """Release r's reservation (migrated into D, or prefill failed)."""
+        if r.rid in self._pd_inflight_rids:
+            self._pd_inflight_rids.discard(r.rid)
+            self._pd_inflight_tokens = max(0, self._pd_inflight_tokens - self._pd_req_kv_tokens(r))
+
+    def _pd_token_gate_closed(self) -> bool:
+        """D-side token-level admission reservation (#1427 follow-up).
+
+        Closed when the inflight reqs' migrate-time footprint plus the
+        decode-growth reserve already covers the D pools' free space --
+        admitting more would just burn P compute on prefills that pile up
+        in _pd_defer holding P KV. Totals across D targets and dp ranks
+        (per-rank exact fit is enforced by the drain-time gate + defer);
+        check-then-build, so one build of overshoot is possible -- also
+        absorbed by the drain gate. On SWA allocators gate on the FULL
+        pool: the swa pool's transient full-seq occupancy during migrate
+        recovers via free_swa, and min()-gating on it would wedge
+        admission for models whose swa pool is window-sized.
+        """
+        if os.environ.get("SGLANG_PD_NO_TOKEN_GATE"):
+            return False
+        avail = 0
+        for alloc in self.d_allocs:
+            f = getattr(alloc, "full_available_size", alloc.available_size)
+            avail += sum(f(dp_rank=rk) for rk in range(self.dp_size))
+        reserved = self._pd_reserved_per * (
+            self._pd_total_d_running() + len(self._pd_inflight_rids)
+        )
+        closed = self._pd_inflight_tokens + reserved >= avail
+        if closed:
+            self._pd_gate_ticks += 1
+            if self._pd_gate_ticks % 512 == 1:
+                logger.info(
+                    "[pd-backpressure] admission paused x%d: inflight=%d reqs "
+                    "%d tok + reserved %d >= D avail %d",
+                    self._pd_gate_ticks,
+                    len(self._pd_inflight_rids),
+                    self._pd_inflight_tokens,
+                    reserved,
+                    avail,
+                )
+        else:
+            self._pd_gate_ticks = 0
+        return closed
+
     def _pd_maybe_push_prefill(self) -> None:
         """Build at most one P batch (round-robin across P slices) and push to
         its prefill_q. Extracted so both the n_decode==1 path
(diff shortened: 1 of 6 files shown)


python/sgl_jax/srt/managers/tp_worker.py
```diff
@@ -475,9 +475,12 @@ def forward_batch_generation(
 
         # Pathways-PD: fuse run_model+sampler+resolve/set into one jit so a
         # decode tick is a single Execute through the ordered dispatch queue.
+        # Any logprob batch falls through to the unfused path below, which is
+        # the only one that computes/materializes logprob fields.
         if (
             self._pd_fuse_sample
```
cjx0709: The client side still selects the fused path from `_pd_fuse_sample` alone and unpacks four values. This batch now falls through to the regular path, which returns three values, so `SGLANG_PD_FUSE_SAMPLE=1` with logprobs will fail during unpacking. The client needs the same batch-level fallback.

cjx0709 (commented):
Two focused comments on the fused logprob fallback and the DP>1 admission gate behavior.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
```diff
@@ -857,12 +952,20 @@ def _pd_maybe_push_prefill(self) -> None:
...
             if self._pd_chunk_pending[i]:
                 continue
             if self._pd_prefill_qs[i].full():
                 continue
+            # Token gate blocks NEW admissions only; a P slice with a pending
+            # chunked req still builds so chunk-N+1 can complete and migrate
+            # (its footprint is already in the ledger from chunk-1). The
+            # builder may admit new reqs alongside a continuation -- bounded
+            # overshoot, see _pd_token_gate_closed docstring.
+            if gate_closed and not any(r is not None for r in self.p_chunked_reqs[i]):
```
cjx0709, line 965: If any DP rank has a chunked request, this still runs the full `get_new_batch_prefill()`. With DP>1, it can admit new requests from the other DP ranks even while the gate is closed, then prefill and defer them. Should we run only existing chunks while the gate is closed, without scanning the waiting queue?

python/sgl_jax/srt/managers/tp_worker.py
zorrofox: Good catch — the client picked the branch (and unpack arity) from the worker-level flag alone. Fixed in e3fccfee: the batch-level predicate is now `ModelWorker._pd_fuse_for_batch(mwb)`, used by both `forward_batch_generation` and the overlap client so the branch choice and tuple arity can never diverge (unit test added).

While validating this on TPU we found `SGLANG_PD_FUSE_SAMPLE=1` is currently broken on main independently of logprobs: the fused jit's future-token resolve hits `ShardingTypeError: select cases must have the same shardings [P('data'), P(None)]` on the first batch (both the PD prefill thread's EXTEND batches and D-side DECODE, dp=1 and dp=2 alike) — the env flag is off by default and not exercised by CI, so it looks like the fused path rotted when the mesh moved to explicit axis types. Two consequences in this PR: `_pd_fuse_for_batch` is additionally gated on `forward_mode.is_decode()` (prefill was never the fusion target), and the unpack fix itself is validated by unit test rather than e2e, since the fused path can't currently run at all. Happy to file a separate issue for the fused-path sharding fix — it's orthogonal to logprobs.

Regression on the fixed head (SGLANG_PD_FUSE_SAMPLE unset, the default): correctness 336/336, C128 r1/r2 15565/15543 tok/s (Med ITL 32.12/32.24), saturated-pool stress 256/256 with zero crashes.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
zorrofox: Right — `get_new_batch_prefill()` scans the waiting queue for every dp rank, so a chunk-continuation build under a closed gate could admit new requests on the other ranks. Fixed in e3fccfee: while the gate is closed, the waiting queue is hidden (swapped for an empty list) for the duration of that build — existing chunks still advance, nothing new is admitted on any rank.

Re-validated at the saturated operating point (dp=2, `--pd-decode-max-tokens 300000`, 16K/4K C128, 256 prompts): 256/256 completed, no crashes, and the `[pd-backpressure]` log shows inflight holding at the pipeline depth (2–4 reqs) across pause episodes instead of creeping up — 14243 tok/s, in line with the pre-fix gate-on run (14117) and the gate-off control (14301).

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
```diff
@@ -883,7 +984,20 @@ def _pd_maybe_push_prefill(self) -> None:
...
                 with self._pd_swap_p_pool(i):
-                    new_batch = self.get_new_batch_prefill()
+                    if gate_closed:
+                        # Chunk-continuation only: hide the waiting queue so
+                        # the builder cannot admit new reqs from ANY dp rank
+                        # while the gate is closed (with dp>1, ranks without a
+                        # chunk would otherwise fill their slots from the
+                        # queue, leaking admissions past the gate).
+                        _saved_q = self.waiting_queue
+                        self.waiting_queue = []
```
cjx0709: Small issue with swapping the queue: `get_new_batch_prefill()` first moves ready grammar requests from `grammar_queue` into this temporary list. After `finally` restores `_saved_q`, a grammar request that was not admitted is in neither queue (and another-rank one can still bypass the gate). Could we skip waiting-queue admission explicitly instead?

cjx0709 (commented):
One additional comment on the closed-gate queue handling.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
zorrofox: You're right on both counts — the moved-but-unadmitted grammar reqs were stranded in neither queue, and the grammar move itself was another way past the gate. Replaced the swap with an explicit pause in 3b3294fc: the PD build sets `_pd_admission_paused`, and `get_new_batch_prefill()` skips both the grammar-queue move and the waiting-queue scan while it's set (existing chunks still advance). The flag is getattr-default-False and never set outside Pathways PD, so the native path is untouched.

Re-validated at the saturated operating point (dp=2, `--pd-decode-max-tokens 300000`, 16K/4K C128, 256 prompts): 256/256 completed, zero crashes, inflight steady at pipeline depth across pause episodes, 14140 tok/s — in line with the earlier gate-on run (14117) and the gate-off control (14301). The GLM-5.2-FP8 saturated config (where the gate engages at standard load) also re-measures unchanged: C64 r1 1067.7 tok/s vs 1068.5/1066.6 before, so the numbers in the PR description stand.

cjx0709 (approved):
LGTM

Issue #1400, opened 2026-06-26
[Bug] Step-3.5-Flash multi-host NEXTN spec-decode: multiple correctness bugs (see numbered Problem comments)


Rodrian7:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version (confirmed present on current `main`).
- [x] 3. Environment + reproduction are provided below (see caveat: observed on a multi-host pod; not reduced to a minimal CI repro).
- [x] 4. This is a bug, not a question.
- [x] 5. English.

### Describe the bug

Under a **multi-host** deployment (`--nnodes >= 2` with `--dp-size >= 2`, so per-token logprob arrays are sharded across hosts), any request with `return_logprob` crashes the scheduler.

`scheduler_output_processor_mixin.py` calls `jax.device_get(...)` **directly** on the per-token logprob arrays. Under multi-host these are global `jax.Array`s that are **not fully addressable by a single process**, and `jax.device_get` on such an array raises.

Single-host deployments are unaffected (the array is fully local to the one process), which is why single-host CI (v6e-1 / v6e-4) never triggers it. The bug is **model-agnostic** — it lives in the shared serving output path, not in any specific model.

Location (current `main`, `python/sgl_jax/srt/managers/scheduler_output_processor_mixin.py`):
- L59 — `host_logprobs = np.asarray(jax.device_get(input_token_logprobs))` (in `_materialize_input_token_logprobs`)
- L132 / L136 — `logits_output.next_token_logprobs = jax.device_get(logits_output.next_token_logprobs)`
- same bare-`device_get`-on-logprobs pattern also around L403 (get-logprobs path)

**Suggested fix:** gather the sharded logprob arrays across hosts before the host transfer — e.g. `jax.experimental.multihost_utils.process_allgather(...)` instead of the bare `jax.device_get` — keeping the gathered order consistent with the per-DP padded compaction in `_materialize_input_token_logprobs`.

### Reproduction

1. Launch a server **multi-host** so logprobs shard across hosts (any model):
   ```
   python -m sgl_jax.launch_server --model-path <model> --device tpu \
     --tp-size <chips> --dp-size <N>=2 --nnodes <hosts>=2 --node-rank $RANK --dist-init-addr $ADDR \
     --host 0.0.0.0 --port 30000
   ```
2. Send a request asking for logprobs:
   ```
[...]

Issue #1010, opened 2026-04-30
Accuracy degradation with retract + small chunked-prefill-size


JamesBrianD:
## Description

When `SGLANG_TEST_RETRACT=1` forces frequent retract (every 3 forward passes) and `chunked-prefill-size` is small (128), MMLU accuracy drops significantly below threshold (0.484 vs 0.5 required). The issue does NOT occur with larger `chunked-prefill-size` (1024) or without retract.

## Reproduction

```bash
# Fails (score ~0.484):
SGLANG_TEST_RETRACT=1 python -m pytest test/srt/test_retract_decode.py::TestRetractDecodeChunkCachePaged -xvs
# Server config: --disable-radix-cache --page-size 16 --chunked-prefill-size 128

# Passes (score ~0.688):
# Same but with --chunked-prefill-size 1024
```

## Test matrix

| Config | SGLANG_TEST_RETRACT | chunked-prefill-size | MMLU Score | Result |
|--------|---------------------|---------------------|------------|--------|
| ChunkCache + page16 | off | 128 | 0.719 | PASS |
| ChunkCache + page16 | on | 1024 | 0.688 | PASS |
| ChunkCache + page16 | on | 128 | 0.484 | FAIL |

## Analysis

The retract path (`reset_for_retract` → re-queue → `prepare_for_extend`) correctly resets KV state. However, when the re-prefill is chunked into very small pieces (128 tokens), the repeated chunk boundaries combined with frequent retraction appear to cause incorrect attention computation or KV cache state corruption.

Possible root causes:
1. `is_chunked` / `already_computed` interaction with retract — after retract, `is_chunked=0` and `already_computed=0` are reset, but the chunked prefill path may not correctly handle the re-prefill of previously partially-chunked sequences
2. Page alignment issues when `page_size=16` and `chunked-prefill-size=128` interact with retract — the page boundary math in `alloc_paged_token_slots_extend` may produce incorrect cache locations after retract resets `kv_allocated_len=0`
3. Forced retract interval (every 3 forward passes) may interrupt mid-chunk prefills, leaving partial KV state that isn't fully cleaned up

## Context

- PR #994 fixed the infinite prefill→retract loop for this config
- The accuracy issue exists independently of the memory management changes
- `TestRetractDecodeChunkCachePaged` is skipped pending this fix

## Related

- PR #994 (memory management unification)
- sglang upstream #12224

JamesBrianD:
[...]

PR #1471, opened 2026-07-20
Fix chunked-prefill request ownership lifecycle

Rodrian7:
## Summary

Fix chunked-prefill requests losing their scheduler owner when an intermediate extend batch is filtered.

- Restore one canonical owner per DP rank before filtering.
- Track pending aborts with the owned request and stop scheduling further chunks.
- Let the consumer of an in-flight result finish logprob and trace bookkeeping before releasing host, KV, and request-pool resources.
- Keep parked abort and retract cleanup explicit, while preserving normal intermediate and final chunk behavior.

Addresses Problem 2 in #1400.

This covers the standard scheduler and multi-process PD paths. `pd == "pathways"` keeps its separate per-P lifecycle and event loop and is intentionally not covered by this change.

## Lifecycle

Before:

1. An intermediate chunk returned with KV and a request slot still allocated.
2. `filter_batch()` excluded it from the completed extend batch.
3. If `scheduler.chunked_reqs` had lost the pointer, the live request became unreachable: no next chunk could be scheduled and abort or retract could not release it.

After:

1. Before filtering, the scheduler restores `last_batch.reqs_info[dp_rank].chunked_req` as the canonical per-DP owner.
2. A non-aborted owner is cached only when it has new KV beyond its matched prefix and remains eligible for its next chunk.
3. A parked abort is released at the scheduler boundary. An overlap in-flight abort is marked pending and is not rescheduled; its result consumer drains the chunk, advances the shared logprob cursor, completes tracing, and then releases the PD host buffer, KV cache, request slot, and owner exactly once.
4. Standard PD-prefill decrements the existing `req.is_chunked` in-flight state in its own result path, so a pending abort is finalized only after that consumer returns.

## Validation

- Focused ownership, overlap-abort, idle-state, and mixed-DP tests: 15 passed plus 12 subtests.
- Standard PD-prefill unit tests: 64 passed.
- Added a registered DP=2 overlap stress test that drives both DP KV pools past 50% utilization under concurrent chunked requests, aborts both active owners, verifies terminal results and full resource recovery, and then sends a follow-up chunked request.
- Falcon `exp-h6189qgrzb` compared the exact PR base `91399b4` and head `be3652f` sequentially on the same v6e-4 worker with TP=4, DP=2, overlap enabled, chunk size 128, and 12 concurrent 3,872-token prompts. Across three post-warmup runs, median input throughput was 20,707 versus 20,579 tokens/s (-0.62%), and median batch time was 2.244 versus 2.258 seconds (+0.62%); no material performance regression was observed.
- In three current-head abort-pressure runs from the same experiment, both active owners returned terminal abort results, all KV and request slots were restored, and the follow-up request completed every time. Median/max terminal abort latency was 17.0/17.1 ms, median full workload-drain and recovery time was 1.929 seconds, and median follow-up latency was 317.4 ms.
- `pre-commit run --all-files` passed.

### DIFF
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -78,6 +78,7 @@ from sgl_jax.srt.managers.tp_worker_overlap_thread import ModelWorkerClient
 from sgl_jax.srt.managers.utils import validate_input_length
 from sgl_jax.srt.mem_cache.base_prefix_cache import MatchPrefixParams
 from sgl_jax.srt.mem_cache.chunk_cache import ChunkCache
+from sgl_jax.srt.mem_cache.common import release_kv_cache
 from sgl_jax.srt.mem_cache.kv_cache_builder import build_kv_cache
 from sgl_jax.srt.mem_cache.radix_cache import RadixKey
 from sgl_jax.srt.mem_cache.swa_radix_cache import SWARadixCache
@@ -449,6 +450,7 @@ class Scheduler(
         if self.chunked_prefill_size <= 0:  # -1 means disable
             self.chunked_prefill_size = None
         self.chunked_reqs = [None] * self.dp_size  # Per-DP chunked requests
+        self._pending_chunked_abort_reqs = [None] * self.dp_size
         self.is_mixed_chunk = (
             self.chunked_prefill_size is not None and server_args.enable_mixed_chunk
         )
@@ -1053,10 +1055,7 @@ class Scheduler(
                         sum(len(i.reqs) for i in self.running_batch.reqs_info),
                     )
             else:
-                # When the server is idle, do self-check and re-init some states
-                self.check_memory()
-                self.check_tree_cache()
-                self.new_token_ratio = self.init_new_token_ratio
+                self.on_idle()
 
                 # Elegant wait if idle
                 if self._comm_backend is not None:
@@ -1149,10 +1148,7 @@ class Scheduler(
                     tmp_batch, tmp_result, batch.launch_done if batch else None
                 )
             elif batch is None:
-                # When the server is idle, do self-check and re-init some states
-                self.check_memory()
-                self.check_tree_cache()
-                self.new_token_ratio = self.init_new_token_ratio
+                self.on_idle()
 
             self.last_batch = batch
             if _pd_iter_trace:
@@ -1426,6 +1422,7 @@ class Scheduler(
         # state for pause/continue generation
         ret["engine_paused"] = self._engine_paused
         ret["waiting_queue_size"] = len(self.waiting_queue)
+        ret["pending_dp_reqs_size"] = len(self.pending_dp_reqs)
         ret["running_batch_size"] = (
             0 if self.running_batch.is_empty() else self.running_batch.batch_size()
         )
@@ -1438,6 +1435,7 @@ class Scheduler(
         ret["cur_batch_is_none"] = self.cur_batch is None
         ret["last_batch_is_none"] = self.last_batch is None
         ret["chunked_req_is_none"] = all(r is None for r in self.chunked_reqs)
+        ret["chunked_req_rids"] = [r.rid if r is not None else None for r in self.chunked_reqs]
 
         # request cache stat
         if isinstance(self.tree_cache, ChunkCache):
@@ -1459,6 +1457,10 @@ class Scheduler(
 
         # physical kv cache stat
         ret["available_kv_tokens"] = self.token_to_kv_pool_allocator.available_size()
+        ret["available_kv_tokens_per_dp"] = [
+            self.token_to_kv_pool_allocator.available_size(dp_rank)
+            for dp_rank in range(self.dp_size)
+        ]
 
         # counters
         ret["num_generated_tokens"] = self.num_generated_tokens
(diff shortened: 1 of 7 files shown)


python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -1742,6 +1755,100 @@ def _get_swa_token_info(self):
...
+                reqs = info.reqs or []
+                keep_indices[dp_rank] = [
+                    i for i, batch_req in enumerate(reqs) if batch_req is not req
+                ]
+                found |= len(keep_indices[dp_rank]) != len(reqs)
+                if info.chunked_req is req:
+                    info.chunked_req = None
+                    found = True
+            if found:
+                batch.filter_batch(keep_indices=keep_indices)
```
JamesBrianD: I’d suggest following SGLang’s pending-abort pattern more closely: model parked chunk abort/retract as an explicit state transition with one unified cleanup path, instead of relying on _remove_parked_req_from_completed_batches() to patch up stale references afterwards.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -1742,6 +1755,100 @@ def _get_swa_token_info(self):
...
+                continue
+            self._remove_parked_req_from_completed_batches(req)
+            self._release_prefill_host_buffer(req)
+            release_kv_cache(
+                req,
+                self.tree_cache,
+                allow_overallocated=(
+                    self.spec_algorithm is not None and not self.spec_algorithm.is_none()
+                ),
+            )
```
JamesBrianD: Abort request should use "is_insert = False"

JamesBrianD:
It would also be good to add a stress test with high concurrency that fills the KV cache, plus a few aborted requests, to verify that the fix works under pressure and does not regress performance or correctness.

python/sgl_jax/srt/managers/scheduler.py
Rodrian7: Addressed. The stale-reference scan and `_remove_parked_req_from_completed_batches()` have been removed. Abort now restores the canonical per-DP owner, records it in `_pending_chunked_abort_reqs`, and prevents another chunk from being scheduled. Parked and result-consumer aborts converge on `_finalize_chunked_abort()`, which performs terminal bookkeeping and releases the host buffer, KV cache, request slot, and owner once. Retract remains explicit because it resets and requeues the request instead of terminating it.

python/sgl_jax/srt/managers/scheduler.py
Rodrian7: Addressed. `_finalize_chunked_abort()` now calls `release_kv_cache(req, self.tree_cache, is_insert=False, ...)`; the parked-retract release path also uses `is_insert=False`.

Rodrian7:
@JamesBrianD Done. The registered `TestChunkedPrefillSize.test_stress_abort_recovers_chunked_owners_and_capacity` E2E test now runs in the v6e-4 suite. It uses 12 concurrent chunked requests, waits until both DP KV pools are past 50% utilization with queued work, aborts the active owner on each DP rank, verifies the aborted and live results, checks full KV/request-slot recovery, and completes a follow-up request.

I also ran a targeted Falcon A/B (`exp-h6189qgrzb`) on the exact PR base `91399b4` and current head `be3652f`, sequentially on the same v6e-4 worker with TP=4, DP=2, overlap enabled, chunk size 128, and 12 concurrent 3,872-token prompts. Across three post-warmup runs:

- Median input throughput: 20,707 tokens/s (base) vs. 20,579 tokens/s (head), -0.62%.
- Median batch time: 2.244 s (base) vs. 2.258 s (head), +0.62%.

No material normal-path performance regression was observed. The broken base cannot provide a meaningful abort-path A/B, so I ran three additional abort-pressure iterations on the head. All three returned terminal aborts for both active owners, restored all KV/request slots, and completed the follow-up request. Median/max terminal abort latency was 17.0/17.1 ms; median full workload-drain and recovery time was 1.929 s; median follow-up latency was 317.4 ms.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -1742,6 +1763,103 @@ def _get_swa_token_info(self):
...
+                continue
+            pending_req = self._pending_chunked_abort_reqs[dp_rank]
+            assert pending_req is None or pending_req is req
+            self._pending_chunked_abort_reqs[dp_rank] = req
+            req.to_finish = FINISH_ABORT()
+
+    def _process_pending_chunked_aborts(self) -> dict[int, Req]:
+        consumed: dict[int, Req] = {}
+        if self.pd == "pathways":
+            return consumed
```
JamesBrianD, line 1808: Why is "pathways" being skipped here?

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -1742,6 +1763,103 @@ def _get_swa_token_info(self):
...
+                req,
+                self.tree_cache,
+                is_insert=False,
+                allow_overallocated=(
+                    self.spec_algorithm is not None and not self.spec_algorithm.is_none()
+                ),
+            )
+            self.chunked_reqs[dp_rank] = None
+            req.reset_for_retract()
+            self._add_request_to_queue(req)
```
JamesBrianD, line 1861: https://github.com/sgl-project/sglang-jax/issues/1010 
I'll use this PR to see if the drop in decode/retract accuracy has been fixed.

JamesBrianD: I test it. Already fixed.

python/sgl_jax/srt/managers/scheduler.py
Rodrian7: Pathways was skipped because it uses a separate per-P asynchronous lifecycle, and applying the generic finalizer would release resources from the wrong pool. With #1469 merged, the proper fix is clearer, but it also requires P-to-D migration and paused-drain changes that would materially expand this PR. The complete Pathways fix is tracked in #1486.

JamesBrianD (approved):

Issue #1425, opened 2026-06-30
HiCache S5 · Recurrent nightly regression gate


MokusMokun:
Standing nightly regression gate for the recurrent radix cache. Implementation: MokusMokun/sglang-jax#10 (rebases onto `main` after PR-1 #1402 + PR-3 #1408).

Gates (single-host v6e-4):
- `recurrent-e2e` — cache-on output byte-identical to `--disable-radix-cache` (determinism)
- `recurrent-ab-perf` — `no-cache` vs `unified-recurrent` A/B (p50 TTFT / tput thresholds)
- `recurrent-cache-aware` — reuse-sweep knee within predicted range
- `recurrent-accuracy` — Qwen3.5-35B-A3B tp4/dp2, cache on (gsm8k greedy / mmlu thinking thresholds)

- Scope: recurrent only (SWA nightly deferred to S5c #1341)
- Depends on: PR-1 #1402 (merged) + PR-3 #1408
- Parent: #1338

PR #1476, opened 2026-07-23
fix(recurrent): stabilize CoW and Track JIT structure

MokusMokun:
Fixes serving-time JIT misses caused by recurrent CoW/Track metadata changing between `None` and arrays. This is the JIT-structure prerequisite for the performance coverage in #1477 and the performance portion of #1425.

## Problem

Three recurrent metadata fields are optional:

- `recurrent_cow_src_indices`
- `recurrent_track_indices`
- `recurrent_track_mask`

Previously, a batch without CoW or a Track boundary used `None`, while a batch performing the operation used arrays. `None` and arrays are different JAX PyTree structures, so the first real cache hit or Track boundary compiled a new executable on the serving path. This added a 30–32 second request-latency spike and failed the cache-miss check.

Precompiling every presence/absence combination removes the serving miss, but increases this profile from 3 executables to 8 and substantially delays server readiness.

## Approach

Whether a server supports recurrent CoW or Track is already known at startup, so the metadata structure is decided once per server:

```text
capability disabled -> None for the server lifetime
capability enabled  -> fixed int32 arrays; zero values mean no-op
```

Concretely:

- CoW-enabled EXTEND batches always carry a source-index array; `0` means no clone.
- Track-enabled EXTEND and DECODE batches always carry index and mask arrays; a zero mask means no boundary.
- Real CoW/Track work changes only array values, not the PyTree structure.
- `CompilationManager` creates dummy batches with the same capability-derived structure as runtime batches.
- Each existing shape bucket is compiled once; no presence/absence matrix is added.

This keeps non-recurrent and capability-disabled servers unchanged.

## Result

The same Qwen3.5-35B-A3B TP4 extra-buffer profile was used throughout.

| Implementation | Executables | Cold startup | Warm startup | First CoW/Track request |
|---|---:|---:|---:|---:|
| all-`None` precompile | 3 | 106.86 s | 53.93 s | 30–32 s JIT miss |
| presence matrix | 8 | 237.77 s | 85.53 s | no miss |
| fixed metadata | 3 | 112.0 s | 54.6 s | no miss |

Real shared-prefix requests reported 2048/4096 cached tokens, confirming that the CoW path was exercised without a serving-time JIT miss.

Fixed metadata restores the original executable count and startup cost. The trade-off is a small steady-state cost from carrying the masked no-op path on enabled servers: approximately 1.6% GSP throughput and 0.9% random-workload throughput versus the presence-matrix revision. Both remain within the 2% nightly gates in #1477.

## Tests

- recurrent compilation/metadata tests: 42 passed locally and on TPU
- repo-pinned pre-commit hooks passed
- full PR CI passed

No user-visible flag, recurrent cache semantics, or memory sizing is changed.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -740,20 +740,17 @@ class ScheduleReqsInfo:
     # Recurrent state indices for hybrid recurrent models (per DP)
     recurrent_indices: np.ndarray | None = None
 
-    # Recurrent CoW src slot per req (0 = no clone), per DP
+    # Recurrent CoW src slot per req (0 = no clone), per DP.
     recurrent_cow_src_indices: np.ndarray | None = None
 
-    # Recurrent track metadata per DP (extra-buffer; set only on a track boundary).
+    # Recurrent track metadata per DP (extra-buffer; mask 0 = no boundary).
     recurrent_track_indices: np.ndarray | None = None
     recurrent_track_mask: np.ndarray | None = None
 
 
-def _build_recurrent_cow_src_indices(reqs: list[Req]) -> np.ndarray | None:
-    """None when no clone is pending, so cold extends skip the donated-buffer
-    CoW scatter in _maybe_apply_recurrent_cow."""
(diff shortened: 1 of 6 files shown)


cjx0709 (approved):
LGTM

PR #1479, opened 2026-07-24
feat(vlm): add cache-aware multimodal scheduling

Yanko-7:
## Motivation

Multimodal prompts use the same placeholder token IDs for different images and videos. Using those raw token IDs as radix-cache keys can therefore make requests with different media appear to share the same prefix. The same mismatch also affects cache-aware DP routing because the routing probe must use the exact key used by the radix cache.

This PR establishes a model-agnostic cache-key contract for in-model VLM requests:

- real model inputs remain unchanged;
- only the cache key replaces multimodal placeholder spans with per-item hash values;
- prefix lookup, cache insertion, and DP affinity use the same effective key.

The contract is implemented on the shared `MultimodalInputs` / `MultimodalDataItem`
representation and applies to image, video, audio, and future modalities. Qwen2.5-VL
is the first in-model integration in this stack, not the boundary of the design.

This is PR 1 of 4 in the VLM stack and is the only PR in the stack that is currently ready for review.

| Order | PR | Scope | Status |
| --- | --- | --- | --- |
| 1 | **#1479** | Cache-aware multimodal scheduling | Ready |
| 2 | #1480 | Generic vision runtime and encoder sharding | Draft; depends on this PR |
| 3 | #1481 | Bucketed vision precompile | Draft; depends on #1480 |
| 4 | #1482 | Qwen3-VL in-model support | Draft; depends on #1481 |

### Cache-key isolation

- Add `Req.cache_input_ids`, derived from `origin_input_ids` by replacing each multimodal placeholder range with the corresponding item's hash-based `pad_value`.
- Keep `origin_input_ids`, `fill_ids`, and the actual model input untouched. The substituted IDs are used only for cache matching and insertion.
- Use the same substituted key in:
  - radix prefix lookup;
  - finished and unfinished request cache insertion;
  - cache-aware DP routing.
- Preserve compatibility with request-like objects and existing test mocks that do not define `cache_input_ids`.

### Placeholder range contract

- Replace inclusive `offsets` with explicit half-open `placeholder_ranges` (`[start, end)`).
- Make the Qwen-VL processor validate:
  - placeholder count against `grid_thw`;
  - contiguous placeholder spans;
  - feature row count against the vision grid;
  - one range assignment per multimodal item.
- Use these ranges for per-item hash substitution instead of replacing every occurrence of a modality token with a single value.

### VLM scheduling defaults

- Keep radix cache and chunked prefill enabled for in-model VLM implementations that
  satisfy the shared placeholder-range and chunk-local routing contracts.
- Keep conservative architecture capability gates for VLM implementations that have
  not yet migrated to or validated those contracts.
- Disable overlap scheduling only for the standalone `--multimodal` multistage pipeline; in-model VLMs continue to use the regular scheduler policy.

## Testing

```bash
python -m pytest \
  python/sgl_jax/test/multimodal/test_qwen_vl_processor.py -q
```

```text
8 passed
```

## Review Notes

- The main invariant is that cache lookup and cache insertion use the same substituted token sequence and length.
- Hash substitution must never reach the model forward path.
- Cache-key substitution is model-agnostic; adding another VLM requires producing the
  shared multimodal item and placeholder-range metadata, not adding Qwen-specific cache logic.
- The standalone multistage pipeline still passes dictionary-based multimodal inputs to its autoregressive stage. This PR does not claim cache-aware multistage radix reuse; it preserves the existing multistage overlap behavior and focuses on the in-model path.
- Review this PR independently. The later draft PRs show cumulative diffs against `epic/vlm`; their diffs shrink as earlier PRs merge.

## Checklist

- [x] The PR description and code are in English.
- [x] The purpose and scope are documented.
- [x] The test plan includes reproducible commands and results.
- [x] Documentation impact was reviewed; no user-facing documentation change is required.

### DIFF
--- a/python/sgl_jax/srt/managers/dp_schedule_policy.py
+++ b/python/sgl_jax/srt/managers/dp_schedule_policy.py
@@ -36,7 +36,7 @@ def req_prefix_match_key(req) -> tuple[list[int] | None, str | None]:
     prefix (an unexpanded batch, an empty/one-token prompt, or a request whose
     reusable prefix clamps to zero), so the caller falls back to load balancing.
     """
-    input_ids = req.input_ids
+    input_ids = getattr(req, "cache_input_ids", None) or req.input_ids
     if not isinstance(input_ids, list) or not input_ids or not isinstance(input_ids[0], int):
         return None, None
 
--- a/python/sgl_jax/srt/managers/dp_schedule_policy.py
+++ b/python/sgl_jax/srt/managers/dp_schedule_policy.py
@@ -17,7 +17,9 @@ BALANCE_REL = 1.1
 CACHE_THRESHOLD = 0.5
 
 
-def req_prefix_match_key(req) -> tuple[list[int] | None, str | None]:
+def req_prefix_match_key(
+    req, cache_input_ids: list[int] | None = None
+) -> tuple[list[int] | None, str | None]:
     """Effective ``(token_ids, extra_key)`` for a cache-affinity prefix probe.
 
     Mirrors the key the request will use for its *real* radix lookup, so the
@@ -36,7 +38,7 @@ def req_prefix_match_key(req) -> tuple[list[int] | None, str | None]:
     prefix (an unexpanded batch, an empty/one-token prompt, or a request whose
     reusable prefix clamps to zero), so the caller falls back to load balancing.
     """
-    input_ids = getattr(req, "cache_input_ids", None) or req.input_ids
+    input_ids = cache_input_ids or req.input_ids
     if not isinstance(input_ids, list) or not input_ids or not isinstance(input_ids[0], int):
         return None, None
 
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -478,7 +478,7 @@ class Req:
                 )
                 match_result = tree_cache.match_prefix(
                     MatchPrefixParams(
-                        key=RadixKey(self.adjust_max_prefix_ids(), self.extra_key, self.dp_rank),
+                        key=RadixKey(self.match_key_ids(), self.extra_key, self.dp_rank),
                         cow_recurrent=(
                             tree_cache.supports_recurrent() and not is_running_recurrent
                         ),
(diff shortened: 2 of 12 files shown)


zhengkezhou1 (approved):

python/sgl_jax/srt/managers/dp_schedule_policy.py
JamesBrianD, line 61: At this point, `req.cache_input_ids` has not yet been computed, which may affect the accuracy of cache-aware matching.

JamesBrianD:
The `offsets -> placeholder_ranges` migration looks incomplete. `qwen_vl.py` now writes `item.placeholder_ranges`, but the in-model Qwen2.5-VL path still reads `item.offsets` in `mm_utils.py` and `vision_metadata/qwen2_5_vl.py`.

With processor output like `placeholder_ranges=[(2, 4)]`, `item.offsets` will be empty/missing, so the embed plan may not scatter visual embeddings into tokens 2 and 3, and metadata row counting can be wrong.

Could we update these consumers to use half-open `placeholder_ranges` consistently?

python/sgl_jax/srt/managers/dp_schedule_policy.py
Yanko-7: > At this point, `req.cache_input_ids` has not yet been computed, which may affect the accuracy of cache-aware matching.

You are right, `req_prefix_match_key` occurs earlier than `handle_generate_request`, which makes it always None here. However, adding a `cache_input_ids` field in `TokenizedGenerateReqInput`, I think, does not align with its semantics. Therefore, in the new commit, I calculate the new `cache_input_ids` in `_select_cache_aware_dp` and pass them into the `req_prefix_match_key`.

Yanko-7:
> The `offsets -> placeholder_ranges` migration looks incomplete. `qwen_vl.py` now writes `item.placeholder_ranges`, but the in-model Qwen2.5-VL path still reads `item.offsets` in `mm_utils.py` and `vision_metadata/qwen2_5_vl.py`.
> 
> With processor output like `placeholder_ranges=[(2, 4)]`, `item.offsets` will be empty/missing, so the embed plan may not scatter visual embeddings into tokens 2 and 3, and metadata row counting can be wrong.
> 
> Could we update these consumers to use half-open `placeholder_ranges` consistently?



> The `offsets -> placeholder_ranges` migration looks incomplete. `qwen_vl.py` now writes `item.placeholder_ranges`, but the in-model Qwen2.5-VL path still reads `item.offsets` in `mm_utils.py` and `vision_metadata/qwen2_5_vl.py`.
> 
> With processor output like `placeholder_ranges=[(2, 4)]`, `item.offsets` will be empty/missing, so the embed plan may not scatter visual embeddings into tokens 2 and 3, and metadata row counting can be wrong.
> 
> Could we update these consumers to use half-open `placeholder_ranges` consistently?

The `mm_utils.py` file mainly contains helper functions for processing images by request rounds. Since the approach was later modified to a more advanced method of sharding images across TP, this file was deleted in the subsequent [PR#1480](https://github.com/sgl-project/sglang-jax/pull/1480). The migration from `offsets` to `placeholder_ranges` has been fully completed, but this needs to wait until all PRs are merged.

JamesBrianD (approved):
LGTM

PR #1480, opened 2026-07-24
feat(vlm): add model-agnostic in-model multimodal runtime

Yanko-7:
## Motivation

The previous in-model vision path was coupled to model-specific metadata and assumed one data-leading encoder lane per DP rank. That made it difficult to add architectures with different patch merging, attention metadata, deepstack behavior, or encoder parallelism. The old path also passed Python-side packed wrappers and placements between the encoder, merge, and cache, creating a second representation that had to be converted before the result could be consumed.

This PR delivers the shared **in-model multimodal runtime**. It defines a small model contract, packs encoder work across the device mesh, returns the final padded/item-ordered JAX array directly, merges it into the text stream, and optionally writes the same array into a paged device embedding cache. It also adds bucketed precompile and DP/TP vision sharding, wires Qwen2.5-VL to the existing TPU variable-length attention kernel introduced by [#1518](https://github.com/sgl-project/sglang-jax/pull/1518), and migrates **Qwen2.5-VL** as the reference adapter.

The varlen attention kernel implementation itself is not part of this PR.

Packed embedding and cache handling intentionally live here rather than in a later cleanup PR. Every model stacked above this PR implements the final array contract from its first commit.

| Order | PR | Scope | Status |
| --- | --- | --- | --- |
| 1 | #1479 | Cache-aware multimodal scheduling | Merged |
| 2 | **#1480** | Shared runtime, final packed/cache path, vision-attention integration, and Qwen2.5-VL | This PR |
| 3 | #1481 | Qwen3-VL, MiMo-V2, and Qwen3.5 adapters | Depends on this PR |
| 4 | #1482 | Gemma 4 support | Depends on #1481 |

All stacked branches target `epic/vlm`, so GitHub displays the cumulative stack. The logical review boundary for this PR is the sixteen commits after the merged #1479, ending with `chore(vlm): embed the deepstack dim into the hidden dim` (`7a3ade3`).

### Final in-model multimodal contract

`InModelMultimodalContract` gives the runtime four architecture-independent hooks:

- `get_input_embeddings()` exposes the text embedding function;
- `get_multimodal_encode_funcs()` registers an encoder per supported modality;
- `precompile_multimodal()` warms model-specific encoder buckets; and
- `get_multimodal_embedding_packed_capacities()` exposes the finite output capacities needed to warm the cache writer.

An encoder returns a two-dimensional JAX array with shape
`[capacity, (1 + deepstack_dim) * hidden]`. Valid item outputs are concatenated in input order; unused bucket rows remain padding. Primary embeddings and optional deepstack planes share the trailing feature dimension.

This is the only encoder-output representation used by the merge and the cache.

### Lane packing and output restoration

`multimodal/in_model/lane_packing.py` provides the reusable device planner:

- greedily balances variable-length items across encoder lanes;
- selects the smallest aligned input bucket, with a power-of-two fallback;
- constructs fixed-shape patch arrays with grid or packed 2D-position metadata;
- validates feature, metadata, placeholder, and merge-unit lengths before execution;
- supports both replicated-DP and tensor-parallel encoders;
- restores lane-major encoder output into the flat, item-ordered array consumed by the runtime.

Packing and output restoration are internal to the shared encoder adapter. Host orchestration therefore sees the same output layout regardless of how work was distributed over the mesh.

### Paged multimodal embedding cache

`EmbeddingPool` is a per-device, hash-addressed cache for encoded image, video, or audio items.

- Pages store primary and deepstack features contiguously as
  `[num_pages, page_size, (1 + deepstack_dim) * hidden]`.
- `write_packed()` consumes the encoder's flat JAX array directly together with item lengths; it does not repack or slice the result into Python objects.
- One donated JIT scatter writes all retained misses into the paged buffer.
- Cache hits are gathered before miss writes are allowed to evict pages used by the current batch.
- `write_mask` avoids retaining an item when the current chunk has already consumed its entire output.
- A zero MiB budget disables the pool; positive budgets are bounded and LRU-evicted.

The same flat gather primitive overlays either freshly encoded rows or cached page rows into token embeddings, including optional deepstack planes.

### Host orchestration and chunk-local merging

`build_multimodal_batch()` converts the placeholders visible in the current EXTEND or MIXED prefill chunk into `ItemTask`s. Each task retains the complete encoder-output length and one or more chunk-local source/destination mappings, so a single item can be consumed correctly across chunk boundaries and DP ranks.

`embed_multimodal_inputs()` then:

1. materializes the text embeddings;
2. groups work by modality;
3. resolves cache hits and misses by item hash;
4. runs each model-provided encoder only for misses;
5. gathers the flat item-ordered rows into placeholder positions; and
6. writes reusable misses back to the embedding pool.

<div align="center">
  In-model multimodal runtime architecture
  <br />
  <sub><i>Shared in-model multimodal runtime</i></sub>
</div>

### Vision encoder sharding

`VisionShardSpecs` centralizes the batch, column-parallel, row-parallel, and attention-head layouts used by vision towers. `resolve_encoder_tp()` selects the encoder mode from the mesh and `--vision-encoder-parallel`:

- `dp`: replicate vision weights over the tensor axis and use those devices as additional encoder lanes;
- `tp`: shard vision weights and attention heads over the tensor axis.

Both modes return the identical flat item-ordered array, so scheduling, merging, caching, and model code above the encoder do not branch on the parallel mode.

### Replication trade-off

The restored encoder output and the `EmbeddingPool` buffer are intentionally fully replicated across the model mesh in this PR.

The runtime currently accepts a flat item list without carrying explicit DP ownership. Lane packing may assign items to any compatible encoder lane, after which replication provides a single item-ordered two-dimensional output for merge and cache writes. The cache likewise uses one host-side LRU and page namespace, shared by all DP ranks. This keeps the model contract independent of encoder DP/TP mode and keeps bucketed JIT/precompile signatures simple.

The trade-off is a data-axis collective when restoring encoder output and a fully replicated embedding-cache allocation on every device. Consequently, `mm_embedding_cache_size_mb` and the derived `embedding_pool_bytes` are per-device budgets and are reserved in full from each device's static-memory
budget.

Supporting a DP-isolated representation would require a different input contract that preserves DP ownership throughout packing, restoration, merging, and cache management. The replicated representation is therefore an
intentional part of the current contract, not an incidental missing sharding annotation.

### Integration with the existing TPU varlen attention kernel

The TPU-native `varlen_attention` operator is provided by [#1518](https://github.com/sgl-project/sglang-jax/pull/1518) and is not implemented or modified in this PR.

This PR adds only the VLM-side integration:

- `VisionVarlenAttentionBackend` adapts batch-sharded vision Q/K/V and `VisionAttentionMetadata` to the existing kernel interface;
- Qwen2.5-VL supplies the windowed and full-attention sequence boundaries;
- TPU execution selects the existing varlen backend, while CPU integration tests continue through the reference-compatible flash backend; and
- host-computed maximum sequence lengths drive integration-side block-size selection through the kernel's public interface.

`VisionAttentionMetadata` is a registered pytree carrying the per-lane sequence boundaries through `jax.jit` and `shard_map`. This replaces the previous packed/ragged metadata path and removes the obsolete reference backend. Kernel design, correctness validation, TPU benchmarks, and kernel-level tuning belong to #1518.

### Qwen2.5-VL reference adapter

Qwen2.5-VL is migrated onto the final contract and wired to the existing varlen attention backend:

- the shared `run_mrope_vision_model()` path packs patches across lanes, calls the vision tower, and restores flat item-ordered embeddings;
- `_build_metadata()` produces window reorder/inverse indices, rotary positions, and shared `VisionAttentionMetadata` for the window/full-attention layouts;
- blocks select the correct attention layout through `fullatt_block_indexes`;
- image/video features support both DP-encoder and TP-encoder modes; and
- output capacities are exposed for encoder and cache-writer warmup.

### Runtime-equivalent precompile

The compilation manager invokes shared host-orchestration helpers to warm multimodal work through the same paths used by real requests:

- text-only and multimodal-input EXTEND variants are compiled separately;
- each model warms its configured vision patch buckets through the shared vision runner;
- flat encoder output capacities are combined with reachable token buckets for merge warmup; and
- the embedding-pool writer is precompiled once per distinct capacity without changing LRU state.

The runtime planner and precompile share `--precompile-vision-patch-paddings`; architecture adapters derive their own merge units and output capacities rather than embedding Qwen-specific formulas in the manager.

### Runtime and scheduler integration

The model runner owns the optional embedding pool and invokes host orchestration before the language model. The scheduler, TP worker, overlap thread, `ForwardBatch`, compilation manager, and cache sizing paths carry the data required by the new runtime. The superseded `mm_utils`, `mm_plan`, and model-specific `vision_metadata` path is removed. Retracted decode requests continue mRoPE positions beyond the original multimodal prompt when they are re-prefilled.

New or consolidated server options:

```text
--vision-encoder-parallel {dp,tp}
--precompile-vision-patch-paddings ...
--mm-embedding-cache-size-mb ...
--mm-embedding-page-size ...
```

Contract detection controls in-model precompile and keeps radix cache and chunked prefill enabled; mixed-chunk support remains architecture-gated.

## How It Works

For each prefill chunk, the scheduler builds item tasks from the visible placeholder ranges. The runtime looks up those items in the embedding pool and sends only misses to the model's modality encoder. The shared adapter may lane-pack work internally, but restores its result to a fixed-capacity array whose valid rows are ordered exactly like the input items.

The merge builds one token-sized index/mask vector and gathers primary and deepstack features from that array directly into the text embeddings. Misses that still have an unconsumed tail are written from the same array into the page pool. Later chunks or requests gather cache hits from the pool with the same overlay kernel.

For Qwen2.5-VL, windowed and full vision-attention blocks use the existing varlen kernel from #1518. This PR supplies the adapter, metadata, and sharding integration; it does not add kernel implementation code. `VisionShardSpecs` determines whether vision weights are replicated for DP encoding or sharded for TP encoding; the runtime contract above the tower is unchanged.

<div align="center">
  Complete process
  <br />
  <sub><i>Workflow</i></sub>
</div>

## Testing

```bash
python -m pytest -q \
  python/sgl_jax/test/test_vlm_stage1_framework.py \
  python/sgl_jax/test/test_embedding_pool.py \
  python/sgl_jax/test/multimodal/test_flash_attention_kernel.py \
  python/sgl_jax/test/multimodal/test_qwen_vl_processor.py
```

Kernel-level varlen attention tests are owned by #1518 and are not repeated in this PR-specific test list.

The complete changed range also passes:

```bash
pre-commit run --from-ref epic/vlm --to-ref HEAD
```

## Review Notes

- The final flat packed-array and direct cache-write design is part of this PR; later model PRs do not introduce a cross-model packing/cache refactor.
- The varlen kernel implementation, accuracy validation, and performance results belong to [#1518](https://github.com/sgl-project/sglang-jax/pull/1518) and are outside this diff. In this PR, review only the VLM-side backend selection, metadata plumbing, and Qwen2.5-VL integration.
- Review the sixteen commits after #1479 rather than treating the cumulative stacked diff as one model change.
- Cache keys are complete multimodal item hashes, not token-prefix keys.
- True item lengths affect host mappings and page allocation, while JIT shapes remain bounded by encoder and token buckets.
- Dummy precompile inputs overlay zero-valued packed or pool rows only to compile runtime-equivalent gather and forward variants; pool-writer warmup does not alter LRU state.

## Checklist

- [x] The PR description and code are in English.
- [x] The final runtime, packed-output, and cache contracts are documented.
- [x] The stack position and logical review boundary are documented.
- [x] The test plan contains reproducible commands.
- [x] New command-line options are exposed through `--help`.


python/sgl_jax/srt/models/qwen2_5_vl.py
```diff
@@ -452,111 +419,209 @@ def __init__(
...
+            grid_thw = np.zeros((num_lanes, 1, 3), dtype=np.int32)
+            grid_thw[0, 0] = grid
+            jax.block_until_ready(self.encode(patches, grid_thw))
+
+    def _build_metadata(
+        self,
+        lane_grids: np.ndarray | jax.Array,
+        capacity: int,
+    ) -> tuple[np.ndarray, np.ndarray, VisionAttentionMetadata, VisionAttentionMetadata]:
+        lane_grids = np.asarray(jax.device_get(lane_grids), dtype=np.int32)
```
lianfang0320: grid_thw is only used by the host-side metadata planning path. _build_metadata() lowers it into indices, position_ids, and the window/full-attention cu_seqlens; the device-side _encode_jit() consumes only those derived arrays and never reads grid_thw directly. Therefore, grid_thw does not need to be placed on the device.
Given this boundary, accepting jax.Array and calling jax.device_get() here is redundant. Current production callers already provide a host-resident np.ndarray; if a future caller passes a device-resident array, this code would unnecessarily copy it back to the host and potentially introduce a synchronization point.
Could we make the host-only contract explicit by changing grid_thw/lane_grids to np.ndarray and removing jax.device_get()?
```
def _build_metadata(
    self,
    lane_grids: np.ndarray,
    capacity: int,
) -> ...:
    lane_grids = np.asarray(lane_grids, dtype=np.int32)
```
The same type narrowing should also be applied to the public encode() method.

python/sgl_jax/srt/managers/schedule_batch.py
```diff
@@ -3046,18 +3037,16 @@ def get_model_worker_batch(
         mrope_positions = _mm["mrope_positions"]
         apply_for_deepstack = _mm["apply_for_deepstack"]
         deepstack_visual_embedding = _mm["deepstack_visual_embedding"]
-        # Build the vision encode/merge plan only for ordinary prefill. Other
-        # forward modes leave the plan unset, and the runner treats a non-None
-        # plan as the sole vision-forward signal.
-        if self.contains_mm_inputs() and self.forward_mode == ForwardMode.EXTEND:
-            mm_embed_plan = build_mm_embed_plan(
+        # Keep items whose placeholder rows intersect the current extend window.
+        if self.forward_mode == ForwardMode.EXTEND:
```
lianfang0320: multimodal_batch is built only for EXTEND, but Qwen2.5-VL supports MIXED, which may also contain multimodal prefill tokens. This can skip vision encoding and embedding merge for mixed batches.
Please handle both EXTEND and MIXED here and add a mixed-chunk multimodal test.

python/sgl_jax/srt/multimodal/in_model/lane_packing.py
```diff
@@ -0,0 +1,217 @@
...
+    valid = np.zeros(num_lanes, dtype=np.int32)
+    output_cap = cap // merge_unit
+    output_starts = np.zeros(len(items), dtype=np.int32)
+
+    for lane_index, lane in enumerate(lanes):
+        input_offset = 0
+        output_offset = 0
+        for item_index in lane:
+            feature = features_np[item_index]
+            end = input_offset + feature.shape[0]
```
lianfang0320, line 124: The patch count is not validated end to end. feature.shape[0], prod(grid_thw), and sum(placeholder lengths) * merge_unit must match; otherwise item boundaries can shift and embeddings may be merged into the wrong tokens.
Please validate these invariants before lane packing.

python/sgl_jax/srt/managers/schedule_batch.py
Yanko-7: Done.

python/sgl_jax/srt/multimodal/in_model/lane_packing.py
Yanko-7: > The patch count is not validated end to end. feature.shape[0], prod(grid_thw), and sum(placeholder lengths) * merge_unit must match; otherwise item boundaries can shift and embeddings may be merged into the wrong tokens. Please validate these invariants before lane packing.

Done.

python/sgl_jax/srt/models/qwen2_5_vl.py
Yanko-7: > grid_thw is only used by the host-side metadata planning path. _build_metadata() lowers it into indices, position_ids, and the window/full-attention cu_seqlens; the device-side _encode_jit() consumes only those derived arrays and never reads grid_thw directly. Therefore, grid_thw does not need to be placed on the device. Given this boundary, accepting jax.Array and calling jax.device_get() here is redundant. Current production callers already provide a host-resident np.ndarray; if a future caller passes a device-resident array, this code would unnecessarily copy it back to the host and potentially introduce a synchronization point. Could we make the host-only contract explicit by changing grid_thw/lane_grids to np.ndarray and removing jax.device_get()?
> 
> ```
> def _build_metadata(
>     self,
>     lane_grids: np.ndarray,
>     capacity: int,
> ) -> ...:
>     lane_grids = np.asarray(lane_grids, dtype=np.int32)
> ```
> 
> The same type narrowing should also be applied to the public encode() method.

Done.

Issue #1118, opened 2026-05-19
[CI] Performance Testing Strategy


Rodrian7:
## Objective

Define the performance testing strategy for SGLang-Jax — which perf tests are Required vs Advisory in PR Gate Stage 3, accuracy/perf baseline thresholds, and baseline update mechanism.

## Decisions Needed

1. **PR Gate Stage 3 — Required vs Advisory**: Which `performance-test-*` jobs must pass to merge (Required), and which only report warnings (Advisory)?
2. **Accuracy baseline thresholds**: Per-model `tolerance_abs` values for accuracy metrics (MMLU, etc.) — how much drift is acceptable before blocking release?
3. **Perf baseline thresholds**: Per-job `tolerance_pct` values for throughput/latency — what regression percentage triggers a block?
4. **Baseline update mechanism**: Auto-update on each release? Manual `[update-baseline]` PR? Hybrid? Who approves baseline changes?
5. **Baseline storage format**: `ci/baseline.json` schema design — model × metric × commit metadata
6. **PR-level perf diff**: When `test:perf` label is applied, how should Advisory warnings surface (PR comment? Check annotation? Both?)

## Blocked Tasks

These CI improvements are blocked pending decisions from this issue:
- **P1-10**: Numerical baseline + accuracy/perf thresholds (`ci/baseline.json` + `baseline_diff.py`)

## Acceptance Criteria

1. Each decision above has a documented resolution with rationale
2. `ci/baseline.json` schema finalized and initial baselines populated
3. Required vs Advisory classification for all perf/accuracy jobs documented
4. Baseline update workflow defined and automated

PR #1484, opened 2026-07-27
perf(scheduler): env-gated per-step phase timing + #293 masking guard

Yajingleo:
## What
Adds env-gated per-forward-step timing of the scheduler overlap loop, split into `recv / schedule / build (get_model_worker_batch + dispatch) / result (forward-wait + D2H)`, plus a distribution analyzer and a masking regression guard.

- `SGLANG_SCHED_PHASE_TIME=1` → logs a running-average `[sched/fwd-step]` breakdown.
- `SGLANG_SCHED_PHASE_CSV=<path>` → also dumps one row per forward step for offline distribution analysis.

**Off by default** — when unset it's a couple of branch checks per loop, no syscalls, zero measurable overhead.

## Why (relates to #293)
#293 asked whether the scheduler thread is the throughput bottleneck. This makes it **measurable**. On current `main` (Qwen3-8B, tp=4, v6e-4):

- Single point (conc 16, 1024/1024): HOST ≈ **2.6–2.9 ms** (dominated by `build` ≈ 2.5 ms) vs forward ≈ **5.7–5.9 ms** → `host/forward ≈ 0.46`, **masked**. Throughput ≈ 2,450 tok/s confirms step ≈ forward, not host+forward.
- Distribution over a **varied** workload (`--random-range-ratio 0.5`, 10,350 forward steps): `host/forward` mean **0.449**, p90 0.486, **p99 0.523**, **100% of steps masked**.

So the scheduler is masked under the forward with ~2× headroom. Contributes a concrete performance check toward #1118.

## Files
| File | Purpose |
|------|---------|
| `python/sgl_jax/srt/managers/scheduler.py` | the env-gated instrumentation (only prod change) |
| `scripts/summarize_phase.py` | distribution (mean/p50/p90/p99 + histogram) **and** guard — exits non-zero if `p99 ≥ margin` |
| `scripts/eval_scheduler_masking.sky.yaml` | managed-spot v6e-4 job: varied workload → CSV → summarize; job status = verdict |
| `scripts/check_sched_masked.sh` | self-contained bash guard (no SkyPilot; runs directly on a TPU host) |

## Testing
Verified end-to-end on v6e-4 (SkyPilot managed spot): guard PASS, distribution as above. The guard gates on `p99 < margin` (default 0.85), robust to the lone batch-drain step where `result ≈ 0`.

Relates to #293, #1118.

### DIFF
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -114,6 +114,16 @@ TEST_RETRACT = get_bool_env_var("SGLANG_TEST_RETRACT")
 TEST_RETRACT_INTERVAL = int(os.environ.get("SGLANG_TEST_RETRACT_INTERVAL", "3"))
 TEST_RETRACT_NO_PREFILL_BS = int(os.environ.get("SGLANG_TEST_RETRACT_NO_PREFILL_BS", str(2**31)))
 RECORD_STEP_TIME = get_bool_env_var("SGLANG_RECORD_STEP_TIME")
+# Debug: log per-phase scheduler-loop timing (recv / schedule / launch / result)
+# as running averages, to see what fraction of the step the scheduler thread
+# actually spends where. Off by default (zero prod overhead).
+SCHED_PHASE_TIME = get_bool_env_var("SGLANG_SCHED_PHASE_TIME")
+SCHED_PHASE_INTERVAL = int(os.environ.get("SGLANG_SCHED_PHASE_INTERVAL", "200"))
+# Optional: dump one row per forward step (step,bs,recv,sched,build,result in ms)
+# to this CSV path, so the host/forward ratio distribution (mean + percentiles)
+# can be computed offline. Implies phase timing. Off by default.
+SCHED_PHASE_CSV = os.environ.get("SGLANG_SCHED_PHASE_CSV")
+SCHED_PHASE_TIME = SCHED_PHASE_TIME or bool(SCHED_PHASE_CSV)
 GRAMMAR_TIMEOUT = float(os.environ.get("SGLANG_GRAMMAR_TIMEOUT", 300))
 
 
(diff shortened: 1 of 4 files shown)


JamesBrianD:
Thanks for looking into this.

Issue #293 was created quite some time ago. Since then, we have already optimized several CPU-side operations in the scheduler, so the original profiling results no longer accurately reflect the current bottlenecks.

Based on our more recent profiling, the main remaining overheads are:

1. The `device_put` operations when converting a `ScheduleBatch` into a `ForwardBatch`.
2. The overhead of launching the TPU JIT execution.

If you would like to continue working on scheduler performance, could you first profile the current `main` branch with a representative workload, and then investigate possible ways to reduce the `device_put` and TPU launch overheads?

The changes in this PR introduce additional instrumentation and scripts while making the scheduler code more complex and harder to maintain. Since they do not directly address the current bottlenecks, we do not plan to merge this PR in its current form.

Yajingleo:
Thanks @JamesBrianD — that makes sense, and it lines up with what I found: the original #293 condition no longer holds. On current `main` the scheduler's per-step host work (~2.6 ms, of which `build`/`ForwardBatch.init_new` is the bulk) stays masked under the ~5.9 ms forward (`host/forward ≈ 0.46`, p99 0.52 over 10k steps). Detailed numbers are on #293.

I agree the instrumentation adds maintenance surface without addressing the current bottlenecks, so I'll close this PR.

I'll follow your suggestion and profile current `main` with a representative workload, focused on the `device_put` (`ScheduleBatch → ForwardBatch`) and TPU JIT launch overhead. As a first step I'll capture a JAX profiler trace to quantify **device-idle time between forwards** — since if the device is already back-to-back, reducing masked host work won't move throughput, whereas launch bubbles would. That'll tell us which regimes (DP / concurrency) the two overheads actually bite in, and bound the achievable gain before proposing any change. Will follow up with findings.

Thanks for the clear pointer.

Issue #1486, opened 2026-07-27
[Bug] Pathways chunked-prefill abort/retract is not tracked across the async P-to-D pipeline


Rodrian7:
### Describe the bug

Pathways PD uses a separate asynchronous prefill lifecycle from the standard scheduler. A request can be owned by `p_chunked_reqs[p_idx][dp_rank]` or be present in a P prefill queue, the prefill thread's local stash, the ready queue, the defer list, or the final P-to-D migration stage.

The generic `abort_request()` path does not inspect those P-side states. In particular, once the final chunk is scheduled, the per-P chunked owner is cleared while `_pd_inflight_rids` retains only the rid, so the scheduler no longer has a request object or `p_idx` with which to mark or finalize the abort.

This is why #1471 deliberately excludes `pd == "pathways"` from its standard-scheduler chunked-prefill cleanup. Removing that guard would be unsafe: the generic finalizer operates on the currently selected scheduler pool, while an in-flight Pathways chunk may still own KV and a request slot on a specific P slice.

### Expected behavior

Pathways should provide the same externally visible lifecycle guarantees:

- An aborted chunked-prefill request must not schedule another chunk.
- Already queued or dispatched P batches must be drained before their resources are released.
- A final chunk that is already in flight must be abortable even after its per-P owner has been cleared.
- Returned logprob spans must still advance the shared cursor, without retaining partial logprobs for the aborted request.
- Precision-trace completion and terminal output must occur exactly once.
- P-side KV, request slots, and D-admission reservations must be released exactly once.
- An aborted final prefill must not remain blocked by D-side capacity.
- Retract/pause must drain outstanding P results before resetting and requeuing requests.
- Non-aborted P-to-D migration behavior must remain unchanged.

### Lifecycle boundaries that need handling

1. Parked per-P chunked owner with no outstanding result.
2. Intermediate chunk in the P queue, prefill-thread stash, forward, ready queue, or defer list.
3. Final chunk after `p_chunked_reqs[p_idx][dp_rank]` has become `None`.
4. A gathered batch containing both live and subsequently aborted requests.
5. Requests already migrated to a D running batch.
6. Pathways pause/retract while P or D results remain in flight.

For mixed gathered batches, the D scatter layout must stay aligned with the original P gather. Aborted page positions cannot simply be removed from `done_per_req`; they need safe sentinel destinations while live requests retain their original positions.

### Suggested implementation direction

[...]

PR #1490, opened 2026-07-28
feat(pd): add optional Raiden-backed KV transfer engine

cjx0709:
## Summary

This PR adds an opt-in [tpu-raiden](https://github.com/google/tpu-raiden) backend for P/D KV transfer. The existing JAX transfer backend remains the default.

This PR transfers KV after request-level prefill completion. Incremental chunked-prefill KV transfer is out of scope; #1496 tracks the existing backend-neutral PD continuation issue and its integration with future chunk-transfer support.

- Introduce a scheduler-facing `TransferBackend` contract and select JAX or Raiden in the backend factory, keeping engine-specific branches out of prefill/decode scheduling.
- Transfer directly from prefill KV pages into decode-owned pages with Raiden.
- Preserve Raiden endpoint/shard metadata through the existing bootstrap service.
- Keep cancelled or timed-out transfers alive until Raiden reports a terminal state, then release request pages and slots.
- Validate incompatible configurations and preload the Raiden native extension before JAX initialization.

Enable Raiden on both workers:

```bash
python -m sgl_jax.launch_server ... \
  --disaggregation-mode prefill \
  --disable-radix-cache \
  --disaggregation-use-raiden

python -m sgl_jax.launch_server ... \
  --disaggregation-mode decode \
  --disable-radix-cache \
  --disaggregation-use-raiden
```

## Raiden build and cache

The validated dependency set is JAX/JAXLIB 0.10.2, libtpu 0.0.42.1, and tpu-raiden `87564793fda8269164e670fa4c4a44548840e153`.

```bash
git clone https://github.com/google/tpu-raiden.git ~/src/tpu-raiden
git -C ~/src/tpu-raiden checkout 87564793fda8269164e670fa4c4a44548840e153

export RAIDEN_SRC=~/src/tpu-raiden
export RAIDEN_CACHE_ROOT=/models/raiden-cache/tpu-raiden-jax
export RAIDEN_COMMIT=87564793fda8269164e670fa4c4a44548840e153
export JAX_VERSION=0.10.2 JAXLIB_VERSION=0.10.2 LIBTPU_VERSION=0.0.42.1
scripts/disaggregation/build_raiden_wheel.sh

wheel_dir="${RAIDEN_CACHE_ROOT}/${RAIDEN_COMMIT}-jax${JAX_VERSION}-jaxlib${JAXLIB_VERSION}-libtpu${LIBTPU_VERSION}"
(cd "${wheel_dir}" && sha256sum -c SHA256SUMS)
wheel=$(<"${wheel_dir}/READY")
python3.12 -m pip install --no-deps "${wheel_dir}/${wheel}"
```

The source checkout stays under `~/src`; only the reusable wheel cache is stored under `/models`.

## Validation

Local tests:

- `PYTHONPATH="$PWD/python" python -m pytest test/srt/disaggregation -q`: `379 passed, 12 skipped`
- Pre-commit checks passed on all changed files.
- Pathways PD was reviewed statically and remains on its independent `PathwaysPDKVTransfer` path.

GKE validation used two physical TPU v6e hosts. Each serving worker ran as one JAX process (`jax.process_count() == 1`) restricted to one serving device (`--device-indexes 0`, TP=1), using `DeepSeek-R1-Distill-Qwen-1.5B`. JAX and Raiden ran sequentially on the same hosts; each workload had a warm-up request, and the Raiden phase reused the persistent JAX compilation cache. Multi-process JAX slices (`jax.process_count() > 1`) were not validated in this PR.

TTFT is reported only for isolated requests (`4096 input / 1 output`, concurrency 1), so request queueing does not hide or inflate the KV-transfer contribution. Five measured requests followed an engine-specific warm-up:

| Metric | JAX transfer | Raiden | Change |
| --- | ---: | ---: | ---: |
| Mean TTFT | 167.31 ms | 124.36 ms | -25.7% |
| Median TTFT | 166.36 ms | 123.91 ms | -25.5% |

GSM8K correctness used the first 200 test examples, 5-shot prompting, temperature 0, and 512 output tokens:

| Engine | Concurrency 1 | Concurrency 8 | Invalid outputs |
| --- | ---: | ---: | ---: |
| JAX transfer | 66.5% | 63.5% | 0 |
| Raiden | 66.5% | 63.5% | 0 |

JAX and Raiden matched on 200/200 predictions at concurrency 1 and 199/200 at concurrency 8.

Abort recovery was exercised with 32 concurrent `4096/256` requests. Raiden interrupted 8 requests with no errors; all transfer queues drained, KV/request-pool availability returned to the pre-test baseline, 16/16 post-abort requests completed, and the concurrency-8 GSM8K run then passed.

A focused single-process v6e API-contract probe against the pinned Raiden wheel additionally verified that non-empty `register_read` calls return `True` even when more requests are registered than `num_slots`; sender timeout and staging-slot rejection both drain through `done_sending`; a forced one-slot collision surfaces as `failed_recving`; and the following transfer reuses the slot with an exact source/destination block match. This confirms that `done_sending` is a sender terminal/release signal, while receiver state determines end-to-end transfer success.

## Checklist

- [x] The PR is written in English.
- [x] The motivation and scope are documented.
- [x] Correctness, API-contract, abort-recovery, and GKE benchmark results are included.
- [x] Build, cache, and launch instructions are included.


cjx0709:
Hello @zorrofox,  I import tpu-raiden as the kv transfer engine. If you have time, could you please review this?

zorrofox (changes requested):
Thanks — the backend abstraction is clean (the `TransferBackend` contract, factory gating, and the `commit()` seam keep the default JAX path at zero behavior change; I verified the release-timing refactor in `prefill.py` is behaviorally equivalent), and the direction is clearly worthwhile given the gather+staging+scatter steps it removes. One blocking issue though, plus a few things I'd want addressed before this lands.

**Blocking — abort/timeout page lifetime on the zero-copy path (silent KV cross-contamination).** Neither `RaidenTransferKVSender.fail()` nor `RaidenTransferKVReceiver.fail()` cancels or deregisters anything engine-side (the wrapper exposes no cancel API), while the scheduler releases pages immediately on abort:

- Prefill side: abort → `_release_prefill_req_resources` → pages go back to the allocator while the engine's registered read slot is still live (only its own `timeout_s` expires it). A reallocated page can then be read by the decode side — silently serving another request's KV. Also `ack_timeout` and the engine `timeout_s` (=`pull_timeout`) are independent flags; safety currently rests on their default values happening to line up.
- Decode side (worse): transfer-queue abort calls `receiver.abort()` then immediately `_release_decode_kv_indices`. The JAX backend is safe because pulls land in a staging buffer and only `commit()` touches the pool; Raiden writes pool pages directly and `abort()` can't stop an in-flight read, so the engine keeps writing into pages the allocator may have already handed to another request.

A zero-copy backend needs "pages are not returned until the engine confirms terminal": either a cancel/deregister call wired into `fail()`, or a quarantine list that holds pages until `poll_stats` reports failed/timed-out. The test suite's `unsafe_skip_buffer_lock=True` assertion is asserting exactly the setting this concern is about, and nothing in the PR explains why skipping the lock is safe. I'd also want an abort-under-load run (aborts issued while transfers are in flight, then correctness-check subsequent requests) as evidence — this class of lifecycle bug doesn't show up in happy-path smoke tests (see #1486 for the same class of issue on the Pathways side).

**Should fix before merge:**

- The native-extension preload only reaches the scheduler subprocess via CPython's `multiprocessing.spawn` re-import of `__mp_main__` — undocumented, untested, and it silently breaks for any entrypoint other than `python -m sgl_jax.launch_server` (Engine API users get a late RuntimeError deep in scheduler init). A comment + a spawn-preload test, or an explicit "launch_server only" error message in the factory.
- Multi-host isn't actually supported: `try_start_decode` synthesizes remote endpoints as `peer_host:base_port+index`, assuming all prefill engines share one host with consecutive ports, and drops the endpoints the peer actually published. Either use the published endpoint strings or add a `jax.process_count() > 1` rejection next to the existing device check.
- `jax.effects_barrier()` in `start_prefill` doesn't guarantee the paged-KV writes are materialized (they're donated-array outputs, not ordered effects; the JAX backend got this implicitly from its gather). If tpu-raiden's read path synchronizes with the TPU stream, cite that; otherwise block on the request's KV buffers explicitly. It's also a global barrier inside a per-request loop, which will serialize the pipeline under overlap scheduling.
- Radix cache + Raiden should be rejected in the factory (the smoke script disables radix, but nothing enforces it; released-into-tree pages can be evicted/reused while an engine slot is still live).
- Now that #1471 has landed, this needs a rebase over the chunked-prefill ownership changes in `prefill.py`/`decode.py`; the release-timing refactor is worth re-verifying against that base.

**Smaller items:** `register_read`'s return value is dropped (no prefill-side slot admission; slot exhaustion behavior undefined); `cleanup_transfer` in the decode abort path uses the decode process's own `jax_process_index` while the metadata was registered under the prefill's (the `receiver.fail()` path gets this right — and the kwarg extension also drifts from the `base/transfer.py` protocol signature); the page-derivation `kv_indices[::page_size] // page_size` silently assumes page-aligned contiguous allocations — worth a fail-fast assertion on both sides since it's now correctness-critical rather than debug-only; ~150 lines of load-bearing comments/docstrings in `jax_transfer/wrapper.py`, `conn.py` and `prefill.py` are deleted without functional need — please restore; old-bootstrap + raiden-worker yields a per-request 404 failure storm instead of a startup capability probe; `commit()` being a no-op for Raiden means transfer-bytes metrics and the PD-KV debug verifier never run on this backend — exactly the tooling one would need to debug the blocking-2 class of issue.

Happy to re-review once the abort/timeout lifetime story is in place and the branch is rebased onto current main.

cjx0709:
@zorrofox Thanks for the detailed review. The branch is now rebased onto current `main` (including #1471), and I have addressed the requested items:

- **Abort/timeout lifetime:** Raiden sender/receiver aborts now enter a pending-failure state and keep direct-HBM pages and request slots owned until Raiden reports terminal. Terminal callbacks perform the release; the shared reaper handles timeouts/orphans, and cleanup is idempotent.
- **Startup and compatibility:** `launch_server` preloads the native extension before JAX. Engine API usage now fails early with an explicit preload instruction. Raiden also probes bootstrap capabilities and rejects radix-cache, D2H, non-TPU, and invalid inflight configurations.
- **Topology and synchronization:** decode consumes the endpoints/shards published by the peer instead of synthesizing ports, validates endpoint topology, and prefill uses `jax.block_until_ready(kv_buffers)` before registering pages.
- **Admission and cleanup:** `register_read` admission is honored, transfer cleanup uses the prefill process identity from metadata, and page derivation has fail-fast alignment/contiguity validation.
- **Observability:** Raiden's `commit()` path now records transferred bytes and runs the direct-receive debug verifier.
- **Tests:** the disaggregation suite passes (`379 passed, 12 skipped`), all current PR CI checks are green, and a two-host TPU v6e abort-under-load run interrupted 8 of 32 concurrent transfers without errors. All queues drained, allocator availability returned to baseline, 16/16 post-abort requests completed, and the subsequent concurrency-8 GSM8K run passed.

The PR intentionally remains request-granularity transfer after prefill completion. Incremental chunked-prefill transfer is out of scope; #1496 tracks the existing backend-neutral PD continuation/recomputation issue and the follow-up chunk-transfer work.

Could you please take another look?

zorrofox:
Re-reviewed the rebased branch — this is a thorough revision. I verified each item against the diff rather than the summary, and they all check out in code: the pending-failure state machine holds direct-HBM pages and request slots until the engine reports terminal (`request_abort` → `poll()`-gated `_finish`, with the reaper mapped onto the same path and idempotent cleanup, plus the three unit tests asserting exactly these semantics); scheduler abort no longer releases pages (`cancel_matching` + terminal-callback release on both sides); endpoints come from the peer's published metadata with topology validation; `block_until_ready` replaces `effects_barrier` outside the per-request loop; `register_read`'s return value, the cleanup process identity, the page-alignment fail-fast, `commit()` metrics + the direct-receive verifier, the bootstrap capability probe, and the radix-cache rejection are all in place. The two-host abort-under-load run is exactly the evidence I was hoping for. Nice work.

Two remaining questions, both about tpu-raiden API semantics that I can't verify from this repo — they're the foundation the lifecycle design rests on, so I'd like them pinned down (a doc/source citation in a comment is enough):

1. **Does `poll_stats` guarantee that a failed or timed-out send eventually appears in `done_sending`?** The sender's terminal check only consults `done_sending`; if the engine reports such slots only via a separate failed/expired list, an aborted sender would stay TRANSFERRING forever and its pages plus an inflight slot would leak (leak-safe rather than corrupting, but permanent under churn). If the guarantee doesn't hold, a bounded fallback (force-FAIL after N× `timeout_s` with the pages quarantined rather than released) would close it.
2. **What does `register_read` returning `False` mean exactly — "transfer not needed" or "no slot available"?** The current handling treats it as success without publishing metadata; if it can also mean slot exhaustion, a burst beyond `num_slots` on the prefill side would turn into silent per-request failures that only surface as decode-side metadata timeouts.

Non-blocking notes: an aborted prefill request on the Raiden path now waits up to `pull_timeout` before the client sees FINISH_ABORT (fine for disconnects, noticeable for abort_all/pause flows — maybe worth a line in the flag help); the rewritten `jax_transfer/wrapper.py` docs kept the load-bearing contracts but dropped a couple of actionable hints (e.g. the per-host-shard pointer in the non-addressable-array error); and multi-host multi-process Raiden still has no e2e evidence — worth stating the validated scope in the PR description.

Happy to approve once the two API-semantics questions are answered.

cjx0709:
@zorrofox Thanks — I pinned both API contracts against tpu-raiden `87564793fda8269164e670fa4c4a44548840e153` and added concise source references in https://github.com/sgl-project/sglang-jax/commit/fe48c6ff3f46c83157d439d809a037e4833b3868.

1. **Failed/timed-out sends do terminalize through `done_sending`.** The JAX API exposes only `(done_sending, done_recving, failed_recving)` ([API](https://github.com/google/tpu-raiden/blob/87564793fda8269164e670fa4c4a44548840e153/tpu_raiden/api/jax/kv_cache_manager.py#L156-L163)). On the sender, `CompleteReadRaw` moves expired send entries to `done_sending`, releases their slot, and erases the entry ([source](https://github.com/google/tpu-raiden/blob/87564793fda8269164e670fa4c4a44548840e153/tpu_raiden/core/kv_cache_manager_with_transfer.cc#L1335-L1355)); staging-slot, D2H, and H2H failures use the same terminal channel ([source](https://github.com/google/tpu-raiden/blob/87564793fda8269164e670fa4c4a44548840e153/tpu_raiden/core/kv_cache_manager_with_transfer.cc#L1900-L1920), [source](https://github.com/google/tpu-raiden/blob/87564793fda8269164e670fa4c4a44548840e153/tpu_raiden/core/kv_cache_manager_with_transfer.cc#L1954-L2051)). A focused v6e probe with the pinned wheel also registered two sends with no receiver; both appeared in `done_sending` after the configured 1.5 s timeout. Therefore the PR's pending-failure sender remains owned until a real engine terminal and then finishes FAILED; no additional bounded fallback is needed while the manager remains alive.

2. **`register_read(False)` means “nothing remains to transfer,” not “no slot available.”** The public contract states exactly that ([API](https://github.com/google/tpu-raiden/blob/87564793fda8269164e670fa4c4a44548840e153/tpu_raiden/api/jax/kv_cache_manager.py#L123-L135)). `NotifyForRead` returns 0 only for empty blocks or an already-arrived ACK; otherwise it records the send and returns the UUID without consulting slot capacity ([source](https://github.com/google/tpu-raiden/blob/87564793fda8269164e670fa4c4a44548840e153/tpu_raiden/core/kv_cache_manager_with_transfer.cc#L629-L669)). In the probe, two non-empty registrations both returned `True` with `num_slots=1`. Forced slot contention happened later: the producer terminalized via `done_sending`, the rejected receiver surfaced `failed_recving`, and a subsequent transfer reused the slot and matched the source block exactly. So the current `False -> transfer_not_needed` handling is correct; it is not an admission/backpressure result.

I also addressed the non-blocking notes in the same follow-up:
- the pull-timeout flag help now states that Raiden FINISH_ABORT visibility may wait for engine terminalization up to that timeout;
- the JAX wrapper again points non-addressable arrays to the per-host shard helper;
- the PR description now states the validated scope precisely: two physical v6e hosts, but one JAX process and TP=1 per worker; multi-process JAX slices were not validated.

The focused local tests passed (`82 passed`), as did all changed-file pre-commit checks.

zorrofox (approved):
Both contracts check out — I verified the cited API docstring and the `CompleteReadRaw` expiry path against the pinned commit, and the receiver-side `failed_recving` behavior for slot contention matches your probe. With the lifetime story grounded in engine-guaranteed terminalization and the scope statement added, this looks good to me. Approving.

Issue #1499, opened 2026-07-30
[Bug] FP8 block-quant linear weight_scale loses its declared sharding at load time — shard_map raises on jax 0.10.x


zorrofox:
### Symptom

Any static-quant checkpoint with compact 2D block-quant linear scales (e.g. GLM-5.2-FP8; DeepSeek-V3-class FP8 checkpoints share the format) fails at the first precompile/forward on the current jax 0.10.2 pin:

```
ValueError: in_specs passed to shard_map: P(None, None, 'tensor') does not match
the specs of the input: P(None, None, None) for arg: float32[16,1,16384].
```

Stack (elided): `models/glm5_moe.py` `self.q_b_proj(...)` → `layers/linear.py` `shard_map(...)` in `QuantizedLinear.__call__`.

Reproduced on current main (`a7353325`) with GLM-5.2-FP8, tp16/ep16, `--device=proxy` (both PD and the plain server take the same load path). The failing check is client-side jax `shard_map` in-spec validation, so native `--device=tpu` is expected to hit it as well; we are verifying that separately.

This never fired before because the combination was untested: jax 0.8.x `shard_map` silently reshards on a textual in-spec mismatch, so the same code runs (with a hidden relayout) on the previous 0.8.1 pin; and bf16 models never enter this code path at all.

### Root cause

`QuantizedLinear.from_linear` declares the scale placeholder with the kernel-boundary sharding, e.g. `NamedSharding(mesh, P(kernel_axes[0], None, kernel_axes[1]))` (`linear.py`).

At load time, `WeightLoader` shards the **compact 2D** scale per the weight mapping, then `_maybe_expand_linear_block_scale` (`weight_utils.py`) expands it 2D→3D via `expand_block_scale`. The expanded array inherits the 2D input's (effectively replicated) layout, and the subsequent `model_param.value = ...` assignment never reshards it to the placeholder's declared sharding. Result: the param reaches `shard_map` as `P(None, None, None)` instead of `P(None, None, 'tensor')`.

This is the same class of issue #1493 fixed at the MoE/MLA kernel boundaries ("restore explicit sharding at kernel boundaries", incl. the jax 0.10 strictness note in the removed FIXME) — the linear block-quant scale path was missed.

### Fix

One-line class of fix, mirroring #1493: reshard the expanded scale to the model placeholder's declared sharding before assignment, e.g. at the end of `_maybe_expand_linear_block_scale`:

```python
[...]

PR #1493, opened 2026-07-28
fix(moe): restore explicit sharding at kernel boundaries

JamesBrianD:
## Summary

- restore the requested target sharding after deferred stacked-MoE weight transposes
- reshard fused MoE v1 token and routing inputs to the full expert-parallel layout before entering the fused kernel
- reshard 4D FP8 EPMoE wi_0/wi_1 scales to the tensor-sharded layout required by shard_map
- reshard Kimi MLA new_k_pe to the data-sharded layout required by the MLA backend shard_map
- add direct wrapper regressions for all strict-sharding boundaries, including the FusedEPMoE wrapper path

Fixes the MoE failures in https://github.com/sgl-project/sglang-jax/actions/runs/30294445553/job/90071850604 and addresses the remaining strict-sharding failures reported in #1488 and #1489.

## Testing

- 10 loader, FP8 EPMoE, MLA backend, and QKV split tests passed
- direct FusedEPMoE wrapper regression passed on a four-device CPU mesh (1 passed, 39 deselected)
- full pre-commit suite passed for all changed files (isort, Ruff, Black, codespell, mypy, and repository checks)

GitHub CI is rerunning for commit a31d6f05e.

Issue #774, opened 2026-02-06
[perf-Pathways] Achieve best performance on Pathways for Qwen/Qwen3-32B


aolemila:
Verl Configuration on GPU: 

[https://github.com/rllm-org/rllm/blob/e24d8c82e0fb5f1f8f9729d73f6722f35d0e980a/examples/swe/train_deepswe_32b.sh#L19](<https://github.com/rllm-org/rllm/blob/e24d8c82e0fb5f1f8f9729d73f6722f35d0e980a/examples/swe/train_deepswe_32b.sh#L19>)

zorrofox:
We benchmarked Qwen/Qwen3-32B on Pathways vs native on identical single-host TPU v7x hardware (tp8, bf16, jax 0.10.2 with the dated Pathways images, current main incl. #1501), same server args on both sides — the full ServerArgs diff between the two deployments is exactly `device=proxy` + `--enable-single-process`.

**Serving profile** (`bench_serving` random 1024/1024, 384 prompts, max-concurrency 256; warm rounds, variance <0.4%):

| side | output tok/s (r2/r3) | median ITL | gsm8k 200q |
|---|---|---|---|
| native | 3648 / 3644 | 74.8 ms | 0.930 |
| Pathways | **4182 / 4169** | **54.6 ms** | 0.945 |

Pathways comes out at **114.5% of native** on this profile. For context, the same A/B on a 1.5B model gives Pathways only ~60% of native (#772-style decomposition: a fixed per-step host-transfer cost dominates when steps are cheap); at 32B the per-step compute amortizes that cost away and the Pathways runtime's faster step execution wins outright.

**RL rollout profile**, anchored to the deepswe-32B verl recipe referenced in this issue (prompt ~4K, response 32768 with ignore_eos, temperature 1.0, n=8 × 8 prompts = 64 concurrent generations per step, radix cache on). Each round generates one full rollout step (2.097M tokens):

| side | rollout tok/s (r1/r2) | mid-step retract cycle |
|---|---|---|
| native | 1570 / 1573 | pause(retract) drains instantly, continue OK, 64/64 complete (1561 tok/s through the cycle) |
| Pathways | 1563 / 1567 | same: instant drain, 64/64 complete (1557 tok/s), zero tracebacks |

[...]

Issue #772, opened 2026-02-06
[perf] to achieve best performance on Pathways for deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B


aolemila:

zorrofox:
Status update from our side, since #1501 (merged) references this issue but never reported back here.

**Where the 1.5B gap stands** (DeepSeek-R1-Distill-Qwen-1.5B, tp4 dp1 bf16, page 256, chunked prefill 2048, radix off; `bench_serving` random 1024/1024, 384 prompts, max-concurrency 256; identical single-host TPU v7x hardware, native vs `--device=proxy --enable-single-process`; warm rounds, variance <1.5%):

| stack | native | Pathways | Pathways / native |
|---|---|---|---|
| jax 0.8.1, before #1501 | 8987 tok/s @ 32.0 ms ITL | 3938 tok/s @ 50.4 ms | 43.8% |
| jax 0.8.1, with #1501 | 8987 | 5365 @ 39.0 ms | 59.8% |
| jax 0.10.x, with #1501 | 9675 @ 29.8 ms | ~5300 @ 38.8 ms | ~55% |

Per-step decomposition (H2D metadata / dispatch / resolve-D2H) showed H2D and dispatch at parity with native; the whole gap was the blocking `device_get` in `resolve_last_batch_result`, ~21 ms/step under proxy vs ~3 ms native, i.e. exactly the `device_get` cost you profiled in March. #1501 starts that D2H on the forward thread (`copy_to_host_async` right after the gather), which brings resolve down to ~8 ms/step. The remaining ~8 ms is the client -> proxy -> worker round-trip itself: a host-side SHM transport experiment moved it by ~1%, within noise, so it is not bandwidth-bound and there is nothing further to gain from the transfer path in sglang-jax code.

**Why this matters less as the model grows**: the residual is a fixed per-step cost, so it only dominates when a decode step is cheap. On Qwen3-32B (tp8, same A/B method, reported in #774) Pathways is at 114.5% of native on the serving profile and at parity on a 32K-response rollout profile.

**Remaining lever we have not measured on current main**: `deploymentMode: colocate_head_with_workers`, which removes one network hop from the round-trip. Our runs used a separate head.

Given that, how would you like to treat this issue: close it with the 1.5B ceiling documented as inherent to the separate-head topology, or keep it open specifically for a colocated-head measurement? Happy to run that if it is the deciding data point.

PR #1501, opened 2026-07-31
perf(pathways): reduce per-step overhead; fix PD abort lifecycle, cpp fastpath and FP8 scale sharding

### LINKED ISSUE #1452: [perf] jitted_run_model cpp fastpath miss every decode step under dp=1 + Pathways proxy backend

zorrofox:
Refs #772, #1486; Fixes #1452, #1499

### Motivation

This is the Stage-4 batch of the Pathways integration roadmap item ("Benchmark and optimize Pathways integration, including the remaining device_get/host-transfer overhead and the decode fast-path miss" — #772, #774, #1452), plus the async P-pipeline lifecycle gap reported in #1486.

We first reproduced the #772 gap with DeepSeek-R1-Distill-Qwen-1.5B (tp4, dp1, bf16, page 256, chunked prefill 2048, radix off; `bench_serving` random 1024/1024, 384 prompts, max-concurrency 256; 3 rounds, warm rounds quoted, variance <1.5%) on identical single-host TPU v7x hardware, native vs `--device=proxy --enable-single-process`, and decomposed each decode step into H2D metadata upload / dispatch / resolve-D2H. The entire gap sits in the blocking `jax.device_get(next_token_ids)` in `resolve_last_batch_result`: ~21 ms per step for a few-hundred-byte array under the proxy backend (pure client→proxy→worker round-trip), vs ~3 ms native. H2D and dispatch are at parity.

Numbers on the current 0.10.x dependency set (see the compatibility note on Pathways image tags):

| config | stack (all jax 0.10.2) | output tok/s | median ITL |
|---|---|---|---|
| native (main baseline) | client 0.10.2 | 9682 (9662–9695 ×3) | 29.9 ms |
| native (this PR) | client 0.10.2 | 9675 | 29.8 ms |
| Pathways (main baseline) | client + Pathways images 0.10.2 | 4047 | 49.5 ms |
| Pathways (this PR) | client + Pathways images 0.10.2 | 5344 | 38.8 ms |

(For reference, on the previously-validated jax 0.8.1 stack the same A/B was native 8987 vs Pathways 3938 → 5365 tok/s, i.e. 43.8% → 59.8% of native.)

### Compatibility notes: first full validation on the jax 0.10.x stack

Pathways serving had previously only been validated up to the jax 0.9 series; this PR's validation is the first on 0.10.x. Two pitfalls worth documenting:

1. **Pathways image tags.** The **static** image tags (`pathways/{server,proxy_server}:jax-0.10.1`, and there is no static `jax-0.10.2` tag) point at older binaries whose Mosaic deserializer tops out at version 13, so a jax 0.10.2 client fails every Pallas kernel under `--device=proxy` with `Failed to deserialize the Mosaic module: expected <= 13 but got 15` (native is unaffected — compile and execute share a process). The **dated daily builds** (e.g. `:20260727-jax_0.10.2`, also tagged `:latest`) are a single multi-version binary and work correctly with the 0.10.2 pin — all Pathways numbers and regression runs in this PR use `20260727-jax_0.10.2`, i.e. the exact pinned dependency set end to end.

2. **FP8 block-quant models crash on 0.10.x without the #1499 fix.** jax 0.8.x `shard_map` silently reshards on a textual in_specs mismatch; 0.10.x validates strictly. Static-quant checkpoints (GLM-5.2-FP8 and other DeepSeek-format FP8 models) load their compact 2D linear scales, expand them 2D→3D, and assign without resharding to the placeholder's declared sharding — so on 0.10.x every such model crashes at the first precompile, native and proxy alike (reproduced on unmodified main both ways; #1493 fixed the same class at the MoE/MLA boundaries but missed the linear path). Fixed here by resharding the expanded scale to the placeholder's sharding (see #1499 for the full analysis, including a static audit of the remaining latent spots — none of which trigger at tp16).

### Changes

1. **Start the next-token D2H on the forward thread** — right after the replicate-gather, `next_token_ids.copy_to_host_async()` so the scheduler's `resolve_last_batch_result` finds the copy in flight instead of paying the full interactive round-trip. Decode-phase resolve drops 21→8 ms/step under proxy (+33% output throughput); native unaffected (3 rounds, within variance).
2. **`device_array`: canonical `NamedSharding` per (mesh, spec, memory_kind)** — pjit's cpp fast-path compares input shardings by pointer; constructing a fresh `NamedSharding` per call defeats it and pays `NamedSharding.__init__` per metadata array per step (+2.5% under proxy on the same config).
3. **Async P-pipeline abort/retract lifecycle (#1486)** — the in-flight rid set becomes a registry `rid → (req, p_idx, reserved tokens, outstanding batch count, pending_abort)`. `abort_request` reaches requests anywhere in the pipeline: idle entries finalize immediately (P KV + req slot + admission token accounting + client abort, exactly once), in-flight entries finalize when their batch drains; chunk continuation stops for aborted owners; requests that already migrated are flagged finished so the standard prefill-result path frees their D KV without touching the mixed batch's page layout; pause/retract quiesces the pipeline so a late P result can't merge beside a requeued copy. The two `scheduler.py` hooks are getattr-guarded no-ops outside pathways PD.
4. **Replace `jax.ShapeDtypeStruct` state placeholders (#1452)** — static-quant checkpoints leave SDS placeholders in module state; jaxlib's `ToPyArgSignature` rejects them (`ComputeCallSignature failed: INVALID_ARGUMENT ... Got type ShapeDtypeStruct`, visible with `TF_CPP_VMODULE=pjit=3,jax_jit=3`), so every `jitted_run_model` call takes the python dispatch path — the per-decode-step cpp cache miss reported in #1452. The placeholders are dead on the forward path (a used leaf with a changed shape would fail to compile); they're swapped for zero-length arrays at `initialize_jit`, with the replaced paths logged. Note the root cause is unrelated to dp/proxy as originally guessed in the issue — the original attribution was a control-group artifact.
5. **Reshard expanded block-quant linear scales to their declared sharding (#1499)** — unblocks every FP8 block-quant checkpoint on jax 0.10.x (crash at first precompile on unmodified main, native and proxy alike); details in the compatibility note above and in #1499.

### Validation

- unit: sharding canonicalization + SDS-fastpath repro/restore (`test/srt/test_jax_utils.py`), lifecycle registry/abort/quiesce (`test/srt/disaggregation/test_pathways_pd.py`, 7 new tests)
- #772 repro: numbers above; greedy outputs byte-identical native vs proxy; gsm8k 200q native 0.600 / proxy 0.600
- #1486 e2e (MiMo tp8 2P1D Pathways): 24 aborts at t=2s → 24/24 terminated, registry marks confirmed in logs for in-pipeline requests, post-abort health 8/8, no leak after 12-abort stress round, 0 tracebacks
- #1452 (GLM static-quant MLA/epmoe tp16 dp1 + proxy): 156 placeholders replaced; decode `#cache_miss` 545/545 → 0 in steady state (only legit first-call-per-shape misses remain); gsm8k before/after 0.820 / 0.840 (parity)
- PD regression matrix (MiMo SWA/GQA tp8 2P1D, Pathways), run on BOTH stacks: jax 0.8.1 — correctness 336/336, bench 15534/15486/15512 tok/s (baseline band), gsm8k 0.855, no crash; jax 0.10.2 end-to-end (first 0.10.x Pathways-PD validation) — correctness 336/336, bench 15778/15713/15769 tok/s, gsm8k 0.890, no crash
- GLM-5.2-FP8 (MLA/epmoe) PD matrix @0.10.2 (tp16 2-slice, Pathways; requires the #1499 fix to even start on this stack): correctness 336/336 (SHORT/MID/LONG × conc 4/8/16), bench C64 16K/4K 192-req — 1276.0 / 1274.7 tok/s (r1/r2, 0.1% variance), median ITL 75.08 / 75.18 ms, no crash. The pre-PR baseline on the same workload/topology (jax 0.8.1 stack) was 1043.6 tok/s at 89.28 ms median ITL — the −14.2 ms/step matches the decode-resolve RTT this PR removes (21→8 ms measured in the #772 decomposition), plus the 0.10.2 stack gain. gsm8k 200q: 0.965 (a second independent bring-up scored 0.955).
- #1499 fix evidence: unmodified main crashes at first precompile on GLM static-quant, reproduced BOTH under `--device=proxy` (PD) and native `--device=tpu` (identical `shard_map in_specs ... does not match` error); with the fix the same configs fire up and pass (native: bench 1024/1024 C256 ×3 rounds = 1106.1 / 1106.1 / 1106.0 output tok/s, <0.01% variance, no crash).

### DIFF
--- a/python/sgl_jax/srt/disaggregation/pathways_scheduler.py
+++ b/python/sgl_jax/srt/disaggregation/pathways_scheduler.py
@@ -7,6 +7,7 @@ carries the wiring hooks. See design in issue #1427.
 from __future__ import annotations
 
 import copy
+import dataclasses
 import logging
 import os
 import threading
@@ -34,6 +35,24 @@ from sgl_jax.srt.mem_cache.chunk_cache import ChunkCache, SWAChunkCache
 logger = logging.getLogger(__name__)
 _PD_DBG = bool(os.environ.get("SGLANG_PD_DBG"))
 _PD_DBG_TIMING = bool(os.environ.get("SGLANG_PD_DBG_TIMING")) or _PD_DBG
+
+
+@dataclasses.dataclass
+class _PdInflight:
+    """Lifecycle handle for one request in the async P pipeline (#1486).
+
+    outstanding counts P batches currently in flight (queued or forwarding or
+    waiting in ready_q/defer) that carry this req: a chunked req can have
+    chunk N draining while chunk N+1 forwards. Abort finalization (P KV +
+    reservation release + client abort) may only run when outstanding == 0 --
+    otherwise the in-flight batch drains later and finalizes exactly once.
+    """
+
+    req: object
+    p_idx: int
+    tokens: int
+    outstanding: int = 0
+    pending_abort: bool = False
 
 
 class PathwaysPDSchedulerMixin:
@@ -225,8 +244,12 @@ class PathwaysPDSchedulerMixin:
         # ready_q items whose D-pool alloc was deferred (avail < need+reserved)
         self._pd_defer: list = []
         # reqs pushed to prefill_q but not yet drained into running_batch;
-        # tracked by rid so a chunked req counts once, not once-per-chunk.
-        self._pd_inflight_rids: set[str] = set()
+        # keyed by rid so a chunked req counts once, not once-per-chunk.
+        # Registry (#1486): each entry keeps a handle on the req across the
+        # whole async P pipeline (queue -> prefill thread -> ready_q/defer ->
+        # migrate), so abort/retract can always find it -- previously a req
+        # past its final chunk existed only as a rid string.
+        self._pd_inflight: dict[str, _PdInflight] = {}
         # D-pool tokens the inflight rids will need at migrate time (page-
         # padded full IL per req) -- the admission-side reservation ledger.
         self._pd_inflight_tokens = 0
@@ -654,6 +677,10 @@ class PathwaysPDSchedulerMixin:
             t_enq_ready,
         ) = item
         p_alloc, p_r2t, p_tree = self.p_allocs[p_idx], self.p_r2ts[p_idx], self.p_trees[p_idx]
+        # This batch is no longer in flight for its reqs (#1486); any
+        # pending_abort entry that reaches outstanding == 0 below is
+        # finalized on the matching path (done / mid-chunk / failure).
+        self._pd_batch_outstanding(batch, -1)
         chunked_excl = {
             dp: info.chunked_req
             for dp, info in enumerate(batch.reqs_info)
@@ -682,7 +709,7 @@ class PathwaysPDSchedulerMixin:
         allocator = self.token_to_kv_pool_allocator
         if done_per_req:
             n_running = sum(len(x.reqs) for x in self.running_batch.reqs_info if x.reqs)
-            reserved = self._pd_reserved_per * (n_running + len(self._pd_inflight_rids))
+            reserved = self._pd_reserved_per * (n_running + len(self._pd_inflight))
             reserved_per_rank = reserved // max(1, self.dp_size)
             need_by_rank = defaultdict(int)
             for r, _, p in done_per_req:
@@ -782,6 +809,15 @@ class PathwaysPDSchedulerMixin:
                     (time.perf_counter() - t0) * 1e3,
                     len(done_per_req),
                 )
+        # Aborted reqs that just migrated: flag them finished so the standard
+        # prefill-result path below caches/frees their D KV and streams the
+        # abort to the client -- keeps the gather/scatter page layout of live
+        # reqs in the same mixed batch untouched (#1486).
+        for r, _, _ in done_per_req:
+            entry = self._pd_inflight.get(r.rid)
+            if entry is not None and entry.pending_abort and not r.finished():
+                r.set_finish_with_abort("Aborted by AbortReq.")
+                logger.info("[pd_async] aborted request dropped after migrate. rid=%s", r.rid)
         batch.req_to_token_pool = self.req_to_token_pool
         batch.token_to_kv_pool_allocator = self.token_to_kv_pool_allocator
         batch.tree_cache = self.tree_cache
@@ -793,6 +829,13 @@ class PathwaysPDSchedulerMixin:
         with jax.profiler.TraceAnnotation("pd_process_prefill"):
             self.process_batch_result_prefill(batch, result, None)
         for r in chunked_excl.values():
+            # Aborted mid-chunk owner whose last in-flight batch just drained:
+            # finalize (frees P KV, clears the owner slot, streams the abort)
+            # instead of caching for a chunk N+1 that must never build (#1486).
+            entry = self._pd_inflight.get(r.rid)
+            if entry is not None and entry.pending_abort and entry.outstanding == 0:
+                self._pd_finalize_aborted(entry)
+                continue
             # Redundant with the prefill thread's own update (same numpy read,
             # done right after forward — see _pd_prefill_loop) but harmless
             # and kept as a safety net. _pd_chunk_pending is NOT touched here:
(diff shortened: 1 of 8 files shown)


python/sgl_jax/srt/disaggregation/pathways_scheduler.py
```diff
@@ -654,6 +677,10 @@ def _pd_drain_ready(self) -> None:
             t_enq_ready,
         ) = item
         p_alloc, p_r2t, p_tree = self.p_allocs[p_idx], self.p_r2ts[p_idx], self.p_trees[p_idx]
+        # This batch is no longer in flight for its reqs (#1486); any
+        # pending_abort entry that reaches outstanding == 0 below is
+        # finalized on the matching path (done / mid-chunk / failure).
+        self._pd_batch_outstanding(batch, -1)
```
cjx0709: Could we decrement `outstanding` only after the item is fully handled? If the capacity check below defers it, the item remains in `_pd_defer` but is already counted as drained. An abort in this window can free the P-side request, and the deferred item may later retry with stale slots.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
```diff
@@ -955,6 +1075,15 @@ def _pd_maybe_push_prefill(self) -> None:
...
         for _off in range(n_p):
             i = (self._pd_next_p + _off) % n_p
+            # Aborted chunked owner with no batch in flight: finalize here so
+            # its P KV frees and no chunk N+1 is ever built (#1486). Owners
+            # with a batch still in flight finalize on that batch's drain.
+            for dp, owner in enumerate(self.p_chunked_reqs[i]):
+                if owner is None:
+                    continue
+                entry = self._pd_inflight.get(owner.rid)
+                if entry is not None and entry.pending_abort and entry.outstanding == 0:
```
cjx0709: Could we skip building a continuation when `entry.pending_abort` is set, even if `outstanding > 0`? The P thread clears `_pd_chunk_pending` before the result is drained, so an abort in this window can still enqueue chunk N+1 and do extra work after cancellation.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
```diff
@@ -869,18 +912,97 @@ def _pd_req_kv_tokens(self, r) -> int:
...
+                any(not q.empty() for q in self._pd_prefill_qs)
+                or any(self._pd_chunk_pending)
+                or not self._pd_ready_q.empty()
+                or bool(self._pd_defer)
+                or any(e.outstanding > 0 for e in self._pd_inflight.values())
+            )
+            if not busy:
+                return
+            time.sleep(0.01)
+        logger.warning(
```
cjx0709: Could we fail closed when quiesce times out instead of only logging a warning? A late P result may still arrive after retract requeues the request, recreating the race this drain is intended to prevent.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -2794,6 +2801,13 @@ def pause_generation(self, recv_req: PauseGenerationReqInput):
         consumed = self._process_pending_chunked_aborts()
         self._retire_chunked_req_batch_owners(consumed)
 
+        # Pathways single-process PD: drain the async P pipeline as well so a
+        # late prefill result cannot merge into running_batch alongside a
+        # requeued copy of the same request after retract (#1486). No-op
+        # unless pathways PD is on.
+        if getattr(self, "_pd_inflight", None) is not None:
```
cjx0709, line 2808: Could retract also clear or requeue idle owners in `p_chunked_reqs`? With Pathways this helper is a no-op, so a request between chunks can remain only in the P pool and resume from its old prefix instead of being recomputed.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
zorrofox: Good catch — this was real: the decrement ran on every pop, so a deferred item's second pass would floor the count at zero while the item was still parked, and an abort in that window could finalize the req and free the P slots under it. Fixed in 6a71dbfc by moving the decrement past the admission gate (and into the failure branch), so a batch is only counted as drained once the drain actually commits to consuming it.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
zorrofox: Agreed — the P thread clears `_pd_chunk_pending` before the result is drained, so that window existed. Fixed in 6a71dbfc: the build loop now skips a slice entirely while its chunked owner has `pending_abort` with a batch still in flight; the owner is finalized on that batch's drain and no continuation is ever built.

python/sgl_jax/srt/disaggregation/pathways_scheduler.py
zorrofox: Agreed, and fail-closed is the only safe direction here: the stranded P pages cannot be reclaimed while the prefill thread may still be executing on them, so neither "free and requeue" nor "proceed and hope" is sound. 6a71dbfc makes `_pd_quiesce` block until the pipeline actually drains, logging a periodic warning (cold-shape compiles on the P thread can legitimately take >30s, which is exactly when the old timeout would have fired). If the P thread dies without reporting, the failure path still enqueues a `result=None` item, so the drain completes through the failure branch.

python/sgl_jax/srt/managers/scheduler.py
zorrofox: Right — `_retract_parked_chunked_reqs` early-returns under pathways PD and the parked owner would have resumed on pre-retract KV (stale if the retract preceded a weight update). 6a71dbfc adds `_pd_retract_chunked_owners`, called from the retract branch after the quiesce: frees the P-side KV and req slot, drops the registry/admission accounting, resets the req and requeues it so it recomputes from scratch. Unit test covers the requeue-not-abort semantics.

cjx0709 (approved):
LGTM

Issue #1297, opened 2026-06-03
[Bug] native attention backend: garbage/empty output on Qwen3-8B + O(batch^2) decode slowness & OOM (TP4, v6e-4)


Rodrian7:
## Summary

On TPU `v6e-4` (TP4), the **`native` attention backend** produces **incorrect generations** (garbage token-salad or empty/degraded output) for `Qwen/Qwen3-8B`, while the `fa` backend on the *exact same* model and server config (only `--attention-backend` changed) is fully correct. Separately, the `native` backend is **pathologically slow** at decode (~0.7 tok/s per request) and **OOMs** at non-trivial context lengths.

Together these block the `qwen3-8b-native` accuracy gate planned in #1200 §1.1.1: the case cannot run a meaningful end-to-end eval within any reasonable time budget, and even when it runs, output correctness is unstable.

---

## Environment

- **Repo:** `sgl-project/sglang-jax` @ commit `07c0bed` (branch `worktree-nightly-4tpu-accuracy`)
- **Python** 3.12.12 · **jax / jaxlib** 0.8.1 · **sgl_jax** 0.0.2
- **Hardware:** TPU `v6e-4` (4 chips, ~32 GB HBM/chip)
- **Model:** `Qwen/Qwen3-8B`, `bf16`, TP4
- `JAX_COMPILATION_CACHE_DIR` warm; model under `--download-dir /dev/shm/`

### Server launch (only `--attention-backend` + the memory knobs differ)

```bash
# native (under test)
python -m sgl_jax.launch_server --trust-remote-code --model-path Qwen/Qwen3-8B \
  --host 127.0.0.1 --port 30011 --tp-size 4 --dp-size 1 --nnodes 1 --node-rank 0 \
  --dist-init-addr 127.0.0.1:35011 --skip-server-warmup --random-seed 3 \
  --download-dir /dev/shm/ --dtype bfloat16 --page-size 256 \
  --attention-backend native \
  --mem-fraction-static 0.6 --max-prefill-tokens 4096 --context-length 3072 \
  --precompile-token-paddings 4096 --precompile-bs-paddings 52 --max-running-requests 52 \
  --device tpu

# fa (baseline)
python -m sgl_jax.launch_server --trust-remote-code --model-path Qwen/Qwen3-8B \
  --host 127.0.0.1 --port 30011 --tp-size 4 --dp-size 1 --nnodes 1 --node-rank 0 \
[...]

PR #1502, opened 2026-08-01
fix(attention): page-align native attention mask boundaries (#1297)

real3stone:
## Motivation

This PR addresses a correctness issue in the native attention backend when `page_size > 1`
with multiple concurrent requests. The problem manifests as garbage / empty output under
these conditions, while the `fa` backend operates correctly and single-batch `native` runs
work fine.

Both mask paths are affected — decode and extend — so this covers the correctness half of
#1297. The per-step cost and the score-tensor OOM reported there share a separate root
cause (the attention logits are materialized as one full `(queries, heads, keys)` matrix)
and are not touched here.

## Root Cause

`cache_loc` is laid out page-aligned per request by `ScheduleBatch._merge_cache_loc`, where
request `i` occupies `ceil(seq_len_i / page_size) * page_size` slots. Both
`_apply_decode_mask` and `_apply_extend_mask` reconstruct the key ranges using a dense
`cumsum(seq_lens)` instead, causing mask windows to slide off the actual keys by the
accumulated page padding ahead of them. Each affected request's mask targets a neighbour's
page-tail slots rather than its own keys, and drops some of its own, including the most
recent one.

The failure condition is exact: a request is wrong iff the total page padding of the
requests ahead of it is non-zero. It is silent — page ids start at 1, so the page-tail
slots hold real allocated indices and `loc > 0` does not remove them. The mask is the only
boundary check on this path.

## Modifications

Two commits, one per mask path. Both take `page_size` as a required keyword-only argument
rather than defaulting it, so a caller that forgets it fails loudly instead of silently
getting the dense layout back.

**decode** — step the block offsets by the aligned lengths, keeping each block's valid span
at the real sequence length so page tails stay masked. This also fixes the sliding-window
bound, which derives from the same `seq_ends`:

```python
aligned_lengths = ((seq_lengths + page_size - 1) // page_size) * page_size
seq_starts = jnp.cumsum(jnp.concatenate([jnp.array([0]), aligned_lengths[:-1]]))
seq_ends = seq_starts + seq_lengths
```

**extend** — the same block offsets, plus its validity bound. That was
`arange(key_len) < sum(seq_lens)`, a single trailing cut which only holds for a dense
layout; with page-aligned blocks the invalid region is holes between blocks, so it becomes
a per-column end:

```python
aligned_lengths = ((seq_lengths + page_size - 1) // page_size) * page_size
k_starts = jnp.cumsum(aligned_lengths, dtype=jnp.int32) - aligned_lengths
k_valid_mask = jnp.arange(key_len) < (k_starts + seq_lengths)[k_batch_ids]
```

## Accuracy Tests

`test/srt/test_native_attention_paged_decode.py` — the first unit tests for this backend,
registered in `unit-test-cpu`. Both mask paths, across page sizes, sequence-length
combinations, MHA/GQA/MQA, the sliding-window bound, prefix lengths, and the bucket padding
serving applies. Compared against an independent per-sequence reference.

```
python -m pytest test/srt/test_native_attention_paged_decode.py
# 17 passed, 40 subtests passed
```

Seven subtests are regression baselines that pass both before and after the change
(`page_size == 1`, batch of one, exact page multiples, padding only on the last request);
reverting the fix turns the other 13 red.

Also checked against a running server on CPU (Qwen3-0.6B, `--attention-backend native
--page-size 4`), instrumented to print the real `seq_lens` and `cache_loc`: the observed
layout matches `cumsum(aligned_lens)` block for block and every mask window lands on that
request's own slots.

An end-to-end check on the configuration from #1297 (TPU v6e-4, `--page-size 256`) is
planned separately.

## Benchmarking and Profiling

No performance change is intended: the mask is computed on the same shapes as before, only
with different boundaries.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/layers/attention/native_backend.py
+++ b/python/sgl_jax/srt/layers/attention/native_backend.py
@@ -297,6 +297,7 @@ def forward_attention(
             is_causal,
             sliding_window_size,
             mesh=mesh,
+            page_size=page_size,
         )
     else:
         attn_logits = _apply_decode_mask(
@@ -341,10 +342,14 @@ def _apply_extend_mask(
     is_causal: bool = True,
     sliding_window_size: int | None = None,
     mesh: Mesh | None = None,
+    *,
+    page_size: int,
 ):
     """
     Applies a block-diagonal and optionally a causal/SWA mask in a unified,
     efficient way, correctly handling padding.
+
+    page_size: KV pool page size; extend `loc` blocks are page-aligned per request.
     """
     query_len, _, key_len = attn_weights.shape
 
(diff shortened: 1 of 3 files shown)


real3stone:
Hi @Rodrian7 @JamesBrianD — friendly ping on this one. It's been open for three weeks (decode fix on 08-01, extend path folded in on 08-07) and still has no CI runs, since first-time-contributor workflows need maintainer approval.

Two small asks:

1.Approve the workflow runs. The tests this PR adds (test/srt/test_native_attention_paged_decode.py) are registered in unit-test-cpu, so this doesn't need TPU runner time. Coverage is 17 tests / 40 subtests; reverting the fix turns 13 subtests red.
2.A first-pass review whenever there's bandwidth — happy to iterate quickly.

One note on validation scope, to be upfront: everything verified so far is CPU-side — the unit tests above, plus a running-server check (Qwen3-0.6B, --attention-backend native --page-size 4) instrumented to confirm the observed cache_loc layout matches cumsum(aligned_lens) block for block. The fix itself is mask/index boundary arithmetic in the native backend, so it isn't device-dependent. I have not run the end-to-end check on the original #1297 configuration (v6e-4, TP4, --page-size 256) because I don't have access to multi-chip v6e hardware — the only TPU I can reach is a single-chip v6e-1 on Colab. If a single-chip end-to-end run (--attention-backend native --page-size 256, GSM8K before/after) would be useful signal, glad to run it and post the numbers; just say the word.

Thanks!

JamesBrianD:
@real3stone Re-submission of https://github.com/sgl-project/sglang-jax/pull/1502 on the latest main because the original workflow runs are pinned to retired self-hosted runner labels.

PR #1504, opened 2026-08-03
fix(overlap): future-token ring wraparound corrupts outputs at high concurrency

### LINKED ISSUE #1317: [Bug] Hybrid linear-attention (GLA) models: recurrent-state accuracy collapses under the overlap scheduler, scaling with decode batch size
## Summary

On hybrid linear-attention models (GLA / "lightning" recurrent layers, e.g. **Ling-2.6**), generation accuracy **degrades monotonically with decode batch size when the overlap scheduler is enabled** (the default). Single-stream is perfect; by batch 128 it is badly degraded. With `--disable-overlap-schedule` (same commit, same model, same config, only that one flag changed) accuracy is flat and healthy across all batch sizes. There is no crash and no warning — the model just emits silently wrong tokens.

Found on TPU v7x multi-host, Ling-2.6-1T, `tp32/dp8/ep32`, `--moe-backend fused`, `main` = `11833934d36fa6f20526d513186c86b9c03e137c`.

> **Correction vs. the original report:** this issue was first filed describing the trigger as *recurrent-slot reuse* (`#prompts > max_concurrency`). That framing is **wrong** — see "What was ruled out" below. Direct host-side instrumentation shows the corruption happens **with zero slot reuse**, and it is driven by **decode batch size**, not reuse. The decisive data has been re-measured and is below.

## Symptom — decisive controls (GSM8K exact-match, greedy, 5-shot, 256 max-new-tokens)

Each request count equals the concurrency (`N == parallel`), so **every request keeps its own recurrent slot for the whole run — no slot is ever reused**. The only knob is batch size:

| batch (`N == parallel`) | overlap ON (`main`, fused decode) | overlap OFF | overlap ON + revert #1110 |
|---|---|---|---|
| 1   | 1.000 | 1.000 | 1.000 |
| 16  | 0.938 | 0.938 | 0.938 |
| 64  | **0.750** | 0.922 | 0.812 |
| 128 | **0.258** | 0.930 | 0.555 |
| 256 | **0.500** | 0.922 | 0.449 |

(The degraded tail has run-to-run variance — e.g. a no-reuse `par256` landed 0.45–0.50 across runs, an interleaved `par64/N256` landed 0.37–0.59 — but it is always badly degraded with overlap ON and always ~0.92 with overlap OFF.)

Three things this table establishes:

1. **Overlap is the variable.** overlap OFF is flat ~0.92 across all batch sizes; overlap ON collapses. Same binary, same weights, same config, only `--disable-overlap-schedule` differs.
2. **It is not slot reuse.** Every column above is `N == parallel` → no recurrent slot is ever freed and reassigned, yet overlap ON still collapses (par128 → 0.258). Reuse is not necessary.
3. **It scales with decode batch size.** Single-stream is exact; the error grows as the decode batch grows.

## What was ruled out

**(a) Slot reuse — refuted by host-side instrumentation.** I instrumented `HybridReqToTokenPool` (alloc/free with rid + req_pool_idx + slot), the per-step `recurrent_indices` (rid→slot pairs), and `ScheduleBatch.copy()` list-aliasing, all env-gated, then ran the degraded `par64/N256` case (accuracy 0.367) and rebuilt the slot-ownership timeline offline:

- `alloc=256, free=256`, **256 distinct `(dp, slot)` pairs — 0 reuse** the entire run.
- **0** requests changed slot mid-life; **0** `(dp,slot)` owned by >1 request; **0** A→B→A cross-batch aliasing; **0** "freed slot still used by old rid" violations.
- **0** `copy()`/`merge_batch` list-aliasing events (`DBGALIAS=0`); **0** double-alloc; within-batch `recurrent_indices` collisions = 0; `has_initial_state` mismatches = 0.

Host-side slot/index/`has_initial_state` bookkeeping is **completely clean**. The corruption is not visible at the host layer.

**(b) The `has_initial_state` lazy-zero path is correct.** A new request on any slot gets `has_initial_state=False`, and `h0 = jnp.where(has_l, h0, 0.0)` zeroes the gathered state before use. The read side is safe; this is not a "forgot to clear a slot" bug.

**(c) PR #1110 (fused DECODE Pallas kernel) is not the root cause.** I reverted #1110 (`cdb173f1`) on top of `main` — decode falls back to the old `fused_recurrent_simple_gla` (`jax.lax.scan`) + separate `get_state` / `set_ssm_state` gather/scatter — kept overlap ON, and re-ran the curve (rightmost column). It **still collapses** (par128 → 0.555). So the bug is **not** in #1110's fused decode kernel; it is in a path shared by both decode implementations. (Revert is marginally less bad at par64/128, so #1110 may *aggravate* it, but it is not the origin. The EXTEND path is unchanged by #1110.)

## Where this points

Putting the controls together: the corruption requires **overlap ON**, affects the **recurrent (GLA) state**, **scales with decode batch size**, is **independent of slot reuse and of the #1110 decode kernel**, and leaves **all host-side bookkeeping correct**. That isolates it to how the **overlap pipeline threads recurrent state across steps on the device** — the part shared by both decode kernels and by EXTEND — rather than to slot lifetime, batch metadata, or the decode kernel math.

I could not dump the device-side recurrent buffer values to pin the exact step, because `jax.debug.callback` does not fire on the overlap worker thread.

Concrete things a maintainer could check, given device-side visibility:

- Under overlap two forwards are in flight and the recurrent pool is threaded functionally (`memory_pools` is `donate_argnames`'d into the jitted forward and the returned `pool_updates` are written back via `replace_all`). Does a launched forward ever consume a **stale recurrent-buffer version** (one that does not yet include the previous in-flight step's writeback) — and does that error grow with the number of sequences packed in the batch?
- Whether the per-DP `recurrent_indices` / `seq_lens` layout + padding fed to the recurrent `shard_map` gather/scatter stays consistent with the rows actually dispatched as batch size grows under overlap.

## Workaround

`--disable-overlap-schedule` — flat ~0.92 across all batch sizes, at the cost of overlap throughput. This is the only safe configuration for hybrid GLA models today.

## Impact

Hybrid linear-attention / GLA models produce **silently wrong output** under the default overlap scheduler for any non-trivial decode batch (already visibly degraded by batch 64, badly by batch 128) — no crash, no warning. Single-stream and overlap-OFF are correct.

zorrofox:
**Likely fixes #1317** — same failure signature (accuracy collapsing with decode batch size under the overlap scheduler, hybrid/linear-attention models hit hardest, host-side bookkeeping clean). This PR reproduces and eliminates that signature on a hybrid-GDN model (A/B below); a Ling rerun on this branch should confirm and close #1317.

### Symptom

With the (default) overlap scheduler, generation accuracy degrades monotonically with server-side batch size, on every model family we tested, while low concurrency and `--disable-overlap-schedule` stay healthy. gsm8k 200q 5-shot greedy, same server, only client parallelism varied:

| model | healthy (c4) | collapsed |
|---|---|---|
| GLM-5.2-FP8 tp16 (v7x, 2 hosts) | 0.955 | **0.700** @ c64 |
| DeepSeek-R1-Distill-Qwen-1.5B bf16 tp4 (single host) | 0.670 | **0.435** @ c128 |

GLM cliff shape (overlap ON): c8 0.955 → c9 0.935 → c16 0.930 → c32 0.860 → c64 0.700. Controls on the same server: `--disable-overlap-schedule` c64 = 0.955; `--max-running-requests 8` with client c64 = 0.955. No retracts, Invalid ≤ 1%, bf16 and quantized models equally affected — this is not a numerics/quantization issue.

The failure signature is a corrupted token right at the start of a request's output (first decode steps), after which greedy decoding derails. A quick probe makes it countable: 128–384 concurrent greedy requests, each asked to echo its own unique codeword; any output that breaks its own codeword is an anomaly. On main: 8/768 anomalies (dense 1.5B) and **44/1152** (Qwen3.5-0.8B hybrid GDN — recurrent state compounds the corrupted token, which is why hybrid models are hit hardest, cf. #1317). With this fix: **0/768** on both.

### Root cause

`ModelWorkerClient.forward_batch_generation` allocates future-token placeholders from a ring buffer and advances the cursor by `bs = len(model_worker_batch.seq_lens)`.

Since the DP-support refactor (#939), `get_model_worker_batch` returns **padded** per-DP arrays — for extend batches padded to the *largest* bs bucket. So the cursor advances by the padded size (e.g. 256) per prefill batch instead of the number of real sequences (2–3), while the ring size (`max_running_requests * 3`) was chosen for real-bs advances. A burst of prefill batches wraps the ring within ~3 batches and overwrites outstanding placeholders **before the owning request's first decode batch resolves them** — that request's first input token silently resolves to whatever a newer batch wrote into the slot. The higher the concurrency, the more prefill batches run between a request's prefill and its first decode, hence the monotonic collapse. Phase-split (PD) deployments and `--disable-overlap-schedule` never consume placeholders, which is why they are immune.

Introduced by #939 (bisected: parent `b787fdef` healthy 0.650 @ c128, `de5287ed` collapsed 0.405; 7-round bisect on the 1.5B single-host setup).

### Fix

Index the future-token map by **`req_pool_idx + 1`** — one pending-token slot per request slot — instead of a moving cursor:

- No cursor, no sizing assumption, no wraparound: a request's slot can only be overwritten by its own next token (forward-thread order guarantees set-before-resolve).
- Placeholder value becomes `-(req_pool_idx + 1)`; `resolve_future_token_ids` is unchanged (it already maps `id -> map[-id]`).
- Padding rows scatter out of bounds and are dropped (`mode="drop"`), so they can never clobber the request that legitimately owns pool idx 0.
- Shapes stay static per bs bucket (no new compilation variants, no AOT impact).
- Applied to the plain overlap path, the fused Pathways-PD decode path (same transformation, slots derived in-jit from `forward_batch.req_pool_indices`/`seq_lens`), and the precompile stub.

### Accuracy after fix (same servers, overlap ON)

| check | main | this PR |
|---|---|---|
| GLM-5.2-FP8 gsm8k c64 | 0.700 / 0.700 | **0.950 / 0.960** |
| GLM-5.2-FP8 gsm8k c4 | 0.955 | 0.955 |
| 1.5B gsm8k c128 | 0.435 / 0.435 | **0.680** |
| 1.5B gsm8k c4 | 0.670 | 0.670 |
| codeword-echo anomalies, 1.5B dense | 8/768 | **0/768** |
| codeword-echo anomalies, Qwen3.5-0.8B GDN | 44/1152 | **0/768** |

### Performance

Same-pod paired A/B, 1.5B tp4, random 1024/1024, 384 prompts, max-concurrency 256, 2 rounds each:

| | output tok/s | median ITL |
|---|---|---|
| main | 8796 / 8850 | 31.9 / 31.6 ms |
| this PR | 8836 / 8805 | 31.6 / 31.9 ms |

Neutral (within run-to-run noise). The per-step op change is a same-size scatter replacing a `dynamic_update_slice`; the map shrinks from `max_running*5` to `req_pool_size+2`.

### Test plan

- [x] Unit tests: `python/sgl_jax/test/managers/test_future_token_map.py` — set/placeholder round trip, padding rows cannot clobber pool-idx-0 slot, placeholder survives a 50-batch padded prefill burst (the exact wraparound scenario)
- [x] gsm8k accuracy matrix before/after on GLM-5.2-FP8 (tp16, 2-host v7x) and DeepSeek-R1-Distill-Qwen-1.5B (tp4, single host), c4 no-regression + high-concurrency recovery
- [x] Codeword-echo contamination probe (3 model families: dense bf16, FP8 MoE MLA, hybrid GDN) — zero anomalies after fix
- [x] Same-pod paired throughput A/B — neutral
- [x] pre-commit (isort/ruff/black/mypy) green on both commits

Notes for reviewers:
- Legacy EAGLE spec paths force `disable_overlap_schedule` (server_args check) and never consume the future map — unaffected. The NEXTN-overlap path shares the fused decode jit updated here; semantics are covered by the unit tests, and we'd appreciate a PD-side CI run to double-confirm.
- #1317 reports the same shape of failure (accuracy collapse scaling with decode batch size under overlap, hybrid/GLA models hit hardest, host-side bookkeeping clean). The Qwen3.5-0.8B GDN A/B above reproduces and fixes that signature; retesting Ling on this branch should allow closing #1317.

### DIFF
--- a/python/sgl_jax/srt/managers/tp_worker.py
+++ b/python/sgl_jax/srt/managers/tp_worker.py
@@ -473,7 +473,7 @@ class ModelWorker:
         sampling_metadata: SamplingMetadata = None,
         forward_metadata=None,
         future_map=None,
-        future_ct=None,
+        future_slots=None,
     ) -> tuple[LogitsProcessorOutput, jax.Array | None, int]:
         # Prepare LoRA batch if LoRA is enabled
         if self.worker.server_args.enable_lora and self.need_prepare_lora_batch:
@@ -509,7 +509,6 @@ class ModelWorker:
             if model_worker_batch.sampling_info:
                 self._update_grammar_vocab_mask(model_worker_batch, sampling_metadata)
             fmap = future_map if future_map is not None else self._pd_dummy_future_map
-            fct = future_ct if future_ct is not None else 0
             (
                 next_token_ids_device,
                 logits_output,
@@ -518,7 +517,7 @@ class ModelWorker:
                 layers_topk_ids,
                 new_future_map,
             ) = self.model_runner.forward_and_sample(
-                forward_batch, logits_metadata, sampling_metadata, fmap, fct
+                forward_batch, logits_metadata, sampling_metadata, fmap
             )
             self.dump_topk_ids(layers_topk_ids, model_worker_batch)
             if launch_done is not None:
--- a/python/sgl_jax/srt/managers/tp_worker_overlap_thread.py
+++ b/python/sgl_jax/srt/managers/tp_worker_overlap_thread.py
@@ -16,7 +16,11 @@ from jax.sharding import NamedSharding, PartitionSpec
 
 from sgl_jax.srt.managers.schedule_batch import ModelWorkerBatch
 from sgl_jax.srt.managers.tp_worker import ModelWorker
-from sgl_jax.srt.managers.utils import resolve_future_token_ids, set_future_token_ids
+from sgl_jax.srt.managers.utils import (
+    future_slot_indices,
+    resolve_future_token_ids,
+    set_future_token_ids,
+)
 from sgl_jax.srt.model_executor.forward_batch_info import ForwardBatch
 from sgl_jax.srt.sampling.sampling_batch_info import SamplingMetadata
 from sgl_jax.srt.server_args import ServerArgs
@@ -45,10 +49,11 @@ class ModelWorkerClient:
         self.device = self.worker.device
         self.cur_sampling_info = None
 
-        # Init future mappings
-        self.future_token_ids_ct = 0
-        self.future_token_ids_limit = self.max_running_requests * 3
-        self.future_token_ids_map = jnp.zeros((self.max_running_requests * 5,), dtype=jnp.int32)
+        # Init future mappings. The map is indexed by req_pool_idx + 1 (one
+        # pending-token slot per request slot; index 0 reserved), not by a
+        # running cursor -- see set_future_token_ids.
+        self.future_map_size = self.worker.model_runner.req_to_token_pool.size + 2
+        self.future_token_ids_map = jnp.zeros((self.future_map_size,), dtype=jnp.int32)
         self.mesh = mesh
         sharding = NamedSharding(mesh, PartitionSpec(None))
         self.future_token_ids_map = jax.device_put(self.future_token_ids_map, sharding)
@@ -109,7 +114,7 @@ class ModelWorkerClient:
         while True:
             (
                 model_worker_batch,
-                future_token_ids_ct,
+                future_slot_indices_np,
                 sampling_metadata,
                 forward_metadata,
             ) = self.input_queue.get()
@@ -131,7 +136,7 @@ class ModelWorkerClient:
                             sampling_metadata=sampling_metadata,
                             forward_metadata=forward_metadata,
                             future_map=self.future_token_ids_map,
-                            future_ct=future_token_ids_ct,
+                            future_slots=future_slot_indices_np,
                         )
                     )
                 self.future_token_ids_map = new_future_map
@@ -158,7 +163,7 @@ class ModelWorkerClient:
             # set_future's cpp-fastpath cache hits; async_gather afterwards.
             self.future_token_ids_map = set_future_token_ids(
                 self.future_token_ids_map,
-                future_token_ids_ct,
+                future_slot_indices_np,
                 next_token_ids,
                 self.mesh,
             )
@@ -284,26 +289,24 @@ class ModelWorkerClient:
             model_worker_batch, self.worker.get_model_runner()
         )
 
+        # Per-request slots: placeholder value -(req_pool_idx + 1) round-trips
+        # through resolve_future_token_ids (map[-id]); padding rows get 0
+        # (a non-negative id, resolved as-is and never consumed).
+        seq_lens_np = np.asarray(model_worker_batch.seq_lens)
+        req_pool_np = np.asarray(model_worker_batch.req_pool_indices)
+        slots = future_slot_indices(seq_lens_np, req_pool_np, self.future_map_size)
+
         # Push a new batch to the queue (JAX handles synchronization automatically)
         self.input_queue.put(
             (
                 model_worker_batch,
-                self.future_token_ids_ct,
+                slots,
                 sampling_metadata,
                 forward_metadata,
             )
         )
 
-        # Allocate output future objects
-        bs = len(model_worker_batch.seq_lens)
-
-        future_next_token_ids = np.arange(
-            -(self.future_token_ids_ct + 1),
-            -(self.future_token_ids_ct + 1 + bs),
-            -1,
-            dtype=np.int32,
-        )
-        self.future_token_ids_ct = (self.future_token_ids_ct + bs) % self.future_token_ids_limit
+        future_next_token_ids = np.where(seq_lens_np > 0, -slots, 0).astype(np.int32)
         return None, future_next_token_ids, 0
 
     def run_precompile(self, only: str | None = None):
--- a/python/sgl_jax/srt/managers/utils.py
+++ b/python/sgl_jax/srt/managers/utils.py
@@ -54,8 +54,31 @@ def resolve_future_token_ids(input_ids, future_token_ids_map, mesh):
     return jax.sharding.reshard(input_ids_global, NamedSharding(mesh, P("data")))
 
 
+def future_slot_indices(seq_lens_np, req_pool_indices_np, map_size):
+    """Per-request future-map slots: real rows -> req_pool_idx + 1 (slot 0 is
+    reserved so 0 stays a valid "no placeholder" input id), padding rows ->
+    map_size (out of bounds, dropped by the scatter)."""
+    import numpy as np
+
+    return np.where(
+        seq_lens_np > 0,
+        req_pool_indices_np.astype(np.int32) + 1,
+        np.int32(map_size),
+    ).astype(np.int32)
+
+
 @jax.jit(static_argnames=("mesh"))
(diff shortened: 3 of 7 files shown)


JamesBrianD:
Validated this with the Ling-2.6-1T cookbook-style launch on v7x 16 chips.

Setup:
- PR head: `79fa0245`
- Model: `inclusionAI/Ling-2.6-1T` mounted from GCSFuse via `/tmp/Ling-2.6-1T` symlink
- Flags included `--tp-size 32 --dp-size 8 --ep-size 32`, `--moe-backend fused_v2`, `--page-size 256`, `--chunked-prefill-size 2048`, `--max-running-requests 512`, `--disable-radix-cache`, `--attention-backend fa`
- Used the no-file-cache GCSFuse mount path; the larger file-cache config was not viable on this node pool because it triggered ephemeral-storage DiskPressure.

Results:
- Server started successfully after weight load and precompile.
- GSM8K `parallel=128`, `num_questions=128`: accuracy `0.938`, invalid `0.000`, latency `20.277s`, output throughput `739.960 tok/s`.
- GSM8K `parallel=256`, `num_questions=256`: accuracy `0.918`, invalid `0.000`, latency `52.631s`, output throughput `594.095 tok/s`.
- I did not see `ERROR`, `Traceback`, `Failed`, or `Killed` in the server logs after the runs.

Based on this validation, the high-concurrency Ling2.6 path that was failing in #1317 looks recovered by this PR.

JamesBrianD:
Could you also add a unit test for the real padded `req_pool_indices` sentinel?

The current padding-drop test uses padded `req_pool_indices=0`, which is a useful adversarial case, but the runtime padded batch uses `-1` for padded `req_pool_indices` and `seq_lens=0`. It would be better to pin that exact contract too, e.g. a padding row with `req_pool_indices=-1` should produce the out-of-bounds future slot and be dropped by `set_future_token_ids(..., mode="drop")`.

Also, please add this test file to the normal test suite/CI path, so the future-token map regression is covered by the suite instead of only being runnable manually.

zorrofox:
Done in 6763b6d1:

- Added `test_runtime_padding_sentinel_rows_are_dropped`: builds the padded batch exactly as the runtime does (`req_pool_indices=-1`, `seq_lens=0`, per `get_model_worker_batch`), asserts the sentinel rows map to the out-of-bounds slot and are dropped by `set_future_token_ids(..., mode="drop")`, and that every slot other than the two legitimate writes is bit-for-bit unchanged. The strict "nothing else changed" check also guards against a `drop` → `clip` regression (clip would corrupt the last slot).
- Registered `python/sgl_jax/test/managers/test_future_token_map.py` in the `unit-test-cpu` suite in `test/srt/run_suite.py` (pytest runner), with the same `USE_DEVICE_TYPE=cpu` platform-pin header the other files in that suite use.

All 4 tests pass on CPU; pre-commit green.

JamesBrianD:
Thank you for fixing this bugs. The pr will be merged after the ci testing.

JamesBrianD (approved):

PR #1506, opened 2026-08-03
feat(pd): support Raiden data-parallel transfer routing

cjx0709:
## Summary

This PR builds on the optional Raiden transfer engine introduced in #1490 and adds data-parallel routing for PD disaggregation.

The previous Raiden path assumed a single DP rank. With multiple ranks, that assumption loses the identity of the Prefill KV owner and the Decode allocator, so endpoint selection, page lookup, admission control, and cleanup can target the wrong rank-local namespace.

This change:

- creates one Raiden KV manager per local DP rank, backed by rank-local zero-copy JAX KV views
- propagates the Prefill source rank and Decode destination rank through request, bootstrap, transfer, scheduler, and cleanup metadata
- selects the source endpoint by `disagg_prefill_dp_rank` and the destination manager/allocator by `dp_rank`
- includes process and source-rank identity in bootstrap keys
- partitions the configured inflight capacity across DP ranks and enforces both global and per-rank admission limits
- lets ranks with available KV capacity continue admission when another rank is blocked, while preserving FIFO within each rank
- preserves abort-safe ownership: KV pages are not released until Raiden reaches a terminal state
- isolates heartbeat, unregister, and destructive Raiden completion polling failures per rank
- validates that KV arrays are physically sharded over the data axis and maps debug gathers to the correct rank-local page shard
- rejects DP > 1 with the legacy JAX transfer engine; this path is supported by Raiden only

## Routing model

Raiden transfers do not require the source and destination ranks to be equal:

```text
request -> Prefill rank i -> Decode rank j
```

The normal MiniLB path currently keeps the ranks aligned (`i == j`). Direct requests can select distinct source and destination ranks, which is used by the exhaustive routing tests. Production cross-rank load balancing and cross-topology resharding are outside this PR.

### Local tests

- `test/srt/disaggregation`: **432 passed, 12 skipped**
- Ruff lint: passed
- Black formatting: passed
- `git diff --check`: passed

### TPU correctness

Validated with Qwen3-8B on two single-host v6e-4 systems, using four TPU devices on each side:

| Topology | Source × destination pairs | Deterministic request checks | Result |
| --- | ---: | ---: | ---: |
| DP1 × TP4 | 1 / 1 | 4 / 4 | PASS |
| DP2 × TP2 | 4 / 4 | 16 / 16 | PASS |
| DP4 × TP1 | 16 / 16 | 64 / 64 | PASS |

Additional checks:

- PD output exactly matched non-PD output at the same DP/TP topology
- cancellation followed by a sentinel request passed for DP1, DP2, and DP4
- normal MiniLB traffic reached every DP rank
- no transfer mismatch, KV extraction, out-of-memory, or leaked-hold errors were observed

### Throughput

Setup: Qwen3-8B, random-token input length 4096, output length 1 or 256, 32 warmup requests, 64 measured requests, concurrency 32, fixed seed, radix cache disabled, and a fixed four-device allocation. Only the Raiden transfer engine is benchmarked for PD.

| Workload | Mode | DP1 × TP4 | DP2 × TP2 | DP4 × TP1 |
| --- | --- | ---: | ---: | ---: |
| 4096 → 1 | Raiden PD | 7.238 req/s | **10.878 req/s** | 7.247 req/s |
| 4096 → 1 | non-PD control | 13.421 req/s | **16.120 req/s** | 10.696 req/s |
| 4096 → 256 | Raiden PD | **5.204 req/s** | 4.808 req/s | 3.137 req/s |
| 4096 → 256 | non-PD control | **6.980 req/s** | 5.387 req/s | 4.200 req/s |

For 4096 → 1, Raiden PD improves by **50.3%** from DP1 to DP2. DP4 then drops **33.4%** versus DP2; the non-PD control drops **33.6%** over the same topology change. For 4096 → 256, DP4 is **39.7%** below DP1 with Raiden PD and **39.8%** below DP1 without PD. Because the same DP4 shape appears without PD, the regression is not specific to the Raiden transfer path; changing TP4 → TP2 → TP1 under a fixed device count also changes dense-model batching and operator efficiency.

## Current limitations

- Prefill and Decode must use the same `dp_size` and compatible per-rank endpoint shard layouts.
- Mixed pre-v4/v4 DP deployments are not supported. After a non-graceful restart, a pre-v4 Prefill registration may remain selectable until the 30-second heartbeat TTL expires; upgrade both sides before enabling DP > 1.
- A v4 DP Prefill paired with a pre-v4 Decode may surface incompatibility as a transfer timeout because the older Decode does not understand the per-rank transfer namespace.
- Normal MiniLB routing remains same-rank; a production cross-rank scheduling policy is not included.
- Cross-topology KV resharding is not included.
- Multi-process (`jax.process_count() > 1`) Prefill/Decode deployments have not yet been validated.
- Raiden remains an optional, ABI-matched installation through the existing #1490 build and deployment flow.

### DIFF
--- a/python/sgl_jax/srt/disaggregation/base/transfer.py
+++ b/python/sgl_jax/srt/disaggregation/base/transfer.py
@@ -51,6 +51,7 @@ class PrefillTransferContext:
     req_id: str
     transfer_id: str
     bootstrap_room: int | None
+    dp_rank: int
     buffer_id: int | None
     payload_factory: Callable[[], dict[str, Any]]
     block_ids_factory: Callable[[], list[int]]
@@ -71,6 +72,8 @@ class DecodeTransferContext:
     req_id: str
     transfer_id: str
     bootstrap_room: int | None
+    local_dp_rank: int
+    source_dp_rank: int
     peer_info: Mapping[str, object]
     kv_indices: Any
     page_size: int
@@ -112,13 +115,14 @@ class TransferBackend(Protocol):
 
     def try_start_decode(self, context: DecodeTransferContext) -> DecodeAdmission: ...
 
-    def prefill_transport_metadata(self) -> dict[str, object]: ...
+    def prefill_transport_metadata(self, dp_rank: int = 0) -> dict[str, object]: ...
 
     def cleanup_transfer(
         self,
         bootstrap_room: int | None,
         *,
         jax_process_index: int | None = None,
+        source_dp_rank: int = 0,
     ) -> None: ...
 
     def inflight_count(self) -> tuple[int, int]: ...
--- a/python/sgl_jax/srt/disaggregation/base/transfer.py
+++ b/python/sgl_jax/srt/disaggregation/base/transfer.py
@@ -22,8 +22,7 @@ def slots_to_page_ids(slots: Any, page_size: int, token_count: int) -> tuple[int
     values = np.asarray(slots).reshape(-1)
     if values.size < token_count:
         raise ValueError(
-            f"KV slot count is smaller than token count: slots={values.size}, "
-            f"tokens={token_count}"
+            f"KV slot count is smaller than token count: slots={values.size}, tokens={token_count}"
         )
     values = values[:token_count].astype(np.int64, copy=False)
     pages: list[int] = []
--- a/python/sgl_jax/srt/disaggregation/base/transfer.py
+++ b/python/sgl_jax/srt/disaggregation/base/transfer.py
@@ -71,8 +71,8 @@ class DecodeTransferContext:
     req_id: str
     transfer_id: str
     bootstrap_room: int | None
-    local_dp_rank: int
-    source_dp_rank: int
+    decode_dp_rank: int
+    prefill_dp_rank: int
     peer_info: Mapping[str, object]
     kv_indices: Any
     page_size: int
@@ -121,7 +121,7 @@ class TransferBackend(Protocol):
         bootstrap_room: int | None,
         *,
         jax_process_index: int | None = None,
-        source_dp_rank: int = 0,
+        prefill_dp_rank: int = 0,
     ) -> None: ...
 
     def inflight_count(self) -> tuple[int, int]: ...
--- a/python/sgl_jax/srt/disaggregation/bootstrap.py
+++ b/python/sgl_jax/srt/disaggregation/bootstrap.py
@@ -871,13 +871,16 @@ class HeartbeatDaemon:
 
     def _loop(self) -> None:
         while not self._stop_event.is_set():
+            self._heartbeat_once()
+            self._stop_event.wait(self._interval_s)
+
+    def _heartbeat_once(self) -> None:
+        for bootstrap_key in self._bootstrap_keys:
             try:
-                for bootstrap_key in self._bootstrap_keys:
-                    self._client.heartbeat(bootstrap_key)
+                self._client.heartbeat(bootstrap_key)
             except Exception:
                 logger.warning(
                     "bootstrap heartbeat for %s failed; will retry",
-                    self._bootstrap_keys,
+                    bootstrap_key,
                     exc_info=True,
                 )
-            self._stop_event.wait(self._interval_s)
--- a/python/sgl_jax/srt/disaggregation/bootstrap.py
+++ b/python/sgl_jax/srt/disaggregation/bootstrap.py
@@ -26,7 +26,7 @@ BOOTSTRAP_CAPABILITIES = ("transfer_metadata",)
 
 # PD wire protocol version. Bump when ``PrefillInfo``
 # or any of the 4 endpoint payloads change shape.
-PROTOCOL_VERSION: int = 3
+PROTOCOL_VERSION: int = 4
 MIN_COMPATIBLE_VERSION: int = 1
 
 
@@ -106,6 +106,8 @@ def check_prefill_compat(
     local_page_size: int,
     local_kv_dtype: str,
     expected_transfer_engine: str | None = None,
+    expected_dp_rank: int | None = None,
+    expected_dp_size: int | None = None,
 ) -> None:
     """Raise ``ValueError`` if the prefill peer's KV layout is incompatible.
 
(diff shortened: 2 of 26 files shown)


zorrofox (changes requested):
Reviewed on top of current main with the diff applied, full-context reads rather than hunks, and the disaggregation suite re-run locally (418 passed reproduced). The rank routing chain is consistent and symmetric end to end — aligned-path invariant at the LB, dual range validation at the tokenizer, rank-carrying registration/publish/cleanup on the sender, the (room, idx, source_rank) bootstrap namespace with re-validation at admission, and rank-tagged release on every teardown path — and the #1490 lifecycle contracts survive the per-rank split since the sender/receiver state machines themselves are untouched. Test breadth is genuinely good (the 4×4 source/destination matrix does exercise i≠j at the manager layer). Two blocking items below, both narrow.

### Blocking 1 — cross-rank head-of-line blocking on the per-rank KV budget path (`decode.py`, `_admit_decode_prealloc`)

The per-rank slot saturation path `continue`s (with a comment saying exactly why), but the per-rank KV budget path still `break`s: one rank's full pool blocks admission for every other rank's queued requests behind it in FIFO. Repro: dp=2, rank0 pool exhausted by long decodes, rank1 idle with a full pool — a rank0 request at the head of the prealloc queue starves rank1 indefinitely (guaranteed under `SGLANG_JAX_PD_FORCE_DP_RANK` skew; room-hash routing only makes it probabilistic). Since the budget is already rank-isolated, `continue` preserves per-rank FIFO and matches the slot path's intent. The per-rank inflight test covers the slot path but not this budget path — worth a twin case.

### Blocking 2 — one rank's engine exception drops other ranks' drained completion events (`raiden_transfer/wrapper.py`, `poll_stats`)

`poll_stats` iterates the per-rank engines with no per-engine isolation: if engine k raises, the sent/received/failed already returned by engines 0..k-1 are discarded with the exception (the caller catches around the whole call). If the engine's `poll_stats` has drain semantics — which the manager's `_done_sending.update()` accumulation pattern suggests, though I can't verify against tpu-raiden — those completion events are lost permanently, so transfers on healthy ranks sit in TRANSFERRING until pull timeout fails them collectively: a single-rank fault defeating exactly the rank isolation this PR builds, and a hole in the done_sending terminalization guarantee from #1490. A per-engine try/except that aggregates partial results and logs the faulty rank closes it. If poll_stats is instead idempotent/snapshot semantics, please correct me and this downgrades to a resilience nit.

### Non-blocking

1. Rolling-upgrade window: the registry key format change (`host:port` → `host:port:dp_N`) means a non-graceful old-P restart leaves a stale entry for up to the heartbeat TTL that `pick_for_room` can still select (defaults pass validation) — transient by TTL, but worth a line in the PR description; likewise a v4 dp>1 P behind a v3 D fails as pull-timeouts rather than a version error since MIN_COMPATIBLE_VERSION is unchanged.
2. `--test-external-dp-routing` validation echoes the request field back (`meta_info["dp_rank"]` comes from the request object, not the scheduler's actual placement), so it cannot catch the scheduler ignoring or reassigning the rank — the case it exists to test. Returning the scheduler-observed rank via the batch output would close the loop.
3. `_abort_decode_request` resolves the source rank inside `suppress(Exception)`; the intake failures most likely to reach this path (missing/invalid source rank) silently skip `cleanup_transfer`, leaking the bootstrap entry until the metadata TTL. A room-wide fallback cleanup (or at least a log) would help.
4. Question, tpu-raiden semantics: prefill-side `register_read` has no admission gate and each rank's engine now has only `ceil(max_inflight/dp)` slots, so skewed load can register more reads on one rank than its slot pool. What is the engine's behavior on slot exhaustion? (#1490 had the same shape globally; the per-rank split tightens the margin by dp×.)
5. `_extract_req_kv` gathers rank-local page ids against the data-sharded global pool, so the `SGL_JAX_PD_DEBUG_KV` digest is spuriously wrong for any dp_rank>0 request — debug-only, but it will mislead exactly the debugging sessions it exists for.
6. `_rank_local_array` is the riskiest new code in the PR (mesh surgery + zero-copy views) and is mocked out in the tests; a real-array unit test plus an explicit assertion that the KV spec actually shards the data axis (a replicated array silently returns every rank the same physical pages) would future-proof it. Smaller notes: the per-rank `ceil` makes the aggregate engine transient-buffer footprint dp×ceil > max_inflight (concurrency is still globally capped, the memory sizing isn't); `_admit_decode_prealloc` reaches into `disagg_transfer_queue._lock/._entries` — a small `count_by_rank()` accessor would keep the lock discipline in one place.

Happy to re-review once the two blocking items are addressed.

cjx0709:
@zorrofox Thanks for the thorough review! Both blockers are fixed now:

- KV/allocator pressure blocks only that Decode rank for the admission pass, so other ranks can keep moving without breaking rank-local FIFO. I added budget and allocator-shortfall tests too.
- Raiden polling is isolated per rank, so one engine error no longer drops events already drained from healthy ranks. I also confirmed `poll_stats` is destructive.

For the non-blocking notes: the rolling-upgrade limits are documented; abort cleanup now logs instead of failing silently; DP>0 debug KV mapping, real-array/data-axis coverage, and `count_by_rank()` are fixed. `register_read` itself does not reserve a slot; exhaustion happens when transfer starts and is surfaced as failure/timeout. The normal aligned MiniLB path stays within the per-rank cap, while production skewed cross-rank scheduling remains out of scope. The ceil over-allocation is bounded and global admission stays exact. The external-routing flag still checks request plumbing only; scheduler-observed placement is a follow-up.

Could you take another look when you get a chance?

zorrofox (approved):
Re-verified against the updated branch rather than the summary: the `capacity_blocked_ranks` approach is strictly better than the plain skip I suggested — it avoids cross-rank blocking while also preserving intra-rank FIFO (a smaller later request of a blocked rank can no longer jump its own queue), and both the budget check and the allocator shortfall feed it, each with a dedicated test. The per-engine poll isolation preserves drained events with a test to pin it, and thanks for confirming the destructive semantics — that makes the isolation load-bearing rather than defensive. The non-blocking follow-through is thorough too: the real-array `_rank_local_array` tests plus the replicated-data-axis rejection close the riskiest gap in the PR, the abort path now logs with a test, and the `register_read` answer (no slot held at registration; exhaustion surfaces at transfer start through the failure path) keeps slot pressure inside the existing leak-safe machinery. Deferring scheduler-observed placement verification is reasonable and clearly stated. `test_pd_decode` + `test_pd_raiden` re-run clean on my side. LGTM.

PR #1508, opened 2026-08-03
feat(dsa): fused sparse-MLA prefill for DeepSeek Sparse Attention (opt-in)

HadarIngonyama:
## Motivation

DeepSeek-family MLA models ship DeepSeek Sparse Attention (DSA) — a lightning indexer selects a top-k set of past tokens per query and attention runs only over that set — but sgl-jax had no attention path that used it for **prefill**: `--attention-backend dsa_sparse` selected tokens at decode but ran a **dense** prefill. On long contexts the prefill (EXTEND) step dominates and scales O(T²), so a dense prefill negates DSA exactly where it matters most.

This PR adds the **sparse prefill** path: a fused self-write + sparse-MLA Pallas kernel that attends only the indexer's selected pages during EXTEND. This cuts the **attention** term on the sparsified layers from O(T²) to O(T·K) (K = the fixed per-query page budget). It does **not** make prefill asymptotically linear: a residual O(T²) remains from the un-sparsified dense `skip_offset` layers (full causal MLA) and from the lightning indexer's dense page-scoring (O(T·S)). The win is a much smaller quadratic constant; a streaming indexer is a follow-up. See Benchmarking.

Scope: initial **single-sequence, single-shot** version. It does not yet support radix/prefix caching, request batching, or chunked prefill; the runtime fails fast with an actionable message if enabled outside that scope. Those are follow-ups.

## Modifications

- **New kernel** `srt/kernels/dsa/sparse_mla_prefill.py`: head-minor chunked sparse-MLA prefill kernel (selected keys on the MXU lane axis, few heads on the sublane axis — efficient under head-TP), `prefill_write_and_attend` (self-write latent into the paged cache, then attend the selected pages), and paged-cache test scaffolding. `sparse_mla_attention` now **requires an explicit `sm_scale`** (raises if `None`): the correct scale is the original `1/sqrt(qk_nope_head_dim + qk_rope_head_dim)` (= 1/√192 for DS/GLM), which the absorbed latent dim `Dk` does not encode.
- **Backend wiring** `srt/layers/attention/dsa_sparse_backend.py`: opt-in (`DSA_PREFILL_SPARSE=1`) routing of EXTEND through the sparse kernel, consuming the indexer page-topk directly and reusing it across shared layers (IndexShare). Default off ⇒ dense prefill, unchanged.
- **Scope guard** `srt/model_executor/model_runner.py`: with `DSA_PREFILL_SPARSE=1`, hard-error unless radix cache is disabled, `max_running_requests == 1`, and `chunked_prefill_size >= context_length`; plus a `DSA_INDEX_TOPK` env override for the selection budget. Gated on the opt-in flag ⇒ CI / dense / decode paths unaffected.
- **Docs** `docs/features/attention_backend.md`: document the `dsa_sparse` backend, enabling flags/env, the sparse-prefill scope + guard, and the O(T·K)-attention / residual-O(T²) asymptotics.

## Accuracy Tests

All numbers are on **GLM-5.2-FP8** (`zai-org/GLM-5.2-FP8`, public, **MIT** — a DSA model: DeepSeek-style sparse attention + IndexShare), served on **TPU v7x (Ironwood), 8 chips / 16 JAX devices, `tp16 ep16 dp1` (head-TP)**, `page_size=128`, `disable_radix_cache`, `max_running_requests=1`, `context_length=8192`, `mem_fraction_static=0.72`, MoE backend `epmoe`, bf16 compute (FP8 weights). Server + eval:

```bash
DSA_PREFILL_SPARSE=<1|0> DSA_INDEX_TOPK=4096 python -m sgl_jax.launch_server \
  --model-path zai-org/GLM-5.2-FP8 --trust-remote-code --device tpu --dtype bfloat16 \
  --tp-size 16 --dp-size 1 --ep-size 16 --moe-backend epmoe --attention-backend dsa_sparse \
  --nnodes 2 --node-rank <r> --dist-init-addr <ip:port> \
  --context-length 8192 --chunked-prefill-size 8192 --max-prefill-tokens 8192 \
  --page-size 128 --mem-fraction-static 0.72 --disable-radix-cache --max-running-requests 1
python3 test/srt/run_eval.py --host 127.0.0.1 --port 30000 --eval-name gsm8k \
  --num-examples 50 --temperature 0 --model zai-org/GLM-5.2-FP8
```

- **CI (CPU) parity** — `test/srt/kernels/dsa/test_sparse_mla_prefill_parity.py` checks the sparse-MLA prefill kernel vs a masked-softmax reference (`interpret=True`), flat and paged modes; no TPU required.
- **GSM8K no-regression control (standard, short prompts).** n=50, greedy (temp 0), **one server flipping only `DSA_PREFILL_SPARSE`**:

  | prefill path | GSM8K (n=50) |
  |---|---|
  | DSA sparse (`DSA_PREFILL_SPARSE=1`) | **47/50 (0.94)** |
  | dense control (`DSA_PREFILL_SPARSE=0`) | **47/50 (0.94)** |

  Identical. Standard GSM8K prompts (~100–300 tokens) fall below `index_topk=4096` (32 pages), so the sparse path selects *all* pages and is mathematically equal to dense — the matched score confirms the fused self-write + kernel introduce **no regression** on the served path (any sparse≠dense here would be a pure plumbing bug).
**Tightened to n=200 (on #1508+#1512).** Re-ran the same `run_eval` gsm8k control at 4× the sample size on the combined single-seq stack (`test/srt/run_eval.py --eval-name gsm8k --num-examples 200`, temp 0, no thinking, `max_running=1`, radix off):

  | prefill path | GSM8K run_eval (n=200) |
  |---|---|
  | DSA sparse (`DSA_PREFILL_SPARSE=1`) | **0.960 (192/200)** |
  | dense control (`DSA_PREFILL_SPARSE=0`) | **0.980 (196/200)** |

  Δ −0.020 (4/200, inside the n=200 noise band) — no meaningful regression at the tighter n=200 CI (same harness as the n=50 row above). Standard prompts stay under the `index_topk=4096` budget, so this is the plumbing/code-path check; selection-under-pruning is the long-context pair below.
- **Long-context recall (NIAH) — selection quality.** NIAH 5×5 grid, ctx ≈ 8k, same server/config:

  | Config | Selected pages | NIAH recall |
  |---|---|---|
  | Dense reference (full coverage) | 57 | 19/25 (76%) |
  | DSA `index_topk=4096` | 32 | 18/25 (72%) |
  | DSA `index_topk=3072` | 24 | 17/25 (68%) |
  | DSA `index_topk=2048` | 16 | 17/25 (68%) |

  DSA tracks dense within 1–2 needles even at ~3.6× page pruning; every miss is coherent text (no garbage). The absolute dense 19/25 is low for a 1M-context model like GLM-5.2 and reflects the strictness of our 25-sample single-needle grid (~4% grid noise), **not** model quality — read this as a *sparse-vs-dense delta*, not an absolute recall benchmark.
- **Long-context GSM8K — paired sparse-vs-dense under pruning.** `gsm8k_longctx.py` builds ~7k-token many-shot prompts so the indexer genuinely prunes (57→32 pages). One server, flipping only `DSA_PREFILL_SPARSE`, `index_topk=4096`, greedy, n=50:

  | prefill path | flag | long-ctx GSM8K (n=50) |
  |---|---|---|
  | DSA sparse | `DSA_PREFILL_SPARSE=1` | 43/50 (0.86) |
  | dense control | `DSA_PREFILL_SPARSE=0` | 42/50 (0.84) |

  Sparse ≥ dense — no accuracy collapse when pages are actively dropped. The low absolutes were a decode-budget artifact, not selection quality: at `max_tokens=900`, ~8/50 generations truncated inside GLM-5.2's thinking block before emitting `Answer:`. Re-running truncation-free (`context_length=12288`, `max_tokens=3072`, same 50 prompts, 1/50 truncated each) gives sparse **50/50** vs dense **48/50** — the pair holds and both jump once decode isn't capped. The n=50 long-ctx pair is a coarse check for the *selection* signal; the tightened **n=200** short-context control (above, on #1508+#1512) shows no regression at 4× the sample size, and the full-`evalscope` (1319) pass rides with the batching PR #1514. Extending this long-context pair to n=200 is the deferred long-context follow-up.

## Benchmarking and Profiling

Single-shot prefill wall time. **Two servers identical except `DSA_PREFILL_SPARSE`** (1 = sparse, 0 = dense), `index_topk=4096`. Config: **TPU v7x, 8 chips / 16 devices, `tp16 ep16 dp1` (head-TP)**, `page_size=128`, `disable_radix_cache`, `max_running_requests=1`, `mem_fraction_static=0.60`; **dense baseline `chunked_prefill_size = max_prefill_tokens = context` (single-shot)**. All three contexts are prompt lengths on **one `context_length=49152` server** (key span S pinned at 49k — see note), with the **prefill token-padding bucket pinned to each prompt length** via `--precompile-token-paddings 16384/32768/49152` — required, since the sparse-MLA prefill kernel grids over the padded bucket, so its cost tracks the bucket, not the actual prompt.

| Context (T) | Dense prefill | DSA sparse prefill | Speedup |
|---|---|---|---|
| 16k | 19.1 s | 10.5 s | 1.82× |
| 32k | 73.0 s | 23.1 s | 3.16× |
| 48k | 155.7 s | 38.4 s | 4.05× |

Dense scales ∝ T² (O(T²) causal MLA on all 78 layers). DSA replaces that with O(T·K) sparse attention on the 72 sparsified layers, but is **not** asymptotically linear — prefill stays O(T²) with a much smaller constant, from (a) the dense `skip_offset` layers and (b) the lightning indexer's O(T·S) page-score einsum. The widening 1.82×→4.05× gap is the smaller quadratic *constant* (6 dense layers vs 78), **not** linear scaling.

**Limitation (single-shot bucketing).** Because the kernel's cost tracks the padded bucket, a prompt that lands between the default paddings (`[64…8192]` + `context_length`) pads up to the next bucket: e.g. a 16k prompt at defaults pads to 49152 and runs at that cost (~34.7 s, slower than dense) purely from padding. Single-shot prefill keeps the bucket under the operator's control (the pinning above); the general fix is **chunked prefill (#1514)**, which splits the prompt to `chunked_prefill_size` so it never lands in an oversized bucket. A native ragged kernel gridding over actual tokens (`cu_seqlens`) is future work.

Per-layer device time, 16k→48k (T 16384→49152 = 3.0×; dense from xprof, attend/indexer measured directly in isolation):

| phase | @16k | @48k | scaling |
|---|---|---|---|
| dense MLA (`MLA-m`) | 224 ms | 2014 ms | **9.0× ⇒ O(T²)** |
| sparse attend | 86 ms | 245 ms | 2.84× (O(T·K)) |
| indexer select | 72 ms | 212 ms | 2.95× (O(T·S), S pinned ⇒ linear in T) |
| self-write (fused, in-place) | <1 ms | <3 ms | ~3× (O(T)) |

**Note — the dense-MLA row is a routing artifact, not a model requirement.** Those `MLA-m` layers are the first `index_skip_topk_offset` (=3) layers, which `config.indexer_types` actually marks as `"full"` (own-indexer **sparse**), plus their shared followers (3–5) that fall to dense only because the dense layers seed them no pages. Dropping the one-line `skip_offset` dense gate runs **all 78 layers sparse** and removes this O(T²) term outright (measured ~25% faster at 48k, GSM8K/needle correctness preserved) — so the dense MLA here is a removable leftover, not inherent to DSA prefill.

× 6 / 72 / 21 layers ⇒ 1.35 / 6.2 / 1.5 s @16k, 12.1 / 17.6 / 4.5 s @48k. **Fixed-S note:** because all points ran on one 49k server, the indexer's key span S is pinned at 49,152, so its O(T·S) reads as linear in T *here*; on a right-sized server (S ∝ T) the indexer becomes a **second O(T²)** term (matched-S: 24.5→212 ms 16k→48k = 8.66×), so this table is a **lower bound** on DSA's short-context speedup (the 16k indexer is ~3× over-provisioned). The **self-write is in-place** — the KV pool is donated (`jitted_run_model(donate_argnames=["memory_pools"])`), and on a 16k trace it is `scatter_custom_fusion` = 25 ms (0.82% of prefill), not a GB-scale copy. 64k single-shot prefill did not fit HBM on this hardware (documented ceiling).

### DIFF
--- a/docs/features/attention_backend.md
+++ b/docs/features/attention_backend.md
@@ -4,13 +4,14 @@ SGL-JAX exposes a small user-facing attention backend switch, while the runtime 
 
 ## User-facing choices
 
-`--attention-backend` accepts three values:
+`--attention-backend` accepts four values:
 
 | Value | Runtime behavior |
 |---|---|
 | `fa` | Default. Uses FlashAttention for MHA/GQA models and the absorbed MLA Pallas backend for MLA models. |
 | `fa_mha` | Forces MLA models through the decompressed MHA FlashAttention path. This is useful for kernel A/B checks, but uses much more KV cache than absorbed MLA. |
 | `native` | Pure JAX/native attention path, mainly for CPU/debugging. If `fa` or `fa_mha` is requested on CPU, the runtime falls back to `native`. |
+| `dsa_sparse` | DeepSeek Sparse Attention (DSA): a lightning-indexer selects the top-scoring pages of past tokens per query and attention runs only over that page set (page-block sparse guided by DSA scores), with IndexShare cross-layer reuse of the selection. For MLA models whose config carries the `index_*` fields only. See [DeepSeek Sparse Attention](#deepseek-sparse-attention-dsa_sparse). |
 
 Example:
 
@@ -28,11 +29,82 @@ python3 -u -m sgl_jax.launch_server \
 |---|---|---|
 | `FlashAttention` | `--attention-backend=fa` for MHA/GQA, or `fa_mha` for MLA fallback | TPU production attention with paged KV cache, SWA metadata, and Pallas kernels. |
 | `MLAAttentionBackend` | `--attention-backend=fa` when `model_config.attention_arch == MLA` | Absorbed MLA path for DeepSeek-family models. |
+| `DSASparseAttentionBackend` | `--attention-backend=dsa_sparse` for MLA models with `index_*` config | DeepSeek Sparse Attention: lightning-indexer top-k + sparse MLA over the selected pages (page-block), with IndexShare. |
 | `NativeAttention` | `--attention-backend=native`, or CPU fallback | Debugging and CPU execution. |
 | `HybridLinearAttnBackend` | Automatic wrapper for hybrid recurrent models | Routes full-attention layers to `FlashAttention`/`MLAAttentionBackend` and linear recurrent layers to KDA/GDN/Lightning backends. |
 | `KDAAttnBackend` | Automatic under `HybridLinearAttnBackend` for Kimi Linear | Kimi Delta Attention recurrent branch. |
 | `GDNAttnBackend` | Automatic under `HybridLinearAttnBackend` for Qwen3.5 hybrid configs | Gated DeltaNet recurrent branch. |
 | `LightningAttnBackend` | Automatic under `HybridLinearAttnBackend` for Bailing MoE V2.5 / Ling-2.6-flash | Lightning / Simple GLA recurrent branch. |
+
+## DeepSeek Sparse Attention (`dsa_sparse`)
+
+DSA cuts the cost of long-context attention by attending to a **selected subset** of past
+tokens instead of the full causal history. A small "lightning indexer" scores past tokens
+for each query and the highest-scoring ones (budget `index_topk`) are chosen. The selection
+is computed on the indexer layers and **reused across the following layers (IndexShare)**,
+so the indexer cost is amortized.
+
+Selection here is at **page granularity** (`page_size` tokens per unit), not exact per-token:
+the indexer's per-token scores are max-pooled to a score per page and the top
+`ceil(index_topk / page_size)` pages are attended in full. This is therefore **page-block
+sparse attention guided by the DSA indexer scores** — a superset of the exact token top-`k`
+set (attending whole pages shifts the softmax denominator), not a bit-exact token-level DSA.
+In practice it tracks dense closely on long-context retrieval (see the PR's NIAH/GSM8K
+results); treat page size and `index_topk` as the accuracy/speed knobs.
+
+Only MLA models whose config carries the DSA `index_*` fields (indexer head dim / heads /
+top-k) are eligible; other models ignore the flag.
+
+### Enabling it
+
+```bash
+python3 -u -m sgl_jax.launch_server \
+  --model-path <deepseek-sparse-MLA-checkpoint> \
+  --device=tpu \
+  --attention-backend=dsa_sparse \
+  --dsa-use-pallas          # use the Pallas kernels (default: the jnp reference)
+```
+
+Two environment variables tune the sparse path:
+
+| Env var | Meaning |
+|---|---|
+| `DSA_INDEX_TOPK` | Override the config's `index_topk` (selection budget, in tokens) at backend construction. Larger = closer to dense, slower. The kernel is compiled for this value, so **changing it requires a relaunch**. |
+| `DSA_PREFILL_SPARSE=1` | Opt in to running **prefill** through the fused sparse-MLA prefill kernel (below). Default off ⇒ prefill runs dense and only decode is sparse. |
+
+### Sparse prefill and its current scope
+
+With `DSA_PREFILL_SPARSE=1`, the prefill (EXTEND) step also attends only to the indexer's
+selected pages via a fused self-write + sparse-MLA Pallas kernel, which is where the
+long-context prefill speedup comes from.
+
+This initial version supports a **single-sequence, single-shot** prefill only. It does
+**not** yet support radix/prefix caching, request batching, or chunked prefill (a cache
+hit / multi-seq / split prompt turns prefill into a partial extend the sparse page table
+cannot represent). The runtime **fails fast with an actionable message** when
+`DSA_PREFILL_SPARSE=1` is combined with an incompatible config, so it must be launched
+with:
+
+```bash
+DSA_PREFILL_SPARSE=1 DSA_INDEX_TOPK=4096 python3 -u -m sgl_jax.launch_server \
+  --model-path <checkpoint> --device=tpu \
+  --attention-backend=dsa_sparse --dsa-use-pallas \
+  --disable-radix-cache \           # radix/prefix cache OFF
+  --max-running-requests 1 \        # single sequence
+  --context-length 49152 \
+  --chunked-prefill-size 49152      # >= context => single-shot (no chunking)
+```
+
+Radix-cache, batching, and chunked-prefill support are follow-ups. The guard is gated on
+`DSA_PREFILL_SPARSE`, so the default (dense-prefill) `dsa_sparse` path, other backends, and
+CI are unaffected.
+
+### Validation
+
+Kernel correctness is covered in CI by
+`test/srt/kernels/dsa/test_sparse_mla_prefill_parity.py`, which checks the sparse-MLA
+prefill kernel against a masked-softmax reference on CPU (`interpret=True`, no TPU needed),
+in both flat and paged-cache modes.
 
 ## Notes for contributors
 
--- a/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill.py
+++ b/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill.py
@@ -1,0 +1,451 @@
+"""TPU Pallas kernel for **sparse MLA-latent attention** (the DSA attention-consuming
+path) — the fused sparse-MLA *prefill* kernel.
+
+Given, per query, a set of selected past tokens (the lightning-indexer top-k), attend
+**only** to those tokens over the MLA *latent* cache. Prefill (S queries) and decode
+(S == 1) share the one kernel; this module is the prefill entry point.
+
+Design (mirrors DeepSeek's ``refs/dsa/tilelang_sparse_mla_fwd.py`` *algorithm*, not
+its GPU code):
+
+* **Sparse == dense-MLA with a gathered KV iteration.** The only structural change
+  from a dense paged-MLA kernel is that the inner loop walks the *selected* KV rows
+  (named by ``indices``) instead of all of them, with the usual flash/online-softmax
+  accumulation.
+* **Head-sharing is what makes it viable on TPU.** MLA caches a single latent vector
+  per token (``[Dk] = kv_lora_rank + qk_rope_head_dim = 512 + 64 = 576``) that is
+  reused by *all* query heads, so each fetched row feeds a fat ``[H, Dk]`` matmul.
+  The "value" is the first ``kv_lora_rank`` dims of the same latent.
+* **One kernel, prefill + decode:** the grid is over ``(batch, query_token)``;
+  decode is just ``S == 1``. Queries are *not* batched into a shared-K matmul because
+  each query has its own selection — parallelism comes from the head dim per query.
+* **Static counts, dynamic addresses.** The number of selected units is fixed
+  (``K``); only the *addresses* (which rows) are data-dependent, read from ``indices``
+  on-chip and used to drive per-unit DMAs from HBM.
+
+Selection granularity is parameterised by ``read_block`` (``RB``): ``indices[b, s, :]``
+are **unit ids**, one unit == ``RB`` contiguous tokens. Page-level selection sets
+``RB == page_size`` and consumes the indexer's page-topk directly; effective attended
+tokens per query == ``K * RB``.
+
+Only the **head-minor chunked** kernel is deployed (few heads/device, i.e. head-TP):
+``G`` selected units are gathered into one ``[CBR, Dk]`` tile with the abundant keys on
+the 128-wide MXU lane axis and the few heads on the sublane axis, so one MXU-filling
+matmul runs per chunk. (The v1 per-unit / non-head-minor / chunk-pipelined A/B
+baselines from the perf-characterization sweep are kept in git history only; they are
+not part of this deployed kernel.)
+"""
+
+from __future__ import annotations
+
+import functools
+
+import jax
+import jax.numpy as jnp
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+
+
+def _sparse_mla_kernel_chunked_hminor(
+    q_ref,  # [1, 1, Hp, Dk]  Hp padded to the bf16 sublane tile (16), NOT to 128
+    idx_ref,  # [1, 1, 1, K]    SMEM
+    pos_ref,  # [1, 1, 1, 1]    SMEM
+    kv_hbm,  # flat: [B, T(+RBF), Dk] HBM;  paged: [1, num_pages*page_size, Dk] HBM
+    pt_ref,  # [1, 1, 1, max_pages] SMEM  page table for this request (paged only; dummy if flat)
+    o_ref,  # [1, 1, Hp, Dv]
+    kv_scratch,  # [CBR, Dk] VMEM  one chunk of gathered latent
+    sem,  # DMA semaphores (G,)
+    *,
+    sm_scale: float,
+    Dv: int,
+    RB: int,
+    RBF: int,
+    T: int,
+    K: int,
+    G: int,
+    CBR: int,  # G*RBF rounded up to the 128-lane tile (score's key axis is the lane axis)
+    paged: bool = False,  # read from the packed 4D paged latent cache via pt_ref
+    page_size: int = 0,  # tokens per page (paged only; static)
+):
+    """Head-minor relayout of the chunked kernel for **few heads/device** (head-TP).
+
+    A default chunked layout would put heads on the 128-wide MXU lane axis (score
+    ``[CBR, Hp]``), wasting ~97% of the MXU when a device holds only ~4 heads. Here we
+    swap the axes: the abundant **selected keys sit on the lane axis** (``CBR`` filling
+    128) and the few **heads sit on the sublane axis** (padded to 16), so 4 heads cost
+    a 4x sublane pad instead of a 32x lane pad. Score is ``[Hp, CBR]``, softmax reduces
+    over the key (lane) axis, value matmul contracts ``CBR``.
+
+    The validity mask lives on the *lane* axis (keys). A lane concat of ``RBF``-wide
+    pieces is not Mosaic-lowerable, so we build the ``[CBR]`` lane mask **arithmetically**
+    from an on-chip ``iota`` with per-unit **static-window** compares (no concat, no
+    dynamic gather, no integer div/mod on a vector): each chunk selects every lane's
+    unit id / row offset with ``G`` broadcasted ``where`` ops, recovers key positions,
+    and turns the boolean into a ``0 / -inf`` additive bias."""
+    b = pl.program_id(0)
+    Hp = q_ref.shape[2]
+    q = q_ref[0, 0]  # [Hp, Dk] bf16
+    qpos = pos_ref[0, 0, 0, 0]
+    NC = (K + G - 1) // G
+
+    lane = jnp.arange(CBR, dtype=jnp.int32)  # [CBR] iota on the key (lane) axis
+
+    def _src(u):
+        """HBM source slice ([RBF, Dk]) for selected unit ``u`` (logical unit id)."""
+        if paged:
+            # unit u == RB contiguous LOGICAL tokens starting at u*RB. RB divides
+            # page_size, so the whole unit lives in one physical page. The 4D cache
+            # is flattened to [num_pages*page_size, Dk]; token (page P, offset o) is
+            # row P*page_size + o. RB%16==0 => RBF==RB => no cross-page over-fetch.
+            #
+            # The token axis is sublane-tiled (8), so a *dynamic* slice offset must be
+            # provably %8. row0 = pp*page_size + o0 is always a multiple of RB(>=16),
+            # but pp (from the page table) is opaque to Mosaic. Express the offset as
+            # an explicit `r8 * 8` (page_size and o0 are both %8) so divisibility is
+            # provable — the same idiom the dense paged kernel uses.
+            lt = u * RB  # logical token start
+            lp = lt // page_size  # logical page (static divisor)
+            o0 = lt - lp * page_size  # in-page offset (multiple of RB)
+            pp = pt_ref[0, 0, 0, lp]  # physical page id
+            r8 = pp * (page_size // 8) + o0 // 8  # (P*page_size + o0) // 8, exact
+            return kv_hbm.at[0, pl.ds(r8 * 8, RBF), :]
+        return kv_hbm.at[b, pl.ds(u * RB, RBF), :]
+
+    # padded lanes (>= G*RBF) are never DMA'd; zero them so a stale/NaN row can't
+    # poison the (masked-out) score before the -inf bias is added.
+    if CBR > G * RBF:
+        pad = CBR - G * RBF
+        kv_scratch[pl.ds(G * RBF, pad), :] = jnp.zeros((pad, kv_scratch.shape[1]), kv_scratch.dtype)
+
+    m0 = jnp.full((Hp,), -jnp.inf, dtype=jnp.float32)
+    l0 = jnp.zeros((Hp,), dtype=jnp.float32)
+    acc0 = jnp.zeros((Hp, Dv), dtype=jnp.float32)
+
+    def chunk_body(c, carry):
+        m_i, l_i, acc = carry
+        base = c * G
+
+        # --- issue G gathers (overlap) into contiguous RBF-aligned lane slots ---
+        for g in range(G):
+            gidx = jnp.minimum(base + g, K - 1)
+            # clamp negative unit ids (topk padding == -1) to 0 for the DMA so the
+            # source offset stays in range; the lane is masked out below anyway.
+            u = jnp.maximum(idx_ref[0, 0, 0, gidx], 0)
+            dst = kv_scratch.at[pl.ds(g * RBF, RBF), :]
+            pltpu.make_async_copy(_src(u), dst, sem.at[g]).start()
+
+        # --- build the [CBR] lane validity bias arithmetically while DMAs run ---
+        # Accumulators are int32 (0/1), NOT bool: Mosaic can't materialise an i1 vector
+        # from a broadcast-scalar select (trunc i8->i1 is unsupported). The i1 predicate
+        # is only formed transiently by the final comparisons feeding the select.
+        u_vec = jnp.zeros((CBR,), jnp.int32)  # per-lane selected unit id
+        row_vec = jnp.zeros((CBR,), jnp.int32)  # per-lane row-within-unit (0..RBF-1)
+        ir_vec = jnp.zeros((CBR,), jnp.int32)  # 1 where the lane's unit id < K
+        for g in range(G):
+            lo = g * RBF  # static lane window for unit g
+            sel = (lane >= lo) & (lane < lo + RBF)  # [CBR] const partition (transient i1)
+            u_g = idx_ref[0, 0, 0, jnp.minimum(base + g, K - 1)]
+            inr = ((base + g) < K).astype(jnp.int32)  # scalar 0/1
+            u_vec = jnp.where(sel, u_g, u_vec)
+            row_vec = jnp.where(sel, lane - lo, row_vec)
+            ir_vec = jnp.where(sel, inr, ir_vec)
+        kp = u_vec * RB + row_vec  # [CBR] key positions
+        # (u_vec >= 0) drops topk padding lanes (unit id -1); their DMA was
+        # clamped to unit 0, so the read is safe and only the mask excludes them.
+        valid = (u_vec >= 0) & (row_vec < RB) & (ir_vec > 0) & (kp <= qpos) & (kp < T)
+        bias = jnp.where(valid, 0.0, -jnp.inf)  # [CBR] fp32
+
+        for g in range(G):
+            gidx = jnp.minimum(base + g, K - 1)
+            u = jnp.maximum(idx_ref[0, 0, 0, gidx], 0)
+            dst = kv_scratch.at[pl.ds(g * RBF, RBF), :]
+            pltpu.make_async_copy(_src(u), dst, sem.at[g]).wait()
+
+        kv_blk = kv_scratch[pl.ds(0, CBR), :]  # [CBR, Dk] bf16
+
+        # score: [Hp,Dk]·[CBR,Dk] -> [Hp, CBR]  (heads on sublane, keys on lanes)
+        s = (
+            jax.lax.dot_general(
+                q, kv_blk, (((1,), (1,)), ((), ())), preferred_element_type=jnp.float32
+            )
+            * sm_scale
+        )
+        s = s + bias[None, :]  # [Hp, CBR] fp32
+
+        # --- flash-softmax over keys = axis 1 (the lane axis) ---
+        m_cur = jnp.max(s, axis=1)  # [Hp]
+        m_new = jnp.maximum(m_i, m_cur)
+        corr = jnp.where(jnp.isneginf(m_new), 0.0, m_new)
+        p = jnp.exp(s - corr[:, None])  # [Hp, CBR]
+        alpha = jnp.where(jnp.isneginf(m_new), 1.0, jnp.exp(m_i - m_new))
+        l_i = l_i * alpha + jnp.sum(p, axis=1)  # [Hp]
+        v_blk = kv_blk[:, :Dv]  # [CBR, Dv]
+        # acc[Hp,Dv] += sum_cbr p[hp,cbr] * v[cbr,dv]
+        acc = acc * alpha[:, None] + jax.lax.dot_general(
+            p.astype(kv_blk.dtype),
+            v_blk,
+            (((1,), (0,)), ((), ())),
+            preferred_element_type=jnp.float32,
+        )
+        return (m_new, l_i, acc)
+
+    m_i, l_i, acc = jax.lax.fori_loop(0, NC, chunk_body, (m0, l0, acc0))
+    out = acc / jnp.where(l_i == 0.0, 1.0, l_i)[:, None]
+    o_ref[0, 0] = out.astype(o_ref.dtype)
+
+
+def sparse_mla_attention(
+    q,  # [B, S, H, Dk]        Dk = kv_lora_rank + qk_rope_head_dim
+    kv,  # [B, T, Dk]           MLA latent cache (value = first Dv cols)
+    indices,  # [B, S, K] int32      selected unit ids (unit == RB tokens)
+    positions,  # [B, S] int32         query positions (causal bound)
+    *,
+    kv_lora_rank: int = 512,
+    read_block: int = 1,
+    block_units: int | None = None,  # units gathered+matmul'd per chunk (G; parallelism knob)
+    sm_scale: float | None = None,
+    interpret: bool = False,
+    page_table=None,  # [B, max_pages] int32: logical page -> physical page. When set,
+    # ``kv`` is the packed 4D paged cache and the kernel gathers from it.
+    page_size: int | None = None,  # tokens/page (required with page_table)
+    seq_len: int | None = None,  # logical T for the causal mask (required with page_table)
+):
+    """Sparse MLA-latent attention (head-minor chunked kernel).
+    Returns ``[B, S, H, kv_lora_rank]`` (float32).
+
+    ``indices`` are *unit ids* (one unit = ``read_block`` contiguous tokens); set
+    ``read_block=1`` for token-granular (exact) DSA, or e.g. 128 for page-sized block
+    reads. Effective attended tokens per query == ``K * read_block``. ``block_units=G``
+    (>=1) sets how many units are gathered into one MXU-filling matmul per chunk
+    (``G>=K`` ⇒ a single chunk, the fastest measured config).
+
+    KV source:
+    * default: ``kv`` is a flat ``[B, T, Dk]`` latent buffer.
+    * paged (``page_table`` set): ``kv`` is the sglang-jax packed 4D MLA cache
+      ``[num_pages, align(page_size,packing)//packing, packing, align(lkv,128)+align(rope,128)]``.
+      Flattened to ``[num_pages*page_size, Dk_pad]``, token (physical page P, offset o)
+      is row ``P*page_size + o``; the kernel resolves each selected unit through
+      ``page_table`` and requires ``read_block % 16 == 0`` (so a unit never over-fetches
+      across a page).
+    """
+    B, S, H, Dk = q.shape
+    K = indices.shape[2]
+    Dv = kv_lora_rank
+    RB = read_block
+    paged = page_table is not None
+    if not block_units or block_units < 1:
+        raise ValueError("sparse_mla_attention requires block_units >= 1")
+    if sm_scale is None:
+        sm_scale = 1.0 / (Dk**0.5)  # note: true (unpadded) Dk
+
+    # TPU/Mosaic wants the lane (last) dim %128; the MLA latent is 576 = 512 + 64
+    # rope, so pad the feature dim with zeros to the next multiple of 128 (640).
+    # Padding the *contraction* dim with zeros leaves q·k unchanged, and the value
+    # is only the first Dv (=512) dims, so the pad never enters the output.
+    Dk_pad = ((Dk + 127) // 128) * 128
+    q = jnp.pad(q, ((0, 0), (0, 0), (0, 0), (0, Dk_pad - Dk))) if Dk_pad != Dk else q
+    if paged:
+        if RB % 16 != 0:
+            raise ValueError("paged KV requires read_block % 16 == 0 (no cross-page over-fetch)")
+        if page_size is None or seq_len is None:
+            raise ValueError("paged KV requires page_size and seq_len")
+        if kv.shape[-1] != Dk_pad:
+            raise ValueError(f"paged cache last dim {kv.shape[-1]} != Dk_pad {Dk_pad}")
+        T = seq_len
+        # flatten [num_pages, ps//packing, packing, Dk_pad] -> [1, num_pages*page_size, Dk_pad]
+        num_pages = kv.shape[0]
+        kv = kv.reshape(1, num_pages * page_size, Dk_pad)
+    else:
+        T = kv.shape[1]
+        if Dk_pad != Dk:
+            kv = jnp.pad(kv, ((0, 0), (0, 0), (0, Dk_pad - Dk)))
+
+    # Add a singleton axis so the per-(b,s) block's trailing two dims are legal
+    # (block over the S axis is 1, which is illegal as a *second-to-last* dim).
+    indices4 = indices.reshape(B, S, 1, K)
+    positions4 = positions.reshape(B, S, 1, 1)
+
+    # head-minor relayout: keys on the lane axis (fills the MXU even with ~4 heads),
+    # heads on the sublane axis (padded to the bf16 sublane tile of 16, not 128).
+    G = max(1, min(block_units, K))
+    RBF = ((RB + 15) // 16) * 16
+    if paged:
+        ps = page_size  # RBF == RB (RB%16==0) => no over-fetch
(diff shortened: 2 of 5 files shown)


zorrofox (commented):
Thanks for taking this on — the EXTEND-sparsity gap is real and this is the right shape for a first cut (opt-in flag, fail-fast scope guard, CPU parity test). A few findings from our side, ordered by importance; only #1 and #2 might change code in this PR, the rest are follow-ups/wording.

### 1. Self-write via functional scatter — please confirm it's not copying the layer cache

`prefill_write_and_attend` writes the chunk's latent with `flat.at[loc].set(...)` on the full per-layer fused KV buffer. Under the jitted forward this is only in-place if XLA donates/aliases that buffer end-to-end through the `shard_map`; if it doesn't, every wired layer copies a GB-scale buffer once per prefill. When we profiled the analogous functional scatter on the (much smaller) indexer-key cache during #1478 development, it showed up as a real copy cost rather than being elided. Could you share an xprof slice (or buffer-donation check) confirming the write is in-place? If it copies, moving the self-write into the kernel via `input_output_aliases` (like the dense MLA path does) is the fix — and it might also explain part of the absolute wall times in the benchmark table.

### 2. Benchmark table needs the config to be interpretable

19.1 s dense prefill at 16k is strongly config-dependent (chip count, TP layout, `chunked_prefill_size` for the dense baseline, page_size). Could you add those to the table? Related: which model are the accuracy numbers from? "744B DeepSeek-style MoE MLA" doesn't match a public checkpoint (V3-family is 671B), and the GSM8K smoke doesn't name the model or the server command at all — as it stands none of the accuracy data is externally reproducible, and dense NIAH at 19/25 (~8k ctx) is low enough that it's hard to tell model quality from setup issues. Two asks: (a) please add one accuracy round on a public DSA checkpoint (sparse vs dense, same server config, flag on/off) — with the current internal-only model there is no reference score to anchor against, so reviewers can't distinguish "expected for this checkpoint" from "regression introduced by the sparse path"; naming the internal model helps but doesn't make the numbers verifiable. (b) add the paired dense-server GSM8K control under the identical config (same server, flag off, n=50) — that's the cheap way to make the "no regression" claim concrete, and it's needed on the public checkpoint from (a) anyway.

### 3. "toward ~linear" overstates the asymptotics (wording + breakdown)

The attend side becomes O(T·K·RB), but `_maybe_index_prefill_pages` still runs the jnp page-topk reference with `one_token_per_seq=False` — an O(T·S) score einsum per full (indexer) layer. So e2e stays quadratic with a smaller constant; the 16k→48k speedup growth is the dense-MLA constant shrinking, and the curves flatten to a fixed ratio once the indexer dominates. Suggest rewording, and a one-line time breakdown (indexer / attend / write) at 48k would show where the next bottleneck is. FWIW we are validating a Pallas streamindex path (scores computed at O(actual kv_len) per query block, prefill/mixed grids supported) that would attack exactly that term — happy to coordinate so the two land compatibly.

### 4. Follow-up: cross-query KV reuse

Each (b, s) grid step re-gathers its own K pages with no reuse, but adjacent queries' page selections overlap heavily in practice. Batching a block of queries against the union of their selected pages (dense-prefill-style bq tiling) is likely the next integer-factor win, and would also amortize the per-step DMA setup. Fine as a TODO — just worth writing down.

### 5. nit: `sm_scale` default uses the latent dim

`sparse_mla_attention` defaults `sm_scale = 1/sqrt(Dk)` with `Dk` = latent (576). In the absorbed domain the score is still the original q·k inner product, so the right default is `1/sqrt(qk_nope_head_dim + qk_rope_head_dim)` (= 1/√192 for DS/GLM shapes). Callers pass it explicitly today, so this is latent-footgun only.

### Things that look right

- The `wrap_negative_indices=False` catch on the self-write (and the comment explaining why `mode="drop"` alone is insufficient) — good find, that one would have silently corrupted the last physical slot.
- IndexShare threading works for the prefill shapes: the model loop threads `topk` and `topk_pages` independently, so `topk=None, topk_pages=[T,k]` on full layers propagates correctly to shared layers.
- CPU `interpret=True` parity test means CI covers the kernel without TPU.
- Scope guard fails fast with an actionable message instead of silently producing wrong results outside the supported envelope.

omershlo:
Thank you @zorrofox for the code review - we will address your comments asap! 
We are currently working on a follow up PR to optimize the indexer and support batching. Aiming for next week. 
We would love to get your guidence !  

cc: @alexnails

HadarIngonyama:
@zorrofox Thanks! Genuinely useful review. Here is what changed, per point:

**One question back on the `skip_offset` dense gate (from #1478).** Looking at the profile I ran for #1 (see image) I noticed the first 6 layers were running dense attention instead of DSA. Looking at the code I saw the backend routes the first `index_skip_topk_offset` layers through `_run_dense` (`dsa_sparse_backend.__call__`: `if layer_id < self.skip_offset ...`). For GLM-5.2 that's layers 0–2, which `config.indexer_types` marks as **`"full"`** (`['full','full','full','shared',...]`, 21 fulls) — i.e. own-indexer *sparse** layers in the reference schedule, not dense. In sparse prefill this also cascades: their shared followers (3–5) then get no threaded pages and all to dense too, so I see **6 dense `MLA-m` layers** up front. Since `index_skip_topk_offset` looks meant to control which leading layers get their **own** indexer (vs IndexShare reuse) rather than to disable sparsity, was the dense routing there intended only for the earlier **decode-only** sparse scope (where EXTEND was dense anyway, so it never mattered), or is there a reason the first full layers must run dense? I have a one-line change that lets them run as full/sparse, and I re-benched TTFT with it for a 25% Improvement at 48k while preserving correctness — happy to fold it into this PR or split it, whichever you prefer since it's your `skip_offset` code.

**1. Self-write is in-place (not a per-layer KV copy).** Confirmed both ways you suggested. *Buffer-donation check:* the forward is `jitted_run_model(..., donate_argnames=["memory_pools"])` and returns the updated pool, so XLA aliases the donated fused-KV buffer to the output — the `flat.at[loc].set(...)` scatter updates it in place (the runtime relies on this; it locks around the donate-dispatch because IFRT marks the input deleted the instant `kDonateInput` dispatches). *xprof slice:* on a 16k single-shot prefill the write compiles to `scatter_custom_fusion` (~0.6 ms each) totaling **25 ms = 0.82%** of the captured (fully device-busy) prefix, on a **0.13 GB/layer** buffer — not GB-scale. So no `input_output_aliases` change is needed; the outer-jit donation already gives in-place semantics. (The #1478 indexer-key-cache copy you saw was a smaller, non-donated buffer — different mechanism.)

**2. Bench/accuracy reproducibility.** The model name was excluded by mistake — I'm running **GLM-5.2** (`zai-org/GLM-5.2-FP8`, public, MIT). The description now names it and adds the full server config (v7x /8 chips / tp16-ep16-dp1 / dense-baseline `chunked_prefill_size` / `page_size` / `mem_fraction` / radix off / `max_running`) plus launch + eval commands. **Dense GSM8K control (your 2b):** compared the actual bench to verify no regression and got — **sparse 47/50 == dense 47/50** (standard GSM8K, n=50, greedy). Standard prompts are <2k < `index_topk`, so selection is full-coverage and this is a tight plumbing no-regression check.

**3. "toward ~linear" → residual O(T²) (wording + breakdown).** You're right — I dropped "~linear"; the PR now states prefill stays O(T²), just with a smaller constant. I added a per-phase, per-layer time breakdown (dense MLA / sparse attend / indexer / self-write) to the PR body. With the dense front-block removed (see the note below), the largest absolute cost is still the sparse-MLA attend (≈O(T·K)), and the remaining *quadratic* term is the lightning indexer's O(T·S) page-scoring — exactly what your Pallas streamindex path targets. Any rough ETA on when you plan to push that new indexer?

**4. Cross-query KV reuse.** Written down as a follow-up: batch a block of queries against the union of their selected pages (dense-prefill-style bq tiling) to amortize the per-step DMA setup and batch the gather+matmul. Agreed it's likely the next integer-factor win.

**5. `sm_scale` default.** Fixed — `sparse_mla_attention` now **requires** an explicit `sm_scale` (raises if `None`) instead of defaulting to `1/sqrt(Dk)` on the latent dim; the correct `1/sqrt(qk_nope_head_dim + qk_rope_head_dim)` isn't recoverable from `Dk`. Callers already pass it, so no behavior change — just removes the latent footgun.
image

zorrofox:
Thanks for the thorough follow-up — the donation evidence on #1 settles it (agreed, no `input_output_aliases` change needed; the #1478 case we profiled was a non-donated buffer, different mechanism), and the updated PR body with the model, config, paired short-context control, and the per-phase breakdown is exactly what we were asking for. Responses on the rest, plus a few things we found re-reviewing the branch as it stands now.

### `skip_offset`: your read is correct — but the change is already on this branch, under the wrong commit message

Confirmed against the reference semantics (`dsa_layer_skips_topk` in the CUDA implementation): `index_skip_topk_offset` schedules which layers compute their own topk vs. reuse the previous layer's — it never meant "run dense". Routing `layer_id < skip_offset` to dense in #1478 was a misread on our side, carried over from the decode-only scope where it happened to be harmless. Good catch, and thanks for tracing the shared-layer cascade.

That said, two things about how it landed:

1. **Commit `716c3731` is titled "fix(dsa): require explicit sm_scale in sparse_mla_attention", but its diff is the `skip_offset` gate removal** — no sm_scale change anywhere in it. And the sm_scale fix itself isn't on the branch: `sparse_mla_prefill.py` still has `sm_scale: float | None = None` with the `1/sqrt(Dk)` fallback at the callsite guard. Looks like the wrong commit got pushed — please push the actual sm_scale change, and untangle the history so each commit says what it does.
2. **Please move the `skip_offset` removal to its own PR** (you offered the split — we'd take it). It changes decode routing semantics for every `dsa_sparse` user, not just this PR's opt-in prefill path, and it flips 3 layers (plus their shared followers during prefill) from dense to sparse — a semantic change that needs its own accuracy validation at a context length where the selection actually binds (well above `index_topk`), with a paired dense control on the same server. Standard short-context GSM8K can't discriminate here for the same full-coverage reason you noted on #2. We're not in a position to run that validation ourselves, but we'll review the PR and results promptly when you post them.

### Long-context accuracy: the paired control is missing exactly where it matters

The short-context 47/50 == 47/50 pair is a good plumbing check (as you said yourself, selection is full-coverage there). But the long-context run — the one where pages are actually dropped (57→32) — reports sparse 43/50 with **no dense number under the identical config**. That's the single comparison that validates the sparse selection semantics, and it's cheap (same server, flag off, n=50). Please add it; ideally it doubles as part of the skip_offset PR's validation matrix.

### GSM8K methodology: n=50 can't support the claim

n=50 is too small to support a "no regression" claim on its own — the 95% binomial CI at 47/50 is roughly ±7 points, so anything short of a catastrophic regression is invisible at that sample size. Two existing references in this repo for calibrating sample size:

- [`benchmark/gsm8k/bench_sglang_jax.py`](https://github.com/sgl-project/sglang-jax/blob/main/benchmark/gsm8k/bench_sglang_jax.py) defaults to (and its own usage example uses) `--num-questions 200` — that's a few minutes per run and tightens the CI to ~±3 points, cheap enough to use for both the short- and long-context paired checks above.
- For a PR-body accuracy claim specifically, the [`model-accuracy-benchmark` skill](https://github.com/sgl-project/sglang-jax/blob/main/.claude/skills/model-accuracy-benchmark/SKILL.md) in this repo establishes `evalscope` against the full GSM8K set as the standard tool for accuracy claims here — a 50-example `run_eval` figure isn't directly comparable to that.

Also worth checking on the long-context run specifically: if GLM-5.2 is served with its chat template's thinking mode active, an insufficient `max_tokens` budget truncates inside the reasoning block and depresses the score independently of selection quality — worth ruling out for the 43/50 before attributing any delta to pruning.


### Streamindex status

Actively in development on our side, and the feasibility work is done: we validated the current tpu-inference `streamindex_topk` kernel (the rewritten scores-in-HBM + external `approx_max_k` version) on both v7x and v6e — upstream's own unit tests pass on both, and at GLM-5.2 shapes (`compression_ratio=1`, k=2048, H=32, page 64) it is numerically correct against an fp32 oracle and 1.0–6.8× faster than the jnp reference at decode shapes. The part relevant to this PR: it computes scores at **O(actual kv_len)** with prefill/mixed query grids supported, so it directly replaces the O(T·S)-with-static-S page-scoring einsum that your breakdown shows as the residual quadratic term. We also verified a bf16 direct-read variant against the native indexer-cache layout (no cache format change needed) as a fallback to the fp8-packed path. Wiring into `dsa_sparse_backend` is what's in progress now — no date to commit to yet, but we'll post the draft PR here when it's up, and it should compose cleanly with your prefill path (your `topk_pages` consumer stays as-is; only the scores producer changes).

HadarIngonyama:
@zorrofox you're right I switched the two files by mistake - now fixed, and will move the `skip_index` fix to it's own PR as you requested.
regarding the extra benches - since this PR does not support batching, each bench actually takes quite a long time (around 30min for 50 prompts). However, I have already a new version that supports batching that runs much faster and I intended to open a new PR for it. Should I instead run the benches on it and add it to the current PR? What do you think is the best course of action?

zorrofox:
Glad the commit mixup is sorted, and thanks for confirming the `skip_offset` split.

On the benching question — the batching version isn't needed for the one number that actually matters here: **the dense counterpart to your existing 43/50 long-context sparse run**, same config, same 50 examples. Batching speeds up concurrent throughput, not the wall-clock of a single long single-shot prefill, so it won't meaningfully cut that ~30min — and more importantly, that check needs to run against the code that's actually landing in *this* PR (single-sequence, unbatched), not a different not-yet-merged implementation. It's one more run at the same n=50 you already have, not a scale-up.

Where the batching version *does* help is the short-context pair — that one's cheap regardless (short prompts, not much overhead from a lack of batching), so scaling it to 200 questions now costs little. Scaling either check further (200+ on long context, or a full evalscope pass) is a reasonable thing to defer until batching lands — that's a proportionality call, not a correctness one, so up to you on timing.

So, concretely: dense long-context control now (blocks the accuracy claim in this PR), short-context bump to 200 now if it's cheap, everything past that can wait for the batched follow-up.

HadarIngonyama:
@zorrofox Thanks — and you were right about the truncation.

**Long-context paired control (the blocking item).** Ran the dense counterpart to the 43/50, same server, flag off, same 50 prompts: **sparse 43/50 vs dense 42/50** — sparse ≥ dense with the selection actually binding (57→32 pages). Following your thinking-mode note I checked `finish_reason`: at `max_tokens=900`, ~8/50 generations truncated inside the reasoning block before `Answer:`, depressing both arms. Re-running truncation-free (`context_length=12288`, `max_tokens=3072`, same 50 prompts, 1/50 truncated each) gives **sparse 50/50 vs dense 48/50** — the pair holds and both jump. Added this paired table to the PR description.

**On sample size / evalscope.** Agreed n=50 can't carry a tight CI. Since you noted a 50-example `run_eval` figure isn't directly comparable to the `evalscope` standard anyway, I'd rather not add a half-measure 200-example `run_eval` here — I'll bring the proper `evalscope` pass (short + long context) with the batching PR, where the runs are fast enough to do it consistently and to the repo standard. Proposing to settle this PR on the paired long-context control above plus the existing short-context 47/50 == 47/50; happy to add more here if you consider it blocking, but I'd like to keep the heavier accuracy matrix on one consistent methodology.

**skip_offset split.** Done — opened as its own PR: #1512. It's stacked on this one (the description asks reviewers to look at the top commit only; I'll rebase onto `main` once this lands) and carries its own TTFT + accuracy validation.

**Batching.** The packed-ragged version is ready; I'll open it as a separate PR shortly, and the n=200 / `evalscope` accuracy matrix will ride with it.

zorrofox:
This all looks solid — the truncation catch is a good find (glad the thinking-budget check paid off), the paired table is in the PR body, and `#1512`'s root-cause writeup is more precise than what we'd worked out ourselves (didn't realize `build_index_share_map` never reads `skip_offset` at all — confirmed, that parameter really has no other runtime effect).

On sample size, though, I want to push back on deferring the larger accuracy pass to the batching PR — not for cost/methodology-consistency reasons, but attribution. If the eventual batching-PR evalscope run shows any dip, there'll be no way to tell whether it's a batching bug or a subtler selection issue already present in this single-sequence code that n=50 simply doesn't have the power to catch — three changes (sparse prefill, the skip_offset fix, batching) would all be entangled in one test run. We'd rather see a n=200 `run_eval` pass — doesn't need to be the full evalscope set, that heavier lift can still ride with batching — run on *this* code (single-sequence, #1508 + #1512 combined) so the sparse-selection mechanism itself is validated in isolation before batching adds another variable on top.

The same attribution gap applies to `#1512`'s own diff (routing layers 0–5 sparse) — right now that's only the n=50 (±1 example) prose claim, worth the same n=200 treatment there regardless of who ends up reviewing that PR.

Once that's in, this looks close to ready from our side.

HadarIngonyama:
Thanks @zorrofox — fair point on attribution, so I ran exactly what you asked: `test/srt/run_eval.py --eval-name gsm8k --num-examples 200` on the shipped single-sequence code (#1508 + #1512) — the actual `(B,S)` single-seq kernel checked out on the pod, not the #1514 ragged build fed one sequence — paired sparse vs dense, identical server/config, flipping only `DSA_PREFILL_SPARSE`.

**#1508+#1512 single-seq · `test/srt/run_eval.py` gsm8k (zero-shot chat, n=200, T=0, max_tokens=2048) · `--max-running-requests 1 --disable-radix-cache`:**

| arm | accuracy | invalid |
|---|---|---|
| sparse (`DSA_PREFILL_SPARSE=1`) | 0.960 (192/200) | 0.000 |
| dense control (`DSA_PREFILL_SPARSE=0`) | 0.980 (196/200) | 0.000 |
| Δ (sparse − dense) | −0.020 | |

No meaningful regression: the 0.960 vs 0.980 gap is 4 problems out of 200, inside the ~±2–3 pt noise band at this sample size, with zero invalid/truncated outputs on either arm. Sparse also sits at/above the 47/50 (0.94) standard-GSM8K baseline already in the PR.

Two honest scope notes so the number isn't over-read:

- Selection is only lightly exercised at short context. Standard GSM8K prompts are short — well under the `index_topk=4096` budget — so page selection is effectively a no-op here (sparse is a mathematical superset of dense, gate G4). So this tightens the sample-size CI on the shipped **code path** (the n=50→n=200 concern you raised); it does not tighten CI on *selection* per se, since selection isn't active at this length. The selection-under-pruning signal stays the long-context pair already in the PR (sparse 50/50 vs dense 48/50 at 12k ctx / 3072-tok budget), and tightening *that* to n=200 is the long-context follow-up you already agreed can defer.
- This `test/srt/run_eval.py` harness drives the same prompting style as evalscope (both are zero-shot chat on `/v1/chat/completions`) — it's just the lighter n=200 subset vs evalscope's full 1319. The two are consistent: evalscope full-1319 gives sparse 0.9674 vs dense 0.9689 (Δ −0.0015), this n=200 gives sparse 0.960 vs dense 0.980 (Δ −0.020) — both a sub-noise gap in dense's favor, i.e. no sparse-side regression at either sample size.

This n=200 covers #1512 too, since it's measured on #1508+#1512 combined — I've folded these numbers into both the #1508 and #1512 descriptions (accuracy sections).

For the full picture: note that I've just updated the **#1514** description (the batching PR stacked on this one) with the complete validation set, so it now supersedes the earlier n=50 / smaller-probe numbers there. It carries paired **full-1319 evalscope** (sparse 0.9674 vs dense 0.9689, Δ −0.0015), CPU-interpret parity gates G0–G6 + A1/A2 (mostly exact against an fp32 oracle), a monotonic batched-throughput curve to **~3.6× at B=32** (radix off, cache_hit=0.000), mixed-length packing (0.92×, 0% padding waste), and on-device chunked-prefill + radix checks (needle recall 6/6 across chunk boundaries, long-ctx GSM8K chunked == single-shot). Taken together — single-seq no-regression here, full-scale no-regression under batching+radix+chunking there — the sparse path holds at every size we've been able to measure.

zorrofox (approved):
Approving. Recapping where this landed, for the record:

- **Self-write in-place** — confirmed via `donate_argnames=["memory_pools"]` + xprof (`scatter_custom_fusion` = 25ms, 0.82% of prefill).
- **Benchmark reproducibility** — public model (`zai-org/GLM-5.2-FP8`), full config, exact server + eval commands.
- **Asymptotics** — correctly qualified as a smaller O(T²) constant, not linear; per-phase breakdown backs it up.
- **`sm_scale`** — now required explicitly, raises instead of silently under-scaling.
- **`skip_offset` dense front-block** — correctly disclosed here as a routing artifact, fix properly scoped to #1512 rather than folded in here.
- **Accuracy** — short-context paired n=200 (0.960 vs 0.980, within noise) run on the actual #1508+#1512 single-sequence code, isolating this mechanism from #1514's batching; long-context paired control (50/50 vs 48/50, truncation-free) added after catching a decode-budget confound in the original run.

The one open item — extending the long-context pair to n=200 — is honestly flagged as deferred to ride with the batching PR rather than claimed as done, which is a reasonable call given how expensive long-context single-shot runs are on this unbatched path.

Nice work tightening this up through the back-and-forth.

zorrofox:
As promised upthread — the indexer-kernel integration PR is now up: #1515.

It re-vendors the new upstream `streamindex_topk` (scores-in-HBM + external `approx_max_k`) with a bf16 direct-read cache branch, wired behind an opt-in flag (default off). It only replaces the token-topk **producer** in `_maybe_index`, so it composes cleanly with the prefill consumer path in this PR. e2e A/B numbers (GSM8K parity, cc=8 +17.6%, cc=64 TPOT 1.28x) are in the PR body.

zorrofox (changes requested):
We attempted an independent reproduction of the 16k single-shot prefill numbers in the benchmark table, and the results isolate a real discrepancy worth resolving before this merges — posting the full data so it's actionable.

**Setup.** Pure #1508 at head `8ebd5092` (verified no #1514/#1515 code present), replicating every documented parameter of the benchmark config: TPU v7x 8 chips / 16 devices, `tp16 ep16 dp1`, `page_size=128`, `--disable-radix-cache`, `max_running_requests=1`, `mem_fraction_static=0.60`, `context_length = chunked_prefill_size = max_prefill_tokens = 49152` (single-shot), `DSA_INDEX_TOPK=4096`, `zai-org/GLM-5.2-FP8`, `--moe-backend epmoe`, `--dtype bfloat16`. Probe: 16384 random `input_ids` direct to `/generate`, `max_new_tokens=1`, e2e timing, warm runs after shape compilation (two repeats each arm, <1% variance).

**Wall-clock, warm 16k, one server flipping only `DSA_PREFILL_SPARSE`:**

| arm | ours (×2) | PR table | ratio |
|---|---|---|---|
| dense (flag=0) | 22.7 / 22.7 s | 19.1 s | 1.19× — reproduces |
| sparse (flag=1) | **34.7 / 34.6 s** | 10.5 s | **3.3× — does not reproduce** |
| sparse vs dense | **1.53× slower** | 1.82× faster | direction inverted |

**XProf, bucket-by-bucket against your own per-layer table.** We captured a full warm 16k prefill window (34.7s, matching the standalone measurement) and aggregated the TPU:0 XLA-op track against your "× 6 / 72 / 21 layers ⇒ 1.35 / 6.2 / 1.5 s @16k" accounting:

| bucket | your @16k | our trace | ratio |
|---|---|---|---|
| dense MLA (6 layers) | 1.35 s | **1.36 s** (`MLA-m` kernel events) | **1.01× — matches to 1%** |
| sparse attend (72 layers) | 6.2 s | **24.16 s** (72 `shard_map` invocations, 335.6 ms avg) | **3.9×** |

The dense bucket matching to 1% rules out environment, compiler, hardware and measurement methodology as explanations — the entire gap sits in the sparse-attend bucket, same 72 invocations on both sides, 86 ms vs 335.6 ms each.

**A methodology question that may explain this.** Your table footnote says the attend/indexer numbers were "measured directly in isolation" (only dense came from xprof). An isolated kernel microbench at 86 ms/layer wouldn't include the in-situ costs our trace attributes to those 72 `shard_map` invocations (dispatch, mesh transitions, interleaving with the indexer's jnp scoring). And 1.35 + 6.2 + 1.5 + 0.03 ≈ 9.1 s sums suspiciously close to the reported 10.5 s. So the key question: **is the 10.5 s sparse wall-time an end-to-end measurement, or a sum of components?** If it's e2e-measured, could you share the exact launch command and the harness/timing method used for the wall-time table (the Accuracy section has commands, but for a different config), plus `LIBTPU_INIT_ARGS` if any? Those are the only variables we could not align from the PR text.

Happy to share our trace (`.xplane.pb`, single-rank) and the probe script if useful.

Given this, I'm withdrawing my earlier approval until the wall-time discrepancy is resolved — the approval was given with the performance table as a core part of the evidence, and that table no longer stands as independently reproducible. To be explicit about scope: the accuracy story in this PR is solid and unaffected, and the correctness work here remains good; this is specifically and only about the performance claim. Once the 10.5 s question is answered (either the missing config variable surfaces and we can reproduce it, or the table is re-measured e2e), re-approving should be quick.

HadarIngonyama:
Thanks for posting the full data — that made this fast to chase. Short version: **we re-ran your exact documented config and reproduced the table** — 10.5 s e2e sparse, 87 ms/layer attend — so the gap isn't in any parameter either of us has written down. Screenshots from our sparse xprof capture attached (client-side wall = 10.5 s; `shard_map` sparse-attend = 87 ms/layer × 72; `MLA-m` ≈ 197 ms/layer).

Same config we ran, matching yours line-for-line: `--context-length 49152 --chunked-prefill-size 49152 --max-prefill-tokens 49152`, `page_size=128`, `mem_fraction=0.60`, `tp16 ep16 dp1`, `--disable-radix-cache`, `max_running=1`, `DSA_INDEX_TOPK=4096`, `zai-org/GLM-5.2-FP8`, epmoe, bf16, single-shot 16k probe direct to `/generate`. To be explicit about our tree: this is #1508 head **plus the #1513 block-quant reshard fix** (required to compile on jax 0.10.2 — more on that below); no #1514/#1515 code present.

**What your own numbers isolate.** `index_topk=4096` / `page_size=128` → 32 pages/query on both sides, and your dense-MLA bucket matching ours to 1% (1.36 vs 1.35 s) proves the prompt is tight on both (O(T²) dense MLA would blow up ~9× under any hidden padding). So: identical work, identical shape, identical 72 invocations — the only thing that differs is **per-invocation efficiency** of that one kernel (335.6 ms vs 87 ms). That's codegen, not computation.

**Also, this settles the "sum vs e2e" question:** our 10.5 s is a client-side round-trip wall (`/generate`, `max_new_tokens=1`, warm, 3/3 identical), *and* the device trace independently shows 87 ms/layer × 72. E2e and trace agree — it isn't a sum of components.

**Hypothesis.** The sparse-attend op is a Pallas/Mosaic kernel whose throughput is set by compiler-selected block sizes (you can see the runtime choosing them: `[GMM kernel] using default block sizes for key: (…): (512, 2048, 2048)`). A different libtpu/Mosaic build — or a cold compilation cache — can pick a 3–4× worse tiling for that kernel while leaving XLA ops untouched. That's exactly the fingerprint here: one bucket 3.9× off, everything else identical. Note this is why "dense-MLA matches to 1% ⇒ environment ruled out" doesn't hold — dense MLA is XLA-compiled and the sparse kernel is Mosaic-compiled, so they diverge independently.

**Our runtime (the un-documented variables):**

| | ours |
|---|---|
| jax / jaxlib | 0.10.2 / 0.10.2 |
| libtpu | 0.0.42.1 |
| `LIBTPU_INIT_ARGS` | unset |
| JAX compilation cache | on (`JAX_COMPILATION_CACHE_DIR`) |

One concrete tell on the runtime: we also carry #1513 (block-quant scale reshard). On jax[tpu] 0.10.2 (Explicit-axis mesh mode) the col-parallel projections (`q_b_proj`) **don't compile** without it — so if you ran pure #1508 without #1513 and it compiled cleanly, you're on a different jax/libtpu than our 0.10.2, which is exactly the block-size-selection variable above. (The fix is a compile-time reshard on the projection layers, a runtime no-op when sharding is already correct — it doesn't touch the sparse-attend kernel, so it's a version indicator, not a perf change.)

Could you share the same three — your `jax`/`jaxlib`/`libtpu` versions + any `LIBTPU_INIT_ARGS`, whether you compiled with a warm cache, and your kernel's block-size key line (the `using default block sizes for key: (…)` log for the sparse-attend/GMM path)? If your block-size key differs from ours, that's the smoking gun and likely a one-line fix. Happy to trade `.xplane.pb` traces both ways.
sparse kernel
sparse full

zorrofox:
Thanks for the detailed environment table and the xprof screenshots — that let us bisect this cleanly. We ran both variables you surfaced, and unfortunately neither closes the gap.

**Runtime versions match exactly.** Verified in our image against your table:

| | ours | yours |
|---|---|---|
| jax / jaxlib | 0.10.2 / 0.10.2 | 0.10.2 / 0.10.2 |
| libtpu | 0.0.42.1 | 0.0.42.1 |

So the "different jax/libtpu → different Mosaic tiling" hypothesis doesn't hold — we're on the same compiler.

**On #1513 being required to compile:** our pure-#1508 baseline (the 34.7s number) *did* compile and serve on jax 0.10.2 without #1513. So on our tree the block-quant scale evidently arrives with a matching sharding already; whatever makes it arrive replicated on your side isn't happening on ours. That itself is a real environment difference worth chasing (weight-loader path? checkpoint layout?), but it means "you compiled without #1513 ⇒ different jax" doesn't follow.

**Two-arm bisect** (fresh v7x-16 deployments, everything else identical to our 08-09 baseline; warm 16k single-shot ×2):

| arm | tree | LIBTPU_INIT_ARGS | warm ×2 | vs our baseline |
|---|---|---|---|---|
| baseline (08-09) | pure #1508 | our 5 async-collective flags | 34.7 / 34.6 s | — |
| **B** | pure #1508 | **unset** | **34.7 / 34.6 s** | **no change** |
| **A** | #1508 + **#1513** | unset | **69.3 / 69.4 s** | **2.0× slower** |

- **`LIBTPU_INIT_ARGS` is not the variable.** Unsetting it reproduces the baseline to the decimal.
- **#1513 is not the missing speedup — on our tree it's a 2× slowdown.** Our guess: since our loader already delivers a correctly-sharded scale, the added `jax.sharding.reshard(...)` is taking a different layout path and dragging extra collectives into every col-parallel projection. (We're capturing an xprof of the 69.3s arm to confirm the attribution — will append.)

So all four alignable variables you listed — jax/jaxlib/libtpu versions, `LIBTPU_INIT_ARGS`, and #1513 — are now ruled out, and arm A (which matches your stated tree + env exactly) lands at 69.3s, not 10.5s.

**What would actually let us close this.** The remaining unknowns are all "how the tree was built", not "what flags were set":

1. `pip freeze` from inside your serving container (or the Dockerfile / image tag) — the jax triple matching doesn't rule out a different Pallas/Mosaic-adjacent dep.
2. The exact `python -m sgl_jax.launch_server ...` command line for the 10.5s run.
3. `git log --oneline` of `/opt/sglang-jax` (or equivalent) from the running container — to rule out any other local hotfix beyond #1508+#1513.
4. If shareable: the raw `.xplane.pb` for your 10.5s capture (we'll diff kernel names / block-size annotations directly against ours instead of eyeballing screenshots).

Happy to share our arm-A/B traces and the exact yaml in return.

zorrofox:
**Correction to the arm-A number above** — the 69.3s was a measurement artifact on our side, not a #1513 effect. Two probe clients ended up hitting arm-A concurrently (a background poller and an in-pod probe), and with `max_running_requests=1` each round's wall-clock became ~2× a single request. Re-measured cleanly (single probe, ×2 warm each):

| arm | tree | LIBTPU_INIT_ARGS | warm 16k ×2 |
|---|---|---|---|
| pure #1508 | — | unset | 34.7 / 34.6 s |
| #1508 + #1513 | — | unset | **34.7 / 34.6 s** |
| #1508 + #1513 | dense control | unset | 22.7 / 22.7 s |

So #1513 is **neutral** on our tree (sparse and dense both unchanged), not a slowdown — apologies for the noise. The corrected picture is actually simpler:

- jax/jaxlib/libtpu: identical (0.10.2 / 0.10.2 / 0.0.42.1)
- `LIBTPU_INIT_ARGS`: unset on both sides → no effect
- #1513: present on both sides → no effect
- JAX compilation cache: **also** enabled on our side (`/tmp/jit_cache`, confirmed from server log) → aligned

i.e. every variable you've surfaced is now aligned and the sparse arm still lands at 34.7s, not 10.5s. The rest of the previous comment stands — the four asks (pip freeze / launch cmdline / container `git log` / raw `.xplane.pb`) are what we need to keep bisecting.

zorrofox:
Found it — and it's neither codegen nor a Mosaic version thing. It's **token-bucket padding**.

**Root cause.** On main (`common_utils.py:39`), `PRECOMPILE_DEFAULT_TOKEN_PADDINGS = [1<<i for i in range(6,14)]` = `[64..8192]`. `_compute_token_buckets` then appends `max_padded_num_tokens = max(max_prefill_tokens, chunked_prefill_size) = 49152`, so the effective bucket list is `[64,128,...,8192, 49152]`. A **16384-token** request falls in the `(8192, 49152]` gap and gets padded to **49152**. The sparse kernel's `grid=(B, S)` walks every one of those 49152 grid steps and runs the full chunk_body (32 DMAs + matmuls) on each — there's no pad-token early-exit — so it does 3× the work. Dense MLA is a *ragged* kernel that skips padded tokens, which is exactly why the dense bucket matched to 1% while sparse was 3.9× off.

**Verified.** One-line hotpatch changing `range(6,14)` → `range(6,16)` (adding 16384 and 32768 to the buckets), same server otherwise:

| tree | buckets | warm 16k sparse (×4) |
|---|---|---|
| pure #1508 | default `[64..8192,49152]` | 34.7 / 34.6 / 34.7 / 34.6 s |
| pure #1508 | **`+[16384,32768]`** | **10.45 / 10.45 / 10.46 / 10.46 s** |

10.46s reproduces your 10.5s to <1%. HLO input shape for the sparse custom-call goes from `bf16[1,49152,16,640]` to `bf16[1,16384,16,640]` — same kernel, same tiling, 1/3 the grid steps.

**So the un-aligned variable was `--precompile-token-paddings`** (or equivalently, whatever puts 16384 in your bucket list). Could you confirm what your launch has there? That closes the reproduction question completely.

**On the PR itself, two things follow from this:**

1. The 10.5s number in the benchmark table is real, but only under a padding config the PR doesn't document — with the repo defaults, sparse@16k is 34.7s and *slower* than dense. The table needs either the `--precompile-token-paddings` value added to the "Setup" section, or the numbers re-measured at repo defaults.
2. More structurally: the sparse kernel's performance is O(padded_S), not O(actual_S) — anyone landing between buckets pays for the whole next bucket. That's a real footgun for the shipped code, not just a repro nuance. Cheapest fix is an early-exit in `_sparse_mla_kernel_chunked_hminor` when `qpos < 0` (padded queries carry `positions = -1` or similar), so the kernel becomes bucket-insensitive like the dense ragged path already is.

Once (1) is addressed I'm happy to flip back to approve; (2) can be a follow-up if you prefer, but it should at least be called out in the PR body as a known limitation.

HadarIngonyama:
Agreed — that's the cause, and your hotpatch reproducing ~10.46 s with the 16384 bucket confirms it. At the default `--precompile-token-paddings` (`[64…8192]` + `context_length`), a 16k prompt pads to **49152**, and the sparse-MLA prefill kernel grids over the padded length, so it does 49152 tokens' worth of work for a 16k prompt. Pinning a 16384 bucket gives the 10.5 s. I've added a bucketing note to the #1508 description so it's reproducible.

On the kernel early-exit you suggested: I prototyped and correctness-verified it, then measured it at the 49152 bucket. It makes the attend kernel bucket-insensitive (**96 ≈ 87 ms/layer**) and recovers most of the padding penalty — **34.7 → 17.4 s**, i.e. ~17 s of the ~24 s of extra over the tight-bucket 10.5 s. The remaining ~7 s is the projections, indexer gathers, and surrounding fusions — XLA ops still running on the full padded buffer that a guard inside the attend kernel can't reach. So it closes most of the gap but not all, and it adds a sentinel operand + control-flow branch to the hottest kernel for a config the production path shouldn't hit — where chunking (below) removes the full ~24 s instead.

The full fix is already in **#1514 (chunked prefill)**: splitting to `chunked_prefill_size` means the prompt never lands in an oversized bucket, so *all* the padded overhead goes away (attend + projections + indexer), bounded to the tail chunk. The clean long-term version is a native ragged kernel gridding over actual tokens via `cu_seqlens` (like the dense ragged MLA), which I'm tracking as future work. Does that split — bucket note here, chunking in #1514, ragged kernel later — work for you?

P.S. — which checkpoint did you run? I'm on `zai-org/GLM-5.2-FP8` (FP8 block-quant — that's the reason I need #1513 to compile). Are you on that one, or the bf16 checkpoint from #1478?

zorrofox:
The updated description covers it — the Setup section now documents `--precompile-token-paddings` as required (with the reason), and the Limitation section spells out the default-bucket 34.7s-slower-than-dense case, the workaround, and the #1514/ragged-kernel path forward. That's exactly the disclosure I was asking for, so I'm re-approving.

On the checkpoint question: we're on **`zai-org/GLM-5.2-FP8`** (the FP8 block-quant one — `--model-path` in our repro yaml points at a local mirror of exactly that HF repo), and it *does* compile and serve without #1513 on our jax 0.10.2 tree. So the "arrives replicated" behavior #1513 fixes isn't universal to that checkpoint — likely a difference in how the loader path lands the scale sharding between our images. Orthogonal to the padding root cause here; just noting it since you asked, and it may be worth a look for #1513's own scoping.

On the early-exit: we independently landed at the same ~17.8s (×5, <0.1% variance) with a 5-line `pl.when(qpos >= 0)` wrap + one-line `positions = jnp.where(loc < 0, -1, positions)` guard in the caller — happy to share the diff if useful. I'd still lean toward including it (it's cheap, and it means #1508 standalone never inverts the sparse<dense direction regardless of bucket config, which matters for the window between this merging and #1514 landing), but the Limitation section already discloses the behavior honestly, so I won't block on it. Your call.

The #1514-as-general-fix framing works for me under one reading: #1508's own guards *require* `chunked_prefill_size >= context_length`, so anyone running #1508 standalone (pre-#1514) is by construction in the single-shot regime and will hit this — which is fine now that the Limitation section tells them to pin the bucket. Just making the dependency explicit.

zorrofox (dismissed):
Re-approving — the description now documents the `--precompile-token-paddings` requirement and discloses the default-bucket limitation (see thread for the root-cause analysis and independent reproduction at 10.45s). Performance claim is now reproducible as documented.

zorrofox:
@HadarIngonyama Please resolve the branch conflicts and ready to merge to Main.

HadarIngonyama:
@zorrofox I rebased and resolved the conflicts - no logic change. CPU sanity tests passed (CPU parity suite - test_sparse_mla_prefill_parity.py, interpret).

zorrofox (approved):
Re-approving after the rebase onto current main.

Verified this is a content-identical rebase: a line-level comparison of the PR diff before (head `8ebd5092`) and after (head `bcf33c67`) shows zero payload changes — the only deltas are blob hashes, hunk-header offsets, and context-line drift from #1515 now being in the base. Same two commits, same scope as the revision approved on 2026-08-10.

Scope of this approval is unchanged: correctness gates plus the documented performance envelope, including the default-bucket limitation now disclosed in the body.

zorrofox:
As promised upthread: the streamindex prefill wiring is now up as #1545 — page-level Pallas top-k for the prefill indexer (O(actual kv) scoring, no [T, max_kv] materialization), with paired single-stream and concurrent-serving numbers in the body. cc @HadarIngonyama

PR #1512, opened 2026-08-06
fix(dsa): route skip_offset "full" layers sparse in decode + prefill (remove unconditional dense front-block)

HadarIngonyama:
> **Stacked on #1508.** This builds directly on the sparse-prefill path added in #1508 — it edits the dense short-circuit that PR introduces, so it cannot stand alone on `main`. Until #1508 merges, the diff below also contains #1508's commits: **please review only the top commit** (`fix(dsa): run skip_offset "full" layers sparse, not dense`). Once #1508 lands I'll rebase onto `main` and the diff collapses to the one-line change.

## Motivation

**Scope — this corrects routing in decode too, not just prefill.** The gate this edits (`layer_id < self.skip_offset or (is_full and q_idx is None)`) sits **above** the `is_decode` branch and is **not** behind the `DSA_PREFILL_SPARSE` opt-in from #1508. So the moment this merges, layers `0..skip_offset-1` switch from dense to **sparse at decode as well — unconditionally, for every existing `dsa_sparse` deployment**, independent of whether anyone ever sets the prefill flag. The TTFT table below is the prefill win; the routing correction itself is unconditional and applies to both paths, so it's the more accurate framing of the blast radius than "prefill TTFT fix."

#1508 makes DSA prefill sparse, but a residual O(T²) term remains from a dense "front-block" of leading layers. That block is a **backend routing artifact, not a model requirement**. GLM-5.2's `config.indexer_types` is `['full','full','full','shared','shared','shared','full', …]`: layers 0–2 are `"full"` — they own a lightning indexer and are meant to run **sparse** (own top-k), while `"shared"` layers reuse the nearest preceding full's top-k (IndexShare).

`DSASparseAttentionBackend.__call__` instead force-routes the first `index_skip_topk_offset` (= 3 for GLM-5.2) layers through dense absorbed-MLA via `layer_id < self.skip_offset`. But `index_skip_topk_offset` only governs which leading layers own their indexer vs share one — `build_index_share_map` (`kernels/dsa/ref.py`) takes it as an argument and **ignores it**; full/shared is derived entirely from `indexer_types`. So the parameter's only runtime effect was this dense gate, which reads like a misinterpretation of "full" (own indexer) as "skip sparse."

The mis-route also cascades. With 0–2 forced dense they emit no page-topk, so their shared followers 3–5 have no `dsa_topk_pages_in` to reuse and hit the "no selection ⇒ dense fallback" branch too — a contiguous **6-layer dense front-block**, the dominant residual quadratic in the #1508 per-layer breakdown (`MLA-m` 224 ms → 2014 ms/layer, 9.0×, O(T²)). It was invisible in #1478 because that PR's sparse path was decode-only and all EXTEND was dense anyway; it only became wasted O(T²) once #1508 made prefill sparse.

## Modifications

- `srt/layers/attention/dsa_sparse_backend.py` (one line): drop `layer_id < self.skip_offset` from the dense short-circuit so a `"full"` layer with a wired indexer always runs its own top-k + sparse attention (0–2 sparse ⇒ 3–5 get threaded pages ⇒ sparse). The dense path now fires only for the genuine "full layer with no indexer wired" case (`is_full and q_idx is None`). Result: **no dense attention on any layer**, matching `indexer_types`.

## Accuracy Tests

Verified with the same controls as #1508 (GSM8K short + long-context `gsm8k_longctx.py`, plus a long-context needle probe), same server/config, flipping only this code fix. Routing layers 0–5 sparse instead of dense holds accuracy within run noise (±1 example) on both short and long-context GSM8K vs the pre-fix baseline, and the ~17k needle probe — now sparse on **every** layer including 0–2 — recalls the planted code exactly, with no dense fallback and no `shared layer … requires dsa_topk_in` assertion in the log. Full paired numbers now match the #1508 no-regression story. The truncation-free long-context re-pair (`context_length=12288`, `max_tokens=3072`, same 50 prompts) gives **sparse 50/50 vs dense 48/50**, and a tightened **n=200 `run_eval` on this exact #1508+#1512 code** (`test/srt/run_eval.py --eval-name gsm8k --num-examples 200`, temp 0, no thinking, `max_running=1`, radix off) gives **sparse 0.960 (192/200) vs dense 0.980 (196/200)** (Δ −0.020, within n=200 noise) — no meaningful regression at 4× the n=50 sample size, measured on the code that actually lands here.

These runs also validate the **decode-side** change, not just prefill: each GSM8K generation is one prefill + many decode steps (more under thinking mode), so the n=50/200/1319 pairs repeatedly exercise the corrected decode routing on layers 0–5 — they are the de-facto validation for the unconditional decode change, it just wasn't called out before.

## Benchmarking and Profiling

Single-shot prefill TTFT, one `context_length=49152` server identical to the #1508 bench server except this code fix (`DSA_PREFILL_SPARSE=1`, `index_topk=4096`, `mem_fraction_static=0.60`); all EXTEND buckets `[16384, 32768, 49152]` compiled with no OOM (the sparse front-block uses less peak HBM than the old dense one):

| Context (T) | Dense | DSA pre-fix (6 dense layers) | DSA this PR (all sparse) | speedup vs dense |
|---|---|---|---|---|
| 16k | 19.1 s | 10.5 s | **9.9 s** | 1.93× (was 1.82×) |
| 32k | 73.0 s | 23.1 s | **19.1 s** | 3.82× (was 3.16×) |
| 48k | 155.7 s | 38.4 s | **28.7 s** | 5.43× (was 4.05×) |

The saving grows with T because it removes an O(T²) block: **+6% @16k → +25% @48k** (9.7 s off the 48k wall). Consistent with the #1508 accounting — the 6 dense `MLA-m` layers were 1.35 s @16k but 12.1 s @48k; replacing them with 6 sparse attends (~1.5 s) + 3 added indexer computations (~0.6 s) nets ≈10 s @48k.

### DIFF
--- a/docs/features/attention_backend.md
+++ b/docs/features/attention_backend.md
@@ -4,13 +4,14 @@ SGL-JAX exposes a small user-facing attention backend switch, while the runtime 
 
 ## User-facing choices
 
-`--attention-backend` accepts three values:
+`--attention-backend` accepts four values:
 
 | Value | Runtime behavior |
 |---|---|
 | `fa` | Default. Uses FlashAttention for MHA/GQA models and the absorbed MLA Pallas backend for MLA models. |
 | `fa_mha` | Forces MLA models through the decompressed MHA FlashAttention path. This is useful for kernel A/B checks, but uses much more KV cache than absorbed MLA. |
 | `native` | Pure JAX/native attention path, mainly for CPU/debugging. If `fa` or `fa_mha` is requested on CPU, the runtime falls back to `native`. |
+| `dsa_sparse` | DeepSeek Sparse Attention (DSA): a lightning-indexer selects the top-scoring pages of past tokens per query and attention runs only over that page set (page-block sparse guided by DSA scores), with IndexShare cross-layer reuse of the selection. For MLA models whose config carries the `index_*` fields only. See [DeepSeek Sparse Attention](#deepseek-sparse-attention-dsa_sparse). |
 
 Example:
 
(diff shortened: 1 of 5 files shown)


zorrofox (commented):
The fix itself is right, and the root-cause diagnosis is more precise than what we'd worked out — confirmed `build_index_share_map` never reads `skip_offset` at all (it's dead-since-#1478 on our end, not something introduced here).

One thing to fix in the description before this is ready to approve: **this changes default decode behavior for every existing `dsa_sparse` deployment, not just the prefill path.** The `is_full and q_idx is None` gate you're editing sits above the `is_decode` branch — it's shared by decode and prefill, and isn't gated behind `DSA_PREFILL_SPARSE`. So layers 0..skip_offset-1 switch from dense to sparse attention at **decode** too, unconditionally, the moment this merges — independent of whether anyone ever sets the opt-in flag from #1508. The current title/motivation frame this entirely as a prefill TTFT fix, which undersells the actual blast radius.

The good news: this is very likely already covered by the accuracy numbers you have. GSM8K generation is one prefill + many decode steps (more so with thinking mode), so the n=50/200/1319 runs already exercised the changed decode path repeatedly — it's just not called out. Could you:

1. Update the motivation/title framing to state plainly that this fixes routing for **both decode and prefill**, and that decode's default behavior changes for all `dsa_sparse` users unconditionally (not opt-in).
2. Add a line to the Accuracy Tests section noting the GSM8K runs exercise the decode-side change too (not just prefill), since that's the de facto validation for it.

Once the description reflects the actual scope, this looks ready to approve from our side.

HadarIngonyama:
thanks @zorrofox, I updated the description as you requested.

zorrofox (approved):
Approving now that #1508 is merged and this collapses to its own delta (`210c6ce1`, single file, +13/−4).

Basis:

- **Root cause verified independently**: `build_index_share_map` never reads `skip_offset` (dead parameter since #1478 — our legacy, not introduced here), so the removed gate's only runtime effect was the erroneous dense front-block. The fix is the minimal correct edit.
- **Scope disclosure is now complete**: the body explicitly frames the unconditional decode-side routing change for existing `dsa_sparse` deployments, and calls out that the n=50/200/1319 GSM8K pairs exercise the corrected decode routing de facto.
- **Independent perf confirmation**: on TPU v7x tp16 (GLM-5.2 FP8) we measured the full stack with this fix at 48k single-shot prefill 28.59s vs 38.4s without it (−25.5%), consistent with the ~25% figure quoted in the body; 16k lands at 9.77s. Accuracy on our side: long-context GSM8K paired runs and a 25-sample needle grid at 76k context show no regression attributable to this change.

One optional follow-up (non-blocking): with the gate gone, `self.skip_offset` becomes a dead field on the backend (constructed and serialized, never read) — fine to drop in a later cleanup.

PR #1514, opened 2026-08-06
feat(dsa): packed-ragged sparse-MLA prefill — batching + radix + chunked prefill

HadarIngonyama:
> **Stacked on #1512 → #1508.** Third in a chain: #1508 (opt-in sparse prefill) ← #1512 (`skip_offset` full-not-dense) ← **this** (batching/radix/chunked). It edits code introduced by both, so until they merge the diff below also contains their commits — **please review the top commit only** (`feat(dsa): packed-ragged sparse-MLA prefill …`). Merge order is #1508 → #1512 → this; I'll rebase down the stack as each lands and the diff will collapse to the batching delta (4 files).

## Motivation

#1508 shipped DSA sparse prefill in a deliberately narrow scope — **single-sequence, single-shot** — with fail-fast guards rejecting request batching, radix/prefix caching, and chunked prefill. That scope was fine to prove the kernel, but it blocks real serving: no concurrent throughput, no shared-prefix reuse, and prompts longer than one chunk can't be admitted.

This PR generalizes the sparse prefill to a **packed-ragged** kernel that attends multiple variable-length requests packed into a single EXTEND grid, and lifts the three guards. The single-sequence path is preserved as the `num_seqs == 1` special case (bit-identical to the #1508 kernel), so there is no regression on the existing path.

## Modifications

- **Kernel** `srt/kernels/dsa/sparse_mla_prefill.py`: packed-ragged sparse-MLA prefill matching the dense FlashAttention contract — per-query causal bounds + per-request page-table base offsets so N ragged requests share one grid; `prefill_write_and_attend_ragged` fuses the self-write and page-sparse attend over the packed batch.
- **Backend** `srt/layers/attention/dsa_sparse_backend.py`: route EXTEND through the ragged path (`q_seq_id` / `seq_lens` / `cu_kv_lens` / `page_indices`); `num_seqs == 1` stays a special case equal to the prior kernel.
- **Guards** `srt/model_executor/model_runner.py`: relax the `DSA_PREFILL_SPARSE` guards so `max_running_requests > 1` (batching), radix cache, and chunked prefill are supported — warn instead of hard-error.
- **Tests** `test/srt/kernels/dsa/test_sparse_mla_prefill_parity.py`: ragged parity + invariant gates (G0–G6) and A1 prefix-equivalence / A2 chunk-equivalence, CPU-interpret, exact vs the single-shot reference.

## Correctness Tests

CPU-interpret parity (`interpret=True`, fp32 oracle, `max|err| < 2e-3`, most gates exact 0.0):

| Gate | Proves | Result |
|---|---|---|
| G0 single-seq equivalence | ragged `num_seqs==1` == existing `(B,S)` kernel | exact |
| G1 batch-of-identical | N identical requests each == single-request output | exact |
| G2 cross-seq isolation | request A unchanged batched beside unrelated B | exact |
| G3 permutation invariance | permuting request order permutes outputs | exact |
| G4 dense==sparse superset | `index_topk ≥ pages(seq)` ⇒ sparse == dense | pass |
| G5 self-write + canary | non-padded slots written; `loc==-1` untouched | pass |
| G6 metadata invariants | packed-batch contract: `cu_q_lens`/page-table width/topk-page-id bounds hold across shapes | pass |
| A1 prefix-equivalence | cached-prefix + extend == single-shot full prefill | exact |
| A2 chunk-equivalence | N-chunk prefill == single-shot prefill of same prompt | exact |

Device (v7x, GLM-5.2-FP8, 8 chips / `tp16-ep16-dp1`):

- **GSM8K — paired full 1319, batched + radix ON**, `max_running=32`, greedy, one server flipping only `DSA_PREFILL_SPARSE` (dense control verified byte-identical except the flag via `/proc/<pid>/environ`): **sparse 0.9674 vs dense 0.9689** (Δ −0.0015 = 2 problems in dense's favor, within noise) — no regression at full scale from batching + prefix caching. 3 concurrent requests answer correctly with no cross-request bleed; burst completion confirms real `max_running`>1 batching. (Standard prompts are <2k, so `index_topk=4096` selects all pages ⇒ sparse is a superset of dense — the clean plumbing/no-regression control, now paired and full-size.)
- **A2 chunked (`chunk=2048 < context=8192`).** Needle recall **6/6** across chunk boundaries (~5.1k-token, 3-chunk prompts; early-chunk needles recalled through two boundaries, selection genuinely sparse at 40 pages vs the 32-page budget). Long-context GSM8K (4-chunk ~7.2k prompts) matches the single-shot baseline (**0.867 vs 0.84**, within n-noise). Byte-exact chunk-invariance is not used as a gate: on repetitive prompts, fp reduction-order differences from accumulating KV chunk-by-chunk flip greedy even for the single-shot reference (output stays coherent) — margin-robust tests (needle, accuracy) are the correct signal.

*Single-sequence selection accuracy is validated in the parent PRs (#1508/#1512): paired n=200 `run_eval` (`test/srt/run_eval.py`, zero-shot chat) 0.960 vs 0.980, plus long-context 50/50 vs 48/50 under active pruning.*

**Long-context paired GSM8K, n=200 (batched).** The n=50 long-context pair from #1508 was deferred to ride with batching; here it is at n=200 on this PR's batched path. `gsm8k_longctx.py` builds ~7.2k-token many-shot prompts so the indexer genuinely prunes (~57→32 pages); one server, both arms identically batched (`--max-running-requests 16`, ctx 12288, `max_tokens=3072`, `index_topk=4096`), flipping only `DSA_PREFILL_SPARSE`:

| arm | accuracy | correct | truncated |
|---|---|---|---|
| sparse | **0.990** | 198/200 | 2 |
| dense  | **0.975** | 195/200 | 5 |

Δ +0.015 (sparse ≥ dense) — no accuracy degradation from selection-under-pruning at 4× the n=50 sample size, measured on the batched path the PR actually ships (both arms batched, so the paired delta isolates selection, not batching).

## Benchmarking and Profiling

Kernel microbench (`deploy/pod/dsa_prefill_phase_bench.py`, single device, per-device H=16, bf16, S=49152, K=32, page 128):

| T | (B,S) indexer / attend ms | ragged n=1 (P0) | n=8 | n=16 (P1 pack vs 1×T) |
|---|---|---|---|---|
| 16 384 | 71.9 / 87.6 | 85.6 (**0.98×**) | 118.7 (1.35×) | 124.4 (1.42×) |
| 32 768 | 143.4 / 165.1 | 167.2 (**1.01×**) | 205.2 (1.24×) | 235.4 (1.43×) |
| 49 152 | 212.1 / 244.9 | 245.2 (**1.00×**) | 285.9 (**1.17×**) | — |

- **P0**: ragged at `num_seqs==1` is within ~2% of the shipping `(B,S)` kernel (16k slightly faster) ⇒ no single-seq regression.
- **P1**: packing N seqs to the same total tokens stays ≤1.2× at realistic sizes (49k n=8 = 1.17×). Overhead grows with request *count* (per-request grid fixed cost), not total tokens — bounded ~1.4× at n=16, and lower the larger each request is (32k n=8 = 1.24×).
- **P5 mixed-length**: a realistic ragged batch `[512, 1024, 2048, 4096, 8192]` (15 872 tokens) attends in **95.2 ms** vs **103.0 ms** for the equal-length pack of the same count/total ⇒ **0.92× (mixed-length is *faster*)**, **0.0% padding waste** — so a real short+long mix packs *more* efficiently than the uniform case.

Batched throughput (`bench_serving` random dataset, in=1024 / out=256, distinct-length prompts `random-range-ratio 0.5`), the A3 payoff of relaxing `max_running==1`. Reported with **radix cache OFF** (`cache_hit_rate=0.000` verified at every point) so the curve is raw batched **compute** scaling, not prefix-cache hits:

| Concurrency B | req/s | total tok/s | mean TTFT | out-tok speedup |
|---|---|---|---|---|
| 1  | 0.096 | 93.9  | 2.86 s | 1.00× |
| 4  | 0.212 | 207.7 | 3.32 s | 2.22× |
| 8  | 0.267 | 259.1 | 4.27 s | 2.89× |
| 16 | 0.306 | 297.0 | 4.94 s | 3.35× |
| 32 | **0.344** | **333.4** | 6.32 s | **3.76×** |

Aggregate throughput scales **monotonically to ~3.6× at B=32** with no saturation regression. Decode dominates the latency growth (TPOT 41→444 ms as the batch fills); TTFT stays modest (2.9→6.3 s). The same sweep with **radix ON** lands within a few % (0.97–1.04×) — A1 caching adds no penalty on this decode-bound workload and only helps genuinely shared-prefix traffic. (An earlier `/v1/completions` shared-prefix probe reported 17.2→90.4 tok/s / 5.3×, but the identical prompts let radix cache the whole prefix and over-state scaling — the radix-OFF curve above is the headline.)

## Notes / follow-ups

- Throughput is measured **radix-OFF with distinct-length prompts** so it reflects raw batched compute; `bench_serving`'s "random" dataset shares prefixes, so radix-ON / identical-prompt runs show high `cache_hit` and over-state scaling.
- P1 n=8 @16k (1.35×) and n=16 (1.42–1.43×) are the points above the ≤1.2× guideline — small-L / high-request-count packing overhead (per-request grid fixed cost), not a concern for typical serving mixes (few large requests); the mixed-length P5 case is 0.92×.
- Cross-query KV page reuse (batch a block of queries against the union of their selected pages) is a further throughput follow-up, not in this PR.
- `bench_serving` tokenizer: GLM-5.2's `tokenizer_config.json` sets `tokenizer_class: TokenizersBackend`, which transformers 4.57 can't load; use a patched fast-tokenizer dir (`tokenizer_class: PreTrainedTokenizerFast`, drop the list-valued `extra_special_tokens`) via `--tokenizer`.

### DIFF
--- a/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill.py
+++ b/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill.py
@@ -319,6 +319,16 @@ def sparse_mla_attention(
         # packed page table: one flat physical-page list; each request's window
         # starts at cu_kv_lens[rid]//ps (the per-query ``base``). Per-query causal
         # bound is seq_lens[rid]. Both are resolved per token below.
+        #
+        # SMEM footprint: the whole flat table (PTW = sum of pages_per_seq over the
+        # packed batch) lives resident in SMEM via ``pt_spec`` at 4*PTW bytes. The
+        # block index map is constant in ``s`` (and ragged is B=1), so it is copied
+        # in once, not re-DMA'd per grid step. PTW is bounded by what fits in HBM:
+        # each page costs page_size*Dk_pad*2 HBM bytes vs 4 SMEM bytes (~1e4x), so
+        # any batch whose KV fits in HBM keeps this block in the low-KB range (~8KB
+        # at ctx 8k / page 128 / max_running 32). Reaching ~1MB here (~270k pages,
+        # e.g. 128k ctx / page 64 / max_running 128) implies ~20GB of resident KV,
+        # i.e. an HBM OOM would hit first.
         PTW = page_indices.shape[0]
         pt_arg = page_indices.reshape(1, 1, 1, PTW).astype(jnp.int32)
         qsid = jnp.clip(q_seq_id, 0, seq_lens.shape[0] - 1).astype(jnp.int32)
--- a/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill.py
+++ b/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill.py
@@ -50,8 +50,10 @@ def _sparse_mla_kernel_chunked_hminor(
     q_ref,  # [1, 1, Hp, Dk]  Hp padded to the bf16 sublane tile (16), NOT to 128
     idx_ref,  # [1, 1, 1, K]    SMEM
     pos_ref,  # [1, 1, 1, 1]    SMEM
+    kvlen_ref,  # [1, 1, 1, 1]  SMEM  per-query causal bound (seq_lens[rid]; == T single-seq)
+    base_ref,  # [1, 1, 1, 1]   SMEM  per-query page-table base (cu_kv_lens[rid]//ps; 0 single-seq)
     kv_hbm,  # flat: [B, T(+RBF), Dk] HBM;  paged: [1, num_pages*page_size, Dk] HBM
-    pt_ref,  # [1, 1, 1, max_pages] SMEM  page table for this request (paged only; dummy if flat)
+    pt_ref,  # [1, 1, 1, PTW] SMEM  page table (per-request slice at base + logical page)
     o_ref,  # [1, 1, Hp, Dv]
     kv_scratch,  # [CBR, Dk] VMEM  one chunk of gathered latent
     sem,  # DMA semaphores (G,)
@@ -60,12 +62,12 @@ def _sparse_mla_kernel_chunked_hminor(
     Dv: int,
     RB: int,
     RBF: int,
-    T: int,
     K: int,
     G: int,
     CBR: int,  # G*RBF rounded up to the 128-lane tile (score's key axis is the lane axis)
     paged: bool = False,  # read from the packed 4D paged latent cache via pt_ref
     page_size: int = 0,  # tokens per page (paged only; static)
+    PTW: int = 1,  # page-table width (max_pages single-seq / total packed pages ragged)
 ):
     """Head-minor relayout of the chunked kernel for **few heads/device** (head-TP).
 
@@ -86,6 +88,8 @@ def _sparse_mla_kernel_chunked_hminor(
     Hp = q_ref.shape[2]
     q = q_ref[0, 0]  # [Hp, Dk] bf16
     qpos = pos_ref[0, 0, 0, 0]
+    kv_len = kvlen_ref[0, 0, 0, 0]  # per-query causal bound (== T in single-seq)
+    base = base_ref[0, 0, 0, 0]  # per-query page-table base (== 0 in single-seq)
     NC = (K + G - 1) // G
 
     lane = jnp.arange(CBR, dtype=jnp.int32)  # [CBR] iota on the key (lane) axis
@@ -103,10 +107,16 @@ def _sparse_mla_kernel_chunked_hminor(
             # but pp (from the page table) is opaque to Mosaic. Express the offset as
             # an explicit `r8 * 8` (page_size and o0 are both %8) so divisibility is
             # provable — the same idiom the dense paged kernel uses.
-            lt = u * RB  # logical token start
+            #
+            # ``u`` is a SEQ-LOCAL logical page id; the request's pages start at
+            # ``base`` in the packed page table (base==0 for the single-seq per-b
+            # table). Clamp base+lp to the table width so a padded/-1 lane can't
+            # gather out of range (the lane is masked out below anyway).
+            lt = u * RB  # logical token start (within the request)
             lp = lt // page_size  # logical page (static divisor)
             o0 = lt - lp * page_size  # in-page offset (multiple of RB)
-            pp = pt_ref[0, 0, 0, lp]  # physical page id
+            pidx = jnp.minimum(base + lp, PTW - 1)
+            pp = pt_ref[0, 0, 0, pidx]  # physical page id
             r8 = pp * (page_size // 8) + o0 // 8  # (P*page_size + o0) // 8, exact
             return kv_hbm.at[0, pl.ds(r8 * 8, RBF), :]
         return kv_hbm.at[b, pl.ds(u * RB, RBF), :]
@@ -149,10 +159,12 @@ def _sparse_mla_kernel_chunked_hminor(
             u_vec = jnp.where(sel, u_g, u_vec)
             row_vec = jnp.where(sel, lane - lo, row_vec)
             ir_vec = jnp.where(sel, inr, ir_vec)
-        kp = u_vec * RB + row_vec  # [CBR] key positions
+        kp = u_vec * RB + row_vec  # [CBR] seq-local key positions
         # (u_vec >= 0) drops topk padding lanes (unit id -1); their DMA was
         # clamped to unit 0, so the read is safe and only the mask excludes them.
-        valid = (u_vec >= 0) & (row_vec < RB) & (ir_vec > 0) & (kp <= qpos) & (kp < T)
+        # ``kv_len`` is the per-request bound (seq_lens[rid]); in the single-seq
+        # path it equals the static T, so this is a strict generalisation.
+        valid = (u_vec >= 0) & (row_vec < RB) & (ir_vec > 0) & (kp <= qpos) & (kp < kv_len)
         bias = jnp.where(valid, 0.0, -jnp.inf)  # [CBR] fp32
 
         for g in range(G):
@@ -208,7 +220,17 @@ def sparse_mla_attention(
     page_table=None,  # [B, max_pages] int32: logical page -> physical page. When set,
     # ``kv`` is the packed 4D paged cache and the kernel gathers from it.
     page_size: int | None = None,  # tokens/page (required with page_table)
-    seq_len: int | None = None,  # logical T for the causal mask (required with page_table)
+    seq_len: int | None = None,  # logical T for the causal mask (required, non-ragged paged)
+    # ── packed-ragged mode (multi-request extend) ────────────────────────────
+    # When ``q_seq_id`` is set, ``q`` is packed as [1, total_tokens, H, Dk] and the
+    # per-query causal bound / page-table base are resolved per token from the same
+    # ragged metadata the dense path uses. ``page_indices`` (flat packed physical
+    # pages) replaces the per-request ``page_table``; ``indices`` are seq-local page
+    # ids relative to each request's window (page_indices[cu_kv_lens[rid]//ps + p]).
+    q_seq_id=None,  # [total_tokens] int32  token -> request id (enables ragged mode)
+    seq_lens=None,  # [num_seqs] int32       per-request kv length (causal bound)
+    cu_kv_lens=None,  # [num_seqs+1] int32    page-aligned kv offsets (page_indices stride)
+    page_indices=None,  # [total_pages] int32  packed physical page ids
 ):
     """Sparse MLA-latent attention (head-minor chunked kernel).
     Returns ``[B, S, H, kv_lora_rank]`` (float32).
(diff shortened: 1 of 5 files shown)


HadarIngonyama:
modified description after running full benchmarks.

python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
```diff
@@ -36,13 +37,21 @@
...
 # to O(budget) flat vs the union path which saturates 512 pages at long ctx.
 _PAGE_TOPK_BUDGET = int(os.environ.get("DSA_PAGE_TOPK", "0"))
 
+# Opt-in: run EXTEND/prefill through the fused sparse-MLA prefill kernel
+# (``sparse_mla_prefill.sparse_mla_attention``) at page granularity instead of the
+# dense fallback. Default OFF ⇒ prefill behaviour is unchanged. Page-level
+# selection (read_block == page_size) consumes the indexer's page-topk directly.
+# Scope: packed-ragged extend — supports batching (max_running>1) via the same
+# ragged metadata the dense path uses. Radix/prefix caching and chunked prefill are
+# still gated off in model_runner (follow-ups A1/A2).
```
zorrofox: Stale comment — this says radix/prefix caching and chunked prefill are "still gated off in model_runner (follow-ups A1/A2)", but this same PR lifts those gates in `model_runner.py` (whose new comment says all three are supported) and ships the A1/A2 gates as tests. Looks like a leftover from an intermediate state of the branch — please update, since this is exactly the comment someone will read to decide whether radix + sparse-prefill is safe to enable.

test/srt/kernels/dsa/test_sparse_mla_prefill_parity.py
```diff
@@ -0,0 +1,759 @@
...
+def _seq_slice(c, i):
+    """Global token indices of request i's REAL tokens."""
+    base = i * c["S_pad"]
+    L = c["seq_lens_list"][i]
+    return np.arange(base, base + L)
+
+
+def test_ragged_parity_varlen():
+    """Ragged multi-request prefill vs masked-softmax oracle (varying seq_lens)."""
+    o, _, c = _run_ragged([384, 200, 512, 129], K=4, seed=1)
```
zorrofox: The varlen gate covers [384, 200, 512, 129] — all multi-page requests. Two boundary lengths worth adding to this list (cheap, same harness):

- **L=1**: a single-token extend packed beside long requests — this is precisely the "decode request as a 1-query extend" case the backend's MIXED-mode comment relies on, and the shape most likely to hit an off-by-one in the per-request base/causal-bound resolution. Nothing currently exercises it.
- **L=128 (== page_size)**: an exact page-boundary length, so the request fills its page with no partial remainder.

python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill.py
```diff
@@ -0,0 +1,586 @@
...
+        kvlen_arg = jnp.full((B, S, 1, 1), T, jnp.int32)  # static causal bound
+        base_arg = jnp.zeros((B, S, 1, 1), jnp.int32)  # per-b table => base 0
+    else:
+        kv = jnp.pad(kv, ((0, 0), (0, RBF), (0, 0)))
+        # dummy 1-wide page table so the kernel signature is uniform (unused when flat).
+        PTW = 1
+        pt_arg = jnp.zeros((B, 1, 1, 1), jnp.int32)
+        kvlen_arg = jnp.full((B, S, 1, 1), T, jnp.int32)
+        base_arg = jnp.zeros((B, S, 1, 1), jnp.int32)
+    pt_spec = pl.BlockSpec((1, 1, 1, PTW), lambda b, s: (b, 0, 0, 0), memory_space=pltpu.SMEM)
```
zorrofox, line 350: Sizing question, non-blocking: in ragged mode this SMEM block is `4 * len(page_indices)` bytes, and `page_indices` spans the whole packed batch. At the bench config here (ctx 8k, page 128, max_running 32) that's ~8KB, but a long-context deployment (e.g. 128k context, page 64, max_running 128) puts it at ~270k entries ≈ 1MB per grid step. Have you compiled/run at a shape in that range, or should the docstring note a practical bound? (The index map is constant in `s`, so the pipeline shouldn't re-copy it per step — the question is purely the resident SMEM footprint.)

test/srt/kernels/dsa/test_sparse_mla_prefill_parity.py
```diff
@@ -0,0 +1,759 @@
...
+            positions[t] = j
+            loc[t] = phys[i, j // _PS] * _PS + (j % _PS)
+
+    cache = np.zeros((num_pages, _PS, 1, _DK_PAD), np.float32)
+    seq_lens = np.asarray([s["L"] for s in seqs], np.int32)
+    cu_q_lens = np.arange(num_seqs + 1, dtype=np.int32) * S_pad
+    cu_kv_lens = np.arange(num_seqs + 1, dtype=np.int32) * (pages_per_seq * _PS)
+    page_indices = phys.reshape(-1)
+
+    # G6 metadata invariants (host-side, fail fast before the kernel).
```
zorrofox, line 396: Nit: the PR table lists G6 as a gate alongside G0–G5, but here it's harness-internal asserts inside `_assemble` rather than a standalone test — either promote it to a `test_gate_G6_*` or relabel it in the table so the gate list matches what CI actually runs.

zorrofox (commented):
Line-level pass on the batching delta (top commit only) — four notes below. All four anchor to code introduced in this PR's own delta, none inherited from the stacked #1508/#1512 commits. Reran the parity suite locally: 13/13 pass.

python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
HadarIngonyama: fixed

test/srt/kernels/dsa/test_sparse_mla_prefill_parity.py
HadarIngonyama: added 1 and 128 - the tests pass.

test/srt/kernels/dsa/test_sparse_mla_prefill_parity.py
HadarIngonyama: turned G6 into its own test - it passes.

python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill.py
HadarIngonyama: Good catch — added a docstring note on the bound rather than force-compiling a 128k/128-way shape, since that config isn't reachable in practice. The flat packed page table is `4*PTW` bytes resident in SMEM (`PTW = Σ pages_per_seq`), and since the block index map is constant in `s` (ragged is B=1) it's copied in once, not re-DMA'd per grid step — so it's footprint, not per-step traffic, as you noted.

The practical bound is that `PTW` is capped by HBM: each page costs `page_size*Dk_pad*2` HBM bytes for its actual KV vs 4 bytes for its SMEM table entry (~1e4×). So any batch whose KV fits in HBM keeps this block in the low-KB range (~8KB at the bench config); the ~1MB figure you computed corresponds to ~270k pages ≈ ~20GB of resident KV, i.e. an HBM OOM on the KV cache hits well before the page table stresses SMEM. Happy to add a large-`PTW` compile smoke check if you'd prefer the demonstration over the argument.

HadarIngonyama:
@zorrofox Ran the deferred long-context pair at n=200 on the batched path this PR ships (not the single-seq build): ~7.2k-token many-shot prompts where the indexer prunes ~57→32 pages, one server with both arms identically batched (max_running=16, ctx 12288, max_tokens 3072, index_topk 4096), flipping only DSA_PREFILL_SPARSE. Sparse 0.990 (198/200) vs dense 0.975 (195/200) — Δ +0.015, sparse ≥ dense, so no degradation under pruning at 4× the n=50 sample size. Since both arms are batched the delta isolates selection rather than batching. That closes the long-context item deferred from #1508.

zorrofox:
@HadarIngonyama 
#1512 is merged, so this is now the last PR of the stack. A single `git rebase --onto origin/main <old-1512-head>` dropping the carried #1508/#1512 commits should clear the conflict and collapse the diff to the pure batching delta. We'll do the final re-review promptly once it's up.

HadarIngonyama:
@zorrofox thanks, I rebased

zorrofox (approved):
Approving — final re-review after the rebase. A `range-diff` of the two carried commits against the previously reviewed revision shows zero content drift (all deltas are base-context movement from #1508/#1512 landing); the diff now collapses to the pure packed-ragged batching increment, and the four round-2 review items remain addressed in `9a6e9bbf`. CI is green.

Independent validation from our side (TPU v7x tp16, GLM-5.2 FP8): with this PR in the stack, chunked prefill at 8192 measures within ~3% of single-shot TTFT at 16k–48k, and it is what first made 110k-token prefill measurable on this hardware (single-shot OOMs). Long-context GSM8K pairs and a 25-sample needle grid at 76k context showed no accuracy regressions attributable to the stack.

alexnails (approved):

PR #1516, opened 2026-08-09
test(dsa): DP parity gates for the packed-ragged sparse prefill

tsvikas:
> **Ready to review — the stack it waited on has landed.** #1508, #1512 and #1514 are all merged, so this is now a **single commit, two files, clean against `main`** (one new test file + one line in `run_suite.py`). No runtime code. CI is green (all 20 checks, including the full TPU suite).

## Motivation

`dsa_sparse` is structurally data-parallel (every `shard_map` in `DSASparseAttentionBackend` uses `P("data", ...)`) and #1514 removed the last `max_running_requests == 1` guard — but nothing exercises `dp_size > 1`. `test_mla_attention.py` hardcodes `dp_size=1` and runs only on a single-chip runner, and `DSASparseAttentionBackend` has no test at all. Every other attention backend has a DP test (`test_flashattention_dp.py`, `test_gdn_attention_dp.py`, `test_kda_attention_dp.py`, `test_mixed_chunk_dp.py`).

**Why the backend and not the kernel:** under `shard_map` the kernel only ever receives its own shard, so `dp=2` with `T/2` tokens is the same call as `dp=1` with `T/2` tokens — a kernel-level DP test would be exercising JAX. What DP can break lives above it: `pages_per_seq` derived from *local* shard shapes, seq-local `topk_pages` resolved against a per-request `page_indices` base, token-axis arrays lining up with the `ql`/`kvc` sections, and page ids that must stay **rank-local** into a page axis sharded `P("data")`.

## Modifications

- **New** `python/sgl_jax/test/test_dsa_sparse_prefill_dp.py`, six CPU tests:
  - `test_dp_invariance_*` (dp2, dp4, multi-request-per-rank, wider head block) — a request gives the same output alone at `dp=1` as inside the `dp=D` batch, with attention TP held **fixed** so only DP can move the result.
  - `test_cross_rank_no_bleed` — changing rank 1's tokens must not move rank 0's output.
  - `test_metadata_shards_evenly` — `page_indices`/`seq_lens` divide by `dp_size`, the per-shard quotient matches `dp=1`, and page ids stay rank-local.
- `test/srt/run_suite.py` — registered under `unit-test-cpu`.

Follows the existing DP tests: `CustomTestCase`, `ForwardBatch.init_new`, per-file fixture (there is no shared builder — `flashattention_common.py` calls `set_mesh()` at import and hardcodes `dp_size=1`). Most cases use `attn_tp=2`: sharding over a size-1 axis is a no-op, so at `attn_tp=1` a spec wrongly reading `P("data", None, None)` would be indistinguishable from the correct `P("data", "tensor", None)`.

## Accuracy Tests

**N/A — test-only, no runtime code changed.** For context, the path this covers was validated on TPU v7x (GLM-5.2-FP8, `DSA_PREFILL_SPARSE=1`, `ep16`): GSM8K n=100 at dp1/dp2/dp4/dp16 = 0.95 / 0.97 / 0.94 / 0.95, and 40/40 long-context needle retrieval at both dp1 and dp16 (serial, ~8k context, `attention_tp` 16 and 1 respectively) with no request ever returning another request's needle.

## Benchmarking and Profiling

**N/A — test-only, no runtime code changed.**

## Threshold rationale

`rtol=atol=1e-5` on fp32. Injecting the classic DP fault — global page ids where the allocator hands out rank-local ones — moves the compared output by **3.4**, five orders of magnitude above the tolerance, so the gates are not near their threshold. Both parity gates also carry inline sensitivity controls asserting the two compared sides are genuinely distinguishable, so a green run cannot be green because it compared identical computations.

## Estimated wall-clock

**~1.0 min measured** (6 tests, CPU, 8 simulated devices); declared `1.5` for runner headroom. `unit-test-cpu` is unpartitioned under a 30-min cap and gates the TPU stages — happy to trim cases if reviewers would rather not spend the budget here.

## Stability

Deterministic: seeded `np.random.default_rng`, no network, filesystem, server or model download. CPU-only via Pallas `interpret=True` plus `XLA_FLAGS=--xla_force_host_platform_device_count=8`. Two module-level patches, both `mock.patch` under `enterContext`: `_PREFILL_SPARSE` (the opt-in flag) and `interpret=True` on the kernel entry point — a documented parameter of that function, not a stubbed internal.

## Checklist

- [x] The test plan: `python -m pytest python/sgl_jax/test/test_dsa_sparse_prefill_dp.py -v`
- [x] (Optional) The necessary documentation update. — N/A

### DIFF
--- a/python/sgl_jax/test/test_dsa_sparse_prefill_dp.py
+++ b/python/sgl_jax/test/test_dsa_sparse_prefill_dp.py
@@ -1,0 +1,579 @@
+"""DP parity for the packed-ragged DSA sparse prefill path.
+
+Run with: JAX_PLATFORMS=cpu XLA_FLAGS=--xla_force_host_platform_device_count=8 \\
+    python -m pytest python/sgl_jax/test/test_dsa_sparse_prefill_dp.py -v
+
+Drives ``DSASparseAttentionBackend``, not the kernel: under ``shard_map`` the
+kernel only ever sees its own shard, so a kernel-level DP test would be
+exercising ``shard_map`` rather than sglang-jax. What DP puts at risk is the
+wiring above it — ``pages_per_seq`` derived from *local* shard shapes, seq-local
+``topk_pages`` resolved against a per-request ``page_indices`` base, token-axis
+arrays lining up with the ``ql``/``kvc`` sections, and page ids that are
+**rank-local** into a page axis sharded ``P("data")``.
+
+Attention TP is held fixed between each reference and its DP run (the meshes
+differ only in how many devices sit on the ``data`` axis), so a difference can
+only come from DP, never from a changed head split.
+
+Most cases use ``attn_tp=2`` rather than 1. Sharding over a size-1 axis is a
+no-op, so at ``attn_tp=1`` a spec that wrongly read ``P("data", None, None)``
+would behave exactly like the correct ``P("data", "tensor", None)`` and the
+tensor half of every spec would go unverified.
+"""
+
+from __future__ import annotations
+
+import os
+
+# Must precede the first jax import: the DP meshes below need >1 CPU device.
+# Append rather than setdefault — XLA_FLAGS set for any unrelated reason
+# (--xla_dump_to=..., a profiling flag) would otherwise drop the device count,
+# leaving one device, and this file is the DP gate in unit-test-cpu.
+if "xla_force_host_platform_device_count" not in os.environ.get("XLA_FLAGS", ""):
+    os.environ["XLA_FLAGS"] = (
+        os.environ.get("XLA_FLAGS", "") + " --xla_force_host_platform_device_count=8"
+    ).strip()
+os.environ.setdefault("JAX_PLATFORMS", "cpu")
+
+import functools
+import unittest
+from unittest import mock
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+from jax.sharding import NamedSharding
+from jax.sharding import PartitionSpec as P
+
+from sgl_jax.srt.layers.attention import dsa_sparse_backend as dsa_mod
+from sgl_jax.srt.layers.attention.dsa_sparse_backend import DSASparseAttentionBackend
+from sgl_jax.srt.layers.radix_attention import RadixAttention
+from sgl_jax.srt.managers.schedule_batch import ModelWorkerBatch
+from sgl_jax.srt.mem_cache.memory_pool import MLATokenToKVPool
+from sgl_jax.srt.model_executor.forward_batch_info import ForwardBatch, ForwardMode
+from sgl_jax.srt.utils.mesh_utils import create_device_mesh
+from sgl_jax.test.test_utils import CustomTestCase
+
+# Fail at import rather than skip: on CPU the device count is forced above, so
+# too few devices means the flag did not take effect, not that the machine is
+# small. Skipping there would report a green run with zero DP coverage.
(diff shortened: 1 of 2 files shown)


tsvikas:
#1514 has landed, so this is unblocked — the diff is now a single commit against `main`: one new test file plus one line in `run_suite.py`. No runtime code, CI green. I have updated the description accordingly (the old "review only the top commit" note no longer applies).

@zorrofox @alexnails — you reviewed the stack this sits on; would either of you have time to take a look? I cannot add reviewers from a fork, hence the mention.

zorrofox (commented):
Thanks for building this — a DP gate for the packed-ragged sparse prefill is exactly the coverage this path has been missing, and the suite runs clean locally (7 tests, ~2 min on CPU). A few notes on gate effectiveness, roughly in order of importance:

**1. The fixture's page-id convention is the opposite of production.** The default `cache_loc` uses rank-local slot ids `0..255` (pages 0..15) and asserts `pages.max() < PAGES_PER_RANK`, but the paged allocator hands out local pages `1..pages_per_rank` and reserves page 0 (`allocator.py` — `arange(1, pages_per_rank + 1)`). So the test populates the page production reserves, never touches the boundary page production actually uses, and would flag a correct implementation that surfaces real ids as a false alarm. This matters in practice: it kept a real boundary bug out of the gate's reach — `_scatter_paged` routed padding rows to `cache3d.shape[0] - 1`, which is an allocatable page (fix just landed in #1585, commit 64122e16). Shifting the fixture ids up by one page would put this class of bug in range.

**2. Nothing asserts the sparse branch was actually taken.** `setUp` patches `dsa_mod._PREFILL_SPARSE` and the kernel entry, but if a later refactor moves the env read or the call site, the whole file silently degrades into a dense-path test that still passes (dense output is nonzero and dense dp=N vs dp=1 parity holds). Cheap fix: wrap the kernel patch in `mock.Mock(wraps=...)` and assert it was called after each `_run`.

**3. The pairwise sensitivity control is a no-op in 3 of 4 invariance tests.** It skips any pair with mismatched shapes (`if x.shape != y.shape: continue`), and in `test_dp_invariance_dp4`, `dp2_with_batching`, and `dp2_tp2_wide_heads` every q_len is distinct, so all pairs are skipped and the control asserts nothing. Comparing overlapping prefixes instead (`n = min(len(x), len(y))`) restores it.

**4. The XLA_FLAGS guard can silently skip everything.** If the environment already sets `xla_force_host_platform_device_count` to 2–7, the append is skipped, the `>= 2` import check passes, `_mesh` raises SkipTest for every DP case, and pytest exits green with only the two metadata tests run — the vacuous pass the comments aim to prevent. Parsing/replacing the existing value (or failing when it is < 8) closes the hole.

**5. Every q_len is a multiple of PAGE_SIZE.** With 96/64/112/80/48/32, the `align_to` page-alignment step in `cu_kv_lens` is indistinguishable from a raw cumsum and the padded-KV-tail path is untested; a single non-aligned length (e.g. 90) in one workload gives the parity check teeth there.

Minor: the metadata re-`device_put` loop duplicates what `get_forward_metadata` already does and would mask a sharding regression inside it (an assertion on the committed sharding would gate it instead); all cases drive `indexer_type='full'`, so the IndexShare shared-layer branch the docstring names as a DP risk has no coverage yet; and the inline DummyModelConfig/DummyRunner stub is now the sixth copy in the tree — might be worth extracting into `test_utils.py` while here.

None of these block the direction — happy to see this land with the fixture convention fixed, the rest can be follow-ups.

PR #1517, opened 2026-08-10
fix(quant): close static-FP8 kernel gaps — gmm_v2 W8A8+block-wise rhs_scale, v7x dense tuned tiles, fused_v2 MoE scale sharding

zorrofox:
## Summary

Serving block-wise static FP8 checkpoints (`weight_block_size=[128,128]`, e.g. GLM-5.2-FP8 / DeepSeek-style) currently hits three gaps:

1. **MoE routed experts fall back to gmm_v1.** `is_supported_by_gmm_v2` rejects W8A8 + block-wise `rhs_scale` (k_blocks > 1), so the static-FP8 MoE path (activation quant enabled from `activation_scheme: dynamic`) drops to gmm_v1 with 128-row alignment — at decode bs=1 that pads 8 topk rows to 128 (16× wasted work) plus extra pad ops.
2. **Dense block-wise `quantized_matmul` has no v7x tuned tiles for GLM-5.2 shapes.** All 469 dense-linear kernel calls per decode step (78 layers × MLA q_a/q_b/kv_a/o_proj, dense-layer MLP, shared experts, indexer) run on clamped fallback tiles: 2.49 ms/step vs ~0.5 ms of pure HBM time.
3. **`moe_backend=fused_v2` + static FP8 crashes at warmup.** The GLM-5 static scale weight-mappings hardcode `sharding=("expert", None, None)` (the epmoe `("expert","tensor")` moe_mesh convention), while fused/fused_v2 run shard_map on the main mesh with `P(("data","tensor"))`:
   ```
   ValueError: in_specs passed to shard_map: P(('data', 'tensor'), None, None, None) does not
   match the specs of the input: P('expert', None, None, None) for arg: float32[256@expert,48,1,2048]
   ```

### Changes

- **`gmm_v2.py`** — support W8A8 + block-wise `rhs_scale` (sub-channel):
  - When the rhs is block-quantized, align the in-kernel lhs activation-quant block to the rhs quant block, and rescale each K block's partial product by its own rhs scale inside the quantized matmul loop (per-token lhs scale × per-block rhs scale, same structure as the existing per-channel path).
  - Remove the W8A8+block-scale rejection from `is_supported_by_gmm_v2`.
  - Keep the W8A16 block-scale slice 2D (`[blk, 0:1, :]`) — the previous 1D slice trips `jax_numpy_rank_promotion='raise'` (latent; surfaces on v7x where int8 rhs has no hardware dot support and W8A8 requests degrade to this W8A16 path).
- **`tuned_block_sizes.py`** — add TPU v7 entries for GLM-5.2 tp16 per-device dense shapes (n_batch 1/8/32/64 × 8 shapes). Tiles are chosen to divide the operand dims exactly so the kernel wrapper does no out-of-kernel `jnp.pad`, with whole-K `in_block` where VMEM allows; `out_block` stays a multiple of the 256-lane MXU column (n_out=576 uses 256 with minimal padding). Each entry compile/run-validated on v7x.
- **`glm5_moe.py`** — static-FP8 MoE scale mappings inherit the expert-dim spec from the weight mapping (`mapping.sharding[0]`) instead of hardcoding `"expert"`, so epmoe keeps the moe_mesh layout and fused/fused_v2 get the main-mesh layout.
- **`gmm_test.py`** — parameterize `test_gmm_v2_activation_weight_quantized` with `block_size=[128, None]` (resolves the in-code TODO for sub-channel coverage) and relax tolerance slightly for the sub-channel combos (bf16 per-K-block rescale reorders roundings vs the f32 reference; single-element margin violations otherwise).

### Why not tune-and-keep gmm_v1?

gmm_v2's 32-row alignment vs gmm_v1's 128 is structural at decode (topk rows ≪ 128), and v2 already owns the per-channel W8A8 path; extending it to sub-channel keeps one code path for all static-FP8 checkpoints.

## Results

GLM-5.2-FP8 (block-wise static, 753B) on TPU v7x, 16 chips / 2 hosts, tp16/ep16, `dsa_sparse` attention, epmoe backend, 1K-context single-stream decode (differential TPOT (t600−t100)/500, ×2 runs, <0.1% variance):

| config | decode TPOT | dense qmm / step | MoE / step |
|---|---|---|---|
| static FP8, before | 17.52 ms | 2.49 ms | gmm_v1 1.21 ms + pad |
| static FP8, **this PR** | **14.00 ms** | **1.19 ms** | **gmm_v2 1.00 ms** |
| W8A8 dynamic quant (BF16 ckpt), same commit | 14.05 ms | — | gmm_v2 (per-channel) |

Static FP8 now matches the dynamic-quant W8A8 baseline while skipping the per-step activation amax (~1.7 ms at these shapes, already included above) and staying deterministic. XProf per-step device module: 13.88 ms vs 14.39 ms (W8A8).

Concurrency sanity: cc=8 steady decode 213 tok/s (pre-PR: 200 tok/s) — no regression.

## Test plan

- [x] `pytest python/sgl_jax/test/kernels/gmm_test.py -k gmm_v2_activation` on v7x: 96/96 pass (int8/fp8 × per-channel/sub-channel × group offsets). Two initial failures triaged: one is the jtu first-test compilation-cache assertion (environment artifact, passes with `JAX_ENABLE_COMPILATION_CACHE=false`), one was the sub-channel tolerance addressed above.
- [x] e2e GLM-5.2-FP8 v7x tp16: decode TPOT 17.52 → 14.00 ms (×2 runs each config)
- [x] GSM8K 200q 5-shot, temperature 0: **0.955** with this PR (pre-PR static FP8: 0.955; W8A8 dynamic baseline: 0.950) — no accuracy change
- [x] cc=8 steady decode throughput: 213 tok/s vs 200 pre-PR
- [x] fused_v2 + static FP8 no longer crashes at warmup (scale sharding fix verified by serving)
- [x] `pre-commit run` (isort/ruff/black/mypy) clean

## Notes for reviewers

- Routing change: W8A8 + block-wise static checkpoints previously ran gmm_v1 and now take gmm_v2 (same numerics family as the existing per-channel W8A8 v2 path; covered by the new sub-channel unit tests + GSM8K). If you'd prefer an escape hatch env to force the old routing during rollout, happy to add one.
- The stock `tune_blockwise_block_sizes.py` sweep times only the Pallas kernel segment from the trace, which ignores the out-of-kernel `jnp.pad` when tiles don't divide the operand — a sweep-picked table initially regressed e2e (pad 0.95→2.39 ms/step). The entries here use exact-divisor tiles instead; a follow-up could add pad-aware timing to the tuner.
- `deepseek_v3.py` and `minimax_m2.py` hardcode the same `sharding=("expert", None, None)` for static MoE scales; they are unaffected today (epmoe-only) but would hit the same shard_map rejection if switched to `moe_backend=fused_v2`. Kept out of this PR to stay minimal — happy to apply the same one-line fix there as a follow-up if preferred.

### DIFF
--- a/python/sgl_jax/srt/kernels/gmm/megablox_gmm_kernel/gmm_v2.py
+++ b/python/sgl_jax/srt/kernels/gmm/megablox_gmm_kernel/gmm_v2.py
@@ -239,7 +239,8 @@ def inner_kernel(
                         tiled_rhs[start_k:end_k, :],
                         preferred_element_type=jnp.float32,
                     ).astype(acc_ref.dtype)
-                    acc += block_acc * tiled_rhs_ref.scale[blk_i, 0, :].astype(acc_ref.dtype)
+                    # Keep the scale slice 2D [1, tile_n] for strict rank promotion.
+                    acc += block_acc * tiled_rhs_ref.scale[blk_i, 0:1, :].astype(acc_ref.dtype)
             else:
                 # Unquantized matmul path.
                 acc = jnp.matmul(
@@ -299,7 +300,16 @@ def inner_kernel(
                         preferred_element_type=preferred_element_type,
                     ).astype(acc_ref.dtype)
 
-                    acc_n += block_acc * block_scale.astype(acc_ref.dtype)
+                    block_acc = block_acc * block_scale.astype(acc_ref.dtype)
+                    if is_block_rhs_scale:
+                        # W8A8 block-wise rhs_scale: q_block_size == rhs_block
+                        # (enforced in make_gmm_configs), so this K block maps
+                        # 1:1 onto rhs scale row start_k // rhs_block. Keep the
+                        # slice 2D [1, col] for strict rank promotion.
+                        block_acc = block_acc * tiled_rhs_ref.scale[
+                            start_k // rhs_block, 0:1, start_n:end_n
+                        ].astype(acc_ref.dtype)
+                    acc_n += block_acc
 
                 acc_list.append(acc_n)
             acc = jnp.concatenate(acc_list, axis=1)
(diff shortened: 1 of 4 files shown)


cjx0709 (approved):
LGTM

PR #1518, opened 2026-08-10
feat(kernel): add native varlen attention for TPU

Yanko-7:
## Motivation

Models with vision encoders and other multimodal inputs often require bidirectional attention over ragged sequences packed into a contiguous token buffer.

The existing encoder FlashAttention kernel supports this layout by running dense attention over the complete packed extent and masking cross-sequence token pairs. This preserves correctness, but does not skip Q/K blocks belonging to different sequences. Its computation therefore remains approximately `O(T²)` for `T` packed tokens.

This PR introduces a TPU-native variable-length attention kernel whose work scales with `sum(sequence_length_i²)` instead of the square of the total packed token count.

## Implementation

The kernel is adapted from RPA3, with paging and KV-cache handling removed for packed encoder self-attention.

The public input layout is:

```text
q:          [token_capacity, num_q_heads, head_dim]
k, v:       [token_capacity, num_kv_heads, head_dim]
cu_seqlens: int32[max_num_sequences + 1]
num_seqs:   int32[1]
```

For each sequence, Q/K/V share the range defined by `cu_seqlens`.

The implementation provides:

- A BF16 MHA fast path using token-major Q/O and interleaved K/V.
- An RPA3-style packed fallback for GQA and F32.
- A single Pallas program that dynamically walks sequences and their Q/KV tiles.
- Double-buffered Q, KV and output DMA.
- Full, causal and sliding-window attention.
- Soft-cap, K/V scaling and attention-sink support.
- A `max_seq_len` argument used only for tuned block-size selection.
- v7x block-size lookup keyed by `(heads, align_to_128(head_dim), next_power_of_two(max_seq_len))`.

## TPU accuracy

Tested on the final commit on TPU v7x with JAX 0.9.0 and BF16 inputs against the pure-JAX reference implementation.

| Case | Sequence lengths | Q/KV heads | Mask | Max absolute error |
|---|---:|---:|---|---:|
| MHA | 63, 129, 257 | 16 / 16 | Full | 0.0078125 |
| GQA | 33, 129 | 8 / 2 | Causal | 0.0156250 |

Both cases pass with `rtol=2e-2, atol=1e-2`. The irregular MHA lengths also exercise the dynamic final KV DMA slice.

## Benchmark

Environment:

- TPU v7x, topology `2x2x1`
- JAX 0.9.0
- BF16, 16 heads, head dimension 128
- 2 warmup iterations; median of 5 measured iterations

The dense FlashAttention baseline uses segment IDs for correctness, but does not skip non-overlapping Q/K blocks. Its tuned block configuration is `(512, 1024, 256)`, with smaller blocks for the 256-token case.

The measured region includes each public operator call, device synchronization and operator-internal layout preparation. Random Q/K/V generation, metadata construction, and the external THD-to-BHTD transpose are excluded.

`Speedup = dense FlashAttention latency / varlen_attention latency`.

| Scenario | Dense FlashAttention | VarlenAttention | Speedup | Max abs error |
|---|---:|---:|---:|---:|
| 256 | 0.1733 ms | 0.1712 ms | 1.01x | 0.001953 |
| 1K | 0.1877 ms | 0.2565 ms | 0.73x | 0.000977 |
| 4K | 1.0353 ms | 1.0259 ms | 1.01x | 0.000977 |
| 8K | 3.5035 ms | 3.1264 ms | 1.12x | 0.000488 |
| 16K | 13.7646 ms | 11.9156 ms | 1.16x | 0.000488 |
| 32K | 53.3739 ms | 42.7728 ms | 1.25x | 0.000244 |
| 128 x 64 | 3.4500 ms | 1.3375 ms | 2.58x | 0.003906 |
| 256 x 64 | 13.5827 ms | 2.4768 ms | 5.48x | 0.007812 |
| 512 x 64 | 52.6324 ms | 4.7776 ms | 11.02x | 0.003906 |
| 8 x 1024 | 3.4480 ms | 0.9935 ms | 3.47x | 0.001953 |
| Mixed 8K | 3.4528 ms | 1.4727 ms | 2.35x | 0.001953 |

Geometric-mean speedup:

- All scenarios: **1.95x**
- Single full sequences: **1.03x**
- Ragged/segmented scenarios: **4.17x**
- Maximum: **11.02x**

## Limitations

- TPU-only Pallas kernel; no interpreter implementation.
- Only the BF16 MHA fast path has v7x tuning entries. GQA and F32 currently use default block sizes.
- Output rows after `cu_seqlens[num_seqs]` are unspecified.

## Tests

```bash
PYTHONPATH=python pytest -q -s \
  python/sgl_jax/test/multimodal/test_varlen_attention_kernel.py
```

- TPU v7x accuracy: 2/2 cases passed.

### DIFF
--- a/python/sgl_jax/srt/multimodal/kernels/varlen_attention.py
+++ b/python/sgl_jax/srt/multimodal/kernels/varlen_attention.py
@@ -1,0 +1,1608 @@
+# Adapted from sgl_jax.srt.kernels.ragged_paged_attention.ragged_paged_attention_v3.
+# Copyright 2025 The tpu-inference Authors. All rights reserved.
+# Modifications Copyright 2026 The SGLang-JAX Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Variable-length ("varlen") attention for TPU.
+
+Sequences are packed contiguously into a single ``[tokens, heads, head_dim]``
+buffer and delimited by ``cu_seqlens`` (the FlashAttention ``flash_attn_varlen``
+convention). There is no KV cache: for every sequence ``i`` the Q, K and V
+share the interval ``[cu_seqlens[i], cu_seqlens[i + 1])``.
+
+Two kernels back the single ``varlen_attention`` entry point, both a single
+``grid=(1,)`` Pallas program that walks every sequence/Q-tile/KV-tile/head with
+double-buffered DMAs:
+
+* **MHA fast path** — ``num_q_heads == num_kv_heads`` and BF16 Q/K/V.
+* **Packed fallback** — GQA (``num_q_heads != num_kv_heads``) or F32.
+
+Only output rows below ``cu_seqlens[num_seqs]`` are defined.
+
+Why the fast path is faster: the fallback keeps the RPA3 layout that word-packs
+``q_per_kv`` for paged-cache compatibility. Without a cache to preserve, the MHA
+path drops that and keeps Q/O token-major (``[tokens, heads, head_dim]``), so
+``load_bq`` is a plain VMEM slice with no dummy lane, no bitcast/strided-gather,
+and write-back is a transpose rather than a repack. Only K/V keep the
+word-interleaved (``K0,V0,K1,V1,...``) packing that lets one DMA carry both.
+GQA (where packing amortizes) and F32 stay on the fallback.
+"""
+
+from __future__ import annotations
+
+import functools
+from typing import Final
+
+import jax
+import jax.numpy as jnp
+from jax import lax
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+
+from sgl_jax.srt.multimodal.kernels.varlen_tuned_block_sizes import (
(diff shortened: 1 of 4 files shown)


python/sgl_jax/srt/multimodal/kernels/varlen_attention.py
```diff
@@ -0,0 +1,1608 @@
...
+    if num_kv_per_block is not None and num_kv_per_block <= 0:
+        raise ValueError("num_kv_per_block must be positive")
+    if vmem_limit_bytes is not None and vmem_limit_bytes <= 0:
+        raise ValueError("vmem_limit_bytes must be positive")
+
+    # Keep names in the signature to make validation calls self-documenting.
+    del sm_scale, mask_value, k_scale, v_scale
+
+
+def dynamic_validate_metadata(
```
lianfang0320: This function seems useless, please confirm it

Yanko-7: Done

python/sgl_jax/srt/multimodal/kernels/varlen_attention.py
```diff
@@ -0,0 +1,1577 @@
...
+from jax.experimental.pallas import tpu as pltpu
+
+from sgl_jax.srt.multimodal.kernels.varlen_tuned_block_sizes import (
+    DEFAULT_KV_BLOCK,
+    DEFAULT_Q_BLOCK,
+    get_tuned_block_sizes,
+)
+
+DEFAULT_MASK_VALUE: Final[float] = -0.7 * float(jnp.finfo(jnp.dtype("float32")).max)
+DEFAULT_VMEM_LIMIT_BYTES: Final[int] = 120 * 1024 * 1024
```
JamesBrianD, line 60: The current default size is 120MB. When benchmarking on v7x, was 64MB or 60MB used?

Yanko-7: When benchmarking, 128 MiB was used.

JamesBrianD (approved):

PR #1521, opened 2026-08-14
fix: [Docs] LoRA usage documentation

### LINKED ISSUE #594: [Docs] LoRA usage documentation
### Checklist

- [x] 1. Make sure you already searched for relevant issues and documentations.

### Documentation Issues or Suggested Changes

We need to add documentation related to LoRA usage.

Mr-Neutr0n:
Fixes #594

Added `docs/features/lora.md` with LoRA usage documentation (dynamic/static modes, server args, adapter selection, examples, limitations) and registered it in `docs/features/index.md` and `docs/index.rst`.

Could not run the suite locally: . Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/docs/features/lora.md
+++ b/docs/features/lora.md
@@ -50,7 +50,7 @@ uv run python -u -m sgl_jax.launch_server \
   --lora-paths adapter1=/path/to/adapter1 adapter2=/path/to/adapter2
 ```
 
-Adapters can also be specified as a JSON dict `{"adapter1": "/path/to/adapter1", ...}`.
+When constructing `ServerArgs` programmatically, adapters can also be specified as a dict `{"adapter1": "/path/to/adapter1", ...}` (the CLI `--lora-paths` only accepts `name=path` / `path` forms).
 
 ### Sending requests with a specific adapter
 
@@ -68,8 +68,10 @@ curl http://localhost:30000/v1/completions \
   }'
 ```
 
-The same field is used in the OpenAI chat completions protocol and in the Python engine API
-(`Engine.generate(..., lora_path="adapter1")`).
+The same field is used in the Python engine API
+(`Engine.generate(..., lora_path="adapter1")`). Chat completions does not yet
+forward `lora_path` to the generation request, so the adapter would be ignored
+there.
 
 ### Server arguments
 
@@ -77,11 +79,10 @@ The same field is used in the OpenAI chat completions protocol and in the Python
 |----------|---------|-------------|
 | `--enable-lora` | `None` | Enable dynamic LoRA (auto-enabled when `--lora-paths` is set). |
 | `--lora-paths` | `None` | Adapters to preload; `name=path` / `path` / dict format. |
-| `--max-loras-per-batch` | `8` | Maximum number of different adapters per batch. |
+| `--max-loras-per-batch` | `8` | Maximum number of different adapters per batch. Preloaded adapters are pinned, so at most `max_loras_per_batch - 1` (7 with the default) can be preloaded. |
 | `--max-lora-rank` | `None` | Maximum LoRA rank; auto-inferred from the loaded adapters when unset. |
 | `--lora-target-modules` | `None` | Modules to apply LoRA to (`q_proj`, `k_proj`, `v_proj`, `o_proj`, `gate_proj`, `up_proj`, `down_proj`, or `all`); auto-inferred from the adapters when unset. |
-| `--max-loaded-loras` | `None` | Maximum adapters kept loaded in memory. |
-| `--lora-eviction-policy` | `lru` | Eviction policy (reserved; no runtime eviction is triggered in the current version). |
+
 
 `qkv` and `gate_up` projections are automatically handled by merging the corresponding per-module
 LoRA weights, and adapters with different ranks are padded to `max_lora_rank`.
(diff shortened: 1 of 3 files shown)


docs/features/lora.md
```diff
@@ -0,0 +1,122 @@
...
+  -d '{
+    "model": "meta-llama/Llama-2-7b-chat-hf",
+    "prompt": "The capital of France is",
+    "max_tokens": 32,
+    "lora_path": "adapter1"
+  }'
+```
+
+The same field is used in the OpenAI chat completions protocol and in the Python engine API
+(`Engine.generate(..., lora_path="adapter1")`).
```
JamesBrianD: Does chat completions support this today? `OpenAIServingChat._convert_to_internal_request()` does not copy `request.lora_path` into `GenerateReqInput`, so the adapter is silently ignored and the request runs against the base model. I think we should remove this claim for now, or fix the request conversion separately.

docs/features/lora.md
```diff
@@ -0,0 +1,122 @@
...
+  --nnodes=1 \
+  --tp-size=1 \
+  --device=tpu \
+  --node-rank=0 \
+  --mem-fraction-static=0.8 \
+  --download-dir=/tmp \
+  --dtype=bfloat16 \
+  --skip-server-warmup \
+  --enable-static-lora \
+  --lora-scaling 1.0
```
JamesBrianD: This example will not get past `check_lora_server_args()`. Static LoRA also requires `--max-lora-rank`, `--lora-target-modules`, and `--max-loras-per-batch 1`. Could we update this to a command that can actually start?

docs/features/lora.md
```diff
@@ -0,0 +1,122 @@
...
+
+`qkv` and `gate_up` projections are automatically handled by merging the corresponding per-module
+LoRA weights, and adapters with different ranks are padded to `max_lora_rank`.
+
+## Static LoRA
+
+Static LoRA is intended for RL scenarios where only a single adapter is used and requests never
+switch adapters at runtime. It is mutually exclusive with `--enable-lora`:
+
+- Supports exactly one adapter (weights are merged into the base model; BGMV is not used)
```
JamesBrianD: This does not match the current static LoRA path. We still keep separate A/B buffers and run them through `BgmvLoRABackend`; the weights are not merged into the base weight. Could we describe the fixed-buffer/external-update behavior used by the RL flow instead?

docs/features/lora.md
```diff
@@ -0,0 +1,122 @@
...
+  --mem-fraction-static=0.8 \
+  --max-prefill-tokens=8192 \
+  --download-dir=/tmp \
+  --dtype=bfloat16 \
+  --skip-server-warmup \
+  --enable-lora \
+  --lora-paths adapter1=/path/to/adapter1 adapter2=/path/to/adapter2
+```
+
+Adapters can also be specified as a JSON dict `{"adapter1": "/path/to/adapter1", ...}`.
```
JamesBrianD: I don't think the JSON dict form works from the CLI. `--lora-paths` is parsed as `type=str, nargs="*"`, so a JSON value is treated as a regular path. Dict input only works when constructing `ServerArgs` programmatically. We should either make that distinction explicit or remove this sentence.

docs/features/lora.md
```diff
@@ -0,0 +1,122 @@
...
+The same field is used in the OpenAI chat completions protocol and in the Python engine API
+(`Engine.generate(..., lora_path="adapter1")`).
+
+### Server arguments
+
+| Argument | Default | Description |
+|----------|---------|-------------|
+| `--enable-lora` | `None` | Enable dynamic LoRA (auto-enabled when `--lora-paths` is set). |
+| `--lora-paths` | `None` | Adapters to preload; `name=path` / `path` / dict format. |
+| `--max-loras-per-batch` | `8` | Maximum number of different adapters per batch. |
```
JamesBrianD: There is a small capacity detail missing here: preloaded adapters are pinned, and the manager reserves one slot by allowing at most `max_loras_per_batch - 1` pinned adapters. With the default value of 8, only 7 adapters can currently be preloaded. Could we clarify that?

docs/features/lora.md
```diff
@@ -0,0 +1,122 @@
...
+| `--max-lora-rank` | `None` | Maximum LoRA rank; auto-inferred from the loaded adapters when unset. |
+| `--lora-target-modules` | `None` | Modules to apply LoRA to (`q_proj`, `k_proj`, `v_proj`, `o_proj`, `gate_proj`, `up_proj`, `down_proj`, or `all`); auto-inferred from the adapters when unset. |
+| `--max-loaded-loras` | `None` | Maximum adapters kept loaded in memory. |
+| `--lora-eviction-policy` | `lru` | Eviction policy (reserved; no runtime eviction is triggered in the current version). |
```
JamesBrianD: These options are not functional yet: `max_loaded_loras` is only validated, and there is no eviction path using `lora_eviction_policy`. Listing them as server capabilities is likely to confuse users. Since the issue discussion also suggested leaving unfinished LRU work out, I would remove these rows for now.

Mr-Neutr0n:
Addressed all six points from your review (21ae363b):

- **53 JSON dict** — now notes the dict form is for programmatic `ServerArgs` construction; CLI `--lora-paths` only accepts `name=path` / `path`.
- **72 chat completions** — removed the claim; now states chat completions does not yet forward `lora_path` and would be ignored there, keeping only the Python engine API.
- **80 capacity** — added the pinned-adapter detail to `--max-loras-per-batch`: at most `max_loras_per_batch - 1` (7 with the default 8) can be preloaded.
- **84 non-functional rows** — removed `max_loaded_loras` and `lora_eviction_policy` (no runtime eviction yet).
- **94 Static LoRA merged weights** — reworded to fixed `A`/`B` buffers updated externally between RL steps, BGMV not used, weights not merged.
- **113 Static example** — added the required `--max-lora-rank`, `--lora-target-modules`, and `--max-loras-per-batch 1`.

Thanks for the thorough review — let me know if the wording needs tweaking.

JamesBrianD (approved):

PR #1523, opened 2026-08-17
feat(pd): overlap chunked prefill KV transfer with Raiden

### LINKED ISSUE #1496: [Bug] PD chunked prefill can recompute committed prompt tokens
## Problem

Native P/D disaggregation can recompute already committed prompt tokens when a request crosses a chunked-prefill boundary.

After #1471, the PD prefill result path decrements `req.is_chunked` when an intermediate chunk is drained. The next `Req.init_next_round_input()` call also uses `is_chunked == 0` to decide that the request is not a continuation and clears `prefix_indices`. These are two different concepts—an in-flight chunk counter and persistent continuation ownership—but they currently share one field.

This affects both the JAX and Raiden transfer engines. Output correctness was preserved in the observed runs, but redundant prefill work increases forward count, queueing, and TTFT under load.

## Reproduction and evidence

Configuration:

- Native PD prefill/decode
- `--disable-radix-cache`
- `--chunked-prefill-size 8192`
- Page size 128
- Random `4096 input / 1 output`, concurrency 8

The scheduler accounts for approximately 4224 tokens per fresh request, so two requests exceed the 8192-token chunk budget and the second request becomes a continuation.

For the same 294-request workload on two TPU v6e hosts:

| Version | Prefill forwards | Dominant batch shape | Throughput |
| --- | ---: | --- | ---: |
| Before #1471 | 173 | 3 sequences / 8192 tokens | ~19.9 req/s |
| Current main with #1471 | 294 | 2 sequences / 8192 tokens | ~10.8 req/s |
| Current main with #1471 reverted | 173 | 3 sequences / 8192 tokens | ~19.8 req/s |

Per-forward time stayed around 88–89 ms and transfer time did not regress. The additional forward invocations and queueing account for the throughput loss.

## Follow-up scope

Incremental chunked-prefill KV transfer is out of scope for #1490, which introduces request-granularity Raiden transfer after prefill completion. Address this issue when adding incremental chunk transfer, while keeping the fix backend-neutral.

The implementation should:

- Separate persistent chunk-continuation ownership from the in-flight chunk counter.
- Preserve committed `prefix_indices` across continuation rounds.
- Avoid recomputing or retransferring committed prompt tokens.
- Preserve #1471's abort/terminal drain guarantees and page lifetime safety.
- Cover JAX and Raiden backends with a continuation unit test and concurrent PD end-to-end test.

cjx0709:
## Motivation

PD disaggregation currently transfers a request's KV cache only after chunked prefill has completed, leaving the full transfer on the TTFT critical path. The continuation state can also lose the committed prefix between chunks and recompute prompt tokens, as reported in #1496.

This PR overlaps each completed prefill chunk's KV transfer with the next chunk's forward pass while preserving strict ownership and failure semantics.

Fixes #1496.

## Modifications

- Add an opt-in `--disaggregation-enable-chunk-prefill-transfer` path for Raiden and ChunkCache.
- Publish DP-scoped per-chunk descriptors through bootstrap protocol v5 and receive each chunk into its exact destination page range.
- Allocate Decode KV once, validate complete non-overlapping coverage, and enqueue Decode only after every child transfer succeeds.
- Bound concurrent child transfers and retain KV ownership until Raiden reaches an engine-terminal state.
- Handle cancellation, retraction, retry isolation, and partial-transfer failure without exposing partial KV or leaking request/KV slots.
- Preserve committed-prefix state across chunked-prefill continuation to avoid recomputation.
- Keep Raiden metadata discovery off the Decode scheduler thread, with bounded retry backoff and no Decode KV allocation before metadata is ready.

The request-level Raiden path remains the default. Mixed protocol-v4/v5 P/D deployments are rejected rather than silently downgraded.

## Accuracy Tests

- Deterministic baseline/chunk-transfer outputs matched across the v6e-1 long-prefill matrix and all Prefill/Decode rank pairs at v6e-4 DP=1/2/4.
- A two-host v7e-8 lifecycle run at `4067bbaf2` covered cancellation, retraction, post-failure sentinel requests, and final Prefill/Decode request-pool usage of zero.
- Review-fix head `303724310`: 217 passed, 2 skipped across the existing PD ownership, Prefill, Decode, and Raiden suites.
- New regression coverage verifies terminal read failures retire every chunk owner before resource release, admission metadata lookup never blocks the scheduler, empty/error lookups back off, pending metadata allocates no Decode KV, and paused P/D transfers drain to terminal.
- No CI workflow or test-matrix entry was added.

## Benchmarking and Profiling

On two v6e-1 hosts with Qwen3-0.6B, Raiden, DP1/TP1, and 216/216 successful measured requests, TTFT changed as follows:

- 2K chunks: -42.95% at 16K input, -32.69% at 32K, and -19.14% at 64K.
- 4K chunks: -34.99% at 16K input, -25.23% at 32K, and -15.13% at 64K.
- Absolute TTFT savings increased from 0.52-0.55s at 16K to 1.71-1.79s at 64K; first-to-second-token ITL was unchanged within noise.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/disaggregation/bootstrap.py
+++ b/python/sgl_jax/srt/disaggregation/bootstrap.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import copy
 import logging
 import threading
 import time
@@ -22,12 +23,14 @@ HEARTBEAT_TTL_SECONDS = 30.0
 # Beat at ~TTL/3 so a single missed beat doesn't evict the entry.
 HEARTBEAT_INTERVAL_SECONDS = HEARTBEAT_TTL_SECONDS / 3.0
 TRANSFER_METADATA_TTL_SECONDS = 300.0
-BOOTSTRAP_CAPABILITIES = ("transfer_metadata",)
+BOOTSTRAP_CAPABILITIES = ("transfer_metadata", "chunk_transfer_metadata")
 
 # PD wire protocol version. Bump when ``PrefillInfo``
 # or any of the 4 endpoint payloads change shape.
-PROTOCOL_VERSION: int = 4
-MIN_COMPATIBLE_VERSION: int = 1
+PROTOCOL_VERSION: int = 5
+# v5 changes request transfer metadata from a single descriptor into a
+# per-chunk envelope. Mixed v4/v5 P/D peers are therefore unsupported.
+MIN_COMPATIBLE_VERSION: int = PROTOCOL_VERSION
 
 
 def _set_registry_size(n: int) -> None:
@@ -173,8 +176,14 @@ class RegisterPrefillRequest(BaseModel):
 class RegisterTransferRequest(BaseModel):
     bootstrap_room: int
     transfer_id: str
+    base_transfer_id: str | None = None
     jax_process_index: int = 0
     prefill_dp_rank: int = 0
+    chunk_index: int = 0
+    # Request-level transfer is represented as one final chunk. Chunked
+    # prefill publishes zero until the final descriptor fixes the total.
+    num_chunks: int = 1
+    chunk_page_offset: int = 0
     transport_metadata: dict[str, object]
 
 
@@ -270,8 +279,61 @@ class _Registry:
             prefill_dp_rank = int(info.get("prefill_dp_rank", 0))
             if not str(info.get("transfer_id", "")):
                 raise ValueError("transfer_id must be non-empty")
+            transfer_id = str(info["transfer_id"])
+            base_transfer_id = str(info.get("base_transfer_id") or transfer_id)
+            chunk_index = int(info.get("chunk_index", 0))
+            num_chunks = int(info.get("num_chunks", 1))
+            chunk_page_offset = int(info.get("chunk_page_offset", 0))
+            if chunk_index < 0:
+                raise ValueError("chunk_index must be non-negative")
+            if chunk_page_offset < 0:
+                raise ValueError("chunk_page_offset must be non-negative")
+            if num_chunks < 0:
+                raise ValueError("num_chunks must be non-negative")
+            if num_chunks and num_chunks != chunk_index + 1:
+                raise ValueError("the final chunk must publish num_chunks == chunk_index + 1")
+            expected_transfer_id = (
+                base_transfer_id
+                if base_transfer_id == transfer_id
+                else f"{base_transfer_id}#c{chunk_index}"
+            )
+            if transfer_id != expected_transfer_id:
+                raise ValueError(
+                    "chunk transfer_id must match base_transfer_id and chunk_index: "
+                    f"expected={expected_transfer_id!r}, got={transfer_id!r}"
+                )
             key = (room, process_index, prefill_dp_rank)
-            self.transfers[key] = info
+            bundle = self.transfers.get(key)
+            if bundle is not None and bundle.get("base_transfer_id") != base_transfer_id:
+                if chunk_index != 0:
+                    raise ValueError(
+                        "a new transfer may replace room metadata only from chunk zero"
+                    )
+                bundle = None
+            if bundle is None:
+                bundle = {
+                    "base_transfer_id": base_transfer_id,
+                    "chunks": {},
+                    "num_chunks": 0,
+                }
+                self.transfers[key] = bundle
+            chunks = bundle["chunks"]
+            assert isinstance(chunks, dict)
+            existing = chunks.get(chunk_index)
+            if existing is not None and existing != info:
+                raise ValueError(f"conflicting transfer metadata for chunk_index={chunk_index}")
+            known_total = int(bundle.get("num_chunks", 0))
+            if known_total and chunk_index >= known_total:
+                raise ValueError(
+                    f"chunk_index={chunk_index} is outside finalized num_chunks={known_total}"
+                )
+            if num_chunks and known_total not in (0, num_chunks):
+                raise ValueError(
+                    f"conflicting num_chunks: existing={known_total}, new={num_chunks}"
+                )
+            chunks[chunk_index] = dict(info)
+            if num_chunks:
+                bundle["num_chunks"] = num_chunks
             self.transfer_last_seen[key] = self.now()
 
     def get_transfer(
@@ -285,7 +347,7 @@ class _Registry:
             info = self.transfers.get(
                 (int(bootstrap_room), int(jax_process_index), int(prefill_dp_rank))
             )
-            return dict(info) if info is not None else None
+            return copy.deepcopy(info) if info is not None else None
 
     def pop_room(
         self,
@@ -673,15 +735,23 @@ class BootstrapClient:
         bootstrap_room: int,
         transfer_id: str,
         *,
+        base_transfer_id: str | None = None,
         jax_process_index: int = 0,
         prefill_dp_rank: int = 0,
+        chunk_index: int = 0,
+        num_chunks: int = 1,
+        chunk_page_offset: int = 0,
         transport_metadata: dict[str, object],
     ) -> None:
         payload = {
             "bootstrap_room": bootstrap_room,
             "transfer_id": transfer_id,
+            "base_transfer_id": base_transfer_id or transfer_id,
             "jax_process_index": jax_process_index,
             "prefill_dp_rank": prefill_dp_rank,
+            "chunk_index": chunk_index,
+            "num_chunks": num_chunks,
+            "chunk_page_offset": chunk_page_offset,
             "transport_metadata": transport_metadata,
         }
         r = self._client.post(
--- a/python/sgl_jax/srt/disaggregation/bootstrap.py
+++ b/python/sgl_jax/srt/disaggregation/bootstrap.py
@@ -354,11 +354,20 @@ class _Registry:
         bootstrap_room: int,
         jax_process_index: int = 0,
         prefill_dp_rank: int = 0,
-    ) -> None:
+        expected_transfer_id: str | None = None,
+    ) -> bool:
         with self.lock:
             key = (int(bootstrap_room), int(jax_process_index), int(prefill_dp_rank))
+            bundle = self.transfers.get(key)
+            if (
+                expected_transfer_id is not None
+                and bundle is not None
+                and bundle.get("base_transfer_id") != expected_transfer_id
+            ):
+                return False
             self.transfers.pop(key, None)
             self.transfer_last_seen.pop(key, None)
+            return bundle is not None
 
 
 def build_app(
@@ -473,8 +482,14 @@ def build_app(
         bootstrap_room: int,
         jax_process_index: int = 0,
         prefill_dp_rank: int = 0,
+        expected_transfer_id: str | None = None,
     ) -> dict[str, str]:
-        registry.pop_room(bootstrap_room, jax_process_index, prefill_dp_rank)
+        registry.pop_room(
+            bootstrap_room,
+            jax_process_index,
+            prefill_dp_rank,
+            expected_transfer_id,
+        )
         return {"status": "popped"}
 
     # Bootstrap runs as a standalone single process and does NOT inherit
@@ -790,14 +805,18 @@ class BootstrapClient:
         *,
         jax_process_index: int = 0,
         prefill_dp_rank: int = 0,
+        expected_transfer_id: str | None = None,
     ) -> None:
+        params: dict[str, object] = {
+            "bootstrap_room": bootstrap_room,
+            "jax_process_index": jax_process_index,
+            "prefill_dp_rank": prefill_dp_rank,
+        }
+        if expected_transfer_id is not None:
+            params["expected_transfer_id"] = expected_transfer_id
         r = self._client.post(
             f"{self._base_url}/pop_transfer",
-            params={
-                "bootstrap_room": bootstrap_room,
-                "jax_process_index": jax_process_index,
-                "prefill_dp_rank": prefill_dp_rank,
-            },
+            params=params,
             timeout=self._timeout_s,
             headers=self._headers(),
         )
--- a/python/sgl_jax/srt/disaggregation/bootstrap.py
+++ b/python/sgl_jax/srt/disaggregation/bootstrap.py
@@ -16,6 +16,11 @@ from fastapi import FastAPI, HTTPException, Request
 from pydantic import BaseModel
 from pydantic import Field as PydanticField
 
+from sgl_jax.srt.disaggregation.base.transfer import (
+    chunk_transfer_id,
+    parse_chunk_transfer_id,
+)
+
 logger = logging.getLogger(__name__)
 
 
@@ -61,6 +66,7 @@ class PrefillInfo:
     protocol_version: int = PROTOCOL_VERSION
     page_size: int = 0
     kv_dtype: str = ""
+    chunk_prefill_transfer: bool = False
     transport_metadata: dict[str, object] = field(default_factory=lambda: {"engine": "jax"})
 
     def to_dict(self) -> dict[str, object]:
@@ -111,6 +117,7 @@ def check_prefill_compat(
     expected_transfer_engine: str | None = None,
     expected_dp_rank: int | None = None,
     expected_dp_size: int | None = None,
+    expected_chunk_prefill_transfer: bool | None = None,
 ) -> None:
     """Raise ``ValueError`` if the prefill peer's KV layout is incompatible.
 
@@ -155,6 +162,14 @@ def check_prefill_compat(
             f"PD topology mismatch: prefill dp_size={peer_dp_size}, "
             f"decode dp_size={expected_dp_size}"
         )
+    if expected_chunk_prefill_transfer is not None:
+        peer_chunk_prefill_transfer = bool(info.get("chunk_prefill_transfer", False))
+        if peer_chunk_prefill_transfer != expected_chunk_prefill_transfer:
+            raise ValueError(
+                "PD chunk prefill transfer mismatch: "
+                f"prefill={peer_chunk_prefill_transfer}, "
+                f"decode={expected_chunk_prefill_transfer}"
+            )
 
 
 class RegisterPrefillRequest(BaseModel):
@@ -170,6 +185,7 @@ class RegisterPrefillRequest(BaseModel):
     protocol_version: int = PROTOCOL_VERSION
     page_size: int = 0
     kv_dtype: str = ""
+    chunk_prefill_transfer: bool = False
     transport_metadata: dict[str, object] = PydanticField(default_factory=lambda: {"engine": "jax"})
 
 
@@ -184,6 +200,7 @@ class RegisterTransferRequest(BaseModel):
     # prefill publishes zero until the final descriptor fixes the total.
     num_chunks: int = 1
     chunk_page_offset: int = 0
+    expected_total_pages: int = 0
     transport_metadata: dict[str, object]
 
 
@@ -284,19 +301,28 @@ class _Registry:
             chunk_index = int(info.get("chunk_index", 0))
             num_chunks = int(info.get("num_chunks", 1))
             chunk_page_offset = int(info.get("chunk_page_offset", 0))
+            expected_total_pages = int(info.get("expected_total_pages", 0))
             if chunk_index < 0:
                 raise ValueError("chunk_index must be non-negative")
             if chunk_page_offset < 0:
                 raise ValueError("chunk_page_offset must be non-negative")
             if num_chunks < 0:
                 raise ValueError("num_chunks must be non-negative")
+            if expected_total_pages < 0:
+                raise ValueError("expected_total_pages must be non-negative")
             if num_chunks and num_chunks != chunk_index + 1:
                 raise ValueError("the final chunk must publish num_chunks == chunk_index + 1")
-            expected_transfer_id = (
-                base_transfer_id
-                if base_transfer_id == transfer_id
-                else f"{base_transfer_id}#c{chunk_index}"
-            )
+            if base_transfer_id == transfer_id:
+                expected_transfer_id = base_transfer_id
+            else:
+                parsed_base, parsed_index = parse_chunk_transfer_id(transfer_id)
+                if parsed_base != base_transfer_id or parsed_index != chunk_index:
+                    raise ValueError(
+                        "chunk transfer_id must match base_transfer_id and chunk_index: "
+                        f"base={base_transfer_id!r}, chunk_index={chunk_index}, "
+                        f"got={transfer_id!r}"
+                    )
+                expected_transfer_id = chunk_transfer_id(base_transfer_id, chunk_index)
             if transfer_id != expected_transfer_id:
                 raise ValueError(
                     "chunk transfer_id must match base_transfer_id and chunk_index: "
@@ -311,10 +337,13 @@ class _Registry:
                     )
                 bundle = None
             if bundle is None:
+                if base_transfer_id != transfer_id and chunk_index != 0:
+                    raise ValueError("a chunk transfer bundle must be created from chunk zero")
                 bundle = {
                     "base_transfer_id": base_transfer_id,
                     "chunks": {},
                     "num_chunks": 0,
+                    "expected_total_pages": expected_total_pages,
                 }
                 self.transfers[key] = bundle
             chunks = bundle["chunks"]
@@ -331,9 +360,17 @@ class _Registry:
                 raise ValueError(
                     f"conflicting num_chunks: existing={known_total}, new={num_chunks}"
                 )
+            known_total_pages = int(bundle.get("expected_total_pages", 0))
+            if expected_total_pages and known_total_pages not in (0, expected_total_pages):
+                raise ValueError(
+                    "conflicting expected_total_pages: "
+                    f"existing={known_total_pages}, new={expected_total_pages}"
+                )
             chunks[chunk_index] = dict(info)
             if num_chunks:
                 bundle["num_chunks"] = num_chunks
+            if expected_total_pages:
+                bundle["expected_total_pages"] = expected_total_pages
             self.transfer_last_seen[key] = self.now()
 
     def get_transfer(
@@ -662,6 +699,7 @@ class BootstrapClient:
         protocol_version: int = PROTOCOL_VERSION,
         page_size: int = 0,
         kv_dtype: str = "",
+        chunk_prefill_transfer: bool = False,
         transport_metadata: dict[str, object] | None = None,
     ) -> None:
         payload = {
@@ -677,6 +715,7 @@ class BootstrapClient:
             "protocol_version": protocol_version,
             "page_size": page_size,
             "kv_dtype": kv_dtype,
+            "chunk_prefill_transfer": bool(chunk_prefill_transfer),
             "transport_metadata": transport_metadata or {"engine": "jax"},
         }
         last_err: Exception | None = None
@@ -756,6 +795,7 @@ class BootstrapClient:
         chunk_index: int = 0,
         num_chunks: int = 1,
         chunk_page_offset: int = 0,
+        expected_total_pages: int = 0,
         transport_metadata: dict[str, object],
     ) -> None:
         payload = {
@@ -767,6 +807,7 @@ class BootstrapClient:
             "chunk_index": chunk_index,
             "num_chunks": num_chunks,
             "chunk_page_offset": chunk_page_offset,
+            "expected_total_pages": expected_total_pages,
             "transport_metadata": transport_metadata,
         }
         r = self._client.post(
--- a/python/sgl_jax/srt/disaggregation/decode.py
+++ b/python/sgl_jax/srt/disaggregation/decode.py
@@ -86,6 +86,7 @@ class DecodeBookkeeping:
     # receiver setup can be deferred to the capacity-gated admission step.
     p_info: dict | None = None
     cancelled: bool = False
+    retracted: bool = False
     created_at: float = field(default_factory=time.monotonic)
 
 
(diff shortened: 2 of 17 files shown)


python/sgl_jax/srt/disaggregation/prefill.py
```diff
@@ -376,17 +426,136 @@ def _pd_mark_time(self: Scheduler, req: Req, name: str) -> None:
...
+            self._on_prefill_transfer_terminal(req_obj, sender_obj)
+
+        self.disagg_prefill_queue.add(
+            req.rid,
+            sender,
+            on_terminal=_on_terminal,
+            req=req,
+        )
+
+    def _retire_failed_chunk_producer(self: Scheduler, req: Req) -> None:
```
zorrofox: **🔴 use-after-release / KV cross-contamination**

`_retire_failed_chunk_producer` clears `chunked_reqs` but leaves `last_batch.reqs_info[dp].chunked_req` pointing at the failed request. On the next tick `_sync_chunked_req_owners` re-installs it (only the consumed-abort path via `_retire_chunked_req_batch_owners` ever clears `info.chunked_req`), and `get_new_batch_to_run` feeds it back into `adder.add_chunked_req`.

Meanwhile `_on_prefill_transfer_terminal` has already run `_release_prefill_req_resources(req)`, freeing the `req_pool_idx` and KV pages. The resurrected request's subsequent forward rounds write `req_to_token` through a slot that can be re-allocated to another request — silent KV corruption for that other request.

Trigger: any mid-chunk transfer failure that leaves the req unfinished without `FINISH_ABORT` — the `has_pending_failure` early-return, or `sender.fail` on the `chunk_cursor` / `chunk_handoff` paths.

Suggested fix: have `_retire_failed_chunk_producer` also clear `last_batch.reqs_info[dp].chunked_req` (mirroring `_retire_chunked_req_batch_owners`), so the owner-sync can't resurrect a request whose resources are gone.

python/sgl_jax/srt/disaggregation/decode.py
```diff
@@ -398,6 +420,13 @@ def process_decode_queue(self: Scheduler) -> None:
                 if entry.kv_indices is not None:
                     self._release_decode_kv_indices(entry.kv_indices, entry.req.dp_rank)
                 continue
+            if entry.retracted:
+                if entry.kv_indices is not None:
+                    self._release_decode_kv_indices(entry.kv_indices, entry.req.dp_rank)
+                advance_disagg_transfer_attempt(entry.req)
```
zorrofox: **🔴 retract desynchronizes P/D transport ids**

`disagg_transfer_attempt` is a process-local counter and never crosses the wire, so the `base#rN` transport id can diverge between the two sides after a one-sided retract:

- **D-side retract after the P sender already reached SUCCESS and popped the room:** the requeued D request looks up metadata under `base#r1`, which no producer will ever publish (the bootstrap lookup is read-only and nothing triggers a re-prefill) → admission defers until the `metadata_timeout` abort.
- **P-side retract while the D request is still in `DecodePreallocQueue`** (its attempt is still 0 — `retract_all` only marks `DecodeTransferQueue` entries): P republishes children as `base#r1#cK`, but `try_start_decode` computes `expected_first_id = base#c0` → `ValueError` → permanent `receiver_init` abort.

Before this PR, retract never touched PD transfers; with this change every asymmetric retract becomes a deterministic user-visible failure. The attempt number probably needs to be part of the handshake (carried in the published metadata and echoed back), or retract needs to reset/renegotiate the transfer id on both sides.

python/sgl_jax/srt/disaggregation/prefill.py
```diff
@@ -518,8 +687,23 @@ def _on_prefill_transfer_terminal(
...
+            # release allocations belonging to a re-admitted request.
+            logger.warning("Ignoring stale prefill sender callback for req_id=%s", req.rid)
+            sender.clear()
+            return
+
+        retract_pending = bool(getattr(req, "disagg_retract_pending", False))
         try:
-            if sender.poll() == KVPoll.SUCCESS:
+            state = sender.poll()
+            if retract_pending:
```
zorrofox: **🔴 AbortReq lost after retract; dead client's request resurrected**

`_on_prefill_transfer_terminal` checks `disagg_retract_pending` before the abort state, and neither `drain_terminal` nor the callback consults `entry.cancelled`. Sequence:

1. `retract_all` sets `disagg_retract_pending` and aborts the sender.
2. While the sender drains, the client disconnects; `abort_request` runs `cancel_matching`, which only sets `entry.cancelled = True` — a flag read solely by `retract_all`'s filter.
3. At terminal, the `if retract_pending: pass` arm wins; the entry (with its `cancelled` flag) is discarded and the req is `reset_for_retract()` + requeued.

`AbortReq` is processed once and never re-examines the waiting queue, so after resume the request re-runs full prefill + transfer and streams a normal SUCCESS to a dead client. Suggest checking `entry.cancelled` in the terminal callback (abort wins over retract), or having the requeue path drop entries whose abort flag is set.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
```diff
@@ -684,13 +951,241 @@ def poll(self) -> KVPoll:
...
+            logger.exception("Raiden chunk layout validation failed for %s", self._req_id)
+            self.request_abort("chunk_layout")
+            return self._finish(KVPoll.FAILED, "chunk_layout")
+        return self._finish(KVPoll.SUCCESS, "raiden_chunks_done_receiving")
+
+    def _discover_and_start_chunks(self, metadata: RaidenChunkedMetadata) -> None:
+        with self._state_lock:
+            if self.state != KVPoll.TRANSFERRING or self._pending_failure_reason is not None:
+                return
+            complete = self._known_num_chunks is not None and self._started_chunks == set(
```
zorrofox: **🔴 blocking HTTP GET in the decode scheduler loop, every tick**

The early-exit guard is

```python
complete = self._known_num_chunks is not None and self._started_chunks == set(range(...))
```

i.e. all-chunks-**started**, not all-metadata-**known**. With `CHUNK_TRANSFER_WINDOW = 2`, chunks beyond the window haven't started, so the synchronous `httpx` GET (`get_transfer_info`, 5s timeout) re-fires on every sub-millisecond event-loop tick for the entire prefill duration of each request — even when every chunk record is already local.

N concurrent transferring requests = N sequential blocking cross-node GETs per tick, executed in the decode scheduler thread. One slow bootstrap response stalls the whole decode loop (ITL spike for all running requests) for up to 5s.

Suggested fix: skip the fetch when all chunk records are already known, and rate-limit discovery (e.g. only re-poll after an interval or when a chunk completes). Longer term this belongs off the scheduler thread.

python/sgl_jax/srt/disaggregation/prefill.py
```diff
@@ -277,12 +306,23 @@ def process_prefill_chunk(self: Scheduler, batch, result) -> None:
...
-            and req.rid not in self.disagg_prefill_queue._entries
-        ]
+        chunk_transfer_enabled = bool(
+            getattr(
+                self.server_args,
+                "disaggregation_enable_chunk_prefill_transfer",
+                False,
+            )
+        )
+        ready_to_transfer = (
```
zorrofox: **🔴 full-pool `block_until_ready` now runs per chunk round**

The chunk-transfer path makes `ready_to_transfer` non-empty on every prefill round, so `prepare_prefill_batch` — which does `jax.block_until_ready` over the entire KV pool — runs once per chunk round instead of once per request. A 64K prompt with 2K chunks means 32 full device joins per request, including rounds where `_pump_pending_chunks` will register nothing because the window is full (the chunk just parks in `_pending_chunks`).

This serializes host and device exactly during long prefills — the case this PR is optimizing. Suggest gating the `block_until_ready` on a read actually being registered this round (i.e. only when the window has a free slot and a chunk will really be handed to the transfer engine).

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
```diff
@@ -684,13 +951,241 @@ def poll(self) -> KVPoll:
...
+                    chunk_index,
+                )
+                self.request_abort("raiden_start_read")
+                return
+            with self._state_lock:
+                self._started_chunks.add(chunk_index)
+                # Bound time since the latest started pull, not total prompt
+                # duration. The reaper still catches a producer that stops
+                # publishing after this chunk.
+                self._transfer_started_at = time.monotonic()
```
zorrofox, line 1310: **receiver pull-timeout counts producer compute time**

The decode receiver keeps `_transfer_started_at` armed between chunk completions, and the reaper doesn't check whether a pull is actually in flight. So if chunk k completes quickly and the prefill side then takes longer than `disaggregation_pull_timeout_seconds` (default 30s) to compute and publish chunk k+1 — an XLA recompile for a new bucket shape, or the parked chunked req queued behind large batches under load — the reaper fails a perfectly healthy transfer on both sides.

Note the sender explicitly nulls its timer for exactly this window; the receiver should do the equivalent — e.g. a separate producer-liveness budget re-armed on metadata discovery, rather than reusing the pull timeout.

python/sgl_jax/srt/disaggregation/factory.py
```diff
@@ -121,9 +155,20 @@ def _create_raiden_backend(
...
+        chunk_transfer_enabled
+        and getattr(scheduler, "tree_cache", None) is not None
+        and _tree_cache_supports_swa(scheduler.tree_cache)
+    ):
+        raise ValueError("Raiden chunk prefill transfer does not support sliding-window KV")
     if bootstrap_client is None or not hasattr(bootstrap_client, "require_capability"):
         raise RuntimeError("Raiden requires a bootstrap client with capability probing")
     bootstrap_client.require_capability("transfer_metadata")
+    if chunk_transfer_enabled:
+        bootstrap_client.require_capability("chunk_transfer_metadata")
```
zorrofox, line 171: **one-sided flag rollout passes startup, then fails 100% of requests**

`require_capability('chunk_transfer_metadata')` only checks the bootstrap **server's** advertised capability list, which is the constant `BOOTSTRAP_CAPABILITIES` regardless of the flag; `PrefillInfo` / `check_prefill_compat` exchange no chunk-transfer flag, and `protocol_version` stays 5 either way.

So a deployment with `--disaggregation-enable-chunk-prefill-transfer` enabled on only one side (a realistic partial rollout) starts up clean, then every request allocates decode KV, hits `Raiden transfer ID mismatch: expected=<base>, got=<base>#c0` in `try_start_decode`, and aborts as `receiver_init` — alloc/release churn plus an error that never names the actual config mismatch.

Suggest carrying the flag in `PrefillInfo` and rejecting at `check_prefill_compat`, the same way `page_size` / `kv_dtype` mismatches are handled.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
```diff
@@ -302,37 +352,68 @@ def try_start_decode(self, context: DecodeTransferContext) -> DecodeAdmission:
...
-                    jax_process_index=peer_process_index,
-                    prefill_dp_rank=context.prefill_dp_rank,
-                    decode_dp_rank=context.decode_dp_rank,
-                    direct_commit=context.direct_commit,
-                    expected_debug=(
-                        metadata.get("kv_debug")
-                        if isinstance(metadata.get("kv_debug"), Mapping)
-                        else None
-                    ),
+            if self.enable_chunk_prefill_transfer:
```
zorrofox, line 467: **chunk mode loses v4 admission validation and kv-debug**

Two regressions vs the v4 path:

1. The admission-time total-block-count validation is gone; under-coverage is only detectable in `_validate_complete_chunk_layout` after **all** data has landed. A prompt-length disagreement (or a buggy peer) that v4 rejected in O(1) at admission now holds decode KV pages, a Raiden slot wave, and per-rank inflight budget for the whole multi-chunk pull before failing with `chunk_layout`; a producer that never publishes the final descriptor is caught only by the reaper. The prefill side knows the total prompt pages at chunk 0 — publishing an expected-total field would restore O(1) rejection.
2. kv-debug verification is silently disabled in chunk mode (`publish` passes `debug_metadata=None`; commit passes `expected_debug=None`) — the correctness tooling goes dark exactly in the new mode that most needs it.

python/sgl_jax/srt/disaggregation/bootstrap.py
```diff
@@ -270,8 +279,61 @@ def register_transfer(self, info: dict[str, object]) -> None:
...
             key = (room, process_index, prefill_dp_rank)
-            self.transfers[key] = info
+            bundle = self.transfers.get(key)
+            if bundle is not None and bundle.get("base_transfer_id") != base_transfer_id:
+                if chunk_index != 0:
+                    raise ValueError(
+                        "a new transfer may replace room metadata only from chunk zero"
+                    )
+                bundle = None
+            if bundle is None:
```
zorrofox, line 339: **chunk N>0 can recreate an orphan bundle after a mid-stream pop**

`register_transfer` recreates a fresh bundle when the room key is absent even for `chunk_index > 0` — the "only from chunk zero" guard fires only against an *existing* bundle with a different base. If the decode receiver fails on chunk 1 and `_finish(FAILED)` pops the room, the still-publishing sender's chunk 2 creates a bundle `{2}` with no chunk 0.

Impact is bounded (a retry's chunk 0 under a new base may replace it, and the sender's ack timeout eventually pops the orphan), but until then the prefill engine keeps computing, registering engine reads, and publishing chunks for a dead transfer. Rejecting a chunk-N>0 registration when no bundle exists would fail fast at publish time.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
```diff
@@ -508,6 +615,116 @@ def attach_block_ids(
...
+                        self._req_id,
+                        pending.chunk_index,
+                    )
+                    self.request_abort("chunk_register")
+                    if raise_on_error:
+                        raise
+                    return
+
+    def _start_pending_chunk(self, pending: _PendingSenderChunk) -> None:
+        child_id = f"{self.uuid}#c{pending.chunk_index}"
```
zorrofox: **nit: extract a helper for the `#cN` wire format**

The child-transfer-id format `f"{base}#c{index}"` is hand-rolled at ~10 sites across `conn.py` and `bootstrap.py`, layered under the equally stringly `#rN` attempt suffix. Since this string *is* the v5 wire protocol (sender registers under it, the registry validates it, the receiver matches on it), any future change requires lockstep edits at all sites, and missing one shows up as a runtime `ValueError` three processes away from the bug. A single `chunk_transfer_id(base, index)` (plus a parse counterpart) next to `slots_to_page_ids` would remove the class of bug.

zorrofox (changes requested):
Did a detailed pass focused on failure/retract/abort lifecycle interactions and on whether the transfer/compute overlap holds up under concurrency. Summary (details in the inline comments):

**Correctness (lifecycle interactions):**

1. **Failed chunked request gets resurrected** — `_retire_failed_chunk_producer` clears `chunked_reqs` but not `last_batch.reqs_info[dp].chunked_req`, so `_sync_chunked_req_owners` re-installs the request on the next tick while the sender's terminal callback has already released its `req_pool_idx` and KV pages. Subsequent forward rounds write `req_to_token` through a slot that may have been re-allocated to another request — silent KV cross-contamination.
2. **Retract desynchronizes P/D transport ids** — `disagg_transfer_attempt` is process-local with no wire propagation, so a one-sided retract (`pause_generation(mode='retract')`) leaves the two sides expecting different `base#rN` ids and turns recoverable in-flight requests into guaranteed aborts (`metadata_timeout` or `receiver_init`).
3. **AbortReq lost after retract** — the terminal callback checks `disagg_retract_pending` before the abort state and never consults `entry.cancelled`, so an abort that lands during sender drain is dropped and the dead client's request is requeued and re-run end to end.

**Performance (both would erode the TTFT win this PR is after, and neither shows up in the DP1/TP1 single-concurrency benchmark):**

4. The chunked receiver issues a synchronous blocking `httpx` GET inside the decode scheduler event loop on every poll tick, because the early-exit guard tests all-chunks-*started* rather than all-metadata-*known*. One slow bootstrap response stalls the entire decode loop for up to 5s (ITL spike for every running request).
5. The chunk path makes `ready_to_transfer` non-empty on every prefill round, so the full-KV-pool `jax.block_until_ready` in `prepare_prefill_batch` now runs once per chunk round (32× for a 64K prompt at 2K chunks) instead of once per request — including rounds where the window is full and no read gets registered. That's exactly the overlap this PR exists to create.

**Robustness / operability:** receiver pull-timeout counts producer compute time and can reap healthy transfers (6); a one-sided `--disaggregation-enable-chunk-prefill-transfer` rollout passes startup checks and then fails 100% of requests with a cryptic per-request error (7); chunk mode drops the v4 admission-time total-block validation and silently disables kv-debug verification (8); a mid-stream registry pop lets the sender keep feeding an orphan bundle with no fail-fast (9); and the `#cN` child-id wire format is hand-rolled at ~10 sites (10).

**Suggestions for the test plan:**

- Lifecycle tests covering (a) retract interleaved with abort, and (b) one-sided retract (P-only and D-only) with transfers in flight — the three most severe findings all live in these interactions.
- A concurrent benchmark (e.g. 16–64 concurrency, mixed prompt lengths) alongside the single-request one; items 4 and 5 are invisible at concurrency 1.

python/sgl_jax/srt/disaggregation/prefill.py
cjx0709: Fixed in f0c9569c4. The failure path now clears the batch-side chunked_req owner before terminal resource release, and the regression test verifies owner sync cannot resurrect the request.

python/sgl_jax/srt/disaggregation/decode.py
cjx0709: Addressed in be9805596. Active P/D transfers now drain across retract with the original wire ID; the process-local retry suffix and one-sided requeue path were removed, with retract coverage added.

python/sgl_jax/srt/disaggregation/prefill.py
cjx0709: Addressed in be9805596. Abort now sets req.to_finish on the queued prefill request, and the terminal callback prioritizes that abort state before success or retract handling. The abort-during-drain path is covered by regression tests.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
cjx0709: Addressed in be9805596. Chunk metadata discovery is asynchronous and rate-limited, and polling stops once all metadata is known. Tests verify receiver polling never blocks and no redundant lookup occurs afterward.

python/sgl_jax/srt/disaggregation/prefill.py
cjx0709: Addressed in be9805596. KV readiness is now attached to the child registration on_ready callback, so the device join happens only when a chunk is admitted into a free sender window, not on every prefill round.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
cjx0709: Addressed in be9805596. Producer-wait time now has a separate watchdog; pull_timeout is armed only while an actual Raiden read is active. A compute-gap regression test was added.

python/sgl_jax/srt/disaggregation/factory.py
cjx0709: Addressed in be9805596. The chunk-transfer flag is now part of PrefillInfo compatibility checks, so one-sided enablement is rejected at startup with a clear mismatch error.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
cjx0709: Addressed in be9805596. Chunk zero now carries expected_total_pages for admission-time validation, and final-chunk KV debug metadata is preserved through commit. Both paths have regression tests.

python/sgl_jax/srt/disaggregation/bootstrap.py
cjx0709: Addressed in be9805596. The registry now rejects chunk N>0 when no parent bundle exists, preventing an orphan bundle from being recreated after pop. A focused registry test was added.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
cjx0709: Addressed in be9805596. Child IDs now use centralized chunk_transfer_id and parse_chunk_transfer_id helpers, with round-trip and validation tests.

python/sgl_jax/srt/disaggregation/prefill.py
zorrofox: Verified the fix on the three synchronous failure paths (`has_pending_failure` early-return, `chunk_cursor`, both `chunk_handoff` branches all route through `_retire_failed_chunk_producer`, which now clears the batch-side owner), and the regression test covers the owner-sync resurrection.

There's a fourth path with the same shape that isn't covered though: a mid-stream chunk **read** failure typically surfaces at sender drain, not at handoff — `poll()` returns FAILED as soon as all started children are terminal + pending failure (with one active read that's the same poll that first observes it), so the next-handoff `has_pending_failure` check never fires. The sequence is then: drain → `_on_prefill_transfer_terminal` → `_release_prefill_req_resources` frees `req_pool_idx`/KV and clears `req.disagg_chunk_sender`, while `self.chunked_reqs[dp]` still owns the req. Next tick `_prepare_chunked_reqs_to_exclude` calls `cache_unfinished_req` on the released req, `add_chunked_req` reschedules it (no finished check there), the forward writes KV into freed pages, and `_raiden_handoff_chunk` even creates a fresh sender since `disagg_chunk_sender` is None. The same gap is reachable when a parked producer's sender fails during pause (the prefill loop drains senders while paused).

Suggested fix: retire scheduler ownership in `_on_prefill_transfer_terminal` itself when `self.chunked_reqs[dp] is req` (i.e. make the terminal callback the single choke point), so no failure path can leave a released req owned.

python/sgl_jax/srt/disaggregation/decode.py
zorrofox: Verified — draining in-flight transfers across retract with the original wire ID removes the desync class entirely (checked `get_disagg_transport_id`, both queues' retract paths, and that `DecodePreallocQueue` entries hold no KV while deferred). Simpler than propagating the attempt number; agreed this is the better shape.

Two notes: (1) the decode-side retract tests were deleted with this change and nothing replaced them — a lifecycle test asserting "retract with transfers in flight → both sides drain and complete after resume" would lock the new semantics in. (2) A pause longer than the sender ack-timeout / the 300s producer watchdog now fails the parked chunk stream on both sides. Bounded, but worth a comment in the code — and on the P side that failure currently lands in the residual path described in the `prefill.py:552` thread.

python/sgl_jax/srt/disaggregation/prefill.py
zorrofox: Verified — `abort_request` sets `to_finish` on queued/chunked entries, and the terminal callback checks it ahead of the success path; no requeue path remains that could resurrect an aborted req. Test asserts the drain-then-abort sequence. Resolving.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
zorrofox: The receiver-side path is fixed — lookup on the executor, scheduler thread only checks `future.done()`, early-exit is now all-metadata-known, and the tests cover both properties.

Two things remain:

1. Admission-time discovery is still a synchronous `get_transfer_info` on the scheduler thread: `_admit_decode_prealloc` → `try_start_decode` calls it per deferred entry per tick, with no rate limit. In chunk mode an entry stays deferred until chunk 0 publishes, so this window is long and the call rate is tick-bound. Pre-existing for non-chunk mode, but chunk mode makes it hot; same treatment (executor + backoff) would close it.
2. The new poller's backoff only grows when a lookup returns data — exceptions/None stay at the 10ms floor, so an unreachable bootstrap gets ~100 GETs/s per receiver against the shared 8-thread pool. Backoff should apply to failures too.

python/sgl_jax/srt/disaggregation/prefill.py
zorrofox: Verified — chunk mode is excluded from `ready_to_transfer`, and the pool join now fires in the per-chunk `on_ready` only when the chunk is admitted into a free sender window; the test's callback-count assertion pins the over-window deferral. Resolving. (Minor, no action needed: the join runs under `_pump_lock`, so a concurrent `abort()` waits behind a device join — latency only.)

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
zorrofox: Verified — pull timeout is armed only while a read is actually active, producer-wait has its own watchdog re-armed on each chunk start, symmetric on both sides, and the compute-gap tests cover the original scenario. Resolving.

One note: `_CHUNK_PRODUCER_TIMEOUT_SECONDS = 300.0` is a hardcoded module constant, and the producer watchdog is silently disabled when `pull_timeout_s <= 0` (it lives inside that branch). Consider promoting it to a server arg next to the pull timeout — it now also bounds how long a pause/retract can hold a parked chunk stream (see the decode.py thread).

python/sgl_jax/srt/disaggregation/factory.py
zorrofox: Verified — flag carried in `PrefillInfo`/registration, enforced in `check_prefill_compat` and at decode intake with an explicit "PD chunk prefill transfer mismatch" abort, both directions parametrized in the test. Rejection lands at first request intake on D rather than process startup (earliest point D can observe P's registration), which is fine — the error now names the config. Resolving.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
zorrofox: Verified — `expected_total_pages` is required on every chunk, conflict-checked in the registry, validated against the decode allocation at admission, and re-checked in the final layout validation; kv_debug metadata rides the final chunk and is recovered at commit. Tests cover all three. Resolving.

python/sgl_jax/srt/disaggregation/bootstrap.py
zorrofox: Verified — non-zero chunk registration without a parent bundle is rejected (and stale-base replacement is chunk-zero-only), with the sender propagating a `chunk_register` abort. Resolving. (Nit, take or leave: the second `transfer_id != expected_transfer_id` raise in `register_transfer` is unreachable after parse validation — `chunk_transfer_id(base, index)` reconstructs it by construction.)

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
zorrofox: Verified — `#c` now appears only inside `chunk_transfer_id`/`parse_chunk_transfer_id`, registry validation included, round-trip tested. Resolving.

python/sgl_jax/srt/disaggregation/prefill.py
cjx0709: Addressed in 303724310. The terminal callback now retires current chunk scheduler, batch, and pending-abort ownership before releasing the request slot or KV pages. A drain-time read-failure regression verifies owner sync and the next scheduling pass cannot resurrect the request.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
cjx0709: Addressed in 303724310. Decode admission now polls bootstrap metadata asynchronously with per-transfer backoff before allocating Decode KV, and both admission and receiver polling back off on empty results or exceptions. The new tests cover non-blocking lookup, retry throttling, and zero KV allocation while metadata is pending.

python/sgl_jax/srt/disaggregation/prefill.py
zorrofox: Verified at head — `_retire_chunk_producer_ownership` is now the single choke point in the terminal callback: all three owners (`chunked_reqs[dp]`, the batch-side `info.chunked_req`, and the pending-abort entry) are cleared with identity checks before `sender.clear()` / `_release_prefill_req_resources`, uniformly for FAILED, SUCCESS, and abort terminals, and stale-sender callbacks are rejected up front. The regression test asserts the ordering at release time (owners already None inside the release call), not just the end state, and the pause-drain path goes through the same choke point with its own end-to-end test. Resolving.

One nit, take or leave: the two synchronous failure branches in `_raiden_handoff_chunk` still call `_abort_prefill_req` before the retire. Same-tick and idempotent, so harmless today, but the "retire before release" invariant is only structurally enforced at the terminal callback — flipping the order in those two branches would make it hold everywhere.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
zorrofox: Verified at head — both remaining points are closed. `get_transfer_info` now has exactly two call sites, both submitted to the executor; `try_start_decode` only reads the cached lookup under the lock; admission polls metadata before `allocator.alloc`, so nothing holds KV (or blocks other entries — same skip semantics as the old deferred path) while metadata is pending; and the doubling backoff applies to exceptions and empty results on both the admission and receiver pollers, with the parametrized tests pinning all of it (never-blocks, no duplicate in-flight GET, zero-alloc-while-pending, backoff-on-failure). Resolving.

FWIW the 50ms backoff ceiling means a down bootstrap still sees ~1 GET/50ms per pending transfer, though the shared 8-thread pool plus the 5s client timeout self-throttles that to a trickle in practice — fine as is, just noting in case you ever want a larger ceiling for the outage case.

zorrofox (approved):
All ten review threads are closed out — verified each fix against the head code rather than the commit messages, including the two follow-ups from the last round (terminal-callback ownership retirement with ordering asserted in the regression test, and the fully async admission/receiver metadata path with failure backoff and zero KV held while pending). The retract-as-drain semantics change also checks out end to end on both sides. The lifecycle test coverage added across these rounds (drain-time read failure, abort-during-drain, retract with in-flight transfers, paused drain on both P and D) is exactly the set this feature needed. Approving.

PR #1524, opened 2026-08-17
fix(scheduler): reserve SWA cache headroom

Yanko-7:
## Motivation

Hybrid sliding-window attention models use separate full-attention and SWA KV-cache pools. The current admission budget accounts for the sliding window but not for allocations accumulated between periodic SWA evictions, the two page-alignment margins, or the remaining decode headroom of requests already running on each DP rank.

As a result, the scheduler can admit more requests than the SWA pool can sustain. Decode then exhausts the SWA pool and retracts an otherwise healthy request even though the full-attention pool budget is valid.

## Modifications

- Add shared helpers for the page-aligned SWA eviction interval and peak live-token occupancy.
- Use the same eviction interval calculation in both decode eviction and admission budgeting.
- Reserve `sliding_window + eviction_interval + 2 * page_size` peak occupancy for newly admitted requests.
- Subtract the remaining SWA decode headroom of already-running requests independently for each DP rank.
- Add focused tests for interval/page-margin accounting and per-DP running-request reservations.

## Accuracy Tests

This change does not modify model weights, kernels, or model-forward numerics.

A model-level A/B reproduction was run with on a TPU v6e-4. It used a two-layer dummy Gemma2 hybrid model (one sliding-attention layer and one full-attention layer), a 64-token sliding window, 16-token pages, and two concurrent 8-token prompts generating 240 tokens each.

| Revision | HTTP status | KV-cache retraction events | Full/SWA pool tokens |
| --- | ---: | ---: | ---: |
| `main` (`32c7a4bb4`) | 200 | 1 | 272 / 224 |
| This change (`64945a53e`) | 200 | 0 | 272 / 224 |

The SWA-only commit removes the avoidable retraction without changing either cache-pool size.

## Benchmarking and Profiling

Not applicable. This is an admission-budget correctness fix; no kernels or model-forward operations are changed.

## Test Plan

```bash
python -m pytest -q \
  python/sgl_jax/test/test_swa_schedule_budget.py \
  python/sgl_jax/test/test_mixed_chunk_dp.py \
  python/sgl_jax/test/mem_cache/test_swa_allocator.py \
  python/sgl_jax/test/mem_cache/test_swa_radix_cache.py
```

## Checklist

- [x] Documentation is not required for this internal scheduler correction.

### DIFF
--- a/python/sgl_jax/srt/managers/schedule_batch.py
+++ b/python/sgl_jax/srt/managers/schedule_batch.py
@@ -801,6 +801,20 @@ def _build_recurrent_track_entries(
     )
 
 
+def swa_eviction_interval(sliding_window_size: int, page_size: int) -> int:
+    """Return the page-aligned decode interval between SWA evictions."""
+    multiplier = float(os.environ.get("SGL_JAX_SWA_EVICTION_INTERVAL_MULTIPLIER", "1.0"))
+    interval = max(page_size, int(sliding_window_size * multiplier))
+    return (interval // page_size) * page_size
+
+
+def swa_eviction_peak_tokens(sliding_window_size: int, page_size: int) -> int:
+    """Return the maximum live SWA tokens held between decode evictions."""
+    return (
+        sliding_window_size + swa_eviction_interval(sliding_window_size, page_size) + 2 * page_size
+    )
+
+
 @dataclasses.dataclass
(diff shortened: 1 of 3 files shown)


cjx0709 (approved):
LGTM

PR #1527, opened 2026-08-17
Add Tenstorrent backend to sglang-jax using libtt

pcmoritz:
## Motivation

I'm the author of https://github.com/pcmoritz/libtt which is a fully open source XLA implementation for https://tenstorrent.com/ devices (it is similar to libtpu for TPUs and also statically links all the dependencies). It is built using a heavily patched version of the tenstorrent software stack (https://github.com/tenstorrent/) to make it as compatible with TPUs/Jax as possible. Most of the Jax unit tests already pass with libtt as a backend.

This PR makes it so sglang-jax can run on the libtt backend out of the box. So far I have only tested the qwen3 family of models, but they run with pretty good speed (e.g. qwen3 8b runs at 30 tok/s decoding speed on a single tenstorrent blackhole P150).

## Modifications

This PR introduces an attention backend which uses StableHLO custom_call to lower the attention computations to the appropriate kernels. I tried to keep modifications to the rest of the code as minimal as possible. Any recommendations on how to improve the code is very welcome!

## Accuracy Tests

So far I have only run simple queries but they do show coherent output. Before finalizing the PR, I'm planning to do more numerical accuracy tests.

## Benchmarking and Profiling

Instructions on how to run this code:

```
# Build libtt and its JAX plugin wheel
git clone https://github.com/pcmoritz/libtt.git
cd libtt
bazel build //:jax_tt_plugin_wheel
export LIBTT_WHEEL="$PWD/bazel-bin/jax_tt_plugin-0.1.0-py3-none-linux_x86_64.whl"
cd ..

# Install SGLang-JAX and the plugin.
git clone --branch agent/tt-backend-clean-minimal https://github.com/pcmoritz/sglang-jax.git
cd sglang-jax
uv venv --python 3.12
uv pip install --python .venv/bin/python -e python jax==0.8.1 jaxlib==0.8.1 "$LIBTT_WHEEL"
```

Start Qwen3-8B:

```
  env -u TT_METAL_RUNTIME_ROOT \
    PYTHONPATH="$PWD/python" \
    JAX_PLATFORMS=tt \
    JAX_USE_SHARDY_PARTITIONER=false \
    JAX_COMPILATION_CACHE_DIR=/tmp/sglang-jax-tt-cache \
    .venv/bin/python -m sgl_jax.launch_server \
      --model-path Qwen/Qwen3-8B \
      --host 127.0.0.1 \
      --port 31000 \
      --device tt \
      --dtype bfloat16 \
      --attention-backend tt \
      --max-running-requests 2 \
      --max-total-tokens 1024 \
      --max-prefill-tokens 256 \
      --chunked-prefill-size 256 \
      --page-size 32 \
      --watchdog-timeout 1200 \
      --disable-precompile \
      --skip-server-warmup \
      --disable-overlap-schedule \
      --disable-radix-cache
```

Test it from another terminal:

```
  curl -sS http://127.0.0.1:31000/generate -H 'Content-Type: application/json' -d '{"text":"The capital of France is","sampling_params":{"temperature":0,"max_new_tokens":128}}'
```

## Checklist


cjx0709:
@pcmoritz  Hello, and thank you for your contribution. I am quite surprised to see that our code can actually run on Tenstorrent hardware. I would like to ask about your upcoming work and integration plans: do you intend to introduce just a few PRs or kernels as an experimental configuration, or do you plan to incorporate more kernels and more complex models later on?
If it is the former, the current approach is fine. However, regarding the latter—since we do not have Tenstorrent hardware and thus cannot perform validation or long-term maintenance—we might need to discuss a strategy for long-term collaborative maintenance.

pcmoritz:
@cjx0709 Thanks a lot for your comment! I'm planning to implement the Tenstorrent backend as close as possible to the TPU backend and handle as much as possible of the complexity in the compiler. I think that's possible because the TPU and Tenstorrent architectures are very similar: They both require/prefer Tensors to be tiled in RAM for performance, both process the data through vector and matrix units, and they both have similar interconnect design / philosophy. That said, it will be inevitable to have some custom kernels (e.g. for attention), and the low level programming between them for the kernels is pretty different. Tenstorrent also has a tile based Python DSL https://github.com/tenstorrent/tt-lang but I haven't studied it in much detail yet and it looks pretty different from pallas. I've mainly implemented kernels with their "tt-metal" C++ programming framework.

It will take some experimentation to get this right and minimize the code differences between the TPU and Tenstorrent backend, but I do think it is possible and if users are interested in using it, I would be interested to see how far we can get in supporting the most important models. To start off with, we should support the Qwen 3 and 3.5 architectures I think. As to hardware access, I can likely help to set up a CI and if more people are interested in contributing, we can try to figure out how to give them hardware access, e.g. https://www.koyeb.com/ has some of the older accelerators that are mostly compatible (that's what I started with), and Tenstorrent is starting to do more to give developers access, and the cards themselves are also pretty accessible.

cjx0709:
@pcmoritz wow, It looks good to add this backend. In this pr, we can simply only add the kernel but without the ci, later if more complex features or significant shared modifications to the scheduler arise later, we can introduce CI and test cases at that stage. 
Additionally, from a long-term maintenance perspective, I recommend adopting a file layout similar to that of sglang：
image
you can move the new tt kernels in new path like: `python/sgl_jax/srt/hardware_backend/tt/attention`

pcmoritz:
Thanks, that's a great suggestion! I'll give this a shot tomorrow (today is already very late for me). In the meantime if you have any other suggestions let me know. I'll also try to get rid of `jitted_greedy_sampler`, it was needed to get it fast, but maybe the existing code path can be made fast via the compiler.

python/sgl_jax/srt/layers/sampler.py
```diff
@@ -178,31 +178,26 @@ def __call__(
...
-        )
-
-        # Apply grammar-constrained vocab mask
-        logits = lax.cond(
-            sampling_metadata.apply_vocab_mask,
-            lambda operands: apply_token_bitmask(operands[0], operands[1]),
-            lambda operands: operands[0],
-            (logits, sampling_metadata.vocab_mask),
-        )
+        logits = logits_output.next_token_logits
```
pcmoritz, line 181: Note: These changes to the sampler are currently needed to make the code fast on the Tenstorrent backend. I feel like they are also good for other backends. I'm happy to split it into a separate PR if that is preferred.

pcmoritz:
@cjx0709 @JamesBrianD Thanks for your feedback so far, I have significantly improved the PR now:

1. Support chunking for prefill
2. Support batching
3. Use the suggested code structure for the new backend
4. Run MMLU and fix the numerical problems (see https://github.com/pcmoritz/libtt/pull/216)
5. Remove the custom greedy sample workaround and go through the normal code path

I wonder if you can have a pass in reviewing it and unblock the CI.

python/sgl_jax/srt/sampling/sampling_batch_info.py
```diff
@@ -68,17 +68,17 @@ def tree_flatten(self):
...
             self.linear_penalty,
-            self.do_penalties,
             self.vocab_mask,
-            self.apply_vocab_mask,
         )
 
         aux_data = {
             "return_logprob": self.return_logprob,
             "top_logprobs_nums": self.top_logprobs_nums,
+            "is_all_greedy": self.is_all_greedy,
```
cjx0709, line 81: Could we include the dynamic-argument PyTreeDef (or these static values) in AotDispatcher's cache key and add a CPU regression test? With SGLANG_JAX_AOT_DISPATCH=1, same-shaped greedy and regular/penalized batches currently share one shape/dtype key, so whichever branch compiles first is silently reused.

pcmoritz: Thanks for pointing this out, I fixed it now!

python/sgl_jax/srt/layers/sampler.py
```diff
@@ -178,31 +178,26 @@ def __call__(
...
+        if sampling_metadata.do_penalties:
+            logits = self._apply_linear_penalty((logits, sampling_metadata))
+        if sampling_metadata.apply_vocab_mask:
+            logits = apply_token_bitmask(logits, sampling_metadata.vocab_mask)
+
+        if sampling_metadata.is_all_greedy:
+            batch_next_token_ids, logprobs = self._greedy_sampling(
+                (logits, sampling_metadata, None)
+            )
```
cjx0709, line 199: Could we replicate the vocab axis before entering the static greedy branch? On v7x-8/TP=8, this repeatedly selected token 10819 even though its own top-logprobs had token 5443 as top-1; this change made plain/logprob/penalty all select 5443 under both stock and AOT dispatch, while retaining the warm throughput gain.
```suggestion
        if sampling_metadata.is_all_greedy:
            logits = jax.sharding.reshard(
                logits, NamedSharding(self.mesh, P("data", None))
            )
            batch_next_token_ids, logprobs = self._greedy_sampling(
                (logits, sampling_metadata, None)
            )
```

cjx0709: Validation setup:

- Hardware/model: v7x-8 (4 chips, 8 devices), BF16 Qwen3-8B, TP=8.
- Functional requests were repeated twice; returned IDs were checked against top-10 and queried-token logprobs.
- Performance used 32 prompts, concurrency 16, 128 input tokens and 128 output tokens; the table reports the fully warm pass.

Functional results:

| Case | Base | PR head | With suggestion |
|---|---:|---:|---:|
| Plain greedy | 5443 | **10819 ❌** | **5443 ✅** |
| Plain greedy + logprobs | 5443; top-1=5443 | **10819; top-1=5443 ❌** | **5443; top-1=5443 ✅** |
| Greedy + penalty | 5443 | 5443 | 5443 |
| Seeded regular | Reference output | Matched base | Unchanged path |
| Grammar JSON | Valid | Valid and matched base | Unchanged path |

The suggested version was also repeated with both stock dispatch and `SGLANG_JAX_AOT_DISPATCH=1`; plain/logprob/penalty all selected 5443 in both modes. For the mismatched pair, token 5443 had logprob -1.875 while the incorrectly selected token 10819 had logprob -21.125.

Performance results:

| Metric | Base | PR head | With suggestion |
|---|---:|---:|---:|
| Greedy output throughput | 1,375 tok/s | 1,546 tok/s | **1,583 tok/s** |
| Greedy mean TPOT | 11.27 ms | 10.00 ms | **9.74 ms** |
| Regular output throughput | 1,446 tok/s | **1,540 tok/s** | — (unchanged path) |
| Regular mean TPOT | 10.69 ms | **10.03 ms** | — (unchanged path) |

sii-xinglong:
@pcmoritz That sounds like interesting work! Would you mind sharing your email address? I have a few questions I’d like to ask

pcmoritz:
@sii-xinglong Thanks for letting me know, if you think the questions are relevant to sglang-jax and / or this PR and of interest to others too, let's discuss on this PR, otherwise feel free to shoot me an e-mail, my email-address is my github handle at gmail dot com.

python/sgl_jax/srt/layers/sampler.py
pcmoritz: Thanks for testing this, I updated it now and pushed the fix (also made sure the linter passes and it works on the Tenstorrent hardware).

cjx0709 (dismissed):

cjx0709:
@pcmoritz  ci fail because the precompile shape miss, I will fix it and push a pr into this branch `pcmoritz:agent/tt-backend-clean-minimal`

pcmoritz:
Thanks, that sounds great to me, thanks for all your help! If it gets more complicated, we could also consider splitting the PR into two, one for the sampler change and one for the TT backend change.

cjx0709:
I opened [pcmoritz/sglang-jax#2](https://github.com/pcmoritz/sglang-jax/pull/2) as a stacked fix PR against this PR's head branch.

The fix keeps `greedy` and `return_logprob` static, moves penalty and grammar predicates inside the compiled graphs, and precompiles the four required `greedy/regular × logprob/no-logprob` variants.

| Validation | Result |
|---|---|
| CPU sampler, precompile, and AOT tests | 37 passed, 1 TPU-only skip, 2 subtests passed |
| v7x-8 16-way sampler matrix | 16/16 passed; 4 JIT cache entries |
| Qwen3-8B TP=8 service matrix with AOT dispatch | 16/16 passed; 0 first-request cache misses; 0 warm cache misses |

Sampler-only specialization trade-off, batch size 64 and vocabulary size 151,936:

| Strategy | Compiled graphs | Compile wall time | Persistent cache size |
|---|---:|---:|---:|
| Fully dynamic | 2 | 14.60 s | 2.26 MB |
| Hybrid (stacked PR) | 4 | 25.25 s | 3.17 MB |
| Fully static (current PR head) | 16 | 100.04 s | 11.75 MB |

The stacked PR contains only the production implementation and regression tests; the experimental mode switches and benchmark harnesses are not included.

pcmoritz:
@cjx0709 I think you are doing the right fix, thanks a lot for writing the code, but it is also more complex on the sampler side than what is currently in `main` now. I think I have an idea how to do the TT backend without needing the sampler change, let me try that in a new PR and I'll get back to you :)

pcmoritz:
@cjx0709 I have been able to remove the sampler changes by optimizing the compiler and made a new PR with just the backend in https://github.com/sgl-project/sglang-jax/pull/1579.

cjx0709:
> @cjx0709 I have been able to remove the sampler changes by optimizing the compiler and made a new PR with just the backend in #1579.

Thanks, maybe we can talk about this merge in #1579

PR #1532, opened 2026-08-18
fix(dsa): budget and report the indexer key cache in the KV pool sizing

### LINKED ISSUE #1533: [Bug] DSA indexer key cache is allocated but not budgeted — KV pool over-provisions by ~5%
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Environment info and a minimal reproduction are included below.
- [x] 4. This is a bug, not a question.
- [x] 5. English.

### Describe the bug

With `--attention-backend dsa_sparse`, `MLATokenToKVPool._create_buffers` allocates a **second** paged buffer for the DSA indexer keys — one slot per "full" layer — but `ModelRunnerKVCacheMixin._compute_cell_size` counts only the MLA latent buffer. The profiler therefore sizes the KV pool as though the indexer buffer did not exist, and the pool over-allocates on every `dsa_sparse` config.

The excess does not come out of the KV budget; it comes out of the `(1 - mem_fraction_static)` activation reserve, which is what actually binds at long context. The failure is silent: the server starts, reports a healthy pool, and has less headroom than requested.

The same blind spot affects reporting — `_buffer_bytes()` sizes only the latent buffer, so `mem_usage`, the `JAX MLA KV Cache allocated` startup log, and `get_kv_size_bytes()` (which feeds the scheduler's `kvcache` gauge) all under-report resident HBM by the indexer's share. That is what let the over-provisioning stay invisible.

Measured on GLM-5.2-FP8 (TPU v7x, tp16 ep16 dp2, page_size 128, bf16), from one server's own startup log:

```
available_kv_cache = 8.70 GiB   <- budget the profiler sizes against
KV buffer          = 8.75 GiB   (17.50 GB global / dp2)
indexer buffer     = 0.47 GiB   (0.94 GB global / dp2)  <- not counted
allocated          = 9.22 GiB   => 5.4% over budget
```

The indexer buffer is laid out by the same `get_kv_cache_shape`, so it costs the same per token with `kv_dim` replaced by the aligned indexer key dim: `align128(128) * 128 * 2 / 128 = 256 B/layer`, and `256 * 21` full layers = 5,376 B/token against the latent's 99,840 — matching the 0.94 GB the pool logs.

Introduced in #1478, which added the indexer key buffer. Still present on `main` at f1f5a513.

### Reproduction

No TPU needed — the broken invariant is "the budget equals what the pool allocates", and both sides run on CPU. Save as `repro_indexer_budget.py` in a checkout and run it with `uv pip install -e "python[cpu]"`:

<details>
<summary><code>repro_indexer_budget.py</code></summary>

```python
"""Minimal repro: the DSA indexer key cache is allocated but not budgeted.

Runs on CPU. Builds the pool the server would build for a `dsa_sparse` config
and compares it against the budget `_compute_cell_size` hands the profiler.

    PYTHONPATH=python python3 repro_indexer_budget.py
"""

import types

import jax
import jax.numpy as jnp

from sgl_jax.srt.mem_cache.memory_pool import MLATokenToKVPool
from sgl_jax.srt.model_executor.model_runner_kv_cache_mixin import ModelRunnerKVCacheMixin
from sgl_jax.srt.server_args import ServerArgs

# GLM-5.2 shapes, scaled to 8 layers so the repro fits on a CPU.
NUM_LAYERS, PAGE_SIZE, SIZE = 8, 128, 256
KV_LORA_RANK, QK_ROPE_HEAD_DIM, INDEX_HEAD_DIM = 512, 64, 128

indexer_types = ["shared"] * NUM_LAYERS
for i in (0, 1, 2, 6):  # "full" layers, each of which gets an indexer buffer
    indexer_types[i] = "full"
NUM_FULL = indexer_types.count("full")


class Runner(ModelRunnerKVCacheMixin):
    """Only the attributes `_compute_cell_size` reads."""

    kv_cache_dtype = jnp.bfloat16
    page_size = PAGE_SIZE
    use_mla_backend = True

    def __init__(self):
        self.server_args = ServerArgs(model_path="dummy", attention_backend="dsa_sparse")
        self.model_config = types.SimpleNamespace(
            hf_config=types.SimpleNamespace(),
            hf_text_config=types.SimpleNamespace(
                kv_lora_rank=KV_LORA_RANK,
                qk_rope_head_dim=QK_ROPE_HEAD_DIM,
                num_hidden_layers=NUM_LAYERS,
                index_head_dim=INDEX_HEAD_DIM,
                index_skip_topk_offset=3,
                indexer_types=indexer_types,
            ),
        )

    def _kv_pool_layer_count(self):
        return NUM_LAYERS


pool = MLATokenToKVPool(
    size=SIZE,
    page_size=PAGE_SIZE,
    dtype=jnp.bfloat16,
    kv_lora_rank=KV_LORA_RANK,
    qk_rope_head_dim=QK_ROPE_HEAD_DIM,
    layer_num=NUM_LAYERS,
    mesh=jax.make_mesh((1,), ("data",)),
    dp_size=1,
    indexer_key_dim=INDEX_HEAD_DIM,
    num_indexer_layers=NUM_FULL,
)

latent = sum(b.nbytes for b in pool.kv_buffer)
indexer = sum(b.nbytes for b in pool.indexer_key_buffer)
# The pool allocates one spare page per DP rank on top of `size` tokens.
budget = Runner()._compute_cell_size() * (SIZE + PAGE_SIZE * 1)

print(f"latent buffers    : {latent:>10,} B  ({NUM_LAYERS} layers)")
print(f"indexer buffers   : {indexer:>10,} B  ({NUM_FULL} full layers)")
print(f"resident total    : {latent + indexer:>10,} B")
print(f"budgeted          : {budget:>10,} B  <- _compute_cell_size")
print(f"reported          : {pool.get_kv_size_bytes():>10,} B  <- get_kv_size_bytes")
over = (latent + indexer) / budget - 1
print(f"\nallocated / budgeted = {over:+.1%}")
assert budget == latent + indexer, "BUG: budget does not cover what the pool allocates"
assert pool.get_kv_size_bytes() == latent + indexer, "BUG: pool under-reports resident bytes"
print("OK: budget and report both match the allocation")
```

</details>

On `main` (f1f5a513) it fails:

```
latent buffers    :  3,932,160 B  (8 layers)
indexer buffers   :    393,216 B  (4 full layers)
resident total    :  4,325,376 B
budgeted          :  3,932,160 B  <- _compute_cell_size
reported          :  3,932,160 B  <- get_kv_size_bytes

allocated / budgeted = +10.0%
AssertionError: BUG: budget does not cover what the pool allocates
```

The overshoot is the indexer's share, so it scales with the fraction of "full" layers: 10% for the 4-of-8 sketch above, 5.4% for GLM-5.2's 21-of-78 measured on hardware.

On a TPU host the end-to-end repro is any `dsa_sparse` launch (e.g. GLM-5.2-FP8): compare `available_kv_cache` in the `TPU Memory profiling:` line against the sum of the `MLA KV cache shape per layer` and `DSA indexer-key cache` lines.

### Environment

sglang-jax `main` (f1f5a513); TPU v7x, jax 0.10.2 (bug reproduces on CPU with `python[cpu]`); model GLM-5.2-FP8, `--attention-backend dsa_sparse`.

tsvikas:
## Motivation

Resolves #1533.

With `--attention-backend dsa_sparse`, `MLATokenToKVPool._create_buffers` allocates a **second** paged buffer for the DSA indexer keys — one slot per "full" layer — but `_compute_cell_size` counts only the MLA latent buffer. The profiler sizes the KV pool as though that buffer did not exist, so the pool over-allocates on every `dsa_sparse` config.

The excess does not come out of the KV budget; it comes out of the `(1 - mem_fraction_static)` activation reserve, which is what actually binds at long context. The failure is silent: the server starts, reports a healthy pool, and has less headroom than requested.

Measured on GLM-5.2-FP8 (TPU v7x, tp16 ep16 dp2, page_size 128, bf16), from one server's own startup log:

```
available_kv_cache = 8.70 GiB   <- budget the profiler sizes against
KV buffer          = 8.75 GiB   (17.50 GB global / dp2)
indexer buffer     = 0.47 GiB   (0.94 GB global / dp2)  <- not counted
allocated          = 9.22 GiB   => 5.4% over budget
```

The indexer buffer is laid out by the same `get_kv_cache_shape`, so it costs the same per token with `kv_dim` replaced by the aligned indexer key dim: `align128(128) * 128 * 2 / 128 = 256 B/layer`, and `256 * 21` full layers = 5,376 B/token against the latent's 99,840 — matching the 0.94 GB the pool logs.

Introduced in #1478, which added the indexer key buffer.

## Modifications

1. **Budget the indexer buffer** (`model_runner_kv_cache_mixin.py`). New helper `_dsa_indexer_cache_params()` returns `(indexer_key_dim, num_indexer_layers)`, or `(0, 0)` when no indexer cache is allocated. Both `_compute_cell_size` (the budget) and the `MLATokenToKVPool` construction (the allocation) now derive the geometry from that one helper, so they cannot drift apart again — the allocation site already computed these numbers and the sizing site simply never asked for them.

   The two terms of the sum use different layer counts on purpose: latent per KV-pool layer (which drops non-attention layers on hybrid-recurrent models), indexer per "full" layer over the whole model. Budget matching *allocation* is the invariant, not the two counts matching each other.

2. **Report the indexer buffer** (`memory_pool.py`). `_buffer_bytes` sized only the latent buffer, so `mem_usage`, the `JAX MLA KV Cache allocated` log line, and `get_kv_size_bytes` (which feeds the scheduler's `kvcache` gauge) all under-reported resident HBM by the indexer's share — an operator reading the startup log saw 8.75 GB where 9.22 GB was resident. The log line now breaks out latent and indexer, and only when an indexer buffer exists (plain MLA deployments get the original line back).

Non-DSA MLA (`fa`) is unaffected: it allocates no indexer buffer and is charged for none.

## Accuracy Tests

Not applicable — no kernel or model forward code is touched. The change affects only how many tokens the pool is sized for and what the pool reports.

## Benchmarking and Profiling

No throughput or latency impact: the same buffers are allocated, they are now counted. The user-visible effect is that `max_tokens` on a `dsa_sparse` config drops by the indexer's share (~5% on GLM-5.2), which is the over-provisioning being removed — the reserve it was taken from is returned.

## Test plan

`python/sgl_jax/test/test_model_runner_kv_cache_mixin.py`, already collected by the `unit-test-cpu` suite — no TPU needed:

```bash
uv pip install -e "python[cpu]"
python -m pytest python/sgl_jax/test/test_model_runner_kv_cache_mixin.py -q   # 11 passed
```

Seven new cases, in two layers:

- **Budget vs. layout** — the indexer is charged on `dsa_sparse` and not on `fa`; the bf16 packing pad is covered at `page_size` 1 and 128; and the latent width is pinned as `align(lora,128) + align(rope,128)`, *not* `align(lora + rope, 128)` (the kernel slices nope and rope as adjacent buffers; `lora=192/rope=64` separates the two formulas, the GLM/DeepSeek 512/64 shape does not). Expected byte counts are derived from `get_kv_cache_shape` — the function `_create_buffers` allocates with — so the tests follow the alignment rule if it changes instead of pinning a stale constant.
- **Budget vs. allocation** — two cases build a real `MLATokenToKVPool` on a single-device CPU mesh and compare the budget against `b.nbytes` of the buffers that actually got allocated, which is the invariant that broke. One pins resident bytes against both `get_kv_size_bytes()` and `cell_size * (size + page_size * dp_size)`; the other reproduces the production over-provisioning without hardware — sizing a DSA pool with the latent-only budget allocates past `_profile_available_bytes`, while the fixed budget stays within it.

Verified by mutation: reverting the budget fix fails 5 of the 7; reverting only the `get_kv_size_bytes` reporting fix fails the allocation test. Both reverts were run against this branch on jax 0.10.2 CPU.

### Note for reviewers (out of scope here)

`_create_buffers` allocates `(size + page_size * dp_size) // page_size` pages — one spare page per DP rank, per layer — which `_compute_cell_size` has never counted. It predates this change and is tiny at production sizes (~0.1%), so it is left alone; the allocation tests name it as explicit slack rather than absorb it into a loose assertion. Happy to fix it here or in a follow-up if you prefer.

## Checklist

- [x] The purpose of the PR, and the issue it resolves (#1533).

### DIFF
--- a/python/sgl_jax/srt/mem_cache/memory_pool.py
+++ b/python/sgl_jax/srt/mem_cache/memory_pool.py
@@ -1392,9 +1392,11 @@ class MLATokenToKVPool(KVCache):
 
     def _calculate_memory_usage(self):
         """Calculate memory usage for the 4D paged MLA cache."""
+        # `get_kv_size_bytes` is the authoritative total (it also feeds the
+        # scheduler gauge); the components below only split it for the log.
+        self.mem_usage = self.get_kv_size_bytes() / GB
         latent_bytes = self._buffer_bytes() * self.layer_num
         indexer_bytes = self._indexer_buffer_bytes() * self.num_indexer_layers
-        self.mem_usage = (latent_bytes + indexer_bytes) / GB
 
         breakdown = (
             f" (latent {latent_bytes / GB:.2f} GB + DSA indexer {indexer_bytes / GB:.2f} GB)"
--- a/python/sgl_jax/srt/mem_cache/memory_pool.py
+++ b/python/sgl_jax/srt/mem_cache/memory_pool.py
@@ -1392,30 +1392,47 @@ class MLATokenToKVPool(KVCache):
 
     def _calculate_memory_usage(self):
         """Calculate memory usage for the 4D paged MLA cache."""
-        total_bytes = self._buffer_bytes() * self.layer_num
-        self.mem_usage = total_bytes / GB
+        latent_bytes = self._buffer_bytes() * self.layer_num
+        indexer_bytes = self._indexer_buffer_bytes() * self.num_indexer_layers
+        self.mem_usage = (latent_bytes + indexer_bytes) / GB
 
         logger.info(
-            "JAX MLA KV Cache allocated. #tokens: %s, KV size: %.2f GB",
+            "JAX MLA KV Cache allocated. #tokens: %s, KV size: %.2f GB"
+            " (latent %.2f GB + DSA indexer %.2f GB)",
             self.size,
-            total_bytes / GB,
+            self.mem_usage,
+            latent_bytes / GB,
+            indexer_bytes / GB,
         )
 
-    def _buffer_bytes(self) -> int:
+    def _buffer_bytes(self, kv_dim: int | None = None) -> int:
+        """Bytes for ONE paged buffer of ``kv_dim`` feature width (default: latent)."""
         total_num_pages = (self.size + self.page_size * self.dp_size) // self.page_size
         from sgl_jax.srt.kernels.mla.v2.kernel import get_kv_cache_shape
 
         shape = get_kv_cache_shape(
             total_num_pages=total_num_pages,
             page_size=self.page_size,
-            kv_dim=self.kv_dim,
+            kv_dim=self.kv_dim if kv_dim is None else kv_dim,
             kv_dtype=self.dtype,
         )
         return shape[0] * shape[1] * shape[2] * shape[3] * jnp.dtype(self.dtype).itemsize
 
+    def _indexer_buffer_bytes(self) -> int:
+        """Bytes for ONE DSA indexer key buffer; 0 when none is allocated."""
+        if not (self.indexer_key_dim and self.num_indexer_layers):
+            return 0
+        return self._buffer_bytes(kv_dim=self.indexer_key_dim)
+
     def get_kv_size_bytes(self):
-        """Calculate KV cache size in bytes."""
-        return self._buffer_bytes() * self.layer_num
+        """Resident bytes for this pool, including the DSA indexer key buffers.
+
+        Counting only the latent buffers under-stated resident HBM by the
+        indexer's share (~5% on GLM-5.2) — the same blind spot that let the
+        pool over-provision without it showing up in the startup log.
+        """
+        latent = self._buffer_bytes() * self.layer_num
+        return latent + self._indexer_buffer_bytes() * self.num_indexer_layers
 
     def get_fused_kv_buffer(self, layer_id: int) -> jax.Array:
         """Return the 4D paged buffer; consumed directly by the MLA v2 kernel."""
--- a/python/sgl_jax/srt/mem_cache/memory_pool.py
+++ b/python/sgl_jax/srt/mem_cache/memory_pool.py
@@ -1396,13 +1396,16 @@ class MLATokenToKVPool(KVCache):
         indexer_bytes = self._indexer_buffer_bytes() * self.num_indexer_layers
         self.mem_usage = (latent_bytes + indexer_bytes) / GB
 
+        breakdown = (
+            f" (latent {latent_bytes / GB:.2f} GB + DSA indexer {indexer_bytes / GB:.2f} GB)"
+            if indexer_bytes
+            else ""
+        )
         logger.info(
-            "JAX MLA KV Cache allocated. #tokens: %s, KV size: %.2f GB"
-            " (latent %.2f GB + DSA indexer %.2f GB)",
+            "JAX MLA KV Cache allocated. #tokens: %s, KV size: %.2f GB%s",
             self.size,
             self.mem_usage,
-            latent_bytes / GB,
-            indexer_bytes / GB,
+            breakdown,
         )
 
     def _buffer_bytes(self, kv_dim: int | None = None) -> int:
@@ -1425,12 +1428,7 @@ class MLATokenToKVPool(KVCache):
         return self._buffer_bytes(kv_dim=self.indexer_key_dim)
 
     def get_kv_size_bytes(self):
-        """Resident bytes for this pool, including the DSA indexer key buffers.
-
-        Counting only the latent buffers under-stated resident HBM by the
-        indexer's share (~5% on GLM-5.2) — the same blind spot that let the
-        pool over-provision without it showing up in the startup log.
-        """
+        """Resident bytes for this pool, including the DSA indexer key buffers."""
         latent = self._buffer_bytes() * self.layer_num
         return latent + self._indexer_buffer_bytes() * self.num_indexer_layers
 
--- a/python/sgl_jax/srt/mem_cache/memory_pool.py
+++ b/python/sgl_jax/srt/mem_cache/memory_pool.py
@@ -1423,8 +1423,6 @@ class MLATokenToKVPool(KVCache):
 
     def _indexer_buffer_bytes(self) -> int:
         """Bytes for ONE DSA indexer key buffer; 0 when none is allocated."""
-        if not (self.indexer_key_dim and self.num_indexer_layers):
-            return 0
         return self._buffer_bytes(kv_dim=self.indexer_key_dim)
 
     def get_kv_size_bytes(self):
--- a/python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
+++ b/python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
@@ -291,15 +291,11 @@ def _build_non_hybrid_memory_pools(token_to_kv_pool) -> MemoryPools:
 class ModelRunnerKVCacheMixin:
 
     def _dsa_indexer_cache_params(self: ModelRunner) -> tuple[int, int]:
-        """``(indexer_key_dim, num_indexer_layers)`` for the DSA indexer key cache.
+        """``(indexer_key_dim, num_indexer_layers)``, or ``(0, 0)`` when no DSA
+        indexer key cache is allocated.
 
-        Returns ``(0, 0)`` when no indexer cache is allocated.
-
-        Single source of truth for the two places that must agree: the memory
-        *budget* (:meth:`_compute_cell_size`) and the actual *allocation*
-        (``MLATokenToKVPool._create_buffers``). They previously derived this
-        independently — in practice only the allocation side did — so the
-        profiler sized the KV pool as though the indexer buffer did not exist.
+        Shared by the KV pool's memory budget (:meth:`_compute_cell_size`) and its
+        allocation (``MLATokenToKVPool._create_buffers``), which must agree.
         """
         if self.server_args.attention_backend != "dsa_sparse":
             return 0, 0
@@ -335,13 +331,18 @@ class ModelRunnerKVCacheMixin:
             per_token = kv_dim * aligned_ps * dtype_size // self.page_size
             cell_size = per_token * num_layers
 
-            # DSA allocates a SECOND paged buffer, for indexer keys, one slot per
-            # "full" layer (`MLATokenToKVPool._create_buffers`). It is laid out by
-            # the same `get_kv_cache_shape`, so it costs the same per token with
-            # `kv_dim` replaced by the aligned indexer key dim. Omitting it made
-            # the profiler over-provision the pool, and the excess came out of the
-            # (1 - mem_fraction) activation reserve rather than out of KV — the
-            # resource that actually binds at long context.
+            # DSA allocates a second paged buffer for indexer keys, one slot per
+            # "full" layer. It uses the same `get_kv_cache_shape` layout, so it
+            # costs the same per token with `kv_dim` replaced by the aligned
+            # indexer key dim.
+            # Note the two terms use different layer counts, deliberately: they
+            # mirror two different allocations. The latent buffer exists per
+            # KV-pool layer (`_kv_pool_layer_count`, which drops non-attention
+            # layers on hybrid-recurrent models); the indexer buffer exists per
+            # "full" layer, counted over the whole model by the same
+            # `build_index_share_map` the pool is constructed from. Budget
+            # matching allocation is the invariant here, not the two counts
+            # matching each other.
             indexer_key_dim, num_indexer_layers = self._dsa_indexer_cache_params()
             if indexer_key_dim > 0 and num_indexer_layers > 0:
                 indexer_per_token = (
@@ -725,8 +726,6 @@ class ModelRunnerKVCacheMixin:
                     f"kv_lora_rank={kv_lora_rank}, qk_rope_head_dim={qk_rope_head_dim}."
                 )
 
-            # Same helper the cell-size budget uses, so the pool can never
-            # allocate a buffer the profiler did not account for.
             dsa_kwargs = {}
             indexer_key_dim, num_indexer_layers = self._dsa_indexer_cache_params()
             if indexer_key_dim > 0:
--- a/python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
+++ b/python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
@@ -634,6 +634,16 @@ class ModelRunnerKVCacheMixin:
         if self.linear_recurrent_config is not None:
             from sgl_jax.srt.mem_cache.memory_pool import HybridLinearKVPool
 
+            # The DSA indexer slot space is built by `build_index_share_map`
+            # over every layer; the hybrid pool indexes full-ATTENTION layers
+            # only. Nothing reconciles the two, and `HybridLinearKVPool` does
+            # not expose `get_indexer_key_buffer` at all, so such a config
+            # would allocate and budget indexer buffers and then die on the
+            # first full-indexer forward. Fail here with the reason instead.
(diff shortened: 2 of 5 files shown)


tsvikas:
@cjx0709 @zorrofox : this restores the budget↔allocation invariant for the DSA indexer key buffer added in #1478; details and a CPU repro in #1533. Would appreciate a look.

zorrofox (commented):
Checked the core invariant end-to-end — budget (`_compute_cell_size`), allocation (`MLATokenToKVPool._create_buffers`), and report (`get_kv_size_bytes`) all agree via the shared `_dsa_indexer_cache_params` helper, and the 128-alignment matches `align_to` on the kernel side. Tests pass locally (11/11 on CPU). No correctness regression found — comments below are structural/minor.

One thing that can't be an inline comment because the file isn't in this diff: `docs/architecture/04-model-executor.md` — the cell-size formula ("Absorbed MLA only caches the latent representation, so there is no ×2") and the pool table are now incomplete for `dsa_sparse`. The indexer term changes the per-token cost by ~5% for a full-indexer config, which matters for anyone sizing `--max-total-num-tokens` from the doc. Fine as a follow-up.

Nice job documenting the spare-page-per-DP-rank slack explicitly instead of hiding it.

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
```diff
@@ -309,7 +331,20 @@ def align128(x: int) -> int:
...
+            cell_size = per_token * num_layers
+
+            # DSA allocates a second paged buffer for indexer keys, one slot per
+            # "full" layer, in the same `get_kv_cache_shape` layout with `kv_dim`
+            # replaced by the aligned indexer key dim. Its layer count is per
+            # full layer over the whole model, not per KV-pool layer.
+            indexer_key_dim, num_indexer_layers = self._dsa_indexer_cache_params()
+            if indexer_key_dim > 0:
+                indexer_per_token = (
+                    align128(indexer_key_dim) * aligned_ps * dtype_size // self.page_size
```
zorrofox, line 343: This term (and the latent term above it) hand-rolls the pool's allocation geometry, while `_create_buffers` derives it from `get_kv_cache_shape` — together with the test helper `_allocated_bytes_per_token` and the `_create_buffers` log line, that's four independent copies of the same layout math that must agree byte-for-byte. If the kernel packing or alignment rule ever changes (e.g. an fp8 KV dtype), the allocation grows automatically but this budget doesn't, silently re-creating the over-allocation bug class this PR fixes.

A pool-owned `bytes_per_token()` helper built on `get_kv_cache_shape` would let budget and allocation share one formula. It would also close a small pre-existing wrinkle this line inherits from the latent term: the truncating `// self.page_size` under-budgets fractionally when `page_size` isn't a multiple of the packing factor (marginal — such page sizes have other problems — but the shared helper fixes it for free).

Fine as a follow-up if you'd rather keep this PR focused.

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
```diff
@@ -686,16 +721,10 @@ def _init_pools(self: ModelRunner, max_num_reqs: int, dp_size: int):
...
-                    hf_text_config.num_hidden_layers,
-                )
-                dsa_kwargs["indexer_key_dim"] = hf_text_config.index_head_dim
-                dsa_kwargs["num_indexer_layers"] = num_full
+            indexer_key_dim, num_indexer_layers = self._dsa_indexer_cache_params()
+            if indexer_key_dim > 0:
+                dsa_kwargs["indexer_key_dim"] = indexer_key_dim
+                dsa_kwargs["num_indexer_layers"] = num_indexer_layers
 
             self.token_to_kv_pool = self._maybe_wrap_hybrid_kv_pool(
```
zorrofox, line 739: Now that the indexer cache is budgeted/reported through the hybrid path too: `HybridLinearKVPool` doesn't delegate `get_indexer_key_buffer` to its inner `MLATokenToKVPool`, so a future hybrid-recurrent + `dsa_sparse` model with full `index_*` config would start up cleanly (allocation, budget, and report all correct) and then hit `AttributeError` on the first forward of a full-indexer layer (`dsa_sparse_backend.py`). Unreachable today only incidentally — current hybrid configs lack `index_head_dim` and fail earlier. Worth either delegating or asserting in `_maybe_wrap_hybrid_kv_pool`.

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -1392,30 +1392,45 @@ def get_indexer_key_buffer(self, slot_id: int) -> jax.Array:
 
     def _calculate_memory_usage(self):
         """Calculate memory usage for the 4D paged MLA cache."""
-        total_bytes = self._buffer_bytes() * self.layer_num
-        self.mem_usage = total_bytes / GB
-
+        latent_bytes = self._buffer_bytes() * self.layer_num
+        indexer_bytes = self._indexer_buffer_bytes() * self.num_indexer_layers
+        self.mem_usage = (latent_bytes + indexer_bytes) / GB
```
zorrofox: `_calculate_memory_usage` and `get_kv_size_bytes` now each compute the identical total independently. Deriving `mem_usage` from `get_kv_size_bytes()` (keeping the component helpers just for the log breakdown) would leave one authoritative sum — the test only pins `get_kv_size_bytes`, so if a future buffer lands in only one of the two copies, the scheduler's kvcache gauge drifts unnoticed.

python/sgl_jax/srt/mem_cache/memory_pool.py
```diff
@@ -1392,30 +1392,45 @@ def get_indexer_key_buffer(self, slot_id: int) -> jax.Array:
...
             page_size=self.page_size,
-            kv_dim=self.kv_dim,
+            kv_dim=self.kv_dim if kv_dim is None else kv_dim,
             kv_dtype=self.dtype,
         )
         return shape[0] * shape[1] * shape[2] * shape[3] * jnp.dtype(self.dtype).itemsize
 
+    def _indexer_buffer_bytes(self) -> int:
+        """Bytes for ONE DSA indexer key buffer; 0 when none is allocated."""
+        if not (self.indexer_key_dim and self.num_indexer_layers):
```
zorrofox: Nit: this guard is redundant — both call sites multiply the result by `num_indexer_layers` (so 0-layer pools already contribute 0), and `_buffer_bytes(kv_dim=0)` returns 0 anyway since `align_to(0, 128) == 0`. Dropping it removes an invariant readers otherwise have to reason about.

python/sgl_jax/test/test_model_runner_kv_cache_mixin.py
```diff
@@ -52,3 +52,202 @@ def test_recurrent_admission_cap_already_aligned_unchanged():
...
+    runner.get_available_device_memory = lambda: available
+    runner.mem_fraction_static = 1.0
+
+    budget = runner._profile_available_bytes(total_device_memory=0)
+    assert budget == available
+
+    mesh = _single_device_mesh()
+    page = runner.page_size
+
+    def pages_for(cell_size):
```
zorrofox: Nit: `pages_for(cell_size)` returns a page-aligned *token* count (it's passed as `size=` to the pool), not a page count. In a test whose subject is page/token/byte accounting, `tokens_for` or `pool_size_for` would be harder to misuse.

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
tsvikas: Agreed, let's do it as a follow up, per your suggestion.
If i understand correctly, `bytes_per_token()` has to be callable before the pool exists, so it wants to be a module-level function or a classmethod over (page_size, dtype, kv_dim), not an instance method, and then the four call sites collapse onto it.

Same for the truncating `// page_size` under-budget, since the shared helper gets it for free.

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
tsvikas: Good catch.
Went with the assert. Two reasons beyond the ambiguity: 
1. the missing method is a symptom, not the gap. Nothing stops a recurrent layer being marked `"full"`, in which case the pool allocates an indexer buffer no forward touches and, after this PR, budgets for it.
2. a pass-through would claim support for a combination that can't be exercised, since no config today has both `linear_recurrent_config` and `index_head_dim`.

`test_hybrid_pool_rejects_a_dsa_indexer_cache` now covers it.

python/sgl_jax/srt/mem_cache/memory_pool.py
tsvikas: Done. `_calculate_memory_usage` now sets `mem_usage` from `get_kv_size_bytes()` and keeps the component helpers only to split the log line, so there's one authoritative total.
The live-pool test also asserts `mem_usage` against measured resident bytes, not just `get_kv_size_bytes()`,
so a future buffer that lands in one path shows up as a test failure rather than a drifting gauge.

python/sgl_jax/srt/mem_cache/memory_pool.py
tsvikas: Dropped it

python/sgl_jax/test/test_model_runner_kv_cache_mixin.py
tsvikas: Renamed to `pool_size_for`, with a docstring saying it returns a page-aligned token count.

tsvikas:
Thanks for the careful read.
All five comments (+ docs) are addressed except the shared `bytes_per_token()` helper, which I've kept as a follow-up per your note.

For the docs, I fixed `04-model-executor.md §4.9.1` and also `07-kv-cache.md §7.2.4.3`

tsvikas:
The two red shards on [run 32758124832](https://github.com/sgl-project/sglang-jax/actions/runs/32758124832) look like runner eviction rather than a test failure. `unit-test-1-tpu (2)` and `unit-test-1-tpu (4)` both ended with the annotation *"The self-hosted runner lost communication with the server."* Neither job uploaded any logs — the log endpoint returns `BlobNotFound` for both — so there is no test output at all, not even a partial run.

Could someone with write access re-run the failed jobs? I don't have the permission, and prefer not to push an empty commit to re-trigger all tests.

zorrofox (approved):
Approving — follow-up commits address the review round: `mem_usage` now derives from `get_kv_size_bytes` (b354349a), the hybrid-recurrent × dsa_sparse gap is closed with a startup assert plus test (c5da88c3), and the architecture docs now document the indexer buffer in the sizing formula (3504f896); the geometry-dedup refactor is agreed as a follow-up. The core invariant (budget/allocation/report agreeing via `_dsa_indexer_cache_params`) was verified end-to-end in the previous review pass.

As the party that introduced this in #1478 and the deployment most exposed to it: this silently costs ~5% of the activation reserve on GLM-5.2 dsa_sparse configs, which is exactly what binds at long context — thanks for the catch. We'll re-derive our long-context memory margins once this lands.

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
tsvikas: Documented in a separate issue: https://github.com/sgl-project/sglang-jax/issues/1553

PR #1540, opened 2026-08-21
fix(fused-moe): pad odd local token extents

Aharrypotter:
## Motivation

The v1 fused-EP MoE kernel derives its launch extent from the global token count
(`local_num_tokens = num_tokens // ep_size`) and rejects the result when it is not a legal
tile. Any precompile bucket whose device-local share is odd therefore fails to compile. At
EP16 the `2032` bucket gives `2032 // 16 = 127`, and bf16 packing (`t_packing=2`) rejects it
deterministically on every rank, before the server ever reaches HTTP readiness:

```text
ValueError: Expected local_num_tokens=127 to be aligned to t_packing=2.
```

This is a launch-shape defect, not a numerical one: `num_tokens`, `ep_size` and `dtype` are
the only inputs, so it reproduces for any model on the fused-MoE v1 path whenever the
scheduler's bucket set and `ep_size` combine to an odd local extent. It was observed on a
v6e-16 host at `TP=16 / DP=4 / EP=16` while validating unrelated work.

## Modifications

- **New `srt/kernels/fused_moe/token_padding.py`** — `align_fused_moe_v1_local_tokens` rounds a
  logical local token count up to a legal v1 launch extent (the small decode tiles `2` and `4`
  are preserved; anything larger goes to `lcm(8, t_packing)`), and
  `align_fused_moe_v1_num_tokens` lifts that to the global-equivalent extent used for tuning and
  launch sizing. The module deliberately has **no JAX import**, so the launch-shape contract can
  be validated on a host with no accelerator.
- **`srt/kernels/fused_moe/v1/kernel.py`** — separate the logical extent from the launch extent.
  `logical_num_tokens` keeps owning the caller-visible contract (`topk_weights` / `topk_ids`
  shape checks); the new `launch_num_tokens` feeds block-config selection,
  `validate_fused_moe_block_config`, and the Mosaic DMA tiling check.
  `_validate_fused_ep_moe_args` gains a `launch_num_tokens` argument and asserts it covers the
  logical extent and stays `ep_size`-aligned.
- **Padding is device-local and symmetric.** Inside `shard_map`, `tokens`, `topk_weights` and
  `topk_ids` are padded to the legal local extent and `local_output` is sliced back on exit.
  Padded rows carry `topk_ids = -1` and `topk_weights = 0.0`, so they route to no expert and
  contribute nothing. The scheduler-visible global shape and the logical bucket are unchanged.
- **Already-legal shapes are untouched.** Both pad and slice are guarded by `pad_local > 0`, and
  the alignment helper is idempotent, so a configuration that compiles today emits no additional
  op and keeps its existing tuned block config.

## Unit Tests

Host CPU, no accelerator required (JAX 0.10.2):

- `python/sgl_jax/test/kernels/test_fused_moe_token_padding.py` — **32 / 32 PASS** (new file).
  Covers the small-tile cases, the `lcm(8, t_packing)` rounding, idempotence on already-legal
  extents, rejection of non-positive and `ep_size`-misaligned contracts, the exact `127 -> 128`
  regression, and that `2032` and `2048` stay distinct logical buckets while both launching at
  `2048`. Includes `test_import_does_not_load_jax`, which asserts the module stays JAX-free.
- `python/sgl_jax/test/test_compilation_manager.py` — **27 / 27 PASS**.

```bash
pytest python/sgl_jax/test/kernels/test_fused_moe_token_padding.py \
       python/sgl_jax/test/test_compilation_manager.py
```

Requires a TPU, registered in `unit-test-tpu-v6e-4` and **not run on this host**:

- `python/sgl_jax/test/kernels/fused_moe_v1_test.py::MoEKernelTest::test_per_rank_padding_boundary`
  — new, parameterized over `local_num_tokens = [127, 128]` and both
  `use_jax_allreduce_metadata` settings; the previously failing extent and its adjacent legal
  neighbour must produce the same result as the reference implementation. The shared `_test_moe`
  helper also gained `assertEqual(actual.shape, a.shape)`, so every existing kernel case now
  asserts padding is invisible to the caller.

```bash
python test/srt/run_suite.py --suite unit-test-tpu-v6e-4
```

Black (`--config python/pyproject.toml`), Ruff and isort pass.

## Accuracy Tests

Not applicable, and deliberately so. This change never alters a tensor that reaches the
reference computation: padded rows are routed nowhere (`topk_ids = -1`, weight `0.0`), the
output is sliced back to the logical extent, and configurations that already compiled take a
byte-identical path. The correctness claim is carried by
`test_per_rank_padding_boundary`, which compares the previously rejected `127` extent against
the same reference `ref_moe` used by every other kernel case.

## Benchmarking and Profiling

Not run. The only configurations whose behaviour changes are ones that previously failed to
compile, so there is no working baseline to regress against and no speedup is claimed. Padding
`127 -> 128` costs at most one extra token row per device; already-legal buckets emit no
additional op.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/kernels/fused_moe/token_padding.py
+++ b/python/sgl_jax/srt/kernels/fused_moe/token_padding.py
@@ -1,0 +1,53 @@
+"""Pure token-extent helpers for fused-MoE wrappers.
+
+This module intentionally has no JAX dependency so the launch-shape contract can
+be validated on a host that does not have, import, or initialize JAX.
+"""
+
+from __future__ import annotations
+
+import math
+
+
+def align_fused_moe_v1_local_tokens(local_num_tokens: int, t_packing: int) -> int:
+    """Round a logical local token count up to a legal v1 launch extent.
+
+    The v1 kernel permits the small decode tiles 2 and 4; larger token tiles
+    must be multiples of 8. The launch extent must also be divisible by the
+    dtype packing factor. The logical token count is not changed by this
+    helper: callers pad each device-local shard and slice the result back.
+    """
+
+    if local_num_tokens <= 0:
+        raise ValueError(f"Expected {local_num_tokens=} to be > 0.")
+    if t_packing <= 0:
+        raise ValueError(f"Expected {t_packing=} to be > 0.")
+
+    for small_extent in (2, 4):
+        if local_num_tokens <= small_extent and small_extent % t_packing == 0:
+            return small_extent
+
+    alignment = math.lcm(8, t_packing)
+    return ((local_num_tokens + alignment - 1) // alignment) * alignment
+
+
+def align_fused_moe_v1_num_tokens(
+    num_tokens: int,
+    ep_size: int,
+    t_packing: int,
+) -> int:
+    """Return the global-equivalent extent for rank-local v1 padding.
+
+    This value sizes v1 tuning and launch internals only. It is not a new
+    global tensor shape or scheduler bucket.
+    """
+
+    if num_tokens <= 0:
+        raise ValueError(f"Expected {num_tokens=} to be > 0.")
+    if ep_size <= 0:
+        raise ValueError(f"Expected {ep_size=} to be > 0.")
+    if num_tokens % ep_size != 0:
+        raise ValueError(f"Expected {num_tokens=} to be aligned to {ep_size=}.")
+
+    local_num_tokens = num_tokens // ep_size
+    return align_fused_moe_v1_local_tokens(local_num_tokens, t_packing) * ep_size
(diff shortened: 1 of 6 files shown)


JamesBrianD:
I’d prefer enforcing the token-bucket constraint before entering the fused-MoE backend instead of adding device-local padding as a fallback. When fused MoE is enabled, every token bucket should be divisible by `ep_size`, and `tokens / ep_size` should satisfy the v1 kernel’s legal local-token extent—at minimum even for BF16, and possibly the stronger existing tiling constraint. Invalid user-provided buckets such as `2032` at EP16 should fail fast with a clear error or be excluded during bucket construction. This keeps the scheduler and kernel shape contracts consistent and avoids silently introducing additional computation inside the kernel.

PR #1544, opened 2026-08-25
feat(vlm): Add Qwen-VL preprocessing and cache-aware request identity

Yanko-7:
## Motivation

This is the first part of the Qwen2.5-VL upstream landing from `epic/vlm`.

Multimodal requests need a typed preprocessing contract and media-aware cache identity before the model runtime can be introduced. Placeholder token IDs alone cannot distinguish requests containing different images or videos, which may produce incorrect Radix Cache hits and DP routing decisions.

This PR only lands request preprocessing and cache identity. It does not add or enable the in-model Qwen2.5-VL runtime.

This consolidates the final functionality developed in #1398, #1439, and #1479.

## Modifications

- Add `MultimodalInputs` and `MultimodalDataItem` as the request-side contract.
- Track per-item placeholder ranges, feature hashes, and cache identity values.
- Add synchronous Qwen-VL image and video preprocessing.
- Compute Qwen MRoPE metadata during preprocessing.
- Add architecture-scoped multimodal processor registration.
- Pass typed multimodal inputs through `TokenizerManager`.
- Use `cache_input_ids` for:
  - Radix Cache lookup and insertion.
  - SWA and unified cache paths.
  - Cache-aware DP routing.
- Add image/video processor, cache fallback, and DP routing tests.

### Compatibility

- Requests without `cache_input_ids` continue using the existing token cache key.
- Text-only requests do not enter the multimodal processor path.
- Models without a registered processor continue using the existing tokenizer path.
- No model is registered and no new model execution path is enabled by this PR.
- Async preprocessing and offloading from #1538/#1542 are intentionally excluded.

## Accuracy Tests

Not applicable. This PR does not modify model forward computation or model weights.

## Benchmarking and Profiling

Not applicable. This PR establishes the request contract and does not claim a performance improvement.

## Test Plan

After integrating all VLM features, the following tests should be run in the development/CI environment.

```bash
python -m pytest -q \
  python/sgl_jax/test/multimodal/test_qwen_vl_processor.py \
  python/sgl_jax/test/mem_cache/test_swa_radix_cache.py \
  python/sgl_jax/test/mem_cache/test_unified_radix_cache.py \
  test/srt/test_dp_schedule_policy.py
```

## Checklist

- [x] The PR description is written in English.
- [x] The purpose and scope are documented.
- [x] A test plan is provided.
- [x] Text and non-multimodal fallback behavior is preserved.
- [ ] Targeted tests passed in the development/CI environment.

### DIFF
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -87,7 +87,6 @@ from sgl_jax.srt.model_executor.forward_batch_info import ForwardMode
 from sgl_jax.srt.model_executor.model_runner_kv_cache_mixin import (
     recurrent_admission_blocked,
 )
-from sgl_jax.srt.multimodal.common.modality_enum import build_cache_input_ids
 from sgl_jax.srt.multimodal.tokenizer_utils import resolve_tokenizer_subdir
 from sgl_jax.srt.precision_tracer import precision_tracer
 from sgl_jax.srt.server_args import PortArgs, ServerArgs
@@ -946,8 +945,7 @@ class Scheduler(
         if not eligible:
             return None
 
-        cache_input_ids = build_cache_input_ids(req.input_ids, req.mm_inputs)
-        token_ids, extra_key = req_prefix_match_key(req, cache_input_ids)
+        token_ids, extra_key = req_prefix_match_key(req)
         matches: dict[int, int] = {}
         prompt_len = len(token_ids) if token_ids else 0
         if token_ids:
@@ -1313,6 +1311,7 @@ class Scheduler(
             recv_req.text,
             recv_req.input_ids,
             recv_req.sampling_params,
+            radix_input_ids=recv_req.radix_input_ids,
             return_logprob=recv_req.return_logprob,
             return_output_logprob_only=recv_req.return_output_logprob_only,
             top_logprobs_num=recv_req.top_logprobs_num,
@@ -1348,7 +1347,6 @@ class Scheduler(
                 req.deepstack_visual_embedding = _extract_mm_value(
                     recv_req.mm_inputs, "deepstack_visual_embedding"
                 )
-            req.cache_input_ids = build_cache_input_ids(req.origin_input_ids, req.mm_inputs)
         # Validate prompt length
         error_msg = validate_input_length(
             req,
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -58,6 +58,7 @@ from sgl_jax.srt.managers.schedule_batch import (
     FINISH_ABORT,
     Req,
     ScheduleBatch,
+    _extract_mm_value,
     acc_global_bid,
     global_server_args_dict,
 )
@@ -86,6 +87,7 @@ from sgl_jax.srt.model_executor.forward_batch_info import ForwardMode
 from sgl_jax.srt.model_executor.model_runner_kv_cache_mixin import (
     recurrent_admission_blocked,
 )
+from sgl_jax.srt.multimodal.common.modality_enum import build_cache_input_ids
 from sgl_jax.srt.multimodal.tokenizer_utils import resolve_tokenizer_subdir
 from sgl_jax.srt.precision_tracer import precision_tracer
 from sgl_jax.srt.server_args import PortArgs, ServerArgs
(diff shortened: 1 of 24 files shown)


python/sgl_jax/srt/multimodal/processors/qwen_vl.py
```diff
@@ -0,0 +1,546 @@
...
+                image_grid_thw=image_grid_thw,
+                video_grid_thw=video_grid_thw,
+                second_per_grid_ts=second_per_grid_ts,
+                vision_start_token_id=self.hf_config.vision_start_token_id,
+                image_token_id=self.hf_config.image_token_id,
+                video_token_id=video_token_id,
+                spatial_merge_size=vision_config.spatial_merge_size,
+                tokens_per_second=getattr(vision_config, "tokens_per_second", None),
+            )
+        mrope_position_delta = np.asarray([[mrope_position_delta]], dtype=np.int32)
```
lianfang0320, line 368: `MultimodalInputs` class defines the `more_positions` and `more_position_delta` as Jax.Array, but here pass with np.ndarray, please confirm the data type.

Yanko-7: Done.

python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -944,7 +946,8 @@ def _select_cache_aware_dp(
         if not eligible:
             return None
 
-        token_ids, extra_key = req_prefix_match_key(req)
+        cache_input_ids = build_cache_input_ids(req.input_ids, req.mm_inputs)
+        token_ids, extra_key = req_prefix_match_key(req, cache_input_ids)
```
JamesBrianD: This introduces two sources of token identity: model `input_ids` and the optional `cache_input_ids`, with each caller responsible for selecting the correct one. This is fragile because lifecycle operations such as truncation can update one without updating the other.

Could we compute one canonical `radix_input_ids` during preprocessing and pass it through the request? It should always exist—equal to model IDs for text requests and hash-substituted for multimodal requests. DP routing, prefix lookup, and insertion should all consume this same field, with an invariant that its length matches the model input IDs. This is similar to upstream SGLang’s canonical `padded_input_ids` approach.

JamesBrianD: This introduces two sources of token identity: model `input_ids` and optional `cache_input_ids`, with callers responsible for selecting the correct one:

```python
cache_input_ids = build_cache_input_ids(req.input_ids, req.mm_inputs)
token_ids, extra_key = req_prefix_match_key(req, cache_input_ids)
```

This is fragile because lifecycle operations such as truncation may update one sequence without updating the other.

Could we compute one canonical cache identity during preprocessing?

```python
cache_ids = build_cache_input_ids(input_ids, mm_inputs)
radix_input_ids = cache_ids if cache_ids is not None else list(input_ids)

tokenized_req = TokenizedGenerateReqInput(
    input_ids=input_ids,               # model semantics
    radix_input_ids=radix_input_ids,   # cache identity
    mm_inputs=mm_inputs,
)
```

Then preserve it explicitly on `Req`:

```python
self.origin_input_ids = recv_req.input_ids
self.radix_input_ids = recv_req.radix_input_ids

assert len(self.origin_input_ids) == len(self.radix_input_ids)
```

DP routing would no longer recompute or select the source:

```python
token_ids, extra_key = req_prefix_match_key(req)

def req_prefix_match_key(req):
    input_ids = req.radix_input_ids
    ...
```

python/sgl_jax/srt/multimodal/processors/qwen_vl.py
```diff
@@ -0,0 +1,546 @@
...
+        tmp.write(payload)
+        return tmp.name
+
+
+def _unwrap_source(source):
+    if isinstance(source, dict) and "url" in source:
+        return source["url"]
+    if hasattr(source, "url"):
+        return source.url
+    return source
```
JamesBrianD: Same as `BaseMultimodalProcessor.unwrap_source()`

Yanko-7: Deleted, directly call unwrap source() at the outer layer.

python/sgl_jax/srt/managers/scheduler.py
Yanko-7: Already unified into radix_input_ids.

JamesBrianD (approved):

PR #1545, opened 2026-08-26
feat(dsa): opt-in Pallas page-topk for the prefill indexer — O(actual kv) scoring, no [T, max_kv] materialization

zorrofox:
This wires the (merged, decode-only) `streamindex_topk` Pallas kernel into the **prefill/extend** indexer path, as previewed in the #1508 thread. Opt-in via `DSA_INDEXER_KERNEL_PREFILL=1`, default OFF ⇒ prefill behaviour unchanged.

### Why

With `DSA_PREFILL_SPARSE=1`, the prefill page-topk currently runs the jnp reference (`streamindex_page_topk_ref`, general path). Two structural costs, measured on TPU v7x tp16 (GLM-5.2 FP8, `page_size=128`, ctx 135168):

1. **O(max_ctx) work regardless of actual length** — the ref gathers and scores all `pages_per_seq` pages for every chunk. Profiled at 33% of a 110k-token chunked prefill (30.6s of 91.4s device time).
2. **[T, max_kv] f32 token scores materialize in HBM** — ~4.4 GB per full layer at chunk=8192/ctx=135k. XLA keeps enough of these alive that `HLO temporaries` alone exceed HBM: chunk 8192 + ctx 135168 fails at startup with `temporaries (98.54G) exceeds available HBM (94.75G)` (≈ 22 full layers × chunk × ctx × 4B). Operators today must drop to chunk 4096, paying ~33% extra TTFT at 16k.

### Modifications

- **Kernel** (`kernels/dsa/streamindex_topk.py`): opt-in *page mode* for `_scores_kernel` — per-KV-block token scores are max-pooled to page granularity in VMEM and stored as one lane-row per block (`bkv_p=128`); only `[T, pages_per_seq]` ever reaches HBM. New public wrapper `streamindex_page_topk` mirrors `streamindex_page_topk_ref(one_token_per_seq=False)` semantics (per-query causal bound, tail-page masking, -1 padding).
- **Compile-time fix**: the python-unrolled per-page DMA loop at `bkv_p=128` cost ~40s of Mosaic compile per kernel instantiation (T=8192), pushing cold-start past the server warmup timeout; a `fori_loop` page fetch compiles in 0.6s (68×) with identical steady-state timing. Decode-sized blocks (`bkv_p<=64`) keep the original unrolled loop, byte-identical to the merged path.
- **Backend wiring** (`dsa_sparse_backend.py`): `_maybe_index_prefill_pages` routes through the kernel under `DSA_INDEXER_KERNEL_PREFILL=1`; ref path untouched and remains the default.
- **Tests**: `test_streamindex_page_topk.py` — TPU parity vs the jnp ref across packed multi-seq, multi-block long extend, prefix (chunked/radix form), and mixed decode+extend batches.

### Performance

Kernel microbench (per-device shapes, tp16 head-TP: H=4, D=128, T=8192, `pages_per_seq`=1056 static; v7x):

| actual kv | jnp ref | kernel | speedup |
|---|---|---|---|
| 8k | 20.7 ms | 4.4 ms | 4.5× |
| 49k | 20.7 ms | 8.0 ms | 2.6× |
| 135k | 20.7 ms | 18.7 ms | 1.1× |

ref is flat (O(max_ctx)); the kernel is O(actual kv), which is what chunked prefill integrates over.

End-to-end, GLM-5.2 FP8 v7x tp16, single stream (TTFT, warm, variance <0.1%):

- **Paired isolation of the kernel effect** (same server, same chunk 4096, flipping only the flag): 76k prefill **62.1s → 44.5s (−28%)**.
- **Newly possible configuration**: chunk 8192 × ctx 135168 (no token-score materialization ⇒ the temporaries OOM term is gone). There, 110k prefill measures **65.5s** (×3, two deployments) and 16k measures **9.3s** — i.e. a single 135k-context server now matches the best short-context single-shot setup (9.75s at ctx 49k) within ~5%, instead of paying the previous ~35-40% small-chunk tax.
- Decode TPOT: no regression between the paired arms (15.3ms at 110k on the kernel arm).

(We deliberately do not quote a 110k ref-vs-kernel percentage: the pre-kernel 110k baseline was measured on an earlier revision of this PR stack, so the clean single-stream speedup claim is the paired 76k row.)

**Concurrent serving (the headline)** — paired precompiled servers, `bench_serving`, same seed/warmup, flipping only the flag:

| load | metric | ref | kernel | gain |
|---|---|---|---|---|
| prefill-heavy (in 4096 / out 128, cc=8) | output tok/s | 2.8 | **29.6** | **10.5×** |
| | median TTFT | 36.0s | 3.0s | 12.2× |
| | median TPOT | 2355ms | 230ms | 10.2× |
| balanced (in 1200 / out 512, cc=8) | output tok/s | 28.1 | **122.4** | **4.4×** |
| | median TTFT | 9.65s | 1.03s | 9.4× |
| | median TPOT | 224ms | 55.7ms | 4.0× |

Under concurrent serving the ref scorer pays its O(max_ctx) cost on **every engine step**, so a steady prefill stream convoys all decode behind ~450ms steps; the kernel collapses the step back to O(actual kv). The single-stream −28% substantially understates the serving-throughput impact.

### Accuracy

- Kernel-level: TPU parity gates vs the jnp ref (4 scenarios, exact page-set equality).
- Long-context paired (flip only `DSA_INDEXER_KERNEL_PREFILL`, chunk 4096): 76k single-needle grid (25 samples, 5 depths) **25/25 vs 25/25**; many-shot long-ctx GSM8K (n=50, `max_new_tokens=3072`) **0.960 vs 0.980** — the same ±1-question band as the #1508/#1512 controls.
- Short-context regression control (200q, 5-shot, temp 0, `parallel=8` against a batching server): kernel **0.965** vs ref **0.955**, both 0 invalid — within the ±1-2 question noise band. Standard GSM8K prompts sit far below `index_topk`, so the kernel selects all pages and is mathematically equivalent to the ref there; this cell is a plumbing no-regression check (same rationale as #1508's short control), and it also exercises the kernel under concurrent packed-extend batches (`max_running=32`).

### Limitations / follow-ups

- Page mode requires `bkv_p=128` (one lane-row per block) and `page_size × 128` block spans; deployments at `page_size=64` work (span 8192) but were not perf-tuned here.
- Page mode is explicitly scoped to `compression_ratio=1` (raises otherwise): the compressed causal bound interacts with page pooling and needs its own parity gate before enabling (DSv4-style caches).
- An independent adversarial review of this change was run pre-submission; its findings (distribution semantics at the call site, double-buffer wait semantics) are addressed as code comments, and the `distribution[2]` usage was verified against `mla_backend`'s EXTEND construction (`[0, 0, num_real_seqs]`).
- The high-kv tail (≥110k actual) approaches ref throughput; the remaining bottleneck is the thin-head (H=4) elementwise epilogue — bf16 intermediates / fused pooling are candidate follow-ups.
- Cold-start (single-stream config) compiles ~4.5min server-wide after the fori fix, same class as ref mode. For batching configs we verified startup-precompile from a **clean compilation cache**: 43min to ready (36min EXTEND grid + 3min DECODE grid), single launch, zero request-path stalls; with a persistent compilation cache the same restart is minutes. Recommend precompile (not on-demand compile) for any concurrent deployment.

### DIFF
--- a/python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
+++ b/python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
@@ -545,7 +545,12 @@ class DSASparseAttentionBackend(MLAAttentionBackend):
                     w_,
                     cache3d.reshape(cache_.shape),
                     seq_lens_,
-                    pi_,
+                    # the kernel indexes page_indices[seq_id * pages_per_seq + p]
+                    # (fixed stride), but sglang packs seq i's pages at
+                    # cu_kv_lens[i]//page_size (variable stride) — repack, same
+                    # as the decode call site, or a multi-request EXTEND batch
+                    # with unequal lengths reads another sequence's pages.
+                    _fixed_stride_pages(pi_, cukv_, page_size, pages_per_seq),
                     cuq_,
                     dist_[2],
                     k_pages=k_pages,
--- a/python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
+++ b/python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
@@ -535,6 +535,11 @@ class DSASparseAttentionBackend(MLAAttentionBackend):
             cache3d = cache_.reshape(cache_.shape[0], page_size, idx_dim)
             cache3d = _scatter_paged(cache3d, k_, seq_lens_, pi_, cuq_, cukv_, pages_per_seq)
             if _INDEXER_KERNEL_PREFILL:
+                # dist_[2] == number of real (seq_len > 0) sequences in this
+                # EXTEND batch: mla_backend builds distribution = [0, 0, N] for
+                # ForwardMode.EXTEND (decode runs as a separate forward), so the
+                # kernel's per-seq grid over [0, N) matches the ref's masked
+                # full-batch loop exactly; padded seqs stay -1 on both paths.
                 topk_pages = streamindex_page_topk(
                     q_,
                     w_,
--- a/python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
+++ b/python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
@@ -25,7 +25,10 @@ from jax.tree_util import register_pytree_node_class
 from sgl_jax.srt.kernels.dsa.ref import streamindex_page_topk_ref, streamindex_topk_ref
 from sgl_jax.srt.kernels.dsa.sparse_mla import compute_topk_pages, sparse_mla_page_level
 from sgl_jax.srt.kernels.dsa.sparse_mla_prefill import prefill_write_and_attend_ragged
-from sgl_jax.srt.kernels.dsa.streamindex_topk import streamindex_topk
+from sgl_jax.srt.kernels.dsa.streamindex_topk import (
+    streamindex_page_topk,
+    streamindex_topk,
+)
 from sgl_jax.srt.kernels.mla.v2.kernel import mla_ragged_paged_attention
 from sgl_jax.srt.layers.attention.mla_backend import MLAAttentionBackend
 from sgl_jax.srt.utils.profiling_utils import named_scope
@@ -49,6 +52,12 @@ _PAGE_TOPK_BUDGET = int(os.environ.get("DSA_PAGE_TOPK", "0"))
 # (sum_h relu(q·k)·w_h, approx_max_k). Token-topk path only; DSA_PAGE_TOPK
 # takes precedence when both are set.
 _INDEXER_KERNEL = os.environ.get("DSA_INDEXER_KERNEL", "0") == "1"
+# Opt-in: prefill/extend page-level indexer top-k via the page-pooled Pallas
+# streamindex kernel instead of the jnp reference. The kernel scores O(actual
+# kv_len) pages (ref is O(max_ctx) padded gather) and never materializes
+# [T, max_kv] token scores in HBM — the dominant HLO-temporaries term at long
+# context. Same selection semantics as the ref (parity-gated). Default OFF.
+_INDEXER_KERNEL_PREFILL = os.environ.get("DSA_INDEXER_KERNEL_PREFILL", "0") == "1"
 # bkv block of 64 pages (4K tokens @ page 64) benched fastest across
 # B=8..64, ctx 8K..128K on v7x/v6e.
 _INDEXER_KERNEL_KV_PAGES_PER_BLOCK = 64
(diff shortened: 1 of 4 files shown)


python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
```diff
@@ -525,19 +534,36 @@ def _run(q_, k_, w_, cache_, seq_lens_, pi_, cuq_, cukv_, dist_):
...
+                # EXTEND batch: mla_backend builds distribution = [0, 0, N] for
+                # ForwardMode.EXTEND (decode runs as a separate forward), so the
+                # kernel's per-seq grid over [0, N) matches the ref's masked
+                # full-batch loop exactly; padded seqs stay -1 on both paths.
+                topk_pages = streamindex_page_topk(
+                    q_,
+                    w_,
+                    cache3d.reshape(cache_.shape),
+                    seq_lens_,
+                    pi_,
```
cjx0709: Could we pass `_fixed_stride_pages(pi_, cukv_, page_size, pages_per_seq)` here? In a packed multi-request EXTEND batch, the kernel indexes `seq_id * pages_per_seq`, so passing `pi_` directly can read another sequence’s pages.

test/srt/kernels/dsa/test_streamindex_page_topk.py
```diff
@@ -0,0 +1,83 @@
...
+    pytest.skip("streamindex_page_topk kernel requires TPU", allow_module_level=True)
+
+
+@pytest.mark.parametrize(
+    "page_size, pages_per_seq, k_pages, bkv_p, bq_sz, q_lens, kv_lens",
+    [
+        # 2-seq packed extend, tail pages partially filled, causal within chunk
+        (8, 16, 3, 128, 8, [5, 12], [37, 61]),
+        # single long extend spanning many bkv blocks
+        (8, 48, 5, 128, 16, [33], [301]),
```
cjx0709, line 24: Could we make this case exceed one KV block? With `page_size=8`, `bkv_p=128`, and `kv_len=301`, `num_bkv` is still 1, so the new multi-block accumulation/prefetch path is not exercised.

python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
zorrofox: Good catch — confirmed, and it was a real correctness bug, not just a hygiene issue. The kernel indexes `page_indices[seq_id * pages_per_seq + p]` while sglang packs seq i's pages at `cu_kv_lens[i] // page_size` (page-aligned cumsum, see `schedule_batch._merge_cache_loc`), so in a packed multi-request EXTEND batch with unequal lengths every query of an offset sequence scored another sequence's pages. Repro on a 2-seq batch (5+13 pages): 12/17 query rows mis-selected vs the ref; with `_fixed_stride_pages` repacking: 0/17.

It slipped through the original gates because (a) the parity harness built `cu_kv_lens` at a fixed stride by construction, and (b) in the accuracy runs every request had ≤ `k_pages` pages, where top-k saturates and selects all pages regardless of scores. Fixed in db67d041 exactly as you suggested (same repack as the decode call site), and the parity gate now includes production variable-stride packing cases so this layout mismatch can't regress silently.

test/srt/kernels/dsa/test_streamindex_page_topk.py
zorrofox: Right — none of the original cases exceeded one KV block. Page mode pins `num_kv_pages_per_block == 128`, so crossing a block boundary needs `pages_per_seq > 128`; added two such cases (138 pages → 2 blocks single-seq, and a multi-seq one where one sequence spans 2 blocks), plus a variable-stride × multi-block combination. The multi-block accumulation/prefetch path passes parity on all of them (db67d041).

cjx0709 (approved):
LGTM

Issue #1543, opened 2026-08-24
[Kernel] mHC Pallas kernel family (Sinkhorn / aggregate / combine)


denis-mil:
Tracking issue for the mHC (manifold-constrained Hyper-Connections, arXiv 2512.24880) portion of "Develop and optimize mHC, CSA, and HCA" (#1460). CSA and HCA are out of scope here.

## Scope

A self-contained Pallas TPU kernel family for the mHC layer boundary, forward pass only, with a pure-JAX reference fallback for non-TPU backends. Three kernels, where `x ∈ R^{n×C}` is one token's pre-mix stream state and `F` is the block (attention or FFN):

| kernel | op |
| --- | --- |
| `mhc_sinkhorn` | `H_res = Sinkhorn(t_res)`, projection toward the Birkhoff polytope |
| `mhc_aggregate` | `u = H_pre @ x` |
| `mhc_combine` | `x_next = H_res @ x + outer(H_post, y)` |

The coefficient path (Eqs. 14-16) starts in ordinary XLA rather than being fused. The paper's five-kernel fusion layout is treated as a measurement question rather than an assumption, and the fusions are tracked as remaining work below.

`H_res` is input-dependent per token and per layer, so the Sinkhorn projection runs on every token at every layer at inference time. That is what makes it worth a kernel rather than an einsum: the XLA lowering is dispatch-bound on `T` independent `n×n` problems.

### Explicitly not in scope

An `srt/layers/` mHC module, model wiring, a `server_args.py` flag, the backward pass, and registration in `kernel-performance-test-tpu-v6e-1`. The only public mHC checkpoint is DeepSeek-V4-Flash-0731 at 304B parameters, so end-to-end wiring means porting all of V4. The package layout leaves room for each of those to drop in later without moving files; the performance-suite registration in particular should happen at adoption time, since nothing calls these kernels until a layer does.

## Grounding

The numerical contract is pinned in a spec derived from the mHC paper, the predecessor Hyper-Connections paper (arXiv 2409.19606), three community implementations, and the `hc_*` keys in the DeepSeek-V4-Flash config — which fix `hc_mult: 4` (= `n`), `hc_sinkhorn_iters: 20`, `hc_eps: 1e-06` against real weights. Points the paper leaves ambiguous are marked unresolved in the spec and surfaced as keyword arguments rather than silently chosen.

Two consequences worth stating up front, because both are easy to get wrong:

[...]

PR #1547, opened 2026-08-26
feat(mhc): add DeepSeek-V4 hyper-connection kernels

KoalaYuFeng:
## Motivation

Add TPU support for DeepSeek-V4 multi-stream hyper-connections (mHC). mHC wraps each attention or FFN sublayer `F`—including HCA and CSA—while keeping four parallel residual streams. Its execution is `X_l → pre → u → F(u) → post → X_{l+1}`.

- **Pre**: collapse the residual streams as
  `u = Σ_i H_pre[i] · X_l[i]`. The resulting `u` is the input passed to HCA, CSA, or FFN; pre also generates `H_post` and Sinkhorn-normalized `H_res` for post.
- **Post**: receive the sublayer output `y = F(u)` and combine it with the original residual streams: `X_{l+1}[j] = H_post[j] · y + Σ_i H_res[j,i] · X_l[i]`.
- **Head**: after the final layer, collapse the four residual streams into one hidden state before the LM head.

Related discussion: #1543. This PR was developed independently around the same mHC kernel-family scope, and we welcome collaboration on further fusion and validation.

## Modifications

- Add Pallas kernels for pre collapse, post expansion/remixing, and head collapse. (pre, post, head kernels)
- Add a float64 NumPy reference and correctness coverage across tuned, ragged, and alternate shapes.
- Add a reproducible operator-level harness at `benchmark/kernels/mhc/benchmark_mhc.py` for both supported hidden sizes and dot-precision modes.

## Accuracy Tests

```bash
PYTHONPATH=python:. python -m pytest -q test/srt/kernels/mhc
```

Result: `55 passed`. DEFAULT and HIGHEST both pass the independent NumPy oracle at `rtol=2e-2, atol=1e-2`.

## Benchmarking and Profiling

On TPU, JAX documents `Precision.HIGHEST` as using six `bfloat16` passes (https://docs.jax.dev/en/latest/201/precision.html). We therefore report `DEFAULT` and `HIGHEST` separately. The table below normalizes `HIGHEST` throughput to `1.00x` and reports `DEFAULT` throughput relative to it.

| N | D=4096 pre | post | head | D=7168 pre | post | head |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 1.14x | 1.04x | 1.09x | 1.19x | 1.06x | 1.13x |
| 128 | 1.25x | 1.77x | 1.21x | 1.39x | 1.82x | 1.31x |
| 256 | 1.35x | 1.84x | 1.28x | 1.47x | 1.94x | 1.35x |
| 512 | 1.46x | 1.93x | 1.36x | 1.53x | 1.97x | 1.38x |
| 1024 | 1.55x | 1.98x | 1.41x | 1.57x | 2.04x | 1.39x |
| 2048 | 1.60x | 2.13x | 1.44x | 1.60x | 2.18x | 1.40x |
| 4096 | 1.62x | 2.15x | 1.45x | 1.61x | 2.18x | 1.40x |
| 8192 | 1.63x | 2.17x | 1.46x | 1.61x | 2.19x | 1.41x |

Both modes pass the independent NumPy oracle at `rtol=2e-2, atol=1e-2`. At fixed `N=128`, HIGHEST reduces the pre/head RMSE by 6--11%:

| N | hidden size (D) | operator | DEFAULT RMSE | HIGHEST RMSE | reduction |
|---:|---:|---|---:|---:|---:|
| 128 | 4096 | pre | 1.883e-4 | 1.769e-4 | 6.1% |
| 128 | 4096 | head | 1.914e-4 | 1.772e-4 | 7.4% |
| 128 | 7168 | pre | 1.975e-4 | 1.811e-4 | 8.3% |
| 128 | 7168 | head | 2.155e-4 | 1.924e-4 | 10.7% |

## Comparison with TPU-Inference

Measured on one TPU v6e-1 with JAX/JAXLIB 0.11.0, TPU-Inference commit `e99a70d`, and vLLM commit `6c95a64`. Both implementations receive the same device arrays, pass correctness before timing, rotate four independent HBM buffers, and use 3 warmups plus the arithmetic mean of 10 device calls per implementation. Calls are interleaved to reduce temporal drift. Device time spans HBM-resident inputs through completed HBM outputs. DEFAULT and HIGHEST are applied to both implementations before compilation.

Speedup is `TPU-Inference latency / SGLang-JAX latency`; values above `1.00x` favor SGLang-JAX. Raw latency is preserved in the benchmark artifacts.

### Hidden size 4096 — DEFAULT

| N | pre | post | head |
|---:|---:|---:|---:|
| 1 | 2.05x | 1.04x | 1.45x |
| 128 | 1.55x | 1.00x | 1.58x |
| 256 | 1.71x | 1.00x | 1.67x |
| 512 | 1.93x | 1.00x | 1.68x |
| 1024 | 2.00x | 1.00x | 2.66x |
| 2048 | 2.58x | 1.38x | 2.73x |
| 4096 | 2.32x | 1.03x | 2.52x |
| 8192 | 2.30x | 1.49x | 2.51x |

### Hidden size 4096 — HIGHEST

| N | pre | post | head |
|---:|---:|---:|---:|
| 1 | 1.81x | 1.01x | 1.31x |
| 128 | 1.37x | 1.00x | 1.44x |
| 256 | 1.39x | 1.00x | 1.35x |
| 512 | 1.49x | 1.00x | 1.40x |
| 1024 | 1.45x | 1.00x | 1.89x |
| 2048 | 1.70x | 1.00x | 1.90x |
| 4096 | 1.51x | 0.77x | 1.75x |
| 8192 | 1.48x | 0.88x | 1.73x |

### Hidden size 7168 — DEFAULT

| N | pre | post | head |
|---:|---:|---:|---:|
| 1 | 1.84x | 1.02x | 1.27x |
| 128 | 1.73x | 1.00x | 1.49x |
| 256 | 1.92x | 1.00x | 1.63x |
| 512 | 2.04x | 1.00x | 1.55x |
| 1024 | 2.64x | 1.00x | 2.59x |
| 2048 | 2.41x | 1.03x | 2.41x |
| 4096 | 2.38x | 1.48x | 2.39x |
| 8192 | 2.39x | 1.49x | 2.40x |

### Hidden size 7168 — HIGHEST

| N | pre | post | head |
|---:|---:|---:|---:|
| 1 | 1.54x | 0.99x | 1.14x |
| 128 | 1.37x | 1.00x | 1.18x |
| 256 | 1.46x | 1.00x | 1.35x |
| 512 | 1.49x | 1.00x | 1.27x |
| 1024 | 1.78x | 1.00x | 1.86x |
| 2048 | 1.58x | 0.77x | 1.74x |
| 4096 | 1.56x | 0.89x | 1.71x |
| 8192 | 1.55x | 0.89x | 1.72x |

All 96 cross-stack operator comparisons pass `rtol=2e-2, atol=1e-2`. HIGHEST post is the only regression: its K=4 contraction currently uses a six-pass high-precision MXU dot; an explicit FP32 VPU reduction or a precision-aware XLA fallback is a follow-up optimization.

## Reproducibility

- **Docker image:** [`koalayufeng/mhc-compare:latest`](https://hub.docker.com/r/koalayufeng/mhc-compare), digest: `sha256:1df0abd8dda16dceb79e04f19e3cddc75ca9ca0a125f04d79a4cf370dd76e444`
- **Pinned versions:** SGLang-JAX `045ff682`, TPU-Inference `e99a70d`, vLLM `6c95a64`, and JAX/JAXLIB
  `0.11.0`

```bash
mkdir -p mhc-results
for hidden in 4096 7168; do
  for precision in default highest; do
    docker run --rm --privileged --network host --shm-size 150g \
      -v "$PWD/mhc-results:/results" \
      koalayufeng/mhc-compare:latest \
      --hidden-size "$hidden" \
      --dot-precision "$precision" \
      --output /results
  done
done
```

## Checklist

### DIFF
--- a/benchmark/kernels/mhc/benchmark_mhc.py
+++ b/benchmark/kernels/mhc/benchmark_mhc.py
@@ -1,0 +1,560 @@
+#!/usr/bin/env python3
+# ruff: noqa: E402
+"""Fair operator-level mHC comparison against TPU-Inference on one TPU.
+
+TPU-Inference and vLLM are optional external checkouts. Point
+``TPU_INFERENCE_ROOT`` and ``VLLM_ROOT`` at their pinned revisions.
+"""
+
+from __future__ import annotations
+
+import argparse
+import gc
+import hashlib
+import inspect
+import json
+import os
+import subprocess
+import sys
+import tempfile
+from datetime import datetime, timezone
+from functools import partial
+from pathlib import Path
+
+SGLANG_ROOT = Path(__file__).resolve().parents[3]
+TPU_INFERENCE_ROOT = Path(
+    os.environ.get("TPU_INFERENCE_ROOT", SGLANG_ROOT.parent / "tpu-inference")
+)
+VLLM_ROOT = Path(os.environ.get("VLLM_ROOT", SGLANG_ROOT.parent / "vllm"))
+
+os.environ.setdefault("VLLM_TARGET_DEVICE", "tpu")
+for source_root in (VLLM_ROOT, TPU_INFERENCE_ROOT, SGLANG_ROOT / "python"):
+    sys.path.insert(0, str(source_root))
+
+import jax
+import jax.numpy as jnp
+import jaxlib
+import numpy as np
+import torchax
+import tpu_inference.layers.vllm.custom_ops.mhc as tpu_source
+import vllm.model_executor.layers.mhc as vllm_source
+from torchax.interop import jax_view, torch_view
+from vllm.config import set_current_vllm_config
+from vllm.config.compilation import CompilationConfig
+from vllm.model_executor.layers.mhc import HCHeadOp, MHCPostOp, MHCPreOp
+from vllm.platforms import current_platform
+
+import sgl_jax.srt.kernels.mhc.mhc as sglang_source
+from sgl_jax.srt.kernels.mhc import (
+    mhc_head_collapse_fused,
+    mhc_post_fused,
+    mhc_pre_fused,
+)
+
+HC = 4
+HIDDEN = 4096
+DOT_PRECISION = jax.lax.Precision.DEFAULT
+DOT_PRECISION_NAME = "DEFAULT"
+ROWS = (2 + HC) * HC
+RTOL = 2e-2
+ATOL = 1e-2
+ROTATING_BUFFERS = 4
+WARMUPS = 3
+TIMED_RUNS = 10
+TIMED_RUNS_PER_BLOCK = TIMED_RUNS // 2
+MEASUREMENT_ORDER = (0, 1, 1, 0)
+INPUT_SEED = 20260825
+
+
+def _inside(module, root: Path) -> bool:
+    return Path(module.__file__).resolve().is_relative_to(root.resolve())
+
+
+if not _inside(sglang_source, SGLANG_ROOT):
+    raise RuntimeError(f"loaded SGLang source outside snapshot: {sglang_source.__file__}")
+if not _inside(tpu_source, TPU_INFERENCE_ROOT):
+    raise RuntimeError(f"loaded TPU-inference outside snapshot: {tpu_source.__file__}")
+if not _inside(vllm_source, VLLM_ROOT):
+    raise RuntimeError(f"loaded vLLM outside snapshot: {vllm_source.__file__}")
+if not current_platform.is_tpu():
+    raise RuntimeError("TPU-inference did not select its TPU platform")
+
+
+class _DispatchConfig:
+    compilation_config = CompilationConfig()
+
+
+with set_current_vllm_config(_DispatchConfig()):
+    _TPU_PRE = MHCPreOp()
+    _TPU_POST = MHCPostOp()
+    _TPU_HEAD = HCHeadOp()
+
+
+def _check_dispatch(op, method: str) -> None:
+    if type(op).__module__ != "tpu_inference.layers.vllm.custom_ops.mhc":
+        raise RuntimeError(f"unexpected TPU implementation: {type(op)}")
+    if op._forward_method.__qualname__ != method:
+        raise RuntimeError(f"unexpected TPU dispatch: {op._forward_method.__qualname__}")
+
+
+_check_dispatch(_TPU_PRE, "VllmMHCPreOp.forward_tpu")
+_check_dispatch(_TPU_POST, "VllmMHCPostOp.forward_tpu")
+_check_dispatch(_TPU_HEAD, "VllmHCHeadOp.forward_tpu")
+
+
+@partial(
+    jax.jit,
+    static_argnames=(
+        "rms_eps",
+        "hc_pre_eps",
+        "hc_sinkhorn_eps",
+        "post_multiplier",
+        "sinkhorn_iterations",
+    ),
+)
+def sglang_pre(
+    residual,
+    fn,
+    hc_scale,
+    hc_base,
+    *,
+    rms_eps=1e-6,
+    hc_pre_eps=1e-6,
+    hc_sinkhorn_eps=1e-6,
+    post_multiplier=2.0,
+    sinkhorn_iterations=20,
+):
+    if hc_pre_eps != hc_sinkhorn_eps or post_multiplier != 2.0:
+        raise ValueError("the common ABI requires equal eps values and post_multiplier=2")
+    layer_input, post_mix, comb_mix = mhc_pre_fused(
+        residual,
+        fn,
+        hc_scale,
+        hc_base,
+        hc_mult=HC,
+        sinkhorn_iters=sinkhorn_iterations,
+        norm_eps=rms_eps,
+        hc_eps=hc_pre_eps,
+        dot_precision=DOT_PRECISION,
+    )
+    return layer_input, post_mix[..., None], comb_mix
+
+
+@partial(
+    jax.jit,
+    static_argnames=(
+        "rms_eps",
+        "hc_pre_eps",
+        "hc_sinkhorn_eps",
+        "post_multiplier",
+        "sinkhorn_iterations",
+    ),
+)
+def tpu_pre(
+    residual,
+    fn,
+    hc_scale,
+    hc_base,
+    *,
+    rms_eps=1e-6,
+    hc_pre_eps=1e-6,
+    hc_sinkhorn_eps=1e-6,
+    post_multiplier=2.0,
+    sinkhorn_iterations=20,
+):
+    post_mix, comb_mix, layer_input = _TPU_PRE(
+        torch_view(residual),
+        torch_view(fn),
+        torch_view(hc_scale),
+        torch_view(hc_base),
+        rms_eps,
+        hc_pre_eps,
+        hc_sinkhorn_eps,
+        post_multiplier,
+        sinkhorn_iterations,
+    )
+    return jax_view(layer_input), jax_view(post_mix), jax_view(comb_mix)
+
+
+@jax.jit
+def sglang_post(x, residual, post_mix, comb_mix):
+    return mhc_post_fused(
+        x,
+        residual,
+        post_mix[..., 0],
+        comb_mix,
+        precision=DOT_PRECISION,
+    )
+
+
+@jax.jit
+def tpu_post(x, residual, post_mix, comb_mix):
+    return jax_view(
+        _TPU_POST(
+            torch_view(x),
+            torch_view(residual),
+            torch_view(post_mix),
+            torch_view(comb_mix),
+        )
+    )
+
+
+@partial(jax.jit, static_argnames=("rms_eps", "hc_eps"))
+def sglang_head(
+    residual,
+    fn,
+    hc_scale,
+    hc_base,
+    *,
+    rms_eps=1e-6,
+    hc_eps=1e-6,
+):
+    return mhc_head_collapse_fused(
+        residual,
+        fn,
+        hc_scale,
+        hc_base,
+        hc_mult=HC,
+        norm_eps=rms_eps,
+        hc_eps=hc_eps,
+        dot_precision=DOT_PRECISION,
+    )
+
+
+@partial(jax.jit, static_argnames=("rms_eps", "hc_eps"))
+def tpu_head(
+    residual,
+    fn,
+    hc_scale,
+    hc_base,
+    *,
+    rms_eps=1e-6,
+    hc_eps=1e-6,
+):
+    return jax_view(
+        _TPU_HEAD(
+            torch_view(residual),
+            torch_view(fn),
+            torch_view(hc_scale),
+            torch_view(hc_base),
+            rms_eps,
+            hc_eps,
+        )
+    )
+
+
+def _inputs(tokens: int, seed: int) -> dict[str, jax.Array]:
+    rng = np.random.default_rng(seed)
+    comb = np.exp(rng.standard_normal((tokens, HC, HC), dtype=np.float32) * 0.1)
+    comb /= comb.sum(axis=-1, keepdims=True)
+    data = {
+        "residual": jnp.asarray(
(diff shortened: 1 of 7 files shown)


python/sgl_jax/srt/kernels/mhc/mhc.py
```diff
@@ -0,0 +1,568 @@
...
+    select_collapse_block_tokens,
+    select_gates_block_tokens,
+    select_post_backend,
+    select_post_block_tokens,
+)
+
+
+def _device_kind() -> str:
+    devices = jax.devices()
+    return devices[0].device_kind if devices else "TPU v6e"
```
denis-mil: Small question about the fallback here — why does this default to "TPU v6e" when jax.devices() is empty?
Would raising (or returning "" and letting _platform_parameters raise its existing message) be preferable?

python/sgl_jax/srt/kernels/mhc/mhc.py
```diff
@@ -0,0 +1,568 @@
...
+    else:
+        pre = jax.nn.sigmoid(mixes[:, :hc] * scale[0] + base[None, :hc]) + hc_eps
+        # Preserve the full projection layout; slicing can change its reduction tree.
+        extra_outs[0][...] = mixes
+
+    # VPU accumulation preserves FP32 gates; Mosaic's MXU dot rounds them to BF16.
+    # Stream each residual into one accumulator instead of widening [BT, hc, d].
+    collapsed = jnp.zeros((bt, d), dtype=jnp.float32)
+    for stream in range(hc):
+        collapsed = collapsed + pre[:, stream, None] * x[:, stream, :].astype(jnp.float32)
```
denis-mil, line 228: I like using VPU accumulation over an MXU dot here — keeps the gates in FP32

denis-mil (commented):
A few questions about how the reported numbers were produced:

What dot_precision was used in the benchmark? The signature defaults to jax.lax.Precision.DEFAULT, which on TPU means bf16 passes for the fp32 dot_general calls. HIGHEST (fp32, 6-pass) is a very different cost, so the numbers are only comparable if the setting is stated. Please record the value alongside the results, and ideally report both DEFAULT and HIGHEST so the accuracy/throughput tradeoff is visible.

What hidden_size was benchmarked — 4096 (DeepSeek V4-flash) or 7168 (V4-PRO)? Correctness is already parametrized over both in [test_mhc.py:133], but the perf numbers appear to be a single width. Both are real deployment targets and they land in different block-size, so should we benchmark and tune both rather than extrapolating from one?

Why is benchmark/kernels/mhc/benchmark_mhc.py not committed? Sibling kernels ship their harness ([bench_mla.py], [bench_flashattention.py], [tune_grouped_topk_bt.py] Without it the next person tuning block sizes has to rewrite the harness.

denis-mil (approved):

python/sgl_jax/srt/kernels/mhc/mhc.py
KoalaYuFeng: Good catch. Silently selecting v6e is unsafe because it may select an invalid VMEM/tile schedule. I’ll make _device_kind() raise a clear error when no JAX device is available. Unsupported device kinds are rejected by _platform_parameters() on the default auto-tuned path.

python/sgl_jax/srt/kernels/mhc/mhc.py
KoalaYuFeng: Thanks! Yes. With hc=4, the reduction is too small to use the MXU efficiently. This keeps the gate multiplication and accumulation in FP32 until the final output cast. We also verified that Mosaic lowers this path to vector mul/add operations, with zero MXU usage.

KoalaYuFeng:
> A few questions about how the reported numbers were produced:
> 
> What dot_precision was used in the benchmark? The signature defaults to jax.lax.Precision.DEFAULT, which on TPU means bf16 passes for the fp32 dot_general calls. HIGHEST (fp32, 6-pass) is a very different cost, so the numbers are only comparable if the setting is stated. Please record the value alongside the results, and ideally report both DEFAULT and HIGHEST so the accuracy/throughput tradeoff is visible.
> 
> What hidden_size was benchmarked — 4096 (DeepSeek V4-flash) or 7168 (V4-PRO)? Correctness is already parametrized over both in [test_mhc.py:133], but the perf numbers appear to be a single width. Both are real deployment targets and they land in different block-size, so should we benchmark and tune both rather than extrapolating from one?
> 
> Why is benchmark/kernels/mhc/benchmark_mhc.py not committed? Sibling kernels ship their harness ([bench_mla.py], [bench_flashattention.py], [tune_grouped_topk_bt.py] Without it the next person tuning block sizes has to rewrite the harness.

Thanks — these are good points. The originally reported numbers used jax.lax.Precision.DEFAULT with hidden_size=4096. I have now rerun the benchmarks with:

 - Both DEFAULT and HIGHEST, applied consistently to our kernels and TPU-Inference. On TPU, JAX
    documents HIGHEST as six BF16 passes (https://docs.jax.dev/en/latest/201/precision.html).
 - Both deployment widths, hidden_size=4096 and 7168.
 - N ∈ {1, 128, 256, 512, 1024, 2048, 4096, 8192}.
 - A committed harness at benchmark/kernels/mhc/benchmark_mhc.py, using identical device inputs,
    correctness-before-timing, four rotating HBM buffers, 3 warmups, and the mean of 10 device-time
    measurements.

All 96 cross-stack comparisons pass rtol=2e-2, atol=1e-2. Relative to HIGHEST, DEFAULT is 1.04–2.19× faster depending on operator and shape; at N=128, HIGHEST reduces pre/head RMSE by 6.1–10.7%. Against TPU-Inference, DEFAULT reaches 1.55–2.64× on pre, 1.27–2.73× on head, and 1.00–1.49× on post. Under HIGHEST, pre/head remain 1.14–1.90× faster; post matches at small N but falls to 0.77–0.89× on the largest shapes, which we identify as a follow-up optimization. The complete per-shape tables are now included in the PR description.

denis-mil:
Hi @KoalaYuFeng, thanks for the thorough updates and benchmark comparisons!

It looks like the Lint / lint (pull_request) check is currently failing on the latest commit. Could you take a look when you get a chance?

KoalaYuFeng:
> Hi @KoalaYuFeng, thanks for the thorough updates and benchmark comparisons!
> 
> It looks like the Lint / lint (pull_request) check is currently failing on the latest commit. Could you take a look when you get a chance?

Ok, I fix this format issue, thanks.

KoalaYuFeng:
Hi @JamesBrianD @cjx0709 @Infra-lab-dev, gentle ping. The lint failure is fixed and CI is green now. denis-mil's review comments have all been addressed (device_kind fallback, DEFAULT/HIGHEST benchmarks for both hidden sizes, committed benchmark harness). Would appreciate a maintainer review when you get a chance. Thanks!

python/sgl_jax/srt/kernels/mhc/mhc.py
```diff
@@ -0,0 +1,574 @@
...
+    comb: jax.Array,
+    *,
+    block_tokens: int | None = None,
+    backend: str = "auto",
+    interpret: bool | None = None,
+    precision=jax.lax.Precision.DEFAULT,
+):
+    """Use XLA while its live set fits VMEM, then aligned Pallas tiles."""
+    if residual_streams.ndim != 3:
+        raise ValueError(f"residual_streams must be [N, hc, d], got {residual_streams.shape}")
```
Infra-lab-dev: Pre accepts [B, T, H, D], but post rejects the same residual shape. Could we keep the shape contract consistent? If this PR only needs flattened [N, H, D] inputs, restricting pre to that shape would also work.

python/sgl_jax/srt/kernels/mhc/mhc.py
```diff
@@ -0,0 +1,574 @@
...
+    mixes = mixes[:n]
+    # Sinkhorn uses a 2048-token tile; collapse is limited near 128 by projection.
+    post, comb = mhc_gates(
+        mixes,
+        hc_scale,
+        hc_base,
+        hc_mult=hc,
+        sinkhorn_iters=sinkhorn_iters,
+        eps=hc_eps,
+    )
```
Infra-lab-dev, line 347: Please pass interpret through to mhc_gates as well. Otherwise, interpret=True only applies to the collapse kernel.

test/srt/kernels/mhc/test_mhc.py
```diff
@@ -0,0 +1,222 @@
...
+        "head_base": jax.random.normal(keys[4], (hc,), jnp.float32) * 0.05,
+        "block_out": (jax.random.normal(keys[5], (n, hidden), jnp.float32) * 0.1).astype(
+            jnp.bfloat16
+        ),
+        "mixes": jax.random.normal(keys[0], (n, mix_hc), jnp.float32),
+    }
+
+
+def _close(got, want, label, tol=PROJECTED):
+    for index, (a, b) in enumerate(zip(got, want)):
```
Infra-lab-dev: use `zip(..., strict=True)` here, otherwise missing outputs can still pass the check.

benchmark/kernels/mhc/benchmark_mhc.py
```diff
@@ -0,0 +1,558 @@
...
+            raise AssertionError(
+                f"output ABI differs: {lhs.shape}/{lhs.dtype} != {rhs.shape}/{rhs.dtype}"
+            )
+        lhs_np = np.asarray(lhs, np.float32)
+        rhs_np = np.asarray(rhs, np.float32)
+        diff = np.abs(lhs_np - rhs_np)
+        max_abs = max(max_abs, float(diff.max(initial=0.0)))
+        relative = diff / np.maximum(np.abs(rhs_np), ATOL)
+        max_rel = max(max_rel, float(relative.max(initial=0.0)))
+        np.testing.assert_allclose(lhs_np, rhs_np, rtol=RTOL, atol=ATOL)
```
Infra-lab-dev, line 307: Check that both outputs are finite before comparing. Matching NaNs or Infs can also pass this check.

python/sgl_jax/srt/kernels/mhc/mhc.py
KoalaYuFeng: Fixed. mhc_post_fused now accepts [..., hc, d], flattens only the leading dimensions internally, and restores them on output.

python/sgl_jax/srt/kernels/mhc/mhc.py
KoalaYuFeng: Agree, interpret is now forwarded to mhc_gates(), so interpret=True consistently applies to both the collapse and Sinkhorn Pallas calls.

test/srt/kernels/mhc/test_mhc.py
KoalaYuFeng: Ok, now use zip(..., strict=True) in the output comparisons

benchmark/kernels/mhc/benchmark_mhc.py
KoalaYuFeng: Fixed. Both output trees are now checked with np.isfinite before computing the error metrics.

KoalaYuFeng:
@Infra-lab-dev Thanks for the review. All four comments are addressed, and the branch is rebased and squashed into one commit. The full mHC suite passes with no performance regression. Could you please take another look?

JamesBrianD:
@KoalaYuFeng Hi, please follow https://github.com/sgl-project/sglang-jax/blob/main/docs/developer_guide/contribution_guide.md#format-code-with-pre-commit to format the codes. After ci passing, the pr will be merged. Thanks!

KoalaYuFeng:
> @KoalaYuFeng Hi, please follow https://github.com/sgl-project/sglang-jax/blob/main/docs/developer_guide/contribution_guide.md#format-code-with-pre-commit to format the codes. After ci passing, the pr will be merged. Thanks!

Hi, @JamesBrianD . Done. I ran pre-commit run --all-files until all hooks passed.

JamesBrianD (approved):

PR #1550, opened 2026-08-26
perf(kda): optimize the KDA prefill kernel on TPU

Junyi-99:
## Motivation

Improve KDA TPU prefill performance without changing the `chunk_kda_fwd` API.

## Modifications

- Replace the bounded-gate Aqk/L and triangular solve with MXU strip GEMMs and a finite-Neumann solve (`safe_gate`).
- Fuse gate cumsum with intra and recurrence with output, keeping `h` and `v_new` in VMEM.
- Remove gather/scatter and transpose glue with unified addressing, then flatten the recurrence grid over packed chunks.
- Add a native `[1, T, H, D]` path that processes eight heads per program and interleaves MXU work across stages.
- Omit backward-only intermediates during inference unless explicitly requested.
- Preserve safe fallbacks and expose each optimization through a static ablation switch.

## Accuracy Tests

- `python/sgl_jax/test/kernels/kda_test.py`: 7/7 passed on TPU v6e.
- All 14 supported ablation-switch combinations match the naive reference in interpret mode and on TPU v6e.
- Applying the Neumann factor chain directly to the RHS has a maximum error of `2.77e-4` against the fp32 oracle (`atol=5e-4`), versus `6.9e-4` for bf16 invert-then-multiply.

## Benchmarking and Profiling

Raw data and figure sources: [kda-tpu-bench/results](https://github.com/Junyi-99/kda-tpu-bench/tree/main/results).

### Protocol and variants

- `original`: all five kernel switches off; `mxu_port`: `safe_gate` only; `optimized`: all five switches on.
- The primary grid, ShareGPT, and utilization benchmarks use one TPU v6e chip, `jax==0.10.2`, BF16 with fp32 strip-GEMM accumulation, and Kimi-K3 per-chip geometry (`H=24`, `K=V=128`, `BT=64`, `lower_bound=-5`). Each point is the median of 100 iterations after 20 warmups; compilation is excluded. The waterfall's middle workload is the labeled `H=16` packed case.
- `S` is the uncached length of each sequence. `B` sequences are packed with `cu_seqlens` into one `[1, T, H, D]` call, where `T=B×S`; there is no tensor batch dimension.

### Primary kernel benchmark

On the core `S=8…1024`, `B=1…64` grid, `optimized` improves all 56 points by **5.14×–32.69×** (geometric mean **13.55×**). `mxu_port` alone gives **1.26×–3.48×** (geometric mean **2.34×**).

The heatmap below extends `S` to 16,384, for 84 points in total. Of those, 78 are measured and every one improves: **5.14×–33.05×**, geometric mean **13.93×** (`mxu_port` alone **1.25×–3.48×**, geometric mean **2.23×**); best point `S=2048, B=64`, 2237.58 → 67.70 ms. The remaining six exceed HBM, and the two cases are marked apart: `OOM` where neither variant fits, and `base OOM` where `original` and `mxu_port` fail to compile but `optimized` still runs. The latter three — `S=4096,B=64`, `S=8192,B=32` and `S=16384,B=16`, all 262,144 tokens — need 51.73G of 31.25G HBM on the unmodified path, while `optimized` completes in 134.48 / 132.95 / 131.51 ms. Those three are the same token count reached by three different `(B, S)` splits and land within **1.1%** of each other.

Latency over the fixed-length grid

Speedup over the extended fixed-length grid

For the ShareGPT prompt-length distribution (52,229 tokenizer-exact lengths, pinned dataset revision and vLLM filter), 100 seed-0 batches per `B` are shared across variants. Geometric-mean speedup rises from **7.88×** at `B=1` to **31.16×** at `B=64`; the overall geometric mean is **14.34×**.

The cumulative ablation is workload-dependent: `safe_gate` contributes most for a short single sequence, while `flat_grid` contributes most at high concurrency.

Cumulative ablation on three representative workloads

Hardware counters show the optimized path raising mean MXU busy utilization from **16.4%** (`mxu_port`) to **38.2%**.

Instruction mix and hardware-counter utilization

### Batch and sequence scaling (`H=8` follow-up)

These figures are from a separate single-chip sweep with `H=8`, `K=V=128`, `BT=64`, and `lower_bound=-5`. They report the mean of 10 iterations after 3 warmups. Use them for scaling trends, not absolute comparisons with the `H=24` results above.

At `B=1`, latency scales approximately as `T^0.89` for `original` and `T^0.57` for `optimized`.

Sequence-length scaling at B=1

With a fixed 8,192-token budget, splitting one sequence into 16 increases latency by **1.67×** for `original`, **2.27×** for `mxu_port`, and **1.27×** for `optimized`.

Constant token budget

At `S=8,192`, per-sequence cost from `B=1` to `B=64` rises **3.2×** for `original`, **5.9×** for `mxu_port`, and only **1.17×** for `optimized`. Eight reproducible `B` values are 19–26% slower than their neighbors; the cause is still unknown.

Per-sequence cost at S=8192

Raw logs and analysis: [`results/bs-sweep/`](https://github.com/Junyi-99/kda-tpu-bench/tree/main/results/bs-sweep).

### End-to-end serving: Kimi-Linear-48B

Real weights, TP=4 on TPU v6e-4; ShareGPT, 160 prompts, concurrency 32. Kimi-Linear has an unbounded gate, so only `fuse`, `unified_layout`, and `flat_grid` apply; `safe_gate` and `head_block` do not.

| Metric | Upstream KDA | This PR | Speedup |
|---|---:|---:|---:|
| TTFT p50 | 244 ms | **164 ms** | **1.49×** |
| TPOT p50 | 36.7 ms | **28.9 ms** | **1.27×** |
| Wall time, 160 requests | 45.2 s | **36.7 s** | **1.23×** |

`_forward_decode` is unchanged; the TPOT improvement comes from reduced prefill interference. GSM8K accuracy is **0.94** over 200 examples.

### End-to-end serving: mini-K3

Same protocol on a 12-layer mini-K3 with K3-faithful geometry and dummy weights. Its bounded gate (`gate_lower_bound=-5`) enables all five switches.

| Metric | Upstream KDA | This PR | Speedup |
|---|---:|---:|---:|
| TTFT p50 | 2,816 ms | **989 ms** | **2.85×** |
| TPOT p50 | 36.4 ms | **15.4 ms** | **2.36×** |
| Wall time, 160 requests | 62.5 s | **33.6 s** | **1.86×** |

In a 30-second steady-state trace, the target KDA kernel falls from **59.3%** to **4.6%** of device time. Achieved BF16 rate rises from **20.9** to **52.2 TFLOP/s/chip**, and HBM bandwidth utilization rises from **27%** to **68%**; MoE becomes the dominant cost.

Raw data and A/B script: [`results/e2e/`](https://github.com/Junyi-99/kda-tpu-bench/tree/main/results/e2e).

## Reproduce

```bash
IMAGE=ghcr.io/junyi-99/kda-tpu-bench:latest
ARGS=(--rm --privileged --net=host -v "$(pwd)/out:/out")

# Kernel microbenchmark: 56-point S×B grid across original, mxu_port, and optimized.
docker run "${ARGS[@]}" "$IMAGE" grid

# The 84-point extended grid behind the heatmap (not wired into a mode; split by
# --seq-lens so a preemption costs one column). Cells above ~262k tokens OOM by design.
docker run "${ARGS[@]}" -it "$IMAGE" shell
#   PALLAS_INTERPRET=0 python bench/bench_kda_grid_wide.py --seq-lens 2048,4096,8192,16384
#   python bench/plot_grid_heatmap.py results/kda_grid.json ~/kda_grid_wide.json \
#     --seq-lens 8,16,32,64,128,256,512,1024,2048,4096,8192,16384 --out heatmap.png

# Kimi-Linear-48B serving A/B: this PR and the upstream KDA path, 160 ShareGPT prompts at concurrency 32.
docker run "${ARGS[@]}" -e E2E_MODEL=/dev/shm/kimi-linear-48b "$IMAGE" e2e

# mini-K3 serving, this PR: 160 ShareGPT prompts at concurrency 32.
docker run "${ARGS[@]}" -e MODEL=mini-k3 "$IMAGE" bench-serving

# mini-K3 serving, upstream KDA path: the matching baseline for the command above.
docker run "${ARGS[@]}" -e MODEL=mini-k3 -e KDA_FORCE_BASELINE=1 "$IMAGE" bench-serving

# Kimi-Linear-48B correctness: GSM8K accuracy over 200 examples.
docker run "${ARGS[@]}" -e MODEL=kimi-linear-48b "$IMAGE" run-eval

# Kernel attribution: instruction mix and MXU/VPU hardware counters for all three variants.
docker run "${ARGS[@]}" "$IMAGE" llo
```

## Checklist

### DIFF
--- a/python/sgl_jax/srt/kernels/kda/kda.py
+++ b/python/sgl_jax/srt/kernels/kda/kda.py
@@ -1,16 +1,6 @@
 # Adapted from https://github.com/primatrix/pallas-kernel (rev 3c691ad3)
 # Vendored to remove external dependency after the upstream repository went private.
-#
-# This file merges the following modules into a single file:
-#   - tops/utils.py (cdiv, align_up, pad_to_multiple, prepare_lens, prepare_chunk_indices, assert_shape, assert_shape_or_none)
-#   - tops/ops/utils.py (exp, exp2, get_interpret)
-#   - tops/ops/common/cumsum.py (chunk_local_cumsum_vector via _chunk_cumsum_kernel)
-#   - tops/ops/kda/chunk_intra_fwd.py (_solve_unit_lower_triangular, _kda_fwd_intra_kernel, kda_fwd_intra)
-#   - tops/ops/common/chunk_delta_h.py (_prepare_chunk_offsets, _chunk_gated_delta_rule_fwd_kernel, chunk_gated_delta_rule_fwd_h)
-#   - tops/ops/gla/chunk.py (_chunk_kda_fwd_o_gk_varlen_kernel renamed to _chunk_kda_fwd_o_gk_pl_kernel, chunk_kda_fwd_o_gk_varlen renamed to chunk_kda_fwd_o_gk)
-#   - tops/ops/kda/gate.py (kda_gate_chunk_cumsum, pallas_kda_gate_cumsum)
-#   - tops/ops/kda/chunk_fwd.py (_align_seqs, _unalign_output, chunk_kda_fwd)
-"""KDA chunked forward pass for variable-length sequences (self-contained)."""
+"""Core KDA kernels and the variable-length forward entry point."""
 
 from __future__ import annotations
 
@@ -24,6 +14,13 @@ import jax.experimental.pallas as pl
 import jax.numpy as jnp
 from jax.experimental.pallas import dslice
 from jax.experimental.pallas import tpu as pltpu
+
+from sgl_jax.srt.kernels.kda.v2.unified_layout import (
+    from_unified_layout,
+    prepare_intra_layout,
+    restore_intra_layout,
+    to_unified_layout,
+)
 
 # ============================================================================
 # Utilities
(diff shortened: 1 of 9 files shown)


JamesBrianD:
Thanks for the substantial optimization work. The PR provides detailed benchmark results, but I would also like to understand how these optimization directions were derived.

Could you add an optimization note describing the analysis process behind this work? In particular:
1. Which optimizations were inspired by existing papers, blogs, or implementations? Could you share the relevant links so we can read more and better understand them?
2. Which optimizations came from your own analysis of TPU hardware, profiling results, or the underlying math? Could you walk us through how you identified the bottlenecks and arrived at these changes?

A before/after pipeline diagram, the important intermediate tensor shapes and memory traffic, and an ablation table mapping each profiling observation to the corresponding optimization would make this much easier to review and maintain.

The current PR body explains the final implementation and performance results well; what is still missing is the reasoning process from the original profile to these five optimization decisions.

cjx0709:
@Junyi-99 hello, thanks for your contribution. Here I adopted a new KDA kernel: https://github.com/sgl-project/sglang-jax/pull/1554, maybe you can help to review and compare with this pr.

Junyi-99:
@JamesBrianD @cjx0709 Thanks for the feedback and for sharing #1554. I’ll take a closer look and also add more details on the optimization reasoning behind this PR.

Junyi-99:
@JamesBrianD @cjx0709 Thanks for the feedback and for pointing me to #1554. I have now reviewed its implementation and benchmarked both PRs on `TPU v6e`.

At a high level, the two PRs share largely the same optimization directions: MXU-oriented intra-chunk construction with a finite-Neumann solve, pipeline fusion, removal of repeated gather/transpose/scatter operations, multi-head-per-program execution, and elimination of the request-axis `N × total_chunks` overhead.

#1554 carries these directions further by fusing all stages into a single inference-only mega kernel, handling packed segment boundaries natively, and optionally using asynchronous DMA to overlap recurrent-state transfers with computation. For `T >= 8192`, it was `4.54–5.81×` faster than #1550 , making it the stronger implementation base.

PR #1552, opened 2026-08-26
fix(server): clamp chunked_prefill_size to max_prefill_tokens

### LINKED ISSUE #1551: [Bug] Scheduler crashes when chunked_prefill_size > max_prefill_tokens
## Checklist

- [x] I searched related issues and did not find a solution.
- [x] This bug is present in the latest revision tested.
- [x] Reproduction and environment details are included below.
- [x] This is a bug, not a usage question.
- [x] This issue is written in English.

## Describe the bug

With `--max-prefill-tokens 2048` and the default `--chunked-prefill-size 4096`, the first request longer than 2048 tokens terminates the scheduler:

```text
ValueError: No bucket >= 4096. Available: [64, 128, 256, 512, 1024, 2048].
```

Padding buckets are capped at `min(max_prefill_tokens, chunked_prefill_size)`, but chunk truncation uses `chunked_prefill_size` alone. The request therefore creates a 4096-token chunk that no bucket can hold.

Expected behavior: normalize or reject the configuration without terminating the scheduler. As a workaround, explicitly set `--chunked-prefill-size <= --max-prefill-tokens`.

## Reproduction

```bash
python -m sgl_jax.launch_server <model arguments> --max-prefill-tokens 2048

# After the server becomes healthy, send a prompt longer than 2048 tokens.
curl http://127.0.0.1:30040/generate \
  -d '{"text": "<4096-token prompt>", "sampling_params": {"max_new_tokens": 1}}'
```

The crash reproduced with two model configurations. Adding `--chunked-prefill-size 2048` served the same workload normally.

## Environment

- Hardware: TPU v6e-4
- sglang-jax revision: `32c7a4bb4ec99a0cdc270bf8f402a39811c7d1e7`
- `max_prefill_tokens`: 2048
- `chunked_prefill_size`: default (4096)

Junyi-99:
## Motivation

Fixes sgl-project/sglang-jax#1551.

When `max_prefill_tokens` is lower than `chunked_prefill_size`, a long prompt can create a prefill chunk with no matching padding bucket and terminate the scheduler.

## Modifications

Clamp `chunked_prefill_size` to `max_prefill_tokens` in `ServerArgs.__post_init__` when both are positive, and emit a warning. The default configuration and `chunked_prefill_size=-1` remain unchanged.

## Accuracy Tests

This change does not modify model or kernel code. The clamped path is equivalent to explicitly setting `--chunked-prefill-size 2048`.

```bash
python3 -m py_compile python/sgl_jax/srt/server_args.py
```

Verified on TPU v6e-4: a 4096-token request that previously crashed with `--max-prefill-tokens 2048` completes after clamping; prompts of 4096 and 8192 tokens were included.

## Benchmarking and Profiling

Not applicable; this change only normalizes server configuration.

## Checklist

- [x] (Optional) The necessary documentation update. No documentation change is needed.

### DIFF
--- a/python/sgl_jax/srt/server_args.py
+++ b/python/sgl_jax/srt/server_args.py
@@ -340,6 +340,15 @@ class ServerArgs:
         # Set chunked prefill size
         if self.chunked_prefill_size is None:
             self.chunked_prefill_size = 4096
+        if 0 < self.max_prefill_tokens < self.chunked_prefill_size:
+            logger.warning(
+                "chunked_prefill_size (%d) > max_prefill_tokens (%d); clamping "
+                "chunked_prefill_size to %d so prefill chunks fit the padding buckets.",
+                self.chunked_prefill_size,
+                self.max_prefill_tokens,
+                self.max_prefill_tokens,
+            )
+            self.chunked_prefill_size = self.max_prefill_tokens
 
         # GGUF
         if (self.load_format == "auto" or self.load_format == "gguf") and check_gguf_file(
--- a/python/sgl_jax/srt/server_args.py
+++ b/python/sgl_jax/srt/server_args.py
@@ -340,15 +340,24 @@ class ServerArgs:
         # Set chunked prefill size
         if self.chunked_prefill_size is None:
             self.chunked_prefill_size = 4096
+
         if 0 < self.max_prefill_tokens < self.chunked_prefill_size:
+            clamped_size = self.max_prefill_tokens // self.page_size * self.page_size
+
+            if clamped_size <= 0:
+                raise ValueError(
+                    f"max_prefill_tokens ({self.max_prefill_tokens}) must be at least "
+                    f"page_size ({self.page_size}) when chunked prefill is enabled."
+                )
+
             logger.warning(
                 "chunked_prefill_size (%d) > max_prefill_tokens (%d); clamping "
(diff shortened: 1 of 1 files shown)


python/sgl_jax/srt/server_args.py
```diff
@@ -340,6 +340,15 @@ def __post_init__(self):
...
             self.chunked_prefill_size = 4096
+        if 0 < self.max_prefill_tokens < self.chunked_prefill_size:
+            logger.warning(
+                "chunked_prefill_size (%d) > max_prefill_tokens (%d); clamping "
+                "chunked_prefill_size to %d so prefill chunks fit the padding buckets.",
+                self.chunked_prefill_size,
+                self.max_prefill_tokens,
+                self.max_prefill_tokens,
+            )
+            self.chunked_prefill_size = self.max_prefill_tokens
```
cjx0709: Could we keep the clamped value divisible by page_size, since max_prefill_tokens=2050 with page_size=128 currently becomes 2050 and then fails check_server_args(), although 2048 would be safe?

Junyi-99: Thanks for pointing this out. I’ll clamp `chunked_prefill_size` to the largest page-aligned value not exceeding `max_prefill_tokens`

cjx0709 (approved):

PR #1554, opened 2026-08-27
perf(kda): add guarded Mega prefill kernel on TPU

cjx0709:
## Motivation

The existing KDA prefill path launches a multi-stage chunked pipeline. This PR adds an inference-only Mega KDA Pallas kernel for both K3's bounded gate and Kimi-Linear's unbounded gate.

The kernel is adapted from a private upstream Pallas-kernel repository. The upstream contact is `pathfinder-pf`.

## Modifications

- Add Mega KDA under `kernels/kda/mega_kda/` without changing training or backward paths.
- Share packed segment handling, gate cumsum, state recurrence, and output fusion between both gate families.
- Keep a static solver specialization: bounded K3 uses the fast Neumann composition; unbounded Kimi-Linear uses overflow-safe decay construction and ordered block substitution.
- Restart the gate prefix at packed boundaries for both gate families so a segment starting inside a tile preserves its own non-zero initial state.
- Use Mega KDA by default for both K3 and Kimi-Linear prefill.
- Fall back to chunked KDA when a 64-token tile intersects more than two live requests.
- Keep `SGLANG_JAX_KDA_PREFILL_KERNEL=chunked` as the explicit escape hatch.

## Accuracy

```bash
PALLAS_INTERPRET=1 JAX_PLATFORMS=cpu uv run --project python --extra cpu --with pytest \
  pytest -q python/sgl_jax/test/kernels/mega_kda_test.py
```

Result: `15 passed in 19.57s` on CPU interpret mode. The suite covers bounded and unbounded gates, BF16 and FP32 beta, extreme unbounded decay, packed request boundaries, padding, layout fallback, and state chaining.

The packed-boundary regression uses one `T=128` input with lengths `(37, 91)`, a bounded gate, and a non-zero initial state for the second segment. It replaces a generic packed parameter, so this PR does not increase the kernel-test invocation count. Before the fix, the focused comparison fails with a maximum absolute output difference of `0.00213`; after the fix it passes at `rtol=2e-2, atol=2e-4`.

The existing backend integration fixtures now construct beta the same way as Kimi-Linear (`sigmoid(...float())`) instead of feeding signed random values. No model-level E2E or nightly test is added to CI.

On TPU v7x, direct unbounded comparisons against chunked KDA with production-like FP32 sigmoid beta covered 64, 128, packed, and 4096 tokens. The aggregate maximum absolute differences were:

- output: `0.0000610352`
- final state: `0.00013836`

## Kernel benchmark

```bash
PYTHONPATH=python python benchmark/kernels/kda/bench_mega_kda.py \
  --tokens 4096 --heads 16 --segments 1 --gate-mode unbounded \
  --warmups 10 --iterations 50 --rounds 2

PYTHONPATH=python python benchmark/kernels/kda/bench_mega_kda.py \
  --tokens 4096 --heads 16 --segments 4 --gate-mode unbounded \
  --warmups 10 --iterations 50 --rounds 2
```

Use `--gate-mode bounded` for the K3 specialization. Values below are the average of the two round medians on TPU v7x:

| Gate | Packed segments | Chunked | Mega KDA | Speedup |
|---|---:|---:|---:|---:|
| bounded | 1 | 16.275 ms | 0.409 ms | 39.8x |
| bounded | 4 | 19.078 ms | 0.418 ms | 45.6x |
| unbounded | 1 | 16.094 ms | 2.773 ms | 5.81x |
| unbounded | 4 | 18.847 ms | 2.768 ms | 6.81x |

Packed segments validate independent request boundaries; they are not sequence-parallel chunks.

## Kimi-Linear end-to-end benchmark

Start the default Mega server:

```bash
unset SGLANG_JAX_KDA_PREFILL_KERNEL
  --model-path moonshotai/Kimi-Linear-48B-A3B-Instruct \
  --trust-remote-code --device tpu --tp-size 8 --dp-size 1 \
  --skip-server-warmup --mem-fraction-static 0.85 \
  --max-prefill-tokens 4096 --chunked-prefill-size 4096 \
  --max-running-requests 8 --dtype bfloat16 \
  --attention-backend fa --page-size 256 --disable-radix-cache \
  --host 0.0.0.0 --port 30040
```

Run the workload:

```bash
python -m sgl_jax.bench_serving \
  --backend sgl-jax --host 127.0.0.1 --port 30040 \
  --model moonshotai/Kimi-Linear-48B-A3B-Instruct \
  --tokenizer moonshotai/Kimi-Linear-48B-A3B-Instruct \
  --dataset-name random --random-input-len 4096 --random-output-len 1 \
  --random-range-ratio 1 --num-prompts 8 --max-concurrency 1 \
  --request-rate inf --warmup-requests 2 --tokenize-prompt
```

For the fallback comparison, restart the same server after setting:

```bash
export SGLANG_JAX_KDA_PREFILL_KERNEL=chunked
```

Results for 8 fixed 4096-token prompts, concurrency 1, and one output token:

| Kernel | Median TTFT | Total token throughput |
|---|---:|---:|
| default Mega | 185.25 ms | 21,960.59 tok/s |
| chunked | 283.32 ms | 14,371.58 tok/s |

Mega reduces median TTFT by `34.6%` and increases total token throughput by `52.8%` in this prefill-focused workload.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/kernels/kda/mega_kda/kernel.py
+++ b/python/sgl_jax/srt/kernels/kda/mega_kda/kernel.py
@@ -84,6 +84,201 @@ def _build_chunk_metadata(segment_ids, chunk_size):
     return chunk_kind, seg_first, seg_last, seg_id
 
 
+def _build_unbounded_intra_terms(q, k, g_cumsum, beta, scale, block_size, precision):
+    """Build causal Aqk/L terms without positive gate exponentials.
+
+    Kimi-Linear's unbounded gate can accumulate far below the BF16/FP32
+    exponent range inside one 64-token tile.  Cross-block products use the
+    first gate value of the row block as a shared reference, which keeps both
+    factors in ``[0, 1]``.  The diagonal block uses the causal gate difference
+    directly because no single shared reference is safe on both sides of its
+    diagonal.
+    """
+    _, tokens, _ = q.shape
+    num_blocks = tokens // block_size
+    aqk_rows = []
+    l_rows = []
+
+    def _dot(lhs, rhs):
+        return jax.lax.dot_general(
+            lhs,
+            rhs,
+            (((2,), (2,)), ((0,), (0,))),
+            precision=precision,
+            preferred_element_type=jnp.float32,
+        )
+
+    for row_block in range(num_blocks):
+        row_start = row_block * block_size
+        row_end = row_start + block_size
+        q_row = q[:, row_start:row_end]
+        k_row = k[:, row_start:row_end]
+        g_row = g_cumsum[:, row_start:row_end]
+        reference = g_row[:, :1]
+        row_decay = jnp.exp2(jnp.clip(g_row - reference, -126.0, 0.0))
+        aqk_blocks = []
+        l_blocks = []
+
+        for col_block in range(row_block):
+            col_start = col_block * block_size
+            col_end = col_start + block_size
+            k_col = k[:, col_start:col_end]
+            g_col = g_cumsum[:, col_start:col_end]
+            col_decay = jnp.exp2(jnp.clip(reference - g_col, -126.0, 0.0))
+            scaled_k_col = k_col * col_decay
+            aqk_blocks.append(_dot(q_row * row_decay, scaled_k_col))
+            l_blocks.append(_dot(k_row * row_decay, scaled_k_col))
+
+        gate_diff = g_row[:, :, None, :] - g_row[:, None, :, :]
+        diagonal_decay = jnp.exp2(jnp.clip(gate_diff, -126.0, 0.0))
+        aqk_diagonal = jnp.sum(
+            q_row[:, :, None, :] * diagonal_decay * k_row[:, None, :, :],
+            axis=-1,
+        )
+        l_diagonal = jnp.sum(
+            k_row[:, :, None, :] * diagonal_decay * k_row[:, None, :, :],
+            axis=-1,
+        )
+        row_iota = jax.lax.broadcasted_iota(jnp.int32, aqk_diagonal.shape, 1)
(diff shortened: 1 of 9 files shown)


python/sgl_jax/srt/kernels/kda/mega_kda/kernel.py
```diff
@@ -0,0 +1,1811 @@
...
+                keepdims=True,
+            )
+            g_cumsum = g_cumsum - first_segment_total * seg_B_mask[None, :, None]
+            Aqk, L, _ = _build_unbounded_intra_terms(
+                q, k, g_cumsum, beta, scale, BC, OUTPUT_PRECISION
+            )
+        else:
+            Aqk, L, _ = _build_bounded_intra_terms(
+                q, k, g_cumsum, beta, scale, BC, OUTPUT_PRECISION, safe_gate
+            )
```
Junyi-99, line 1008: the log-prefix restart at `:1001` is only applied for `lower_bound is None`, so B rows carry `+g_A_total` into `k_eg_beta` (:1018) and `b_qg` (:1131), and the `w @ h0_B / q @ h0_B` terms underflow to ~0.

Junyi-99 (commented):
there is a small bug in `_boundary()` (see inline comment at `:1005`), a segment that starts mid tile with a non-zero state seems to lose its `h0` contribution

I put a repro and a possible fix here, hope it helps: https://gist.github.com/Junyi-99/a44b942146a1a66ec851094aba6af4d8

It would also be nice to have a test for `bounded gate` with a `packed boundary` and a `non-zero initial state`.

python/sgl_jax/srt/kernels/kda/mega_kda/kernel.py
cjx0709: thanks for your comment, I solved it in 715cb552f, and add a acc test unit case.

JamesBrianD (approved):

PR #1555, opened 2026-08-27
feat(scheduler): log per-DP queue-wait time at prefill admission

Yajingleo:
## Motivation

`Req.queue_time_end` was declared but never assigned, so the time a request spends waiting for admission was measured nowhere. With DP that wait is worth having per-rank: `#queue-req` shows the queue is deep, but not *whose* queue is deep or how long a request actually sits in it.

This is the request-scheduling analogue of Linux run-queue latency — time spent runnable but not yet running, reported per queue (here a DP rank rather than a CPU), which is what exposes an imbalanced load balancer.

## Modifications

- Add `record_queue_wait_times()` to `scheduler_metrics_mixin.py`: closes the timing for each admitted request and groups the waits by DP rank. `queue_time_end` is set only once, so requests re-admitted after retraction are not double-counted.
- Call it from `log_prefill_stats` and append `queue-wait per DP (ms)` with `avg`/`max` per rank, next to the existing `#prefill per DP` / `#running per DP`. Only logged when `dp_size > 1`; ranks that admitted nothing log `None`.
- Move the `Req` / `ScheduleBatch` / `PrefillAdder` imports under `TYPE_CHECKING` so the module can be imported without pulling in the whole scheduler — needed for the unit test.

Sample line (dp_size=2):

```
... #prefill per DP: [4, 2], #running per DP: [7, 5],
queue-wait per DP (ms): [{'avg': 18.4, 'max': 31.2}, {'avg': 204.7, 'max': 311.9}], #queue-req: 9,
```

One caveat worth knowing, since it isn't visible in the diff: `queue_time_start` has two pre-existing start points. The common path stamps it on append to `waiting_queue` (runnable, unadmitted) — true run-queue latency. The grammar path stamps it on entry to `grammar_queue`, where the request is still blocked on compilation, and that span carries through unchanged. For grammar-constrained requests the number is therefore total time-to-admission, i.e. an upper bound. This PR does not change those start points.

## Accuracy Tests

Not applicable — logging only; no scheduling decision or model output changes.

## Benchmarking and Profiling

Not applicable. One `perf_counter()` read and a dict append per admitted request, in a path that already formats a log line.

## Test Plan

```
cd python && PYTHONPATH=. python -m pytest sgl_jax/test/managers/test_scheduler_metrics.py -q
```

`3 passed` — covers per-DP grouping, the retracted/unstarted skip cases, and the rendered log line.

## Checklist

https://claude.ai/code/session_01LQvxU2ovvhC5hujdjUqNiZ


JamesBrianD:
I agree that queue time is a useful signal to record, especially for diagnosing DP load imbalance. However, I’m not sure it belongs in the per-prefill batch info log.

The batch-local `avg`/`max` values are difficult to aggregate, query, or alert on over time, and the log becomes quite verbose as the DP size grows. A metrics histogram labeled by DP rank would be a better long-term interface and would allow us to inspect percentiles such as p50/p95/p99.

This would also align with upstream SGLang, which exports queue latency through the `sglang:queue_time_seconds` histogram. Could we record this as a metric instead of extending every prefill batch log line?

Yajingleo:
Thanks for the suggestion. I replaced the per-prefill batch log output with a Prometheus histogram named `sglang:queue_time_seconds`, labeled by `dp_rank`. The metric name and bucket boundaries mirror upstream SGLang's queue-time histogram, and each request is observed only on its first prefill admission so retracted requests are not double-counted. I also updated the focused unit tests; they pass locally (2 passed). The change is in a57b4a29.

JamesBrianD:
Thanks for switching this to a per-DP histogram. The direction looks right, but there is one export issue with `--enable-single-process --enable-metrics`.

`scheduler_metrics_mixin.py` imports `prometheus_client` before `http_server.launch()` sets `PROMETHEUS_MULTIPROC_DIR`. The client chooses its storage backend at import time, so in single-process mode the histogram stays in the normal in-process registry, while `/metrics` only reads the multiprocess registry. The code records the sample, but the metric is missing from the endpoint.

Could you move the Histogram import/creation until after Prometheus has been initialized, or otherwise make the registry choice consistent for single-process mode? It would also be good to add a test that records one queue sample and then scrapes the same registry used by `/metrics`; the current mock test does not cover the export path.

The grammar timing difference can remain as the #1556 follow-up.

Yajingleo:
Thanks for the detailed reproduction. I reproduced the import-order issue and have a fix ready locally.

The fix removes the module-level `Histogram` import and creates the queue histogram only during scheduler metrics initialization, after `PROMETHEUS_MULTIPROC_DIR` has been configured. The `/metrics` setup also selects the registry matching prometheus-client's active value backend, so an embedder that imported prometheus-client early does not hide in-process metrics behind a `MultiProcessCollector`.

I added an isolated fresh-interpreter regression that covers both launch shapes (single-process and spawned scheduler). It explicitly verifies that importing `scheduler_metrics_mixin` does not import `prometheus_client`, sets `PROMETHEUS_MULTIPROC_DIR`, confirms the mmap-backed value class is selected, observes a DP-labelled queue sample, and scrapes the actual `/metrics` ASGI endpoint to assert the histogram count is present. The focused suite passes (`3 passed`).

I'll push the implementation next.

Yajingleo:
Pushed the fix and regression coverage in `01c25d2d`. The focused test passes in both single-process and spawned modes (`3 passed`).

python/sgl_jax/test/managers/test_scheduler_metrics.py
```diff
@@ -0,0 +1,105 @@
...
+    metric = Mock()
+
+    record_queue_wait_times([already_recorded, no_start], metric, now=10.0)
+
+    metric.labels.assert_not_called()
+    assert already_recorded.queue_time_end == 2.0
+    assert no_start.queue_time_end is None
+
+
+def test_queue_time_histogram_is_exported_in_single_and_spawned_modes(tmp_path):
```
JamesBrianD: Please remove this test file. The nested subprocess scripts are too difficult to maintain for the coverage they provide.

python/sgl_jax/srt/utils/common_utils.py
```diff
@@ -210,12 +210,26 @@ def set_prometheus_multiproc_dir():
...
-    from prometheus_client import CollectorRegistry, make_asgi_app, multiprocess
-    from starlette.routing import Mount
+def get_prometheus_registry():
+    from prometheus_client import REGISTRY, CollectorRegistry, multiprocess, values
+
+    # prometheus_client fixes ValueClass at first import. Normally launch() sets
+    # PROMETHEUS_MULTIPROC_DIR before that import, but embedders may have already
+    # imported the client. Scrape the registry matching the selected backend so
+    # in-process metrics are never hidden behind a MultiProcessCollector.
+    if values.ValueClass.__name__ == "MutexValue":
```
JamesBrianD: Please remove this helper and keep the existing registry setup. Checking a private `ValueClass` name is brittle and still does not handle spawned workers correctly.

Yajingleo: Done.

python/sgl_jax/test/managers/test_scheduler_metrics.py
Yajingleo: Done

Yajingleo:
Thanks for the careful review! Comments are addressed. Added `set_prometheus_multiproc_dir` upfront in the main before importing prometheus client.

JamesBrianD (approved):

PR #1548, opened 2026-08-26
feat(vlm): Add in-model multimodal runtime and Qwen2.5-VL implementation

Yanko-7:
## Motivation

This is the second part of the Qwen2.5-VL upstream landing and depends on #1544.

The request contract and media-aware cache identity are provided by the previous PR. This PR adds the model-agnostic in-model multimodal runtime and the Qwen2.5-VL implementation, while intentionally keeping the model unreachable from the serving path.

Separating the runtime/model implementation from serving activation allows the model code, vision sharding, and embedding orchestration to be reviewed without changing Scheduler or ModelRunner behavior.

This consolidates the final implementation developed in #1433, #1444, and #1480.

## Modifications

- Add `InModelMultimodalContract`.
- Add multimodal lane packing and host orchestration.
- Add the paged multimodal embedding pool.
- Add vision DP/TP sharding utilities.
- Add variable-length vision Flash Attention support.
- Add Qwen2.5-VL:
  - Vision transformer.
  - Image and video encoding.
  - MRoPE support.
  - DeepStack embedding support.
  - Weight loading and sharding.
- Add embedding pool and vision attention tests.
- Preserve the existing text-only Qwen2 path when multimodal positions are absent.

### Dormant-by-design boundary

This PR intentionally does not export:

```python
EntryClass = Qwen2_5_VLForConditionalGeneration
```

Therefore Qwen2.5-VL is not registered by `ModelRegistry` and cannot be selected or reached by the serving path in this PR.

Scheduler, ForwardBatch, ModelRunner, HBM accounting, and serving arguments are not activated until the following PR.

### Compatibility

- Vision-specific attention and sharding are only called by the new VLM runtime.
- Existing serving models cannot reach the new runtime.
- Generic preprocessing performance changes from #1538/#1542 are excluded.

## Accuracy Tests

- [x] Paged multimodal embedding pool behavior and placement.
- [x] Variable-length vision Flash Attention numerical accuracy.
- [x] RoPE/MRoPE float32 phase computation.

Results:

- Embedding pool: 10 tests passed.
- Vision Flash Attention numerical comparison: 1 test passed.
- RoPE/MRoPE precision regression: 4 test cases passed.

## Benchmarking and Profiling

No serving performance claim is made in this PR because the runtime remains dormant.
The follow-up serving PR will report end-to-end compilation, memory, and latency results.

## Test Plan

Commit-level pre-commit hooks passed locally.

```bash
python -m pytest -q \
[...]

PR #1558, opened 2026-08-27
feat(vlm): add Qwen2.5-VL support and cookbook

Yanko-7:
## Motivation

This is the final serving landing for Qwen2.5-VL and builds on the request/cache contract from #1544 and the in-model multimodal runtime from #1548.

The earlier PRs intentionally left `Qwen2_5_VLForConditionalGeneration` unreachable from the serving path. This PR activates that implementation end to end and includes the preprocessing work needed to benchmark the final request path rather than a temporary synchronous baseline.

The functional serving integration rebuilds #1480. The non-blocking preprocessing path rebuilds #1538 and #1542.

Additionally, I have added the cookbook for Qwen2.5-VL.

### Serving activation

- Register `Qwen2_5_VLForConditionalGeneration` through `EntryClass` and detect in-model multimodal capability through `ModelRegistry`.
- Mark supported Hugging Face architectures as multimodal.
- Build multimodal work only for placeholder ranges visible in the current prefill chunk.
- Carry the host-side multimodal batch through `ScheduleBatch`, `ModelWorkerBatch`, `ForwardBatch`, and `ModelRunner`.
- Run vision encoding and embedding merge before the language backbone.
- Allocate and account for the multimodal embedding pool.
- Add vision precompile buckets and DP/TP encoder configuration.
- Warm up multimodal embedding and packed-write shapes.
- Add scheduler, compilation, embedding-pool, and ModelRunner regression coverage.
- Update the multimodal architecture documentation.

### Parallel preprocessing

- Move image/video loading and Hugging Face processor work off the asyncio event loop.
- Load images and videos concurrently through bounded I/O workers.
- Run processor work in a bounded executor with one processor instance per worker.
- Default Qwen-VL to 4 I/O workers and 2 processor workers; expose `--mm-io-worker-num` and `--mm-processor-worker-num` for explicit tuning.
- Fall back to one processor worker when a processor cannot be cloned or does not declare concurrency support.
- Avoid redundant multimodal chat-template encode/decode work.
- Shut down multimodal executors with the tokenizer manager and HTTP server lifecycle.
- Add processor-threading, concurrent media-loading, and OpenAI chat-path regression tests.

### Activation and compatibility guards

The in-model serving path is enabled only when both conditions hold:

1. The resolved model implements `InModelMultimodalContract`.
2. The request contains typed `MultimodalInputs`.

Consequently:

- Text-only requests do not build a multimodal batch.
- Non-VLM models do not allocate a multimodal embedding pool.
- Existing model compilation paths do not receive vision buckets.
- Existing Radix Cache behavior is preserved when no multimodal cache identity is present.
- Processor parallelism remains bounded and configurable.

## Accuracy Tests

| Item | Value |
| --- | --- |
| Model | `Qwen/Qwen2.5-VL-32B-Instruct` |
| Hardware | Single-host TPU v7x-8, topology `2x2x1` |
| Parallelism | DP4 x effective TP2 (`--tp-size 8 --dp-size 4`) |
| Precision | BF16 model, vision tower, and KV cache |
| Context / max sequence | 32768 / 32768 |
| Evaluator | EvalScope 1.5.1, OpenAI-compatible API |
| Generation | `max_tokens=8192`, temperature 0, seed 42 |
| Evaluation batch size | 24 |

| Dataset | Metric | Result | Samples |
| --- | --- | ---: | ---: |
| MMMU Val | Mean accuracy | **62.67%** | 900 |
| MMMU-Pro Vision | Mean accuracy | **47.75%** | 1,730 |

### Setup

The run uses `Qwen/Qwen2.5-VL-32B-Instruct` on a single TPU v7x-8 (`2x2x1`), DP4 x effective TP2, with BF16 model/vision/KV and no quantization. It sets `--mm-io-worker-num 4 --mm-processor-worker-num 16`, performs one warmup request, prewarms the multimodal shape, and flushes cache before each case.

The calibrated text control averages 1,422 LM-visible input tokens. The multimodal request uses 1,024 random source-text tokens plus one random 512x512 JPEG; after chat templating, it averages 1,097.23 text tokens and 326.00 vision tokens, or 1,423.24 total. The two workloads therefore differ by only 1.24 input tokens per request.

### Results

The unbounded run uses 1,000 requests and 500 output tokens per request.

| Metric | Text | Multimodal | MM vs text |
| --- | ---: | ---: | ---: |
| Mean TTFT | 16,276.11 ms | 19,145.39 ms | **+2,869.29 ms / +17.63%** |
| Median TTFT | 12,036.01 ms | 15,025.80 ms | **+2,989.79 ms / +24.84%** |
| P99 TTFT | 50,698.61 ms | 55,411.70 ms | **+4,713.09 ms / +9.30%** |
| Duration | 61.600 s | 66.641 s | **+8.18%** |
| Request throughput | 16.234 req/s | 15.006 req/s | **-7.57%** |
| Input token throughput | 23,084.55 tok/s | 21,356.74 tok/s | **-7.48%** |
| Output token throughput | 8,116.93 tok/s | 7,502.88 tok/s | **-7.57%** |
| Mean TPOT | 66.67 ms | 69.91 ms | **+4.85%** |
| Median TPOT | 70.89 ms | 74.01 ms | **+4.40%** |
| P99 TPOT | 88.94 ms | 99.35 ms | **+11.70%** |
| Median E2E latency | 47,422.30 ms | 52,066.95 ms | **+9.79%** |
| P99 E2E latency | 61,443.00 ms | 66,011.26 ms | **+7.43%** |

## Test Plan

Unit and integration tests:

```bash
python -m pytest -q \
  python/sgl_jax/test/test_vlm_stage1_framework.py \
  python/sgl_jax/test/test_compilation_manager.py \
  python/sgl_jax/test/test_model_runner_kv_cache_mixin.py \
  python/sgl_jax/test/multimodal/test_qwen_vl_processor.py \
  test/srt/openai_server/basic/test_serving_chat.py
```

Required TPU serving validation:

- [x] DP=1 image request.
- [x] DP=1 video request.
- [x] DP=2 concurrent image requests.
- [x] Mixed text-only and multimodal requests.
- [x] Radix Cache hit and miss with different images.
- [x] Chunked/retracted prefill covering a split placeholder.

## Checklist

- [x] The PR description is written in English.
- [x] The dependency and combined scope are documented.
- [x] Serving activation is capability- and request-gated.
- [x] Multimodal I/O and processor work are bounded and configurable.
- [x] Legacy runtime removal remains an independent follow-up.
- [x] Unit and integration tests passed (`88 passed, 7 skipped`).
- [x] Reference TPU accuracy results are included.
- [x] Reference TPU benchmark results are included.


python/sgl_jax/srt/entrypoints/engine.py
```diff
@@ -302,8 +302,14 @@ def rerank(
...
         """Shutdown the engine"""
+        tokenizer_manager = getattr(self, "tokenizer_manager", None)
+        if tokenizer_manager is not None:
+            tokenizer_manager.shutdown()
         kill_process_tree(os.getpid(), include_parent=False)
-        if self.server_args.enable_single_process:
+        if (
+            getattr(self, "server_args", None) is not None
+            and self.server_args.enable_single_process
+        ):
```
JamesBrianD: `getattr(self, "server_args", None) is not None` This syntax isn't very concise. See if you can confirm if `server args` exists, or save the `enable_single_process` attribute.

test/srt/multimodal/test_in_model_multimodal.py
```diff
@@ -0,0 +1,1106 @@
...
+    build_multimodal_batch,
+)
+from sgl_jax.srt.multimodal.in_model.interface import InModelMultimodalContract
+from sgl_jax.srt.multimodal.in_model.lane_packing import (
+    balance_lanes,
+    encoder_num_lanes,
+    pack_vision_inputs,
+    replicate_across_mesh,
+    run_mrope_vision_model,
+)
```
JamesBrianD, line 29: The test file is incorrectly named; it's not located under test/srt. Also, I noticed many test cases can be simplified or deleted.

python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -55,6 +58,33 @@
...
+    multimodal_model=None,
+) -> int:
+    """Per-device byte budget reserved for the multimodal embedding pool."""
+    enabled = (
+        getattr(model_config, "is_multimodal", False)
+        and ModelRegistry.is_in_model_multimodal(model_config.hf_config.architectures)
+        and not is_draft_worker
+        and not server_args.multimodal
+        and not server_args.enable_lora
+    )
```
JamesBrianD, line 80: Should open when disaggregation_mode != "decode"

python/sgl_jax/srt/entrypoints/engine.py
Yanko-7: Done. I placed the assignment of server args before `atexit.register(self.shutdown)` to ensure it definitely exists at shutdown time.

python/sgl_jax/srt/multimodal/processors/base_processor.py
```diff
@@ -65,11 +64,53 @@ def _normalize_image_source(source) -> bytes | str:
...
+        if self.mm_processor_worker_num > 1 and not self.supports_mm_processor_concurrency:
+            logger.warning(
+                "%s does not support concurrent multimodal processing; using one worker.",
+                type(self).__name__,
+            )
+            self.mm_processor_worker_num = 1
+        try:
+            self.mm_processor_executor = MultimodalProcessorExecutor(
+                processor, self.mm_processor_worker_num
+            )
```
JamesBrianD: Setting `mm_processor_worker_num` to 1 only serializes work submitted to `MultimodalProcessorExecutor`. For `max_workers == 1`, the executor uses the original `processor` object, while `apply_chat_template()` bypasses the executor and calls the same object directly on the event-loop thread.

As a result, request A may be running multimodal preprocessing on the executor thread while request B concurrently calls `apply_chat_template()` on the same processor. This violates the fallback's intended contract for processors that do not support concurrency or cannot be cloned.

Upstream SGLang avoids sharing the original processor with a background worker: it only creates the executor for multiple independently cloned processors; otherwise, it leaves the executor as `None` and runs preprocessing synchronously. Could we follow the same ownership model here, or otherwise ensure chat-template rendering and preprocessing cannot concurrently access the same processor?

python/sgl_jax/srt/configs/model_config.py
```diff
@@ -224,7 +224,9 @@ def __init__(
                 self.quantization_config.ignored_layers = ignored
         # Check model type
         self.is_generation = is_generation_model(self.hf_config.architectures, is_embedding)
-        self.is_multimodal = False
+        self.is_multimodal = any(
+            architecture in multimodal_model_archs for architecture in self.hf_config.architectures
```
lianfang0320, line 228: There are several similarly but different concepts: `model_config.is_multimodal,` `server_args.multimodal`, and `is_in_model_multimodal`. It would be confusing, please document the semantics.

Yanko-7: I added the explanation to 12-multimodal.md, but this is still not a very good practice. I hope to unify it later on. For example,
```python
class MultimodalRuntimeMode(Enum):
    TEXT
    LEGACY_SRT
    IN_MODEL_SRT
    MULTISTAGE
def resolve_multimodal_runtime(server_args, model_config):
    if server_args.multimodal:
        return MultimodalRuntimeMode.MULTISTAGE

    if model_config is None or not model_config.is_multimodal:
        return MultimodalRuntimeMode.TEXT

    if ModelRegistry.is_in_model_multimodal(
        model_config.hf_config.architectures
    ):
        return MultimodalRuntimeMode.IN_MODEL_SRT

    return MultimodalRuntimeMode.LEGACY_SRT
```

test/srt/multimodal/test_in_model_multimodal.py
Yanko-7: The redundant tests have been removed and the file has been renamed to `test in model multimodal.py`.

python/sgl_jax/srt/model_executor/model_runner.py
Yanko-7: Done.

python/sgl_jax/srt/multimodal/processors/base_processor.py
Yanko-7: Done.Keep consistent with SGLang.

docs/architecture/12-multimodal.md
```diff
@@ -2,11 +2,13 @@
...
+sglang-jax's standalone multimodal subsystem is an independent parallel architecture supporting text-to-image / video (Wan), vision-language understanding (Qwen2.5-VL), omni-modal (Qwen3-Omni), and audio (MiMo Audio) modalities. Unlike the single-Scheduler architecture used for text inference, multimodal uses a `GlobalScheduler` to orchestrate a multi-stage pipeline, with each stage owning its own Scheduler, Model Executor, and device mesh. The reason for adopting an independent multi-stage architecture rather than reusing the text engine is that multimodal tasks have computational patterns fundamentally different from autoregressive text generation: Diffusion requires fixed-step iterative denoising, ViT processes the full image patch grid in one pass, VAE needs large memory blocks for latent decoding — these stages have different scheduling strategies, memory management, and parallelism patterns, and cannot be unified into the text engine's prefill-decode loop.
```
JamesBrianD: Edit docs. Regular autoregressive in-modol path is new and valid now

Yanko-7: Added. 
"The newly added in-model path is now a supported execution path for regular autoregressive multimodal models through the standard SRT prefill-decode loop. "

JamesBrianD (dismissed):
Others LGTM!

JamesBrianD (approved):

PR #1559, opened 2026-08-28
model: Kimi-K3 support (MXFP4 LatentMoE, KDA/MLA interleave, SITU, AttnRes)

vlasenkoalexey:
## What

Adds **Kimi-K3** model support: `KimiK3ForCausalLM` / `KimiK3ForConditionalGeneration`, the
K3-specific layers (SITU activation, the AttnRes depth protocol), MXFP4 quantization support for
the 896-expert LatentMoE (weights + streaming loader + megablox fp4 GMM kernels), configs, weight
mappings, registry wiring, and a test suite.

27 files, +6,212/−14 — additive; existing code is barely touched.

## Architecture notes (K3 vs Kimi-Linear)

K3 is close enough to Kimi-Linear that subclassing looks free, and far enough that five things
break if assumed symmetric. All five are handled and pinned by checkpoint evidence in the tests:

1. **KDA output gate is full-rank** (`use_full_rank_gate`) — single `g_proj [12288, 7168]`;
   the **forget** gate stays low-rank (`f_a_proj`/`f_b_proj`).
2. **The MoE is a LatentMoE** — experts live in `routed_expert_hidden_size = 3584`, not the 7168
   hidden; router reads the full hidden; latent down-projection on the way out.
3. **MLA Q is LoRA-factored** (`q_a_proj [1536, 7168]` → `q_b_proj [18432, 1536]`).
4. **MLA has its own output gate.**
5. **`A_log` is padded to `head_dim` (128)**, with 96 (`num_heads`) meaningful entries — the
   loader must narrow, not reshape.

The released checkpoint is multimodal, so the text-weight prefix (`language_model.`) is detected
from shard key names rather than hardcoded.

## Validation

- **Semantics parity vs the reference implementation**: the K3-specific computations (SITU,
  AttnRes, A_log handling) are tested against functions extracted verbatim from Moonshot's torch
  reference and executed side by side (`test_kimi_k3_torch_parity.py`).
- **Plumbing equivalence at full depth**: pointing `architectures` at `KimiK3ForCausalLM` while
  loading Kimi-Linear-48B (every K3 config switch off) runs a real 27-layer model through the new
  decoder layer / model class / mappings / registry. Result: **5/5 prompts byte-identical
  `output_ids`** vs the existing Kimi-Linear class at temperature 0, coherent text. This separates
  plumbing bugs (proven here) from K3-semantics bugs (proven by the parity suite).
- **Real-weight integration**: a 4-layer truncated K3 config loads real MXFP4 weights and serves
  `/generate` on TPU v7x (4 chips). A depth truncation proves the integration surface (layer
  assembly, MXFP4 load, KDA/MLA interleave, AttnRes) — not output coherence, which requires full
  depth.
- **Registry guard**: K3's `text_config` declares `model_type: kimi_linear`, so config-based
  routing would silently load Kimi-Linear (no AttnRes, no SITU) and still produce fluent text.
  `test_kimi_k3_registry.py` pins top-level-architecture routing to prevent exactly that.
- **Test suite**: 12 test files, registered in the `unit-test-cpu` suite (CPU-safe: TPU- and
  checkpoint-dependent cases self-skip; torch parity uses `importorskip`). Local run on current
  main: **67 passed, 18 skipped**.

## Known limitation

Full-depth (93-layer) serving needs per-host weight staging of the 1.42 TiB checkpoint, which does
not fit a `tpu7x-standard-4t` host's tmpfs. Making full depth practical needs expert-sharded
byte-range weight streaming (each EP rank fetching only its experts) — planned as a follow-up;
this PR contains the complete model implementation and its verification.


cjx0709:
@vlasenkoalexey hello, thanks for you contribution. But since `vllm-torchtpu` has not been open sources, did you obtain the author's permission before adopting the code?

vlasenkoalexey:
Good point, let me double-check

vlasenkoalexey:
Ok, now confirmed. The only substantial private code in the PR was a vendored fp4 GMM package a fork of the tokamax kernel that was public until 2026-08-11, plus internal fp4/SITU/tuning additions that I did myself. It's been removed from the tip and from all branch history in both public repos, and replaced with a dequant-to-bf16 reference path. 

All private changes are now scraped from history, and I can either generate kernel from scratch or introduce one back once vLLM TorchTPU become public.

In current shape model won't longer run on 16 v7x chips (32 chiplets).

cjx0709:
> Ok, now confirmed. The only substantial private code in the PR was a vendored fp4 GMM package a fork of the tokamax kernel that was public until 2026-08-11, plus internal fp4/SITU/tuning additions that I did myself. It's been removed from the tip and from all branch history in both public repos, and replaced with a dequant-to-bf16 reference path.
> 
> All private changes are now scraped from history, and I can either generate kernel from scratch or introduce one back once vLLM TorchTPU become public.
> 
> In current shape model won't longer run on 16 v7x chips (32 chiplets).

Thanks for the update and for removing the private code. My understanding is that TorchTPU and related repositories, including vllm-torchtpu, may be open-sourced around October. If that timeline holds, it may be more practical to wait for the upstream FP4 LatentMoE kernel rather than spend time on a clean-room reimplementation beforehand.  
This is just a suggestion—please feel free to proceed with your preferred plan and keep us updated here.

PR #1509, opened 2026-08-04
perf: use in-place recurrent state cloning

yilunyu:
## Summary
- add a Pallas recurrent-state slot clone kernel with input/output aliasing
- use it for recurrent copy-on-write while preserving donated pool buffers
- add correctness/donation coverage and a TPU microbenchmark

PR #1560, opened 2026-08-28
perf: use in-place recurrent state cloning

JamesBrianD:
## Summary
- add a Pallas recurrent-state slot clone kernel with input/output aliasing
- use it for recurrent copy-on-write while preserving donated pool buffers
- add correctness/donation coverage and a TPU microbenchmark

## Context
This replaces #1509. Its existing workflow attempts request retired runner labels and never acquired a runner. The original four commits were cherry-picked onto the latest main without changing their authorship, so this PR receives a fresh pull_request event using the current CI runner labels.

Supersedes #1509.


JamesBrianD:
This is a latest-main replacement for #1509 so CI can use the current zonal runner labels. The original four commits and yilunyu authorship are preserved. GitHub only permits formal review requests for repository collaborators, so no repository access was added; @yilunyu, please review this replacement PR here.

PR #1561, opened 2026-08-28
feat: add Ling-3 serving with replicated MoE and recurrent radix cache

cjx0709:
## Purpose

This draft is a shared checkpoint for Ling-3.0 Tiny bring-up and follow-up testing. It is not being prepared for upstream merge yet.

## Scope

- add `BailingMoeV3` configuration and Ling-3 Tiny/Flash model definitions
- load the public `inclusionAI/Ling-3.0-tiny` Hugging Face checkpoint, including the hybrid KDA + MLA path
- run the v7x-8 target with attention DP=8 and EP=1
- keep all 128 routed experts fully replicated on every DP rank while retaining the existing `epmoe` backend interface
- support recurrent `UnifiedRadixCache`, overlap scheduling, and cache-first DP affinity for multi-turn Agent traffic
- integrate the Ling-3 reasoning parser and tool-call parser
- preserve the model-native long-context limit; no serving-side context-length override is added

## Runtime contract

- Hardware: one TPU v7x-8 slice (`2x2x1`, 8 devices)
- Parallelism: `--tp-size 8 --dp-size 8 --ep-size 1`
- MoE: `--moe-backend epmoe`, with Ling-3 Tiny experts fully replicated rather than EP-sharded
- Scheduling: `--dp-schedule-policy cache_aware`; an eligible rank holding the longest reusable prefix is preferred before load balancing
- Radix cache: unified recurrent radix tree with the recurrent extra buffer enabled
- Overlap scheduling: enabled by default; do not pass `--disable-overlap-schedule`
- Long context: do not pass `--context-length`; the checkpoint declares a native 131072-token context window
- Model source: Hugging Face, persisted once and mounted at `/models/Ling-3.0-tiny`

Replicated quantized experts are not implemented in this checkpoint.

## Prepare the model

Use the Hugging Face checkpoint, not ModelScope. For a local machine with a writable `/models` directory:

```bash
python -m pip install -U "huggingface_hub[cli]"

export MODEL_DIR=/models/Ling-3.0-tiny
mkdir -p "$MODEL_DIR"
hf download inclusionAI/Ling-3.0-tiny --local-dir "$MODEL_DIR"
```

For Falcon/GKE, upload this Hugging Face snapshot to persistent object storage once, then mount that bucket prefix read-only at `/models/Ling-3.0-tiny`. Subsequent tasks should start directly from the mounted path and must not download the checkpoint again.

## Known working TPU environment

From the repository root:

```bash
python3 -m pip install -q uv
uv venv --python 3.12 .venv-ling3
source .venv-ling3/bin/activate

uv pip install "jax[tpu]==0.10.2" \
  -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
uv pip install -e python
uv pip install --upgrade "libtpu==0.0.42.1" \
  -f https://storage.googleapis.com/jax-releases/libtpu_releases.html

export PYTHONPATH="$PWD/python${PYTHONPATH:+:$PYTHONPATH}"
```

## Start Ling-3.0 Tiny on v7x-8

```bash
export MODEL_DIR=/models/Ling-3.0-tiny

python -u -m sgl_jax.launch_server \
  --model-path "$MODEL_DIR" \
  --served-model-name ling3-tiny \
  --trust-remote-code \
  --reasoning-parser ling3 \
  --tool-call-parser ling3 \
  --device tpu \
  --tp-size 8 \
  --dp-size 8 \
  --ep-size 1 \
  --moe-backend epmoe \
  --attention-backend fa \
  --dp-schedule-policy cache_aware \
  --enable-cache-report \
  --enable-unified-radix-tree \
  --enable-recurrent-extra-buffer \
  --page-size 256 \
  --chunked-prefill-size 512 \
  --recurrent-track-interval 512 \
  --max-recurrent-state-size 256 \
  --max-running-requests 64 \
  --precompile-bs-paddings 8 16 32 64 \
  --mem-fraction-static 0.6 \
  --dtype bfloat16 \
  --random-seed 42 \
  --host 0.0.0.0 \
  --port 30000
```

There is intentionally no `--disable-overlap-schedule` and no serving-side `--context-length` argument in this command.

Check readiness and the effective server configuration:

```bash
curl -fsS http://127.0.0.1:30000/health
curl -fsS http://127.0.0.1:30000/get_server_info > /tmp/ling3-server-info.json
```

Thinking-mode OpenAI request example:

```bash
curl -sS http://127.0.0.1:30000/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "ling3-tiny",
    "messages": [{"role": "user", "content": "Solve the problem carefully and give the final answer."}],
    "temperature": 1.0,
    "top_p": 0.95,
    "top_k": 20,
    "max_tokens": 128000,
    "chat_template_kwargs": {"enable_thinking": true}
  }'
```

`max_tokens` above is a request-side completion budget, not a serving-side context cap. For real prompts, keep prompt plus completion within the model-native context window.

## Follow-up probes

Multi-turn Agent quality, Radix reuse, and DP-rank affinity:

```bash
python benchmark/hicache/bench_ling3_multiturn_agent.py \
  --server-url http://127.0.0.1:30000 \
  --model ling3-tiny \
  --sessions 32 \
  --parallel 32 \
  --max-tokens 4096 \
  --expected-dp-size 8 \
  --min-quality-rate 0.95 \
  --output-json /tmp/ling3-multiturn-agent.json
```

Concurrent prefix diversity and Radix correctness:

```bash
python benchmark/hicache/bench_ling3_radix_correctness.py \
  --server-url http://127.0.0.1:30000 \
  --parallel-list 8 32 64 \
  --families 8 \
  --branches 8 \
  --expected-dp-size 8 \
  --output-json /tmp/ling3-radix-correctness.json
```

Bounded GPQA Diamond thinking-mode check:

```bash
python test/srt/run_eval.py \
  --base-url http://127.0.0.1:30000 \
  --model ling3-tiny \
  --eval-name gpqa \
  --num-examples 32 \
  --num-threads 16 \
  --temperature 1.0 \
  --top-p 0.95 \
  --top-k 20 \
  --max-tokens 128000 \
  --enable-thinking true \
  --checkpoint-path /tmp/ling3-gpqa32-checkpoint.jsonl
```

The checkpoint is appended and flushed after every completed example. Re-run the same command with `--resume-unextracted-only` to reuse completed answers and schedule only missing or unextracted examples. Avoid using a 32768-token completion budget to draw quality conclusions for thinking mode: long reasoning may not reach its final answer within that budget. Remove `--num-examples 32` when a full evaluation is desired.


python/sgl_jax/test/test_bailing_moe_v3.py
```diff
@@ -0,0 +1,348 @@
...
+from sgl_jax.srt.layers.attention.linear.kda_backend import KDAAttnBackend
+from sgl_jax.srt.layers.moe import EPMoE
+from sgl_jax.srt.model_loader.arch import get_model_architecture
+from sgl_jax.srt.models.bailing_moe_v3 import (
+    BailingKDAAttention,
+    BailingMLA,
+    BailingMoeV3DecoderLayer,
+    BailingMoeV3ForCausalLM,
+)
+from sgl_jax.srt.utils.mesh_utils import create_device_mesh
```
JamesBrianD, line 28: Remove this test file?

JamesBrianD:
help to add dvfs flag in the commands

python/sgl_jax/srt/layers/moe.py
```diff
@@ -72,25 +79,46 @@ def __init__(
...
+                axis_names=("expert", "tensor"),
+                axis_types=(jax.sharding.AxisType.Explicit, jax.sharding.AxisType.Explicit),
+            )
+
+            abstract_mesh = self.mesh.abstract_mesh
+            self.updated_mesh = abstract_mesh.update(
+                axis_sizes=(self.ep_size, self.tp_size), axis_names=("expert", "tensor")
+            )
+            wi_sharding = P("expert", None, "tensor")
+            wo_sharding = P("expert", "tensor", None)
```
JamesBrianD, line 121: The original tp size here is the world size //ep size. We should change it so that the tp size can be set independently.

python/sgl_jax/srt/configs/bailing_moe_v3.py
```diff
@@ -0,0 +1,236 @@
...
+        topk_group=4,
+        topk_method="noaux_tc",
+        score_function="sigmoid",
+        scoring_func="sigmoid",
+        routed_scaling_factor=2.5,
+        router_dtype="fp32",
+        # Ling-3-Tiny is served with attention DP. Keeping the complete expert
+        # table on every DP rank avoids EP dispatch/all-to-all for this small
+        # model and lets each rank evaluate its local tokens independently.
+        replicate_moe=True,
```
JamesBrianD, line 62: Could we avoid making `replicate_moe=True` the model-family default? This is reasonable for the Ling-3 Tiny DP=8/EP=1 serving setup, but the same config class is also used for Ling-3 Flash. Since the upstream Flash config does not define this private field, it will inherit `True` and replicate all 512 experts on every rank, defeating TP/EP weight sharding and likely causing OOM. Please default this to `False` and enable replication explicitly only for the Tiny serving configuration (or through an explicit runtime option).

cjx0709:
@JamesBrianD Thanks for comments, it is still a draft status, with the replicated moe and force cache aware mode. I think I can split it into 3 prs in this week: 1. support ling3 tiny and plus; 2. support replicated moe; 3. support force cache aware mode.(or 1/2 into one pr).

cjx0709:
close this pr according to https://github.com/sgl-project/sglang-jax/issues/1565 this RFC

PR #1562, opened 2026-08-28
refactor(kv-cache): centralize MLA pool geometry

### LINKED ISSUE #1553: [Feature] Derive the MLA/DSA KV budget from get_kv_cache_shape instead of restating the layout in _compute_cell_size
### Checklist

### Motivation

Follow-up to #1532, from @zorrofox 's review, which suggested it as a follow-up to keep that PR focused.

`ModelRunner._compute_cell_size` (`model_runner_kv_cache_mixin.py:315`) hand-rolls the MLA v2 pool's allocation geometry — the 128-alignment, the `kv_packing = 32 // dtype_bits` page padding, and the per-token divide:

```python
kv_dim = align128(cfg.kv_lora_rank) + align128(cfg.qk_rope_head_dim)                        # :326
per_token = kv_dim * aligned_ps * dtype_size // self.page_size                              # :333
indexer_per_token = align128(indexer_key_dim) * aligned_ps * dtype_size // self.page_size   # :342
```

while the pool derives the same numbers from the kernel ABI. As zorrofox put it on the review, that was four independent copies of the same layout math that must agree byte-for-byte: the two budget terms above, the test helper `_allocated_bytes_per_token`, and the `_create_buffers` log line. #1532 closed two of them: `_buffer_bytes`/`get_kv_size_bytes` and the test helper both now go through `get_kv_cache_shape`. The budget is the copy that's left.

**The drift:** if the kernel packing or alignment rule changes, the allocation follows `get_kv_cache_shape` automatically and the budget does not. The pool then allocates more than the scheduler budgeted, and the excess comes out of the `1 - mem_fraction_static` activation reserve rather than out of KV, silently re-creating an over-allocation bug.

**A truncation wrinkle:** `// self.page_size` under-budgets fractionally when `page_size` isn't a multiple of the packing factor, exactly as described. For the GLM-5.2 latent width (640) under bf16: `page_size=3` budgets 1706 against 1706.67 allocated; `page_size=7`, 1462 against 1462.86; `page_size=33`, 1318 against 1318.79. Marginal, and such page sizes have other problems. The shared helper fixes it for free.

**A second way the same duplication bites,** found while checking the above: the two formulas agree exactly whenever `kv_dim` is already a multiple of 128, and disagree whenever it isn't. That condition holds today only because both sides pre-align their components. So a width reaching `_compute_cell_size` un-pre-aligned (a third latent component, or a future backend passing a raw `head_dim`) parts `align128(a) + align128(b)` from `align128(a + b)`, and the budget under-counts. Note this is a separate trigger from the packing change above: varying the packing alone does not break agreement (640 and 128 match under both float32 and bf16 at every power-of-two page size).

**Proposal**, per the review thread: a `bytes_per_token()` helper built on `get_kv_cache_shape`, with the call sites collapsing onto it. It has to be callable before the pool exists (`_compute_cell_size` runs while deciding how large to make the pool) so it wants to be a module-level function or a `classmethod` over `(page_size, dtype, kv_dim)` rather than an instance method. `_buffer_bytes` becomes a thin wrapper, and the test helper drops its private copy. Alignment and the page divide then happen in exactly one place.

Not urgent: no in-tree configuration hits any of these today. This is about making the budget and the allocation structurally unable to diverge, rather than merely aligned by hand.

### Related resources

[Original thread](https://github.com/sgl-project/sglang-jax/pull/1532#discussion_r3819047654)

JamesBrianD:
## Summary

- make `MLATokenToKVPool` the owner of latent MLA and DSA indexer physical geometry, following SGLang's pool-centered ownership
- expose one pre-allocation `profiled_bytes_per_token()` interface while keeping role-specific shape and alignment helpers private to the pool
- reuse the same pool-owned calculations for capacity sizing, allocation, resident-byte reporting, pytree restoration, and runtime buffer offsets
- preserve logical indexer widths across JAX pytree round trips while deriving aligned physical dimensions in one place
- reject unsupported hybrid-recurrent + `dsa_sparse` configurations at pool initialization with a clear error

## Motivation

This is a code-quality follow-up to #1532. That fix made DSA indexer cache memory visible to sizing and reporting, but the packing/alignment formula was still duplicated across the model runner and memory pool.

The allocation owner now provides the sizing coefficient directly, so layout changes cannot update allocation without also updating profiling. The implementation intentionally stays inside `MLATokenToKVPool` rather than introducing a standalone layout type for one attention family; DSA reuses the MLA latent pool geometry and adds an explicit indexer role, matching SGLang's ownership model.

This PR does not change the attention kernel or claim a TPU performance improvement.

Closes #1553.

## Tests

- `python -m pytest -q python/sgl_jax/test/test_model_runner_kv_cache_mixin.py` (14 passed)
- `python -m pytest -q python/sgl_jax/test/mem_cache/test_kv_cache.py python/sgl_jax/test/test_kv_cache_builder.py` (14 passed, 4 subtests passed)
- `pre-commit run --from-ref origin/main --to-ref HEAD` (all hooks passed)

`test_mla_attention.py` contains TPU-device-specific cases that call `pltpu.get_tpu_info()` and cannot run on the local CPU environment; no target-TPU performance or numerical claim is made here.


JamesBrianD:
Hi @tsvikas, could you please review this PR?

cjx0709 (approved):
LGTM

Issue #1565, opened 2026-08-31
[RFC] Split Ling-3.0 Tiny/Flash serving into three reviewable PRs


cjx0709:
## Summary

PR #1561 currently combines Ling-3.0 Tiny/Flash modeling, replicated MoE, recurrent Radix serving, cache-first DP scheduling, parsers, benchmarks, and unrelated eval changes. This RFC proposes landing the feature in three reviewable slices.

### PR 1: Ling-3.0 Tiny and Flash modeling

- Extend the existing `bailing_hybrid` config surface for Ling-3 KDA + MLA fields, matching the upstream SGLang layout.
- Add one shared `bailing_moe_v3.py` model implementation for Tiny and Flash.
- Support Tiny Q-LoRA, Flash flat Q projection, head-wise MLA gating, KDA lower-bound gating, Flash SwiGLU limits, checkpoint loading, and recurrent-state metadata.
- Add Ling-3 reasoning/tool parsing in a dedicated detector.
- Keep replicated MoE disabled by default and validate this slice with the existing EP/TP path.

### PR 2: Explicit replicated MoE

- Add an opt-in runtime setting; never enable replication as a model-family default.
- Make MoE TP size explicit and independent from EP/world size.
- Replicate expert weights across DP ranks while preserving tensor sharding and TP reduction.
- Initially require `ep_size == 1` for replicated mode; quantized replicated experts remain out of scope.
- Validate Ling-3.0 Tiny with `tp=8, dp=8, ep=1` and a second configuration with per-replica TP greater than one.

### PR 3: Forced cache-aware DP scheduling

- Add a separate `force_cache_aware` DP scheduling policy.
- Preserve the existing soft `cache_aware` behavior.
- Prefer the longest eligible cached prefix, break ties by load, and fall back to shape-aware scheduling on a full miss.
- Avoid adding scheduler diagnostics to the public OpenAI response schema unless a separate API decision is made.

## Validation

- Per-PR CPU/no-checkpoint unit tests and config/weight-mapping coverage.
- Ling-2.x/Kimi regression checks for shared config and KDA code.
- Use Falcon-managed GKE `v6e-1` capacity for PR 1 checkpoint startup and short correctness tests, reusing the shape of recent successful Falcon jobs after current cluster admission/routing checks.
- Add `LIBTPU_INIT_ARGS=--xla_tpu_dvfs_p_state=7` to TPU validation commands.
[...]

PR #1566, opened 2026-08-31
feat(model): add Ling 3 Tiny modeling

cjx0709:
## Motivation

Land PR 1 of #1565 as a reviewable Ling-3 modeling change. The original #1561 mixed modeling, replicated MoE, cache-aware DP scheduling, diagnostics, and unrelated eval changes.

This PR is scoped to Ling-3.0-tiny modeling and the minimum runtime support it needs. Ling-3.0-flash, replicated MoE, and forced cache-aware DP scheduling remain follow-up work.

## Modifications

- Extend `BailingHybridConfig` with the Tiny KDA/MLA layer schedule, safe-gate fields, head-wise MLA gating, and recurrent-state metadata.
- Add `BailingMoeV3ForCausalLM` with Tiny Q-LoRA checkpoint loading, hybrid KDA + MLA decoder layers, dense/shared/routed expert paths, head-wise MLA gates, and Ling-3 KDA lower-bound gating.
- Add a no-op pre-`o_proj` MLA extension point to DeepSeek V3; Ling-3 uses it for its head-wise gate while the shared MLA implementation retains ownership of `o_proj`.
- Propagate the KDA safe-gate configuration through extend/decode and the hybrid recurrent cache path.
- Add Ling-3 reasoning/tool-call parsing while preserving the existing GLM parser behavior.
- Support routed experts with EPMoE only in this Tiny PR; fused and fused-v2 MoE backends are intentionally rejected.
- Remove Flash-only flat-Q/clamp handling, one-off Ling-3 eval and tensor-dump hooks, and the large diagnostic component test from this PR.

### SGLang cookbook-aligned GSM8K

The primary accuracy check follows the current SGLang Ling-3.0-tiny cookbook recipe exactly: full GSM8K, zero-shot chat, thinking enabled, one sample per question, `temperature=1.0`, and `top_p=0.95`.

Environment:

- source commit: `4592bc09ae4769994db30112040b3def92cd905a`
- model: `inclusionAI/Ling-3.0-tiny`
- model revision: `b61f4338de3e68ffc9c0bc1ed5e902981a4a929e`
- evaluator: `sgl-eval` at `a231b7a439b235090ff7baa30778fa2b514309ae`
- runtime: JAX `0.10.2`, BF16, 1x TPU v6e, TP=1
- server parsers: `deepseek-r1` reasoning + `glm45` tool calls, matching the Tiny cookbook

```bash
sgl-eval run gsm8k \
  --base-url http://127.0.0.1:30000/v1 \
  --num-threads 32 \
  --temperature 1.0 \
  --top-p 0.95 \
  --thinking
```

| Runtime | Questions | Score | Stop rate | Truncated | Errors |
| SGLang-JAX, 1x TPU v6e, BF16 | 1,319 | **94.39%** | 100.00% | 0.00% | 0.00% |
| SGLang cookbook reference, 1x H200, BF16 | 1,319 | **94.01%** | - | - | - |

The TPU result is +0.38 percentage points relative to the published SGLang BF16 result under the same stochastic recipe. This does not show a dataset-level quality regression from the fused TPU KDA path. The SGLang reference is published in the [Ling-3.0-tiny cookbook benchmark config](https://github.com/sgl-project/sglang/blob/main/docs/src/snippets/configs/inclusionAI/ling-3.0-tiny-benchmarks.jsx).

### Additional MMLU sanity check

```bash
python3 test/srt/run_eval.py \
  --base-url http://127.0.0.1:30000 \
  --model ling3-tiny \
  --eval-name mmlu \
  --num-examples 1000 \
  --num-threads 16 \
  --temperature 0.0 \
  --max-tokens 8192 \
  --enable-thinking true
```

| Metric | Score |
|---|---:|
| Overall | **0.7110** |
| STEM | 0.8527 |
| Humanities | 0.5229 |
| Social sciences | 0.7685 |
| Other | 0.7846 |

These dataset evaluations validate the Tiny checkpoint. They were collected before the final Tiny-only structural refactor at `800afeddea4ab1ee5dfed7adc666f2c65ebe4539`; that refactor removes Flash-only paths and changes the MLA reuse boundary without changing Tiny's math. The final reviewer follow-up at `fe4b4de732799e1d07a4303bb2ea3836cc1d491f` removes the unvalidated fused-MoE mappings; the evaluated EPMoE execution path is unchanged.

### Targeted tests

```text
59 passed, 149 warnings in 262.97s
```

The Tiny-only structural commit `800afeddea4ab1ee5dfed7adc666f2c65ebe4539` was checked out by SHA and tested on JAX `0.10.2` with a single TPU v6e device.

Coverage includes the full KDA attention suite on TPU, the Bailing hybrid config/layer schedule, Tiny checkpoint mapping, the pre-`o_proj` gate boundary, the hybrid recurrent cache path, Ling-3 reasoning parsing, and Ling-3 tool-call parsing.

- local CPU-compatible targeted suite: `29 passed`
- repository-pinned Ruff 0.13.3: passed on all changed Python files
- `git diff --check`: passed

### Final EPMoE E2E validation

The final source tree for `fe4b4de732799e1d07a4303bb2ea3836cc1d491f` was reconstructed from the pushed base plus the local patch before the commit was pushed, then verified against Git tree `7ee0cc600b60a6c02b6daf2572f73d8665fa7f47`.

- runtime: JAX `0.10.2`, BF16, 1x dedicated TPU v6e, TP=1
- existing TPU suite: `40 passed, 149 warnings in 263.74s`
- checkpoint: revision `b61f4338de3e68ffc9c0bc1ed5e902981a4a929e`, 32 shards, 15,787,992,416 bytes
- server: explicit `--moe-backend epmoe`, FA attention, Ling-3 reasoning/tool parsers
- E2E: health and model endpoints passed; 3/3 deterministic OpenAI Chat requests returned valid completions
- server log scan: no traceback, fatal, OOM, resource-exhaustion, check-failure, or NaN markers

## Benchmarking and Profiling

This PR is scoped to modeling correctness. No performance claim is made here.

## Checklist

- [x] English PR body.
- [x] Linked the split RFC (#1565).
- [x] Included reproducible test commands and exact revisions.
- [x] Removed Flash-only and one-off eval/dump diagnostics from the final diff.
- [x] Ran full-dataset accuracy validation on dedicated non-CI TPU capacity.
- [x] Validated the final refactor with the full KDA suite on a dedicated TPU v6e.
- [x] Validated the final EPMoE-only source tree with the real Tiny checkpoint and OpenAI-compatible E2E requests.


cjx0709:
Direct accelerator validation completed on an isolated temporary v6e-1 pool in `ainfer-tpu-bench`.

- tested commit: `1bf984bc4baf73b2c7b9754f76bf43843ef2bc40` (current follow-up commit is formatting-only)
- runtime: JAX 0.10.2, TPU backend, 1 v6e chip
- result: **58 passed** in 240.56s, including the full KDA/Pallas, Bailing MoE, Ling3 modeling/parser, reasoning parser, and KV-cache mixin suites
- cleanup: the test Job, Pod, TPU node, and temporary node pool were all deleted and verified absent

No extra CI TPU rerun was requested; the normal PR checks are left to the shared CI capacity.

cjx0709:
Tiny accuracy validation update (exact head `243164171367d9d4220d290fe9d7fc5243eccbd6`):

- added the official-HF Torch CPU golden path requested in review, plus deterministic JAX tensor manifests and layer/router dumps
- changed per-step top-k comparison to teacher-force the same HF prefix; free-running logprobs after the first mismatch are no longer reported as if they shared a context
- final TPU component suite: **62 passed** in 285.41s on an isolated `ainfer-tpu-bench` v6e-1
- strict four-prompt accuracy gate: **failed**, so this PR remains Draft/not ready for merge
- 28/32 teacher-forced greedy steps match; the four mismatches occur at one step per prompt
- first layerwise expert-set divergence is layer 4, where about `1.5e-3` router-score drift exceeds an approximately `1.0e-3` group cutoff margin and flips a MoE group
- recomputing grouped top-k from each side's dumped scores plus the same expert bias reproduces its selected experts, so no top-k logic mismatch is currently indicated
- Torch BF16 CPU goldens are hardware-sensitive; canonical generation is now documented/pinned to `n2-highmem-16` rather than claiming `ATEN_CPU_CAPABILITY=default` is portable

The PR body has the per-case cosine/RMSE/top-20 table. All temporary CPU/TPU Jobs and pools were deleted after artifacts were copied; `ainfer-tpu-bench` is back to only `default-pool`.

python/sgl_jax/srt/models/bailing_moe_v3.py
```diff
@@ -0,0 +1,947 @@
...
+                        ("up_proj", "w3_shared"),
+                        ("down_proj", "w2_shared"),
+                    ):
+                        mappings[f"{prefix}.mlp.shared_experts.{source_name}.weight"] = (
+                            WeightMapping(
+                                target_path=f"{target_prefix}.experts.{target_name}",
+                                sharding=(None, None),
+                                transpose=True,
+                            )
+                        )
```
JamesBrianD: In this mapping, fused moe is none sharding. Please check it.

cjx0709: I removed the fused moe v2 in ling3, now only support epmoe, because ling3-tiny is 9B a small scale model.

python/sgl_jax/srt/models/bailing_moe_v3.py
```diff
@@ -0,0 +1,895 @@
...
+    def _pre_o_proj(
+        self,
+        attn_output: jax.Array,
+        hidden_states: jax.Array,
+    ) -> jax.Array:
+        gate, _ = self.g_proj(hidden_states)
+        gate = jax.nn.sigmoid(gate.astype(jnp.float32)).astype(attn_output.dtype)
+        num_tokens = attn_output.shape[0]
+        gated = attn_output.reshape(num_tokens, self.num_heads, self.v_head_dim)
+        gated *= gate[:, :, None]
```
JamesBrianD: Does the num_heads need tp sharding?

python/sgl_jax/srt/models/bailing_moe_v3.py
```diff
@@ -0,0 +1,895 @@
...
+                covered.update(k for k in expert_keys if k in ckpt_keys)
+                continue
+            if "*" in mapping_key:
+                pattern = re.escape(mapping_key).replace(r"\*", r"(.*?)")
+                covered.update(k for k in ckpt_keys if re.fullmatch(pattern, k))
+            else:
+                if mapping_key in ckpt_keys:
+                    covered.add(mapping_key)
+
+        auto_skipped = {k for k in ckpt_keys if loader._is_excluded_layer_weight(k)}
```
JamesBrianD: Could you incorporate all the logic here into the weight loader?

python/sgl_jax/srt/entrypoints/openai/serving_chat.py
```diff
@@ -902,6 +902,12 @@ def _get_reasoning_from_request(self, request: ChatCompletionRequest) -> bool:
             return kwargs.get("enable_thinking") is not False
         if parser == "mimo":
             return kwargs.get("enable_thinking") is True
+        if parser == "ling3":
+            thinking = kwargs.get("thinking")
+            enable_thinking = kwargs.get("enable_thinking")
```
MokusMokun: why there are two model reasoning/thinking switches ? seems like unnecessary redundancy.

python/sgl_jax/srt/function_call/ling3_detector.py
```diff
@@ -0,0 +1,22 @@
...
+        re.DOTALL,
+    )
+
+    def __init__(self):
+        super().__init__()
+        self.func_detail_regex = re.compile(
+            r"<tool_call>\s*(.*?)(?:(?:\\n|\n)\s*|(?=<arg_key>)|(?=</tool_call>))"
+            r"(<arg_key>.*?)?</tool_call>",
+            re.DOTALL,
+        )
```
MokusMokun, line 22: `Ling3Detector`, including its parent `Glm4MoeDetector`, seems like having a issue for non-streaming inference path: it passes exact tool-name lookup without stripping whitespace

For example, "<tool_call>execute_bash  \n..." produces "execute_bash  " and is rejected as undefined. Please apply .strip() somewhere.

python/sgl_jax/srt/models/bailing_moe_v3.py
cjx0709: align to `["data", "tensor"]` style

python/sgl_jax/srt/models/bailing_moe_v3.py
cjx0709: move to weight loader

python/sgl_jax/srt/function_call/ling3_detector.py
cjx0709: sovled

python/sgl_jax/srt/entrypoints/openai/serving_chat.py
cjx0709: good notice, solved

JamesBrianD (approved):

PR #1567, opened 2026-08-31
[Perf] feat(dsa): query-block sparse-MLA prefill kernel (2.2x TTFT @ 110k, 2.6x @ 16k)

zorrofox:
### Motivation

The DSA sparse prefill kernel (`sparse_mla_prefill.py`) runs one program per query: each of the `S` queries independently DMAs its `K` selected pages from HBM. At long context the per-query selections overlap heavily (attention sinks + local windows), so the same physical page is re-fetched by thousands of programs. At 110k ctx (880 pages, 8192-query chunk, K=32) each page is fetched ~298x: **38.7 GB of HBM reads per layer against a 129 MB KV cache**. The score matmul also feeds only `H` (~4 with head-TP) sublane rows, idling the MXU. Profiling shows this attend kernel at ~47% of the 110k TTFT.

This is the same redundancy that GPU sparse-MLA kernels (e.g. DeepSeek's `flashmla_sparse`) eliminate with warp-level sharing of gathered KV; this PR is the TPU analogue.

### Design

Within the existing `DSA_PREFILL_SPARSE=1` opt-in surface, the sparse extend now runs blocked by **default** — no new flag to discover. The semantics are parity-gated identical to the per-query kernel (same masked-softmax math; only the execution shape changes), the default `DSA_PREFILL_SPARSE=0` path and CI are untouched, and `DSA_PREFILL_QBLOCK=0` remains as an escape hatch back to the per-query kernel. Three parts:

1. **jnp preprocessing** (`build_block_unit_tables`): group queries into blocks of `QB=64` (`DSA_PREFILL_QBLOCK_QB`), compute each block's selected-page **union** (sorted, compacted) plus a per-query membership bitmap. Scatter-free (broadcast-compare + sort): ~0.5 ms/layer at the 110k shape.
2. **Pallas kernel** (`sparse_mla_prefill_qblock.py`): grid over query blocks; the block's union pages are DMA'd **once each** through an `NBUF=4` prefetch ring (the per-page fetches are latency-bound, not bandwidth-bound). Per-query selection is restored inside flash-softmax with a `-inf` bias = membership AND causal AND kv_len. Score layout is key-major (`[page_rows, QB*H]`), which fills the MXU lane axis (4 head rows -> 256 rows) and keeps all bitmap reads on the sublane axis (Mosaic-friendly: aligned 32-row window + iota select).
3. **Ragged/paged mode**: seq-local page ids are lifted to global page-table keys, so one block may hold queries from different requests (packed multi-request extend, same metadata as the dense path). Wired behind the existing `DSA_PREFILL_SPARSE` ragged call site.

### Prior art

The macro-structure (query-block x KV-block flash softmax, ring-buffered async page DMA, clamped always-legal copies with mask-side correctness) matches tokamax's public MLA ragged-paged-attention kernel ([`openxla/tokamax`, `_src/ops/experimental/mla`](https://github.com/openxla/tokamax/tree/main/tokamax/_src/ops/experimental/mla)). The deltas are workload-driven: RPA walks dense contiguous KV, so a depth-2 double buffer with multi-page batched DMAs saturates bandwidth, while top-k sparse gather fetches discontiguous 128-token pages (latency-bound, ~1.5 us each) and needs a deeper (`NBUF=4`) prefetch ring; and RPA has no page-dedup problem, which is the core of this kernel (block-union + membership bias). To my knowledge there is no public TPU kernel for top-k sparse paged MLA prefill. In-kernel KV cache write-back (fused in RPA) is a natural follow-up kept out of this PR to stay minimal.

### Results (v7x, GLM-5.2 fp8, tp16, page 128, ctx 135k, chunk 8192, single stream)

Paired on the same deployment, flag flipped in place; OFF arm reproduces the #1545 numbers exactly.

| metric | OFF (per-query kernel) | ON (qblock QB=64) | delta |
|---|---|---|---|
| TTFT 16k | 9.28 s | **3.58 s** | **-61.4% (2.59x)** |
| TTFT 110k (x3) | 65.53 / 65.54 / 65.54 s | **30.37 / 30.40 / 30.41 s** | **-53.6% (2.16x)** |
| decode TPOT 110k | 15.25 ms | 15.34 ms | unchanged (decode path untouched) |

Kernel microbench (110k shape, per-layer-per-chunk, bf16): per-query 43.3 ms -> qblock 8.1 ms (**5.4x**) on a locality-flavoured selection (union ~69 pages/block); 8.9x when all queries share the same pages.

### Cross-generation validation (v6e-64, GLM-5.2 BF16 + dynamic W8A8, tp64/dp8/ep64, ctx 8k)

Same paired protocol on TPU v6e (previous generation, and the first multi-`dp` validation of the sparse extend path). Kernel parity 32/32 unmodified; paired throughput (1k-in/1k-out random):

| concurrency | default (qblock) | DSA_PREFILL_QBLOCK=0 | ratio |
|---|---|---|---|
| 64 | 578.7 tok/s | 389.0 | **1.49x** |
| 256 | 1121.6 | 845.3 | **1.33x** |
| 460 | 1739.5 | 1240.1 | **1.40x** |

GSM8K 200q paired: 0.965 / 0.955; needle 25/25 both arms; zero Mosaic/runtime errors across both arms. Note this is a short-context saturated-selection workload — the kernel still wins on pure page-DMA dedup even when sparsity itself saves nothing.

### Known boundary (documented; why an escape hatch exists)

Blocked execution does more FLOPs per query (each query scores the whole block union). It wins whenever neighbouring queries' selections overlap — measured break-even at union ≈ 520 of 880 pages per 64-query block; a pathological **uniform-random** selection (union ≈ 795/880, no sinks, no locality) is ~0.5x vs the per-query kernel. Real DSA selections are sink+local-window dominated (same assumption as the GPU sparse kernels), and the e2e numbers above confirm the win at both 16k and 110k. `DSA_PREFILL_QBLOCK_QB` tunes the tradeoff.

### Accuracy gates

- CPU parity (interpret): 19/19 — preprocessing set-reconstruction vs python-set oracle (incl. cc=8/32 packed multi-request shapes); kernel vs masked-softmax oracle AND vs the deployed per-query kernel (<2e-3 fp32); ragged qblock vs deployed ragged (blocks straddling requests; caches bit-identical).
- TPU parity: 6/6 (bf16/fp32, flat + ragged; |err| ~1e-4).
- Needle 76k, 5 depths x 5 values, paired: **ON 25/25 vs OFF 25/25** (answer digits identical in all samples).
- GSM8K 200q paired (5-shot, temp 0, parallel 32): **ON 0.950 vs OFF 0.970** — both above the 0.93 regression bar; **193/200 predictions byte-identical**, the 7 divergent generations are downstream token divergence from bf16-level numeric perturbation (same signature as prior paired runs on this path). Same run shows the concurrency win: 200 questions in 116 s (ON) vs 234 s (OFF).
- Compile envelope: precompile grid bs {1,8,32} x tokens {512..4096} passes with qblock ON; live requests at cc=1 / 8 / 32 all served with zero tracebacks / Mosaic errors.

### Test plan

- [x] `test_qblock_preprocess.py` (13 cases, CPU)
- [x] `test_qblock_kernel_parity.py` (6 cases, CPU interpret, incl. ragged)
- [x] `test_qblock_kernel_tpu.py` (6 cases, TPU)
- [x] GSM8K 200q paired ON/OFF (0.950 / 0.970, 193/200 identical)
- [x] cc=1/8/32 compile envelope (precompile grid + live requests at each cc)

### DIFF
--- a/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill_qblock.py
+++ b/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill_qblock.py
@@ -1,0 +1,92 @@
+"""Blocked (query-batched) sparse-MLA prefill — P1 of the DSA attend path.
+
+The deployed per-query kernel (``sparse_mla_prefill.py``) runs one program per
+query: each of the ``S`` queries independently DMAs its ``K`` selected pages. At
+long context the selections overlap heavily (sinks + local windows), so the same
+physical page is re-fetched by thousands of programs — ~298x redundant HBM
+traffic per layer at 110k — and the score matmul feeds only ``H`` (~4) sublane
+rows, idling the MXU.
+
+The blocked kernel amortises both: ``QB`` queries share one program, the block's
+selected-page **union** is DMA'd once per page, and per-query selection is
+restored with a ``-inf`` membership bias (bitmap AND causal AND kv_len) inside a
+flash-softmax over ``QB*H`` score rows.
+
+This module hosts (in build order):
+
+1. ``build_block_unit_tables`` — jnp preprocessing (outside the kernel): turn
+   ``topk_units [T, K]`` into per-block union tables + membership bitmaps.
+2. the Pallas query-block kernel itself (v1: single-sequence flat).
+"""
+
+from __future__ import annotations
+
+import jax
+import jax.numpy as jnp
+
+_SENTINEL = jnp.iinfo(jnp.int32).max
+
+
+def build_block_unit_tables(
+    topk_units,  # [T, K] int32 selected unit ids per query, -1 padded
+    *,
+    query_block: int,  # QB: queries per block
+    u_max: int,  # static cap on the per-block union size
+):
+    """Per-query-block union + membership tables for the blocked kernel.
+
+    Returns ``(blk_units, blk_member, blk_counts)``:
+
+    * ``blk_units [nQB, u_max]`` int32 — the block's selected units, sorted
+      ascending, ``-1`` padded. Uniques beyond ``u_max`` are dropped.
+    * ``blk_member [nQB, query_block, u_max]`` int8 — 1 where query ``q`` of the
+      block selected ``blk_units[b, u]`` (the kernel's -inf bias source).
+    * ``blk_counts [nQB]`` int32 — the TRUE union size, **uncapped**: a value
+      ``> u_max`` means the block overflowed and its tables are incomplete; the
+      caller must gate (pick ``u_max >= min(query_block*K, num_units)`` to make
+      overflow impossible).
+
+    ``T`` need not divide ``query_block``; trailing queries are padded with
+    ``-1`` rows (all-zero membership). Unit ids are opaque keys: callers with
+    packed multi-request (ragged) inputs lift seq-local ids to global keys
+    (e.g. ``base + page``) before calling, so blocks may straddle requests.
+    """
+    T, K = topk_units.shape
+    qb = query_block
+    n_blk = -(-T // qb)
+    pad = n_blk * qb - T
+    tk = jnp.pad(topk_units.astype(jnp.int32), ((0, pad), (0, 0)), constant_values=-1)
+    flat = tk.reshape(n_blk, qb * K)
+
+    # sort with -1/padding mapped to a +inf sentinel so invalids sink to the end
+    vals = jnp.where(flat < 0, _SENTINEL, flat)
+    svals = jnp.sort(vals, axis=1)
+    is_new = jnp.concatenate([jnp.ones((n_blk, 1), bool), svals[:, 1:] != svals[:, :-1]], axis=1)
+    uniq = is_new & (svals != _SENTINEL)
+    blk_counts = uniq.sum(axis=1).astype(jnp.int32)
+
+    # compact: scatter each first-occurrence to its rank; overflow ranks drop
+    rank = jnp.cumsum(uniq, axis=1) - 1
+    rank = jnp.where(uniq, rank, u_max)  # non-unique/sentinel -> out of range
+    rows = jnp.broadcast_to(jnp.arange(n_blk, dtype=jnp.int32)[:, None], rank.shape)
+    blk_units = jnp.full((n_blk, u_max), -1, jnp.int32)
+    blk_units = blk_units.at[rows, rank].set(svals, mode="drop")
+
+    # membership: binary-search every (query, k) selection in the compacted
+    # table (valid prefix ascending; -1 pad remapped to the sentinel so the
+    # array is globally sorted). O(T*K*log u_max) and no [.., K, u_max]
+    # broadcast materialisation.
+    search_tbl = jnp.where(blk_units < 0, _SENTINEL, blk_units)
+    pos = jax.vmap(jnp.searchsorted)(search_tbl, vals)  # [n_blk, qb*K]
+    pos_c = jnp.minimum(pos, u_max - 1)
+    hit = (
+        (vals != _SENTINEL)
+        & (pos < u_max)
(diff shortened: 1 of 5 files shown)


python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill_qblock.py
```diff
@@ -0,0 +1,453 @@
...
+    Dv = kv_lora_rank
+    RB = read_block
+    QB = query_block
+    ragged = q_seq_id is not None
+    if ragged:
+        if B != 1:
+            raise ValueError(f"ragged mode packs all requests into B=1 (got B={B})")
+        if seq_lens is None or cu_kv_lens is None or page_indices is None:
+            raise ValueError("ragged mode requires seq_lens, cu_kv_lens and page_indices")
+        if page_size is None or page_size != RB:
```
cjx0709, line 299: Could we keep the existing `read_block % 16 == 0` guard for ragged/paged mode? `_copy` uses `PS // 8` and reads `RBF = align_up(RB, 16)` rows, so an unsupported page size can otherwise resolve the wrong physical offset or cross a page instead of failing clearly.

python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill_qblock.py
```diff
@@ -0,0 +1,453 @@
...
+
+    # membership -> slot-major [B,nQB,U_pad,QBHp] int8, lanes H-expanded so the
+    # kernel's lane r = q*H + h reads member[q] directly.
+    memt = jnp.repeat(blk_m.transpose(0, 1, 3, 2), H, axis=3)  # [B,nQB,u_max,QBH]
+    memt = jnp.pad(memt, ((0, 0), (0, 0), (0, U_pad - u_max), (0, QBHp - QBH)))
+    units4 = jnp.pad(blk_u, ((0, 0), (0, 0), (0, U_pad - u_max)), constant_values=-1)
+    units4 = units4.reshape(B, nQB, 1, U_pad)
+    # clamp: if a block overflowed u_max (impossible with the default cap), only
+    # the retained prefix of the union is walked.
+    counts4 = jnp.minimum(blk_c, u_max).reshape(B, nQB, 1, 1)
```
cjx0709, line 364: Could we warn when an explicitly reduced `u_max` may truncate the union? The trade-off is reasonable, but this `minimum` silently drops selected pages while the API advertises parity; a one-time warning and docstring note would expose the possible accuracy loss without adding a fallback.

python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill_qblock.py
zorrofox: Good catch — restored in ca545606. Ragged mode now rejects `read_block % 16 != 0` up front, with the same contract as the per-query paged kernel: `_copy` resolves the page start as `(PS//8)*8` and reads `RBF = align_up(RB, 16)` rows, so an unsupported page size would resolve the wrong physical offset or cross a page instead of failing clearly. Added `test_ragged_rejects_non16_page_size` to pin the guard.

python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill_qblock.py
zorrofox: Done in ca545606. An explicitly reduced `u_max` below the lossless bound `min(query_block * K, num_units)` now logs a one-time warning, and the docstring spells out the truncation semantics: a block whose true union exceeds the cap silently drops the tail of its sorted union, so outputs may diverge from the per-query kernel. The default path stays lossless and silent. Added `test_umax_reduced_warns_once` (warning fires exactly once; default `u_max=None` stays quiet).

cjx0709 (approved):

PR #378, opened 2025-11-13
Feat/eagle support

pengchengneo:
https://github.com/sgl-project/sglang-jax/issues/192
complete Eagle Worker main process

`python3 -u -m sgl_jax.launch_server   --model-path Qwen/Qwen3-32B   --trust-remote-code   --device=tpu    --random-seed=3   --mem-fraction-static=0.8   --max-prefill-tokens=4096   --max-running-requests=16  --attention-backend=fa   --dtype=bfloat16  --port 30000 --host 0.0.0.0 --disable-overlap-schedule --tp-size 4 --speculative-algorithm EAGLE3 --speculative-draft-model-path  AngelSlim/Qwen3-32B_eagle3 --page-size 64 --speculative-eagle-topk 1 --speculative-num-steps 3 --speculative-num-draft-tokens 4 --speculative-draft-model-revision 67caf31f9062d7ab64872e0a111d499bc16cd205`

image

PR #1569, opened 2026-08-31
perf(rpa-v3): store the speculative tree mask with kv on the lane axis

real3stone:
EAGLE speculative decoding drafts a **tree** of candidate tokens, not a chain. Sibling
branches sit next to each other in the KV cache but represent mutually exclusive futures
and must not see each other — a causal mask cannot express that, so target-verify passes
an explicit boolean mask over (draft token, kv position).

Today each bit of that mask is blown up to `head_dim` int32s — **512 bytes per bit** —
and the kernel reads back only lane 0 of each. The wasted bytes are the visible half of
the problem. The expensive half is that this layout lands kv on the VMEM **sublane**
axis while the consumer needs it on the **lane** axis, so every `(bq, bkv)` block pays a
scalarized sublane→lane transpose. On v6e that makes the masked path **61x slower than
causal verification at the same shape**, and ~98% of the gap is one line.

This PR stores the mask as `[total_q_rows, 1, W]` — one int32 per bit, kv on the lane
axis. The transposing read becomes a plain 2-D slice, the per-q-row DMA loop collapses
into two tiled copies, and the VMEM double buffer shrinks ~128x.

**9.2280 ms → 0.1677 ms (55x), output bitwise identical.**

## Motivation

`custom_mask` exists only for `topk > 1` — both producers are gated on it
(`eagle_draft_worker.py:130` and `:415`) — and the E2E speculative test has pinned topk to
1 since EAGLE support landed in #378 (2025-11-19). `test/srt/test_speculative_decoding.py:52`:

```python
# FIXME(pc) topk > 1 has poor performance now, change it when build_tree_mask_for_draft_decode kernel is  implemented
"--speculative-eagle-topk",
"1",
```

So the masked path is not reached by the default configuration today, which is also why a
61x regression on it could sit here unnoticed.

**This PR does not claim to unblock `topk > 1`.** That FIXME points at a different
component — the host-side `build_tree_mask_for_draft_decode`, still pure numpy — while the
cost measured in this PR is a separate one inside the target-verify attention kernel. There
is at least one further obstacle beyond both. This change removes the one that was located
and measured; whether that is sufficient is not established, and the change stands on its
own terms either way.

### Why the old layout is slow

`custom_mask` is expanded along `head_dim` on the way in (`ragged_paged_attention_v3.py:1799`),
so the VMEM scratch is `(2, bq_sz, bkv_sz, head_dim)` int32. TPU tiles the two minormost
dims, which decides which axis everything lands on:

```
BEFORE   bkvmask_ref[2, bq, bkv, head_dim] int32        tile (8,128) over (bkv, head_dim)

                     lane ────────────────────────►  head_dim: 128 copies of ONE bit
                                                        (one q row shown; bq is a
                                                         separate leading dim)
                        ┌─────┬─────┬─────┬ ─ ─ ┬─────┐
   s   kv = 0           │ b0  │ b0  │ b0  │     │ b0  │
   u   kv = 1           │ b1  │ b1  │ b1  │     │ b1  │
   b   kv = 2           │ b2  │ b2  │ b2  │     │ b2  │
   l     ...            │ ... │ ... │ ... │     │ ... │
   a   kv = 511         │b511 │b511 │b511 │     │b511 │
   n                    └─────┴─────┴─────┴ ─ ─ ┴─────┘
   e                       ▲
   │                       └── the kernel reads only this ONE column:  [..., 0]
   ▼

   but the consumer needs   s = [bq*heads, bkv_csz]   with kv on the LANE axis,
   i.e. exactly the transpose of what is stored.  So `[..., 0]` is not "read one
   element" — it is a sublane→lane relayout:

        8 target VREGs (32 KiB)   ←   touching 1024 source VREGs (4 MiB)
        amplification = head_dim = 128        Mosaic has no vector op for this
                                              shape, so it scalarizes.


AFTER    bkvmask_ref[2, bq, 1, bkv] int32               tile (1,128) over (1, bkv)

                     lane ────────────────────────►  kv: 128 DIFFERENT positions
                        ┌─────┬─────┬─────┬ ─ ─ ┬─────┐
       q row 0          │ b0  │ b1  │ b2  │     │b511 │  ← same 512 bits as above
       q row 1          │ c0  │ c1  │ c2  │     │c511 │  ← next q row's bits
         ...            │ ... │ ... │ ... │     │ ... │
                        └─────┴─────┴─────┴ ─ ─ ┴─────┘

   already in the layout the consumer wants.  The read becomes a plain 2-D slice;
   indexing 0 on the size-1 axis is a squeeze, not a relayout.
```

Supporting arithmetic: 143 µs to move 4 MiB implies ~29 GB/s, which is ~50x below this
chip's *HBM* bandwidth and orders below VMEM. No streaming read is that slow. This is
also the only place in the kernel that indexes the minormost dim with a scalar; every
other VMEM read goes through `strided_load` (`:796-807`), reshaping to 128 lanes and
slicing only sublanes.

An isolation run pins ~98% of the gap to this one read. The load-bearing arm keeps the
mask DMA, the whole 8 MiB scratch and the dynamic semaphore and query offsets, and changes
only which axis the values are read from — it lands within noise of removing the mask
branch entirely. So the DMA, the HBM traffic, the VMEM footprint and the dynamic
addressing are each independently exonerated.

## Modifications

* **The q row index moves to the leading dim.** Mosaic's slice-alignment check only
  inspects the last `tile_rank` dims, so a dynamic, unaligned row start is legal by
  construction. The rank-2 alternative needs a `pl.multiple_of` promise on a row offset
  derived from SMEM, and under `disable_bounds_checks=True` a wrong promise is a silent
  miscompile rather than a fault — not needing one is worth more than the line count
  suggests. (Verified on hardware: the rank-3 form compiles with an unaligned row start;
  the rank-2 equivalent is rejected.)
* **`_fetch_mask` collapses** from a per-q-row `fori_loop` issuing `2*load_q_sz` async
  copies into two tiled DMAs.
* **`get_vmem_estimate_bytes` now charges `2*bq*bkv*32` bits.** A `(R, 1, W)` int32
  scratch was measured to reach exactly `R*W*4` bytes, i.e. the size-1 second-minor axis
  takes a `(1, 128)` tile and costs no sublane padding. Leaving the old formula alone
  would have kept the block search charging 128x for a buffer that no longer exists and
  silently handed back much of the win.
* **Host side**, `flashattention_backend` packs every row to one power-of-two width
  `>= max(aligned_seq_lens)`. A rectangle needs a single width; rounding to a power of two
  keeps the number of distinct traced shapes at `O(log max_context_len)` rather than one
  per batch.
* Since kv now lives on the lane axis, **`bkv_alignment` is raised to 128** on the masked
  path, and `run_rpa_kernel` rejects non-128-aligned kv blocks outright — a tuned-table
  hit bypasses `get_default_block_sizes`, and failing loudly beats miscompiling.

### Behavior changes worth flagging

* **One shared buffer outside the mask path changes shape.** `zero_mask` is materialized
  and passed on every call, mask or not, and goes from `(bkv_sz, head_dim)` to
  `(bq_sz, 1, bkv_sz)` — e.g. 512 KiB to 64 KiB at `bkv=1024`. It is never read without a
  mask, so this is a smaller unconditional allocation and no semantic change, but it is the
  one place this PR touches the non-mask path. Making it conditional (as `custom_mask`
  already is) would be a non-mask-path change needing its own justification, so it is left
  alone here.
* **`custom_mask` is now rank-3 and validated at entry, and one producer does not match it.**
  The draft-decode path (`eagle_draft_worker.py:415`) sets `forward_metadata.custom_mask`
  directly in the old flat form, bypassing the backend repack, and would now hit a clear
  `ValueError`. Worth being precise that this is not a regression, because that mask has never
  reached the kernel by two independent routes: the call raises `TypeError` first (it passes
  `parents_list=None` into a builder whose first statement is `len(parents_list)`,
  `eagle_util.py:96`), and even without that, the object it writes to is replaced three lines
  before the first forward — `draft_forward` does
  `attn_backend.forward_metadata = metadata_per_step[i]` (`:512`), and
  `get_eagle_multi_step_metadata` builds a fresh `FlashAttentionMetadata()` that never sets
  `custom_mask`. Reviving that path needs a per-step restructure of `draft_forward`, not an
  ABI patch, so updating the producer here would only add a third dead form.
* **The trace-timing helper now warns on a `task` regex miss.** That string has to equal the
  `pallas_call` name; on a miss the extractor quietly fell back to host-side MARKER events
  instead of `device_duration_ps`. The two are not comparable — timing one configuration
  through each path produces a ratio that looks like a real effect and is not. Found the hard
  way while adding the masked benchmark in the first commit.

## Accuracy Tests

Three independent layers:

1. **Bitwise identity against the previous commit.** Same inputs, same bits carried, only
   the storage changed — so `max|delta| == 0.0` is the acceptance criterion, not a tolerance.
   Holds on uniform, **ragged** (`kv_len` 512..2048 within one batch — the only shape where
   a pad-to-max packing bug can show), and `q_len=16`.
2. **Existing masked accuracy tests** (`test_flashattention_mha.py`, `test_flashattention_gqa.py`,
   `-k custom_mask`) pass. These compare against `ref_ragged_paged_attention`, which still
   consumes the **flat** mask and recomputes its own `cumsum(q_len*kv_len)` offsets. The test
   helper derives the rank-3 view separately on purpose: if the oracle also learned about
   rectangles, the tests would be comparing a padding bug against itself.
3. **New CPU test for the host repack** (`test_verify_mask_packing.py`, in `unit-test-cpu`).
   That repack is the only production producer of this array and had no coverage at all,
   while its failure mode — one sequence reading another's mask — is silent. The test pins
   the invariant the kernel relies on (row index == per-DP-rank cumulative q-token index) and
   covers padding slots between live slots, `dp=2`, and the width bucketing.
4. **New 4-chip test** (`test_flashattention_custom_mask_dp.py`, in `unit-test-tpu-v6e-4`).
   The mask is sharded `P("data")`, so rank `r` gets a contiguous block of
   `per_dp_bs * draft_token_num` rows while the kernel addresses rows through the
   *rank-local* `cu_q_lens`. If those disagree one rank reads another's mask, silently.
   Worth stating plainly: **`dp > 1` with a mask had no hardware coverage before this PR
   either** — `test_flashattention_dp.py` is not registered in any suite, and
   `unit-test-tpu-v6e-4` contained no flashattention test at all. This PR changes the rank
   of the sharded array (1-D to 3-D), which is exactly the kind of thing that contract can
   break on, so it should not stay untested. The test runs the same ragged batch at `dp=4`
   and on a single chip with block sizes pinned, and requires the outputs to agree; every
   sequence gets its own random mask, so any cross-rank mix-up changes the result.

The split between the two new tests is deliberate. With no padding slots the packed array
is byte-identical at `dp=1` and `dp=4`, so the 4-chip test exercises the **sharding**
contract; padding slots, where the two layouts do diverge, are what the CPU test covers.
Both build their mask through `_pack_verify_mask` rather than by hand, so neither is
checking the layout against the test's own idea of it.

## Benchmarking and Profiling

v6e-1, jax 0.10.2 / libtpu 0.0.42.1. `bs16 x q_len8 x prefix2040`, page 128,
32 q-heads / 8 kv-heads, head_dim 128, bf16, blocks `(16,512,16,512)` (64 blocks).
Times are `device_duration_ps` from the trace, via this repo's own tuner.

| | time | µs / (bq,bkv) block | VMEM estimate |
|---|---|---|---|
| before (`HEAD~1`) | 9.2280 ms | 144.2 | 18.38 MiB |
| **after (`HEAD`)** | **0.1677 ms** | **2.6** | **10.44 MiB** |
| causal reference (`causal=1`) | 0.1519 ms | 2.4 | 10.38 MiB |

**55.0x.** The masked path now costs **10% more than causal verification** instead of
61x more. (The causal row is a reference point, not a no-op: it still applies a mask,
just one computed in registers with `tpu.iota`, already in the right layout.)

The VMEM estimate lines up exactly with the buffers: `18.38 − 10.44 = 7.94 MiB` is the
old mask double buffer (`2*16*512*128*4` = 8.0 MiB), and `10.44 − 10.38 = 0.06 MiB` is
the new one (`2*16*1*512*4` = 64 KiB).

### Reproducing

The benchmark lands in the **first commit** precisely so the before/after comes from one
command. No benchmark in this repo previously exercised a non-None `custom_mask`:
`bench_flashattention.py`, `tune_target_verify_v3.py` and `get_block_spec_config_v3.py`
all passed `custom_mask=None`, so the tuned block-size tables are measured on causal
verification only and this gap was invisible.

```bash
cd benchmark/kernels/flash_attention
python tune_target_verify_v3.py --batch-size 16 --draft-token-num 8 \
  --prefix-len 2040 --page-size 128 --q-head-num 32 --kv-head-num 8 \
  --head-dim 128 --candidate 16,512,16,512 --custom-mask
```

Run it at `HEAD~1` for the baseline and at `HEAD` for the result; drop `--custom-mask`
for the causal reference.

Two things are **not** in those numbers, and are not claimed:

* The `jnp.repeat` this PR removes runs *outside* `pallas_call`, so its cost was never
  part of the measured kernel time. On this config it materialized `262144 x 128` int32
  = **128 MiB** in HBM per call. That disappears too.
* The host-side repack and its blocking `device_get` were not measured. The new packing
  allocates `padded_bs * draft_token_num * W * 4` bytes, where `W` is the power-of-two
  round-up of the batch's largest aligned kv length (`mask_row_width`). The old packing
  concatenated per-slot chunks at each slot's own aligned length. On a uniform full batch
  the two are within 2x; on a ragged or heavily padded batch the rectangle is larger,
  because padding slots still occupy rows and one long request sets `W` for every row.
  What that buys is shape stability: `W` has one traced shape per context-length bucket,
  whereas the old `max_len` moved with batch composition, so its shape count was bounded
  only by the number of length combinations a batch could take. Nothing warms these up --
  `precompile_spec_decode` varies batch size only, over a dummy batch with `seq_lens = 1`
  -- so every distinct width costs a first-request compile.

## Follow-ups

Each of these lands as its own PR right after this one. They are split out because bundling
them would make this diff harder to review or would confound the measurement above — not
because they are optional. Listed in the order they will be sent:

1. **Drop `cu_seq_mask_lens` from the scalar prefetch tuple.** It has no consumer after this
   change (the rectangle is uniform, so a row index is just the global q-token index), but
   removing it means renumbering `input_output_aliases`, whose failure mode is silent buffer
   mis-donation. That is far easier to review as a diff containing nothing else.
2. **Remove the `bq_sz <= 16` cap for custom masks** (`:1563-1566`). Its stated reason —
   *"Custom-mask intermediates exceed v5/v6 scoped VMEM with a 32-row query tile"* — no longer
   holds once that intermediate drops from 16 MiB to 64 KiB. Lifting it changes the chosen
   block sizes, so it needs its own before/after rather than moving the numbers above.
3. **Have the tree kernel emit the final layout, and drop the host round-trip.**
   `build_eagle_tree_structure_kernel.py:367-371` allocates the mask `(capacity, 128)` int32
   and returns `tree_mask[:, 0]` at `:407`, discarding 127 of every 128 words: a Pallas store
   wants a full 128-lane minor dim, and a flat `(capacity,)` output is not directly
   expressible. The flat result is then pulled to the host (`flashattention_backend.py:391`,
   a blocking `device_get`), repacked into the rectangle with numpy (`:398`), and pushed back
   (`:401`), once per target-verify step.

   Kept separate because it changes the output contract of a kernel whose golden test
   (`test_eagle_tree_build.py`) gates merges, and because the host round-trip predates this
   PR -- folding it in would mix a pre-existing cost into this diff's measurement.
4. **The v1/v2 mirror kernel** (`ragged_paged_attention.py`) carries the same code and the same
   cost. It follows once this shape has settled in review.

**Beyond these.** The rectangle's width is set by the batch's longest sequence, but the
prefix columns carry no information: `build_eagle_tree_structure_kernel.py` initialises the
mask to all ones (`:20`, `:42`) and then only ever clears and re-sets the `draft_token_num`
draft columns (`:127-138`), so every draft token sees the whole committed prefix
unconditionally. Sending only the `q x q` corner would make `W` a constant 128 rather than
`pow2(max_kv)` -- less host memory, and, more usefully, a single traced shape instead of one
per context-length bucket, which the existing `seq_lens = 1` precompile batch would then
cover exactly. Upstream SGLang ships this as `TreeMaskMode.QLEN_ONLY`, with a bit-packed
variant on top. I have not prototyped it on this backend, so this is a direction, not a
commitment.

## Checklist

### DIFF
--- a/python/sgl_jax/test/test_flashattention_custom_mask_dp.py
+++ b/python/sgl_jax/test/test_flashattention_custom_mask_dp.py
@@ -1,0 +1,241 @@
+"""DP=4 coverage for the rank-3 speculative custom mask (TPU, 4 chips).
+
+What this pins down
+-------------------
+The verify mask is a rectangle ``[total_q_rows, 1, W]`` sharded with ``P("data")``,
+so rank ``r`` receives a contiguous block of ``per_dp_bs * draft_token_num`` rows.
+Inside ``shard_map`` the kernel addresses a row as
+``cu_q_lens_ref[seq_idx] + bq_idx * bq_sz``, where ``cu_q_lens`` is the *rank-local*
+cumsum produced by ``_per_dp_cumsum``.
+
+If those two disagree, one rank reads another rank's mask. Nothing crashes: the
+model simply accepts wrong draft tokens. ``test_verify_mask_packing`` pins the
+packing arithmetic on the host, but only this test puts the real sharded array
+through the kernel.
+
+Method: run the same logical batch at dp=4 and at dp=1 with block sizes pinned, and
+require the outputs to agree. Every sequence gets its own random mask, so any
+cross-rank mix-up changes the result.
+"""
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+from jax.sharding import NamedSharding
+from jax.sharding import PartitionSpec as P
+
+from sgl_jax.srt.kernels.ragged_paged_attention.ragged_paged_attention_v3 import (
+    ragged_paged_attention,
+)
+from sgl_jax.srt.layers.attention.flashattention_backend import (
+    _pack_verify_mask,
+    _per_dp_cumsum,
+)
+from sgl_jax.srt.utils.mesh_utils import create_device_mesh
+from sgl_jax.test.test_utils import CustomTestCase
+
+PAGE_SIZE = 128
+HEAD_DIM = 128
+DTYPE = jnp.bfloat16
+GQA = 4
+DRAFT_TOKENS = 8
+BLOCKS = (16, 512, 16, 512)  # pinned so dp=1 and dp=4 make identical block choices
+
+
+def _cdiv(a, b):
+    return -(-a // b)
+
+
+def _build(kv_lens, n_kv_heads, dp_size, seed=0):
+    """Per-rank inputs for a target-verify step, plus the rank-3 mask.
+
+    ``kv_lens`` is the whole logical batch; it is split evenly across ranks. Page
+    indices are rank-local, because under ``shard_map`` each rank sees only its own
+    slice of the KV cache.
+    """
+    bs = len(kv_lens)
+    assert bs % dp_size == 0
+    per_dp_bs = bs // dp_size
+    q = DRAFT_TOKENS
+    packing = 32 // (jnp.dtype(DTYPE).itemsize * 8)
+    aligned = [_cdiv(kv, PAGE_SIZE) * PAGE_SIZE for kv in kv_lens]
+    pages_per_seq = max(_cdiv(kv, PAGE_SIZE) for kv in kv_lens)
+    pages_per_rank = per_dp_bs * pages_per_seq
+    width = max(128, 1 << max(max(aligned) - 1, 1).bit_length())
+
+    # Build the FLAT tree-kernel output, then hand it to the production packer.
+    # Doing the rectangle by hand here would make the test agree with whatever
+    # mental model wrote it -- the same trap as teaching an oracle about the
+    # layout it is supposed to check.
+    rng = np.random.default_rng(seed)
+    blocks = []
+    for kv in kv_lens:
+        prefix = kv - q
+        assert prefix > 0
+        block = np.zeros((q, kv), np.int32)
+        block[:, :prefix] = 1  # committed prefix: every draft token sees all of it
+        corner = np.tril(np.ones((q, q), np.int32))
+        corner *= (rng.random((q, q)) < 0.7) | np.eye(q, dtype=bool)
+        block[:, prefix:] = corner
+        blocks.append(block.reshape(-1))
+    cm = np.concatenate(blocks)
+
+    seq_lens_np = np.asarray(kv_lens, np.int32)
+    aligned_np = np.asarray(aligned, np.int32)
+    cm_kl = np.where(seq_lens_np > 0, seq_lens_np, q - 1).astype(np.int64)
+    cm_off = np.concatenate([[0], np.cumsum(q * cm_kl)])
+    mask = _pack_verify_mask(cm, seq_lens_np, aligned_np, cm_off, q, dp_size, per_dp_bs)
+    assert mask.shape == (bs * q, 1, width), mask.shape
+
+    extend = np.full((bs,), q, dtype=np.int32)
+    cu_q = _per_dp_cumsum(extend, dp_size, per_dp_bs)
+    cu_kv = _per_dp_cumsum(np.asarray(aligned, np.int32), dp_size, per_dp_bs)
+    # Rank-local page indices, repeated per rank.
+    page_indices = np.tile(np.arange(pages_per_rank, dtype=np.int32), dp_size)
+    distribution = np.tile(np.array([0, per_dp_bs, per_dp_bs], np.int32), dp_size)
+
+    total_q = bs * q
+    keys = jax.random.split(jax.random.PRNGKey(seed + 1), 4)
+    n_q_heads = GQA * n_kv_heads
+    # The cache must hold real values, not zeros. With k=0 every committed
+    # position scores 0, contributing exp(0)=1 to the denominator and nothing to
+    # the numerator, so (a) the output collapses to the eight new tokens diluted
+    # by ~2000 ones, which puts the whole signal below any usable tolerance, and
+    # (b) a wrong page index reads zeros just like a right one, so the rank-local
+    # page addressing this test also exercises would go unchecked.
+    #
+    # Both arms allocate the same total page count (dp=4: 4*32, dp=1: 1*128) and
+    # sequence s reads global pages [s*16, (s+1)*16) either way, so seeding from
+    # the same key gives both arms identical per-sequence cache contents.
+    return dict(
+        q=jax.random.normal(keys[0], (total_q, n_q_heads, HEAD_DIM), DTYPE),
+        k=jax.random.normal(keys[1], (total_q, n_kv_heads, HEAD_DIM), DTYPE),
+        v=jax.random.normal(keys[2], (total_q, n_kv_heads, HEAD_DIM), DTYPE),
+        cache=jax.random.normal(
+            keys[3],
+            (dp_size * pages_per_rank, PAGE_SIZE, n_kv_heads * 2 // packing, packing, HEAD_DIM),
+            DTYPE,
+        ),
+        kv_lens=jnp.asarray(kv_lens, jnp.int32),
+        page_indices=jnp.asarray(page_indices),
+        cu_q_lens=jnp.asarray(cu_q),
+        cu_kv_lens=jnp.asarray(cu_kv),
+        distribution=jnp.asarray(distribution),
+        custom_mask=jnp.asarray(mask),
+    )
+
+
+def _run(case, mesh):
+    """Mirror the backend's shard_map wiring, tp=1 so only the data axis splits."""
+    jax.sharding.set_mesh(mesh)
+    in_specs = (
+        P("data", "tensor"),  # queries
+        P("data", "tensor"),  # keys
+        P("data", "tensor"),  # values
+        P("data", None, "tensor", None, None),  # kv_cache_fused
+        P("data"),  # kv_lens
+        P("data"),  # page_indices
+        P("data"),  # cu_q_lens
+        P("data"),  # cu_kv_lens
+        P("data"),  # distribution
+        P("data"),  # custom_mask: [total_q_rows, 1, W], dim 0 split
+    )
+
+    def body(q, k, v, cache, kv_lens, page_indices, cu_q, cu_kv, dist, mask):
+        return ragged_paged_attention(
+            q,
+            k,
+            v,
+            cache,
+            kv_lens,
+            page_indices,
+            cu_q,
+            cu_kv,
+            dist,
+            mask,
+            causal=0,
+            sm_scale=HEAD_DIM**-0.5,
+            m_block_sizes=BLOCKS,
+        )
+
+    # Same construction as flashattention_backend's call site.
+    fn = jax.jit(
+        jax.shard_map(
+            body,
+            in_specs=in_specs,
+            out_specs=(P("data", "tensor"), P("data", None, "tensor", None, None)),
+            check_vma=False,
+        )
+    )
+    args = [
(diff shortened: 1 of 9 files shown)


lingebeng:
Very Good. Thank you for your work!

real3stone:
The three failing checks here are unrelated to this PR — root causes filed as #1597 (the JAX 0.11.1 sharding tightening and the Flax unpin, both from #1574).

python/sgl_jax/test/test_flashattention_custom_mask_dp.py
```diff
@@ -0,0 +1,241 @@
...
+    cm_off = np.concatenate([[0], np.cumsum(q * cm_kl)])
+    mask = _pack_verify_mask(cm, seq_lens_np, aligned_np, cm_off, q, dp_size, per_dp_bs)
+    assert mask.shape == (bs * q, 1, width), mask.shape
+
+    extend = np.full((bs,), q, dtype=np.int32)
+    cu_q = _per_dp_cumsum(extend, dp_size, per_dp_bs)
+    cu_kv = _per_dp_cumsum(np.asarray(aligned, np.int32), dp_size, per_dp_bs)
+    # Rank-local page indices, repeated per rank.
+    page_indices = np.tile(np.arange(pages_per_rank, dtype=np.int32), dp_size)
+    distribution = np.tile(np.array([0, per_dp_bs, per_dp_bs], np.int32), dp_size)
```
JamesBrianD, line 107: The issue here might be with the KV page table constructed for the test, causing the DP to read different data on both sides.

For a simple example: 4 sequences require [2, 1, 2, 1] pages respectively, with 2 physical pages reserved for each sequence.
- DP=1: The page table is a continuous range, with cumulative indexing based on the actual number of pages. The 3rd sequence starts reading from global page 3.
- DP=2: Each rank is allocated 4 physical pages. The 3rd sequence is the first in rank 1, so it starts reading from global page 4.

Even if the cache is initialized with the same seed, the KV content read on both sides will be different, resulting in inconsistent output.

real3stone: You're right, that's exactly it. Fixed in 04eade18 — page_indices is now built from the same cu_kv the kernel indexes with, so both arms read the same pages.

Test-only: production's cache_loc is already packed that way (schedule_batch.py:2420).

JamesBrianD (approved):

PR #1575, opened 2026-09-02
fix(rpa): guard int16 span math by kv capacity

### LINKED ISSUE #1541: [Bug] RPA v3: int16 span overflow silently truncates attention history past absolute position 32767 (prefill/extend only; decode is unaffected)
### Summary

`ragged_paged_attention_v3.py` narrows its span arithmetic to `int16` when three
conditions hold, none of which bounds the sequence length:

```python
# ragged_paged_attention_v3.py:491-493
int_ty = jnp.int32
if get_dtype_packing(q.dtype) != 1 and tpu_version >= 6 and use_causal_mask:
    int_ty = jnp.int16
```

The spans it then builds are **absolute sequence positions**, not block-local offsets:

| value | line | expression |
|---|---|---|
| `processed_q_len` | `:943` | `kv_q_gap + bq_idx * actual_bq_sz`, where `kv_q_gap = kv_len - q_len` |
| `processed_kv_len` | `:984` | `bkv_idx * bkv_sz` |
| `effective_kv_len` | `:948` | `min(kv_len, processed_q_len + actual_bq_sz)` |

`int16` tops out at 32767, so past that the casts at `:494-496` wrap negative and both
the causal comparison (`:506`) and the KV bound (`:513-514`) go wrong. There is no crash,
no NaN and no warning — the model silently stops attending to its early history.

**This is not a new discovery.** Upstream fixed the same defect in
[vllm-project/tpu-inference#2755](https://github.com/vllm-project/tpu-inference/issues/2755) /
[#2756](https://github.com/vllm-project/tpu-inference/pull/2756) (commit `6f0f60b0`,
2026-06-12). sglang-jax imported RPA v3 on 2026-03-24 (`38469dca`) with the pre-fix
condition and has not followed the fix. This repo's condition has also diverged — it carries an
extra `use_causal_mask` term added by #207 — so the upstream patch cannot be
cherry-picked verbatim.

### Reproduction

TPU v6e (single chip), `jax[tpu]==0.10.2`, bf16, `causal=1`, on `71f72a31`
(re-verified against `cddbf6f1` on 2026-08-23 — the kernel file is byte-identical across both),
chunked-prefill shaped batch (`q_len=256`, `distribution=[0,0,1]` → `RpaCase.MIXED`).

The KV cache is built so the answer is readable by eye: `K = 0` everywhere (uniform
softmax), `V = +1` below position 32768 and `V = -1` at/above it. The output of the last
query row is then exactly `mean(V)` over whatever range the kernel actually attended to.

**Mode 1 — silent prefix loss.** Comparing the kernel as shipped against the same kernel
with the single condition at `:492` disabled (identical block sizes, schedule and DMA —
only `int_ty` differs):

| `kv_len` | as shipped (int16) | `int_ty` forced to int32 | expected, full history |
|---|---|---|---|
| 32256 | +1.0000 | +1.0000 | +1.0000 |
| 33024 | **−1.0000** | +0.9844 | +0.9845 |
| 40960 | **−1.0000** | +0.6016 | +0.6000 |
| 49152 | **−1.0000** | +0.3340 | +0.3333 |

Output range is `[-1, +1]`; the gap is 1.33–1.98. A query at absolute position `p > 32767`
can only see KV positions that have *also* wrapped, i.e. `>= 32768`; everything before is
dropped. Right at the cliff the effective context is 32 tokens.

**Mode 2 — whole-block zero output** (not covered by the upstream report). When a `bq`
block lands on `effective_kv_len == 32768` exactly, `trunc16(32768) = -32768` is the int16
*floor*, so `k_span < effective_kv_len_int` is false for every `k`, `v` is zeroed, and
because `mask_value` is `-0.7 * fp32max` (finite, not `-inf`) the softmax denominator stays
non-zero — the block returns exact zeros without any NaN:

| `kv_len` | last-row output | all-rows | rows that are exactly 0 |
|---|---|---|---|
| 32736 | +1.0000 | +1.0000 | 0 / 256 |
| **32768** | **+0.0000** | **+0.8750** | **32 / 256** |
| 33024 | −1.0000 | −0.9996 | 0 / 256 |

`0.8750 = 224/256` — exactly the 32 rows of the one affected `bq` block (`bq_idx=7`,
`bq_sz=32`). Mode 2 has a *lower* trigger threshold than mode 1: any prompt crossing
32768 hits it, because `bq_sz` is a power of two (`MAX_BQ_SZ = 32`, `:1522`) and therefore
divides 32768 exactly.

### Scope

| path | status | why |
|---|---|---|
| prefill / chunked prefill / extend | **affected** | `q_len > 1` → PREFILL/MIXED → `use_causal_mask = True` |
| speculative target verify | **affected** | MIXED |
| **single-token decode** | **not affected** | `:385-386` forces `use_causal_mask = False` for `RpaCase.DECODE`, and the int16 branch is gated on that term |

Chunked prefill does not help: it splits `q`, while `kv_q_gap ≡ prefix_len` grows linearly
with the chunk index. With `--chunked-prefill-size 2048`, chunk 16 onward starts past 32768.

All three trigger conditions hold under default settings (`--dtype auto` resolves to bf16
on bf16 checkpoints, cookbooks pass `--dtype bfloat16` explicitly; v6e/v7x; `fa` backend
with `causal=1`). `docs/cookbook` ships `--context-length 32768` in 9 files and
`--context-length 262144` in 3.

### Why CI does not catch it

Two independent reasons, either alone would be enough:

1. `util.py:get_tpu_version()` returns `-1` off-TPU, so `tpu_version >= 6` is false and the
   CPU suites always take the int32 path — `unit-test-cpu` is structurally blind to it.
2. The TPU suite does exercise the kernel, but `test_flashattention_{mha,gqa,misc}.py` top
   out around `kv_len = 1024`.

### Proposed fix

Restore the length bound while keeping the local `use_causal_mask` guard (the latter is a
workaround for a Mosaic miscompile with custom-mask shapes, per the comment at `:489-490`,
and must not be dropped). `page_indices` is 1-D here (`:1381` asserts it), and the host
already computes the quantity at `:1468`; inside the kernel both inputs are static shapes,
so this stays a trace-time decision:

```python
int_ty = jnp.int32
max_kv_len = (num_page_indices // kv_lens_ref.shape[0]) * page_size
if (
    get_dtype_packing(q.dtype) != 1
    and tpu_version >= 6
    and use_causal_mask
    and max_kv_len <= jnp.iinfo(jnp.int16).max
):
    int_ty = jnp.int16
```

Verified on v6e with a trace-time probe: at a per-sequence capacity of 32640 the MIXED
variant still selects `int16`; at 65536 it falls back to `int32`. The fix does not give up
the fast path where it is safe.

One property worth stating explicitly: the bound is on `pages_per_seq * page_size`, i.e.
the **allocated capacity**, not the actual sequence length. A server launched with a large
`--context-length` therefore loses the int16 path on every request, including short ones.
That is upstream's choice and it is the conservative one.

I am also happy to add the missing regression coverage — a long-context accuracy case in
`test_flashattention_gqa.py` (already registered in `unit-test-tpu-v6e-1`), roughly 20 MB
of KV at `kv_len ≈ 40K`, which fits a v6e-1 runner. Upstream #2756 shipped no test.

### Appendix — how much does the int16 branch actually buy?

Neither upstream #2756 nor this repo has ever published a number for it, so I measured.
Same kernel, single condition flipped, identical block sizes. Methodology: order
counterbalanced (ABA/BAB), plus a **null control** where both arms are int16 — the null
must read 0, and it measures the rig's own bias.

v6e, bf16, MIXED, `kv_len = 32256`, `q_len = 256`:

| q heads / kv heads | null control | measured effect |
|---|---|---|
| 8 / 1 | −0.01% | **−1.50%** (int16 *slower*) |
| 16 / 2 | −0.04% | below resolution |
| 32 / 4 | +0.06% | **+0.42%** |
| 64 / 8 | +0.01% | below resolution |
| 32 / 4 (`kv_len = 16384`) | +0.09% | **+0.46%** |

Best case for int16 is **+0.46%**; most shapes are a wash and the 1-KV-head shape is
1.5% slower. A likely mechanism is visible in the Mosaic IR: the row iota is built in
`int32` and put through the full floor-division expansion before being truncated to
`int16`, so the int16 path adds a whole-tile `arith.trunci` and only narrows the three
downstream `addi`/`cmpi` ops.

**Caveat: this is v6e only.** I have no v7 access. Also note that the commit which
introduced the int16 branch upstream (`b2e416f0`, "[RPA3] Separate kernel to 3 calls and
many optimizations") was a bundle, so it may never have been measured in isolation.

This PR does not change whether the branch exists — it only restores the missing length
guard, which keeps us aligned with upstream. If maintainers would rather delete the branch
outright, that is a strictly smaller diff and I am happy to do that instead.

dajiaohuang:
## Summary
- bound the RPA v3 int16 span fast path by full per-sequence KV capacity
- keep the existing causal-mask / TPU v6+ guard that avoids the Mosaic custom-mask bug
- add a CPU unit test that locks the dtype selection at the 32767-token boundary

## Testing
- `uv run --system-certs --python 3.12 --with pytest --with 'jax[cpu]==0.10.2' --with numpy~=2.2.6 --with psutil~=7.0.0 --with pyzmq~=27.0.1 --with fastapi~=0.116.1 --with orjson~=3.11.1 env PYTHONPATH=/Users/bytedance/Documents/trae_projects/repostew/sglang-jax/python python -m pytest /Users/bytedance/Documents/trae_projects/repostew/sglang-jax/python/sgl_jax/test/test_kernel_utils.py -q`

Fixes #1541.

### DIFF
--- a/python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention_v3.py
+++ b/python/sgl_jax/srt/kernels/ragged_paged_attention/ragged_paged_attention_v3.py
@@ -59,6 +59,26 @@ def _semaphore_kwargs(disable_semaphore_checks: bool) -> dict:
     return {}
 
 
+def _get_span_int_dtype(
+    q_dtype: jnp.dtype,
+    *,
+    tpu_version: int,
+    use_causal_mask: bool,
+    pages_per_seq: int,
+    page_size: int,
+) -> jnp.dtype:
+    """Select the narrow span dtype only when the full per-sequence capacity is safe."""
+
+    if (
+        get_dtype_packing(q_dtype) != 1
+        and tpu_version >= 6
+        and use_causal_mask
+        and pages_per_seq * page_size <= jnp.iinfo(jnp.int16).max
+    ):
+        return jnp.int16
+    return jnp.int32
+
+
 class RpaCase(Enum):
     """Represents the different cases for Ragged Paged Attention.
 
@@ -486,11 +506,10 @@ def _ragged_paged_attention_kernel_loop(
         if soft_cap is not None:
             s = soft_cap * jnp.tanh(s / soft_cap)
 
-        # Use int16 for span computations when safe: non-f32 dtype on TPU v6+
-        # with causal mask. Custom mask shapes can trigger a Mosaic compiler bug.
-        int_ty = jnp.int32
-        if get_dtype_packing(q.dtype) != 1 and tpu_version >= 6 and use_causal_mask:
-            int_ty = jnp.int16
+        # Use int16 only when the entire per-sequence KV capacity fits in int16.
+        # Custom mask shapes can trigger a Mosaic compiler bug, so the causal-mask
+        # guard remains part of the fast-path check.
+        int_ty = span_int_ty
         processed_q_len_int = processed_q_len.astype(int_ty)
         processed_kv_len_int = processed_kv_len.astype(int_ty)
         effective_kv_len_int = effective_kv_len.astype(int_ty)
@@ -1814,6 +1833,13 @@ def ragged_paged_attention(
 
     use_causal_mask = causal == 1
     tpu_version = get_tpu_version()
+    span_int_ty = _get_span_int_dtype(
+        q.dtype,
+        tpu_version=tpu_version,
+        use_causal_mask=use_causal_mask,
+        pages_per_seq=pages_per_seq,
+        page_size=page_size,
+    )
 
     def run_rpa_kernel(
         q,
--- a/python/sgl_jax/test/test_kernel_utils.py
+++ b/python/sgl_jax/test/test_kernel_utils.py
@@ -1,6 +1,11 @@
 import unittest
 from unittest.mock import MagicMock, patch
 
+import jax.numpy as jnp
+
+from sgl_jax.srt.kernels.ragged_paged_attention.ragged_paged_attention_v3 import (
+    _get_span_int_dtype,
+)
 from sgl_jax.srt.kernels.ragged_paged_attention.util import get_tpu_version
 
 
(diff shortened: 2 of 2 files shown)


JamesBrianD:
Hello, this issue https://github.com/sgl-project/sglang-jax/issues/1541 had fixed, please help to check it.

PR #1578, opened 2026-09-02
perf(moe): optimize EPMoE unpermute

Junyi-99:
## Motivation

> **Stacked on #1577.**

The einsum put `top_k` on the VPU's sublane axis, which for top_k < 8 cost a full relayout copy of the gathered values.

```python
reshaped_intermediate = jnp.reshape(unsort_intermediate, (weights.shape[0], top_k, -1))
output = jnp.einsum("BKE,BK -> BE", reshaped_intermediate.astype(jnp.float32), weights_fp32)
```

Reshaping the index array rather than the values, and summing over `top_k` with an unrolled loop instead of an einsum, means no `(tokens,top_k, hidden)` tensor is built. 

Related PR: https://github.com/sgl-project/sglang-jax/pull/1058

## Modifications

`python/sgl_jax/srt/layers/moe.py` — in `EPMoE._unpermute`, reshape
`argsort_indices` (`int32[tokens*top_k]`, ~128 KB) into `(tokens, top_k)` and gather
column `k` directly, then accumulate over `range(top_k)` in float32.

`test_unpermute_sums_over_top_k`, parametrised over `top_k ∈ {1, 4, 8}`. (The existing test covers `top_k = 2` only)

## Accuracy Tests

For `top_k ≤ 2`: bit-identical.

For `top_k ≥ 4`: the unrolled sequential float32 accumulation differs from XLA's reduction order by about one bfloat16 ulp. Comparing `_unpermute` output against the parent commit on identical inputs, TPU v6e-1:

| shape | max abs delta | bit-identical |
|---|---:|---|
| Grok-2 `top_k=2 hidden=8192` T=2048 | 0 | yes |
| Grok-2 `top_k=2 hidden=8192` T=8192 | 0 | yes |
| Qwen3-235B `top_k=8 hidden=4096` T=2048 | 0.0078125 | no |
| Qwen3-235B `top_k=8 hidden=4096` T=8192 | 0.03125 | no |

Please flag if bit-identity at `top_k ≥ 4` is required.

```bash
python3 -m pytest python/sgl_jax/test/layers/test_moe_unpermute.py \
                  python/sgl_jax/test/kernels/moe_block_quant_test.py \
                  python/sgl_jax/test/test_moe_weight_loader_sharding.py -q
# 12 passed
```

## Benchmarking and Profiling

`EPMoE._unpermute` alone, TPU v6e-1, bf16, median of 20 runs.

| model shape                      | tokens |    parent |       this PR |   speedup |
[...]

PR #1585, opened 2026-09-04
perf(dsa): pallas paged write-back for qblock self-write + QB=256 default (110k TTFT -28%)

zorrofox:
Follow-up to #1567, shipping its two listed natural extensions together:

## 1. Paged write-back kernel (in-place self-write)

The qblock wrapper's self-write scattered new kv rows through a flat `[Pn*ps, D]` view of the 4D cache. XLA materializes TWO full relayout copies of the pool per call for that round-trip (the 4D pool and the 2D view get different tilings) — measured at 13.8% of 110k prefill device time. 4D-scatter rewrites do not help: the flatten/unflatten pair is generated by XLA's scatter normalization itself.

`paged_write_back` is a small pallas kernel that writes rows in place: a run table over `out_cache_loc` (contiguous-run detection host-free), HBM-direct row copies, `input_output_aliases` for the pool, and a `cond` fallback to the XLA path when the run budget is exceeded. Edge tokens whose rows straddle a packing-tile word boundary go through a word-granular read-modify-write (last commit) so the flat token view's inherited packing tiling stays bit-correct.

## 2. QB default 64 -> 256

Saturated attend work scales as `(T/QB) * ctx_pages`; the QB sweep is uniformly positive from 16k to 135k with accuracy gates clean at every step, and 512 projects <2% further. `DSA_PREFILL_QBLOCK_QB` still overrides.

## Measured (GLM-5.2, TPU v7x tp16, chunked 8192 / ctx 135k)

| stage | 110k TTFT | 16k |
|---|---|---|
| #1567 as merged (QB=64) | 30.2s | 2.71s |
| + write-back | 24.43s (-19%) | 2.71s |
| + QB=256 | **21.89s (-28% cumulative)** | **2.40s (-11%)** |

- decode TPOT unaffected (decode does not run the qblock path).
- v6e-64 (tp64/dp8, W8A8): same kernels zero-change; paired throughput ON/OFF 1.49/1.33/1.40x at cc 64/256/460.
- radix cache interop: write-back pages preserve prefix-reuse semantics — 32-session x 4-round agentic probe hits 32/32 every round, warm-round latency flat, needle 25/25 on reused prefixes.

## Accuracy

- evalscope gsm8k full 1319, thinking 4-shot: **0.9757** (write-back + QB256 stack, v6e).
- paired 200q gsm8k four arms (ON/OFF x QB64/256): 0.940-0.965 within noise, 190+/200 responses byte-identical to the OFF arm.
- needle 25/25 at every stage; cc=1/8/32 compile surface clean.

## Tests

- [x] CPU (jax 0.11.1): full dsa kernel suite 49 passed / 3 skips
- [x] TPU v7x (jax 0.11.1, post-#1567 image + this branch): full dsa suite 49 passed / 3 skips
- [x] pre-commit clean

### DIFF
--- a/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill_qblock.py
+++ b/python/sgl_jax/srt/kernels/dsa/sparse_mla_prefill_qblock.py
@@ -459,9 +459,20 @@ def prefill_write_and_attend_ragged_qblock(
     row = jnp.zeros((T, Dk_pad), cache.dtype)
     row = row.at[:, :Dv].set(kvc.astype(cache.dtype))
     row = row.at[:, Dv : Dv + rope].set(kpe.reshape(T, rope).astype(cache.dtype))
-    flat = cache.reshape(Pn * ps, Dk_pad)
-    flat = flat.at[loc].set(row, mode="drop", wrap_negative_indices=False)
-    cache_new = flat.reshape(Pn, pspk, pk, Dk_pad)
+    # Scatter directly into the 4D cache. The flat-view round-trip
+    # (reshape [Pn*ps, D] -> scatter -> reshape back) materialises TWO full
+    # relayout copies of the cache per call (the 4D pool and the 2D view get
+    # different tilings) — measured 13.8% of 110k prefill device time. 3D
+    # index math on ``loc`` is free; padded loc == -1 stays dropped (negative
+    # page index with wrap_negative_indices=False).
+    cache_new = paged_write_back(
+        cache,
+        row,
+        loc.astype(jnp.int32),
+        page_size=ps,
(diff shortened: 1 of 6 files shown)


zorrofox:
The three failing 4-tpu lanes are the repo-wide main breakage tracked in #1599 (jax 0.11.1 fallout from #1574 in `test_lightning_backend` / wan2.1 e2e — #1578 fails identically with a disjoint diff), unrelated to this change.

This PR's own surface is green: CPU and 1-tpu lanes pass, and the full dsa kernel suite with this branch applied passes 49/49 on TPU v7x under jax 0.11.1. Will re-run the 4-tpu lanes once #1599 is resolved.

PR #1586, opened 2026-09-04
feat(metrics): report the speculative accept length in /get_server_info

real3stone:
## TL;DR

Speculative decoding has no health metric anyone can read: the accept length is computed
but reachable only from a log line that resets every interval, and the two benchmark
clients that ask for it get `n/a`. This exports it on `/get_server_info`, so a run that
has silently stopped benefiting from speculation becomes visible.

## Why it matters

Speculation is lossless, so if the accept length collapsed to 1.0 — every speculated token
rejected, the draft model contributing nothing but latency — accuracy would not move and
nothing in CI would notice: `test/srt/test_speculative_decoding.py` is the only end-to-end
speculative test, and its only assertion is an MMLU score (`:92`). The readers already
exist and always print `n/a` (`bench_one_batch_server.py:265`,
`bench_serving.py:2391-2392`); nothing in the tree writes the key. I also want the number as
the baseline for judging whether tree drafting (`--speculative-eagle-topk > 1`) would pay
for itself here.

## The change

- `scheduler_output_processor_mixin.py:392,395` — increment `cum_spec_accept_length` and
  `cum_spec_accept_count` alongside the per-interval pair they mirror. Both were
  initialised in `init_metrics` and never touched by anything.
- The block holding them moves out of `process_batch_result_decode` into
  `account_spec_decode_tokens` (`:373`), unchanged apart from the two new lines. It is a
  pure move, and it is what makes the increments reachable from a CPU test — speculative
  decoding otherwise needs a TPU and a draft model.
- `scheduler.py:1515` — expose the ratio in `get_internal_state`. It returns `None`, not
  `0.0`, when no verify step has run: the bench clients print any float verbatim and only
  fall back to `n/a` on `None`.
- `init_metrics` — drop `spec_num_total_accepted_tokens` / `spec_num_total_forward_ct`,
  which have no reader and no writer anywhere in the tree.

The exported value is the same quantity as the logged `accept-len`, averaged over the
whole run instead of the last interval. Floor 1.0, since every verify step commits the
bonus token; ceiling `--speculative-num-draft-tokens`.

## Test

`python/sgl_jax/test/test_spec_accept_metrics.py`, in `unit-test-cpu`: the `None`-at-zero
contract the bench clients depend on, that `account_spec_decode_tokens` moves both pairs
by the same amounts and skips retracted slots, and that `log_decode_stats` resets the
per-interval pair without touching the cumulative one.

Mutation-checked locally: returning `0.0` instead of `None`, dropping either increment,
counting retracted slots, and resetting the cumulative pair in `log_decode_stats` each
fail at least one test.

### DIFF
--- a/python/sgl_jax/srt/managers/scheduler.py
+++ b/python/sgl_jax/srt/managers/scheduler.py
@@ -71,7 +71,10 @@ from sgl_jax.srt.managers.schedule_policy import (
     PrefillAdder,
     SchedulePolicy,
 )
-from sgl_jax.srt.managers.scheduler_metrics_mixin import SchedulerMetricsMixin
+from sgl_jax.srt.managers.scheduler_metrics_mixin import (
+    SchedulerMetricsMixin,
+    compute_avg_spec_accept_length,
+)
 from sgl_jax.srt.managers.scheduler_output_processor_mixin import (
     SchedulerOutputProcessorMixin,
 )
@@ -1515,6 +1518,9 @@ class Scheduler(
     def get_internal_state(self, recv_req: GetInternalStateReq):
         ret = dict(global_server_args_dict)
         ret["last_gen_throughput"] = self.last_gen_throughput
+        ret["avg_spec_accept_length"] = compute_avg_spec_accept_length(
+            self.cum_spec_accept_length, self.cum_spec_accept_count
+        )
         ret["memory_usage"] = {
             "kvcache": round(self.token_to_kv_pool_allocator.get_kvcache().mem_usage, 2),
             "token_capacity": int(self.max_total_num_tokens),
--- a/python/sgl_jax/srt/managers/scheduler_metrics_mixin.py
+++ b/python/sgl_jax/srt/managers/scheduler_metrics_mixin.py
@@ -17,13 +17,24 @@ logger = logging.getLogger(__name__)
 RECORD_STEP_TIME = get_bool_env_var("SGLANG_RECORD_STEP_TIME")
 
 
+def compute_avg_spec_accept_length(cum_accept_length: int, cum_accept_count: int) -> float | None:
+    """Tokens committed per verify step; >= 1.0, since the bonus token always lands.
+
+    None rather than 0.0 before the first verify step: the benchmark clients
+    print any float verbatim and only fall back to "n/a" on None.
+    """
+    if cum_accept_count <= 0:
+        return None
+    return cum_accept_length / cum_accept_count
+
+
 class SchedulerMetricsMixin:
     def init_metrics(self: Scheduler):
         self.last_gen_throughput: float = 0.0
         self.last_input_throughput: float = 0.0
         self.step_time_dict = defaultdict(list)  # Dict[batch size -> step time]
-        self.spec_num_total_accepted_tokens = 0
-        self.spec_num_total_forward_ct = 0
+        # accept_token / spec_num_forward_ct count the same thing, but log_decode_stats
+        # zeroes them every interval; these two are never reset.
         self.cum_spec_accept_length = 0
         self.cum_spec_accept_count = 0
         self.total_retracted_reqs = 0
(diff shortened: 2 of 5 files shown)


python/sgl_jax/srt/managers/scheduler.py
```diff
@@ -1509,6 +1512,9 @@ def move_ready_grammar_requests(self):
     def get_internal_state(self, recv_req: GetInternalStateReq):
         ret = dict(global_server_args_dict)
         ret["last_gen_throughput"] = self.last_gen_throughput
+        ret["avg_spec_accept_length"] = compute_avg_spec_accept_length(
```
lingebeng, line 1521: Non-blocking: avg_spec_accept_length includes warmup requests and previous benchmark cases because its counters are process-lifetime cumulative. 
This is fine for now, but we should either document that scope or add per-run snapshot/reset support in a follow-up.

lingebeng (approved):

python/sgl_jax/srt/managers/scheduler.py
real3stone: Agreed. It's sharper in `bench_one_batch_server.py` than in `bench_serving.py`: that one sweeps `itertools.product(batch_size, input_len, output_len)` against a single server (`:341`) and reads `/get_server_info` at the end of every case (`:264`), so every row of the printed table is the running average over warmup plus all earlier rows, while it reads like a per-row number.

For the follow-up I'd rather expose the two raw counters next to the ratio than add a reset:

```python
ret["cum_spec_accept_length"] = self.cum_spec_accept_length
ret["cum_spec_accept_count"] = self.cum_spec_accept_count
```

A client snapshots before and after a case and divides the deltas — exact, no new endpoint, and still correct when more than one client polls, whereas a reset would race between them. The benchmark scripts would then difference around each case. I'd keep `avg_spec_accept_length` as the lifetime value so existing readers don't change meaning.

Happy to take this as a follow-up once this lands.

real3stone:
@lingebeng thanks for the review. Unrelated ask while I have you.

This metric is groundwork for tree speculative decoding (`--speculative-eagle-topk > 1`), which doesn't run end to end today — I have a bug report ready to file with the concrete failures.

The hard part is verifying a fix. `test/srt/test_speculative_decoding.py` is the only e2e speculative test, it runs `--tp-size 4`, and it lives in `accuracy-test-tpu-v6e-4` (`run_suite.py:498`) — so anything I change there needs 4 chips, and I only have a single-chip v6e. It also pins `--speculative-eagle-topk 1` today, behind the FIXME at `:52`.

Following `docs/developer_guide/tpu_resources_guide.md`, I asked for the SkyPilot API server endpoint in the SGL-JAX Slack channel but haven't had a reply — possibly I asked in the wrong place. Is that still the route for an outside contributor, or is there another one now? Even a few hours of v6e-4 would be enough to validate a fix, and I'm happy to work within whatever constraints there are on it.

python/sgl_jax/srt/managers/scheduler.py
lingebeng: Great. Thanks a lot!

lingebeng:
> @lingebeng thanks for the review. Unrelated ask while I have you.
> 
> This metric is groundwork for tree speculative decoding (`--speculative-eagle-topk > 1`), which doesn't run end to end today — I have a bug report ready to file with the concrete failures.
> 
> The hard part is verifying a fix. `test/srt/test_speculative_decoding.py` is the only e2e speculative test, it runs `--tp-size 4`, and it lives in `accuracy-test-tpu-v6e-4` (`run_suite.py:498`) — so anything I change there needs 4 chips, and I only have a single-chip v6e. It also pins `--speculative-eagle-topk 1` today, behind the FIXME at `:52`.
> 
> Following `docs/developer_guide/tpu_resources_guide.md`, I asked for the SkyPilot API server endpoint in the SGL-JAX Slack channel but haven't had a reply — possibly I asked in the wrong place. Is that still the route for an outside contributor, or is there another one now? Even a few hours of v6e-4 would be enough to validate a fix, and I'm happy to work within whatever constraints there are on it.

cc @cjx0709

PR #1603, opened 2026-09-07
[DSv4 M1.0] Bring mHC hyper-connection kernels onto epic/dsv4 (cherry-pick #1547)

JamesBrianD:
## Motivation

Stack entry point for **INFERENCE-98 (M1) — DeepSeek V4 Flash 0731 model integration**.

M1.2 (the mHC layer) needs the multi-stream hyper-connection kernels. Those are already
implemented and benchmarked in #1547, so this stack consumes them rather than reimplementing.
This PR brings #1547 (`e701b5c3c`) onto `epic/dsv4` unchanged so the rest of the stack can
build on a stable `sgl_jax.srt.kernels.mhc` API without waiting on the #1547 review cycle.

## Modifications

Cherry-pick of #1547, byte-identical — `pre` / `post` / `head` Pallas kernels, the float64
NumPy oracle, the operator-level benchmark harness, and the `run_suite.py` entry.

No new code in this PR.

## Notes for reviewers

- **This is not a substitute for reviewing #1547.** If #1547 changes during review, this stack
  rebases onto the merged version. The stack only depends on the four public symbols
  (`mhc_gates`, `mhc_pre_fused`, `mhc_post_fused`, `mhc_head_collapse_fused`).
- #1547 ships **pre / post / head** but no fused post+pre kernel. The reference implementation's
  production path is the fused seam; M1.2 therefore implements the *unfused* sequential semantics
  (which the reference explicitly documents as the equivalent semantics). Fusing the seam is
  follow-up optimization work (INFERENCE-83), deliberately out of M1 scope.

## Accuracy Tests

```bash
PYTHONPATH=python:. python -m pytest -q test/srt/kernels/mhc
```

- Local CPU (jax 0.11.1): `55 skipped` — the suite is `skipif(default_backend() != "tpu")`,
  so CPU cannot validate it. Not claiming a pass from this.
- TPU v7x: see the falcon run linked in the first review comment.

## Checklist

- [x] Please use English, otherwise it will be closed.
- [x] The purpose of the PR, or link existing issues this PR will resolve.
- [x] The test plan, such as providing test command.
- [ ] (Optional) The necessary documentation update.

PR #1604, opened 2026-09-07
[DSv4 M1.1a] DeepSeek-V4 config + the layer-classification contract

JamesBrianD:
> **Stacked PR — base is `dev/dsv4-m1.0-mhc-kernels` (#1603), not `epic/dsv4`.**
> Review only the top commit. Stack: #1603 → **this** → M1.1b → M1.2 → M1.3a → M1.3b → M1.4.

## Motivation

M1.1a of INFERENCE-98 (DeepSeek V4 Flash 0731 model integration).

Two things have to exist before any V4 code can be written: a config class, and
an agreed answer to *"what kind of attention layer is layer i?"*.

The second one is the reason this is its own PR. Three subsystems need that
answer — the model (M1), the KV/state pools (C1.1), and the attention metadata
derivation (M2.1) — and they are being built by different tasks. If each derives
it from `compress_ratios` independently they will drift, and the failure mode is
a layer attending over the wrong cache, which does not raise. So it ships once,
exported, with tests as the contract.

## Modifications

- `python/sgl_jax/srt/configs/deepseek_v4.py`
  - `DeepseekV4Config` — defaults are the shipped Flash 0731 values, so
    `DeepseekV4Config()` is a usable fixture with no checkpoint present.
  - `classify_layers(config)` — **the exported contract.** Semantics follow
    upstream vLLM's `_layer_type_for`: `<=1` → SWA-only, `4` → CSA, `128` → HCA.
    Any other ratio raises; a silent fallback would be a miscompiled layer.
  - `trunk_compress_ratios(config)` — truncates to `num_hidden_layers`.
  - `hash_moe_layer_flags(config)` — first `num_hash_layers` layers route by token id.
  - `mhc_param_shapes(config)` — for M1.1b (weight mapping) and M1.2 (the layer).
- Registered in `hf_transformers_utils._CONFIG_REGISTRY` (+2 lines).
- Registered the test in the `unit-test-cpu` suite (+5 lines).

## Two things worth knowing, both found by writing the tests

**1. `compress_ratios` is longer than the trunk.** The shipped list has **46
entries against `num_hidden_layers=43`**. The three trailing zeros are not
layers. A consumer that iterates the raw list gets three phantom SWA-only
layers *and* misreads the real last layer — which is C4A — as SWA-only.
`trunk_compress_ratios` is the guard, and `test_shipped_ratios_are_longer_than_the_trunk`
is the regression test. Resulting classification: 2 SWA-only + 21 C4A + 20 C128A.

**2. `layer_types` is a reserved name.** transformers' `PretrainedConfig`
validates and rewrites `layer_types` (it expects HF's
`"sliding_attention"`/`"full_attention"` strings), and sgl-jax's
`ModelConfig.get_hybrid_layer_counts` reads it with exactly those semantics.
Exposing our enum tuple under that name makes the config **fail to construct**.
Ours is `attention_layer_types`; there is a test pinning it.

Also: `rope_scaling` defaults to the shipped YaRN dict rather than `None`,
because transformers rewrites a `None` into `{"rope_type": "default"}`, which
silently drops the 16x context extension (1M → 64K).

## Scope

`dspark_*` and `num_nextn_predict_layers` are accepted and round-tripped but not
consumed — DSpark is INFERENCE-84, MTP is out of M1. Not dropping them keeps a
config round-trip lossless.

## Accuracy Tests

```bash
PYTHONPATH=python:. python -m pytest -q python/sgl_jax/test/configs/test_deepseek_v4_config.py
```

`25 passed` on CPU (jax 0.11.1). Includes
`test_defaults_match_the_pinned_upstream_config_if_present`, which diffs **every**
default against a committed copy of the real
`deepseek-ai/DeepSeek-V4-Flash-0731@7872f01b` `config.json` — so a wrong default
cannot pass. No TPU needed for this PR.

## Checklist


JamesBrianD:
Closing — moved to `primatrix/sglang-jax`: this PR is now primatrix#342, stacked on primatrix#341. See #1603 for the rationale.

Issue #1460, opened 2026-07-13
Development Roadmap (2026 Q3)


JamesBrianD:
Here is the development roadmap of SGL-JAX for 2026 Q3.

Contributions and feedback are welcome!

- Open a new [discussion](https://github.com/sgl-project/sglang-jax/discussions/new)
- Join the [SGL-JAX Slack Channel](https://sgl-fru7574.slack.com/archives/C09EBE5HT5X).

## Roadmap semantics

- This roadmap reflects the capability of the `main` branch. Items stay unchecked until the corresponding implementation and validation are merged.
- Remaining work from the [2026 Q2 roadmap](https://github.com/sgl-project/sglang-jax/issues/909) was re-audited against `main` on 2026-07-26. Only work that is still incomplete is carried below.
- Linked issues define scope and acceptance criteria. Broad roadmap entries without a dedicated issue should get one before implementation starts.

## Focus

- **Runtime and Architecture:** Upgrade the JAX/libtpu stack and improve scheduler, data-parallel, and cache architecture.
- **PD Disaggregation and HiCache:** Adopt `tpu-raiden` and expand compatibility with major serving features.
- **Model and VLM Support:** Complete key text and multimodal models on TPU.
- **Kernel Optimization:** Develop and optimize attention, collective-matmul, and model-specific kernels for TPU v6e/v7.
- **Community and Infrastructure:** Improve evaluation, nightly CI, developer workflows, and reusable AI skills.

## Runtime

- [x] Upgrade to JAX `0.10.1` and libtpu `0.0.42.1`; update the scope of [#1025](https://github.com/sgl-project/sglang-jax/issues/1025) from the older 0.9.2 target and resolve the logprob blocker [#584](https://github.com/sgl-project/sglang-jax/issues/584). https://github.com/sgl-project/sglang-jax/pull/1461
- [ ] Validate the new stack across supported TPU platforms, Pallas kernels, model accuracy, and CI.
- [ ] Continue the Q2 TPU-CPU transfer-pattern refactor: reduce blocking `device_get` paths in scheduler/speculative decode and validate the Pathways proxy fast path ([#772](https://github.com/sgl-project/sglang-jax/issues/772), [#774](https://github.com/sgl-project/sglang-jax/issues/774), [#1452](https://github.com/sgl-project/sglang-jax/issues/1452)).

## PD Disaggregation and HiCache

**Owner: @cjx0709**  
[...]

PR #1607, opened 2026-09-08
fix(dsa): streamindex_topk compressed visibility follows completed groups (ratio > 1)

zorrofox:
## Motivation

`streamindex_topk` bounds the causal window in compressed space with `k_span <= q_pos // compression_ratio`. For `compression_ratio > 1` that admits the entry the query itself sits inside. The compressor only writes an entry when its group completes, so inside a prefill chunk a query at position `p` with `(p + 1) % ratio != 0` could select a slot whose content is produced by tokens *after* `p`, and the same query selected different entries depending on where its chunk happened to end. Nothing in the repo hits this today because the only caller (`dsa_sparse_backend`, GLM-5.2) passes `compression_ratio=1`, where both rules coincide.

DeepSeek-V4's reference implementation (`deepseek-ai/DeepSeek-V4-Flash`, `inference/model.py`, `Indexer.forward`) masks with `arange(seqlen // ratio) >= arange(1, seqlen + 1) // ratio`: entry `e` is visible only if `e < (p + 1) // ratio`, i.e. once its last token `(e + 1) * ratio - 1` is at or before the query. This PR makes the kernel use that bound so it can serve as the V4 CSA indexer (`compression_ratio=4`). Context: sgl-project/sglang-jax#1460 (V4 roadmap item); the V4 indexer module being built on `epic/dsv4` currently reimplements the scoring in plain JAX precisely because of this mask difference and notes it can become a thin wrapper over this kernel once the bound is fixed.

## Modifications

- `python/sgl_jax/srt/kernels/dsa/streamindex_topk.py`: `bq_pos_compressed = (q_pos + 1) // compression_ratio - 1` in both grids (was `q_pos // compression_ratio`); docstring states the visibility rule. `kv_len = seq_len // compression_ratio` is unchanged. Identity for `compression_ratio == 1`.
- `test/srt/kernels/dsa/test_streamindex_topk.py`: the shared NumPy oracle and the fp8 packed-cache case now encode the same rule (the fp8 case previously had no causal mask at all and relied on every entry being visible). New tests: `test_streamindex_topk_compressed_visibility_is_completed_groups` (ratio 4; prefill chunk, mixed decode+prefill, chunk end on a group boundary, prefill from position 0; asserts the exact visible set per query) and `test_streamindex_topk_selection_does_not_depend_on_chunk_end` (same query, chunk ending at 17 vs 19, identical selection).
- `streamindex_page_topk` (page mode) is untouched; it already rejects `compression_ratio != 1`.

## Accuracy Tests

TPU v7x, single host (8 devices), jax/jaxlib 0.11.1:

```
# before the kernel change (tests only): 7 failed, 24 passed
#   every failure = one extra entry selected, the query's own in-progress group
# after:
python -m pytest test/srt/kernels/dsa/test_streamindex_topk.py test/srt/kernels/dsa/test_streamindex_page_topk.py -q
31 passed in 27.21s
```

Full `test/srt/kernels/dsa/` directory on the same host (`JAX_PLATFORMS=tpu,cpu`, the qblock/ref cases pin a CPU backend): `48 passed, 3 skipped in 59.29s`.

`pre-commit run --files <changed>` clean (isort / ruff / black).

## Benchmarking and Profiling

No performance change: one scalar expression per query block differs; no kernel structure, tiling, or memory traffic changes.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/kernels/dsa/streamindex_topk.py
+++ b/python/sgl_jax/srt/kernels/dsa/streamindex_topk.py
@@ -425,7 +425,10 @@ def _scores_kernel(
                     + bq_idx * bq_sz
                     + jnp.arange(bq_sz, dtype=jnp.int32)
                 )
-                bq_pos_compressed_vec.append(q_pos // compression_ratio)
+                # Last visible compressed entry: the one whose final token
+                # (entry+1)*ratio-1 is at or before q_pos, i.e. entries
+                # [0, (q_pos+1)//ratio). Identity for ratio == 1.
+                bq_pos_compressed_vec.append((q_pos + 1) // compression_ratio - 1)
 
             # Wait for cur bq if not ready yet
             wait_fetch_bq(batch_start_seq_idx, bq_idx, bq_sem_idx)
@@ -496,7 +499,10 @@ def _scores_kernel(
                     + bq_idx * bq_sz
                     + jnp.arange(bq_sz, dtype=jnp.int32)
                 )
-                bq_pos_compressed_vec.append(q_pos // compression_ratio)
+                # Last visible compressed entry: the one whose final token
+                # (entry+1)*ratio-1 is at or before q_pos, i.e. entries
+                # [0, (q_pos+1)//ratio). Identity for ratio == 1.
+                bq_pos_compressed_vec.append((q_pos + 1) // compression_ratio - 1)
 
             wait_fetch_bq(batch_start_seq_idx, bq_idx, bq_sem_idx)
             bq_vec = load_bq(bq_sem_idx)
@@ -657,7 +663,10 @@ def streamindex_topk(
         sequences[i:j] are chunked-prefill-only, and sequences[j:k] are mixed. The
         k is also the total number of sequences.
       k: Number of top-K elements to retrieve.
-      compression_ratio: KV cache compression ratio.
+      compression_ratio: KV cache compression ratio. Compressed entry ``e``
+        covers original positions ``[e*ratio, (e+1)*ratio-1]`` and is visible
+        to a query only once its last token is at or before the query
+        position (``e < (q_pos+1)//ratio``, the DeepSeek-V4 indexer rule).
       num_kv_pages_per_block: number of kv pages to be processed in one block in
         the pallas kernel. This is a tuple of (decode, prefill, mixed) cases.
       num_queries_per_block: number of queries to be processed in one block in the
--- a/test/srt/kernels/dsa/test_streamindex_topk.py
+++ b/test/srt/kernels/dsa/test_streamindex_topk.py
@@ -93,7 +93,11 @@ def _numpy_topk_oracle(
             q_abs_pos = (S_total - T_seq) + t_idx
 
             for s_idx in range(S_valid):
-                if s_idx * comp_ratio > q_abs_pos:
+                # Compressed entry ``s_idx`` covers original positions
+                # [s_idx*ratio, (s_idx+1)*ratio-1]; it is visible only once its
+                # last token is at or before the query (DeepSeek-V4
+                # ``inference/model.py`` Indexer mask).
+                if (s_idx + 1) * comp_ratio - 1 > q_abs_pos:
                     continue
 
                 score = 0.0
@@ -455,7 +459,10 @@ def test_streamindex_topk_quantized():
     naive_scores = np.full((T_seq, max(S_valid, k)), -np.inf, dtype=np.float32)
 
     for t_idx in range(T_seq):
+        q_abs_pos = (S_seq - T_seq) + t_idx
         for s_idx in range(S_valid):
+            if (s_idx + 1) * comp_ratio - 1 > q_abs_pos:
+                continue
             score = 0.0
             for h in range(H_I):
                 h_kv = h // (H_I // H_KV)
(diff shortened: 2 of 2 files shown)


zorrofox:
The `unit-test-cpu` failure here is the pre-existing main breakage tracked in #1608 (`test_spec_accept_metrics.py` vs `SchedulerMetricsMixin.init_metrics` reading `server_args`, from #1555 x #1586); this PR touches only `kernels/dsa/streamindex_topk.py` and its TPU-only test. I'll hold off on rebasing so the queued TPU jobs are not restarted; happy to rebase once #1608 lands if you prefer that over overriding the CPU check.

zorrofox:
@cjx0709 could you take a look at this one when you have a moment? It is a correctness fix to the `streamindex_topk` kernel from #1515/#1545: for `compression_ratio > 1` the causal bound admitted the query's own in-progress compressed group, which does not match DeepSeek-V4's reference indexer mask. For `compression_ratio == 1` (GLM-5.2, the only current caller) the two rules are identical, so there is no behaviour change on main. CI is fully green after the rebase onto #1609, and the v7x results are in the description.

zorrofox:
@KoalaYuFeng your `integration/deepseek-v4` branch (270dfcbd, `kernels/dsa/streamindex_topk.py`) switches the compressed visibility rule to `(q_pos + 1) // compression_ratio - 1` under `numerical_mode="v4"`, which is exactly the change in this PR (and the official `inference/model.py` rule `e < (p + 1) // ratio`). Since you arrived at the same rule independently on a full 43-layer deployment, would you be willing to review this PR so the fix lands on main and your branch can drop the local patch? Two data points from your side would be especially useful: whether your GSM8K/HumanEval runs were on the completed-groups rule, and whether you saw any case where the old `q_pos // compression_ratio` rule was intentional.

Related: #1642 replaces the kernel's exit stage with an exact SparseCore radix-select behind `topk_backend="auto"`. Your `v4` mode uses `jax.lax.top_k` at that exit; on TPU7x at E=262144, k=512 we measured `lax.top_k` at 6.5 ms vs 4.1 ms for the current `approx_max_k(recall_target=1.0)` and 0.86 ms for the SparseCore path, all three returning the same index set. With #1642 the exact ordering you want comes for free, so the `lax.top_k` branch should not be needed.

The bf16 rounding points you insert in `compute_scores` to match the official numerics look like a good small follow-up on top of #1607 (a `numerical_mode` or similar static flag on the kernel). If you plan to upstream that, or the CSA kernels more broadly, I am happy to review; splitting kernels from the model/pool integration, as was done for #1547/#1549, would make that path much faster.

JamesBrianD:
Thanks for you fixing

JamesBrianD (approved):

KoalaYuFeng:
> @KoalaYuFeng your `integration/deepseek-v4` branch ([270dfcb](https://github.com/sgl-project/sglang-jax/commit/270dfcbd43631728fb8b58621dee84b49284d710), `kernels/dsa/streamindex_topk.py`) switches the compressed visibility rule to `(q_pos + 1) // compression_ratio - 1` under `numerical_mode="v4"`, which is exactly the change in this PR (and the official `inference/model.py` rule `e < (p + 1) // ratio`). Since you arrived at the same rule independently on a full 43-layer deployment, would you be willing to review this PR so the fix lands on main and your branch can drop the local patch? Two data points from your side would be especially useful: whether your GSM8K/HumanEval runs were on the completed-groups rule, and whether you saw any case where the old `q_pos // compression_ratio` rule was intentional.
> 
> Related: #1642 replaces the kernel's exit stage with an exact SparseCore radix-select behind `topk_backend="auto"`. Your `v4` mode uses `jax.lax.top_k` at that exit; on TPU7x at E=262144, k=512 we measured `lax.top_k` at 6.5 ms vs 4.1 ms for the current `approx_max_k(recall_target=1.0)` and 0.86 ms for the SparseCore path, all three returning the same index set. With #1642 the exact ordering you want comes for free, so the `lax.top_k` branch should not be needed.
> 
> The bf16 rounding points you insert in `compute_scores` to match the official numerics look like a good small follow-up on top of #1607 (a `numerical_mode` or similar static flag on the kernel). If you plan to upstream that, or the CSA kernels more broadly, I am happy to review; splitting kernels from the model/pool integration, as was done for #1547/#1549, would make that path much faster.

@zorrofox Thanks for upstreaming this! I’ve reviewed the PR, and it looks good to me. Our full GSM8K/HumanEval runs used the same completed-groups rule via numerical_mode="v4", and we haven’t found any V4 case where the old bound was intentional. Since this is now merged, we’ll adopt the shared implementation and remove our duplicate patch after regression testing.

Regarding #1642, thanks for your contribution and we’d be happy to reuse the shared selector. On v6e, I reproduced the ~38% decode / ~62% prefill latency reductions at 1M context. However, our 8K-context, 128-query full-indexer test took 0.334 ms with XLA versus 0.657 ms with SparseCore, with profiling pointing to per-row cooperation and synchronization overhead. The 2048 valid candidates are padded to 8192 columns, triggering the current auto threshold. Have you observed similar short-context regressions on your side? We’d appreciate your thoughts on how best to handle these cases when adopting the shared selector. 

Regarding #1607, we’ll follow your suggestion to split the kernels from the model/pool integration. We plan to upstream the BF16 rounding adjustments as a small follow-up with a static numerical_mode flag, then submit the CSA kernels separately with standalone correctness tests and benchmarks. Thanks for offering to review!

PR #1610, opened 2026-09-08
perf(vlm): simplify preprocessing and flatten vision inputs

Yanko-7:
## Motivation

Reduce CPU preprocessing and host dispatch overhead in Qwen2.5-VL and Qwen3-VL serving. Media loading and Hugging Face processing currently use separate execution stages, while vision inputs and layout metadata require additional preparation around encoder execution. This change combines preprocessing in one worker task and prepares flat, explicitly sharded encoder inputs on the host.

Additionally, processing `bfloat16` in `pack_lanes` was observed to be slower on CPU compared to `float32`. Therefore, image preprocessing on the host is kept in `float32`, and the tensors are cast to the target model weight `dtype` at the encoder input after `device_put` to the accelerator.

## Modifications

- Load media and run the Hugging Face processor in the same multimodal processor worker. Remove the separate I/O executor and `--mm-io-worker-num`; worker concurrency is controlled by `--mm-processor-worker-num`.
- Use TorchCodec for Qwen image decoding with a Pillow fallback, pybase64 for base64 decoding, and xxhash for multimodal item hashing. Add the required dependencies and use the fast Hugging Face processor.
- Pack vision patches and metadata into flat buffers with explicit sharding. Build layout metadata on the host, use Numba for output-index construction, and restore the original item order with padding rows masked to zero.
- Keep image tensors in `float32` during host preprocessing/`pack_lanes` to avoid CPU `bfloat16` performance penalties, deferring the cast to the model `dtype` at the vision encoder input after device placement.
- Adapt Qwen2.5-VL, Qwen3-VL, attention, and vision precompilation to the flat input layout, retaining token-major outputs and Qwen3-VL DeepStack features.

## Benchmarking and Profiling

Qwen3-VL-32B-Instruct on a single TPU v7x-8, DP4 × effective TP2, with BF16 model and KV cache. The single-image workload uses one random 512×512 JPEG per request, a 2K context, an unbounded request rate, and 500 output tokens per request. Both cases complete 1,000 requests with identical server-verified input and output token totals.

Single-image multimodal serving reaches **8,527.41 output tok/s**, or **95.13%** of text-only throughput.  These are text-only versus multimodal measurements from the same comparison. TTFT includes scheduler queueing under the burst workload.

| Metric | Text | Single-image multimodal |
|---|---:|---:|
| Successful requests | 1,000 | 1,000 |
| Input tokens (server-verified) | 1,344,206 | 1,344,206 |
| Output tokens (server-verified) | 500,000 | 500,000 |
| Duration | 55.777 s | 58.634 s |
| Request throughput | 17.928 req/s | 17.055 req/s |
| Input token throughput | 24,099.43 tok/s | 22,925.18 tok/s |
| Output token throughput | 8,964.19 tok/s | 8,527.41 tok/s |
| Total token throughput | 33,063.62 tok/s | 31,452.59 tok/s |
| Mean TTFT | 13,820.48 ms | 15,937.96 ms |
| Median TTFT | 10,933.37 ms | 13,425.55 ms |
| P99 TTFT | 46,881.43 ms | 49,192.05 ms |
| Mean TPOT | 64.89 ms | 65.44 ms |
| Median TPOT | 68.54 ms | 68.36 ms |
| P99 TPOT | 82.26 ms | 90.18 ms |
| Mean E2E latency | 46,201.97 ms | 48,591.19 ms |
| Median E2E latency | 45,290.66 ms | 47,598.53 ms |
| P99 E2E latency | 55,445.41 ms | 57,846.17 ms |
| Mean ITL | 66.32 ms | 66.64 ms |
| Median ITL | 32.77 ms | 33.06 ms |
| P95 ITL | 88.20 ms | 62.41 ms |
| P99 ITL | 371.05 ms | 358.61 ms |

## Checklist

### DIFF
--- a/python/sgl_jax/srt/models/qwen2_5_vl.py
+++ b/python/sgl_jax/srt/models/qwen2_5_vl.py
@@ -9,7 +9,7 @@ import jax.numpy as jnp
 import numpy as np
 from flax import nnx
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
-from jax.typing import ArrayLike
+from numba import njit, types
 from transformers import modeling_flax_utils
 
 from sgl_jax.srt.configs.model_config import ModelConfig
@@ -31,7 +31,6 @@ from sgl_jax.srt.multimodal.in_model.lane_packing import (
     run_mrope_vision_model,
 )
 from sgl_jax.srt.multimodal.layers.attention.flash_attention_backend import (
-    VisionAttentionMetadata,
     make_vision_attention_backend,
 )
 from sgl_jax.srt.multimodal.layers.vision_sharding import (
@@ -53,7 +52,7 @@ def _apply_rotary_pos_emb_vision(
     cos: jax.Array,
     sin: jax.Array,
 ) -> jax.Array:
-    """Apply precomputed vision RoPE to ``x[B, T, heads, head_dim]``."""
+    """Apply precomputed vision RoPE to ``x[tokens, heads, head_dim]``."""
     half_dim = x.shape[-1] // 2
     x_real, x_imag = x[..., :half_dim], x[..., half_dim:]
     return jnp.concatenate(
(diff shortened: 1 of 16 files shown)


python/sgl_jax/srt/models/qwen2_5_vl.py
```diff
@@ -426,62 +419,195 @@ def __init__(
...
+                                            metadata[lane, position] = y * merge + dy
+                                            metadata[lane, position + 1] = x * merge + dx
+                                            patch_offset += 1
+                                    unit_offset += 1
+                            metadata[lane, window_start + window_segment] = patch_offset
+                            window_segment += 1
+                    metadata[lane, full_start + frame_segment] = patch_offset
+                    frame_segment += 1
+            # Repeated final ends describe empty segments in the bucket padding.
+            metadata[lane, window_start + window_segment : full_start] = patch_offset
```
lianfang0320: Combining metadata into a single buffer makes sense for reducing the overhead of passing and handling separate arrays. 
One potential concern is that the current packing and unpacking code relies on matching offsets and field order, which can make write/read consistency harder to maintain and the data flow harder to follow.
The logical field structure could be kept separate from the flat transfer format. Would it be worth considering a model-specific `NamedTuple` for the unpacked fields? This could improve readability without changing the buffer format or adding a larger layout abstraction.

Yanko-7: modified to a variable-length parameter dictionary and expanded it in the encode function.

lianfang0320 (approved):
LGTM

PR #1614, opened 2026-09-09
Add experimental Raiden PD scheduler overlap

cjx0709:
## Changes

- Add **default-off**, per-role `--disaggregation-enable-overlap-schedule` for Raiden PD: immutable prefill chunk handoffs, decode future-token pipeline, and safe KV ownership through cancellation/retraction.
- Keep a whole-pool fence before new native receive admission. Regular decode is nearly contiguous; admission and prefill chunk gaps remain. This is **single JAX process per role only**; multi-host/SPMD overlap is not validated or enabled.
- Retain output-only logprob delta/batch-alignment fixes: PD tokenization selects this path, and resending history duplicates streamed logprobs. Remove timing changes, debug counters, feature docs, and development/Falcon scripts; keep a small CPU UT set and manual reproduction tools under `scripts/disaggregation/`.

## Evidence and limits

Qwen3-30B-A3B; one v7x-8 pod, TP4/DP1/EP4 per role. Three paired runs, 60 s continuous warmup + 180 s measurement each; throughput counts in-window output tokens.

| Input/output | Concurrency | PD overlap OFF tok/s | ON tok/s | Paired gain | TTFT p95 OFF → ON |
| 4K/256 | 16 | 1047.7 | 1409.1 | +34.5% | 286 → 220 ms |
| 4K/256 | 32 | 1805.1 | 2218.5 | +23.3% | 251 → 844 ms |
| 16K/256 | 16 | 404.6 | 424.8 | +5.0% | 6.437 → 7.959 s |
| 16K/256 | 32 | 404.4 | 423.1 | +4.6% | 15.950 → 17.211 s |
| Mixed 4K/16K × 256/1024 | 16 | 906.2 | 1180.2 | +30.2% | 1.376 → 1.360 s |

- Throughput/latency entries are medians; gains are medians of paired ratios. Higher throughput can worsen TTFT. Mixed assigns equal persistent slots, not equal completion counts.
- Prior 20-minute mixed soak: 3,523 requests matched non-PD references, full KV/slot recovery. These device results precede diagnostic-only cleanup; revised manual scripts have not been rerun on TPU.
- **Single process per role only. Multi-host/SPMD remains unvalidated.** Regular decode is nearly contiguous, but admission creates 5–10 ms gaps. No physical DMA-overlap claim.

Device forward continuity — actual XPlane intervals

Blue: model forward; orange: other JITs; red: host admission fence. The baseline is PD overlap OFF, not non-PD. Device events cover 2.401 s of the roughly 5 s host trace; the missing tail is not idle, and op spans cannot exclude internal DMA waits.

## Reproduce / use

After building/installing Raiden below, run this **manual** entry point on an idle, exclusively assigned v7x-8 host/pod:

```bash
python scripts/disaggregation/pd_overlap_benchmark.py \
  --model /path/to/Qwen3-30B-A3B --host-ip "$(hostname -I | awk '{print $1}')" \
  --output /tmp/pd-overlap-repro --repetitions 3 --warmup 60 --duration 180 --soak-seconds 1200
```

It starts/stops its own bootstrap/P/D/router processes, checks baseline vs overlap token equality and KV recovery, runs 30 steady windows in alternating order, then a mixed C32 soak. Outputs: `summary.md`, `results.json`, exact commands, environment and logs. No resource allocation or CI registration. Allow several hours; the soak checks stream completion/recovery, not non-PD token equality. Profile collection and multi-host tests are outside this entry point.

<details>
<summary>1. Build and install the matching Raiden wheel</summary>

Use a Linux x86_64 TPU environment with Python 3.12, Git, curl and C++ build prerequisites; allow ample disk space for Bazel. The existing build script from #1602 applies the JAX compatibility patch and bootstraps Bazel. These commands build locally; an existing persistent cache can be reused after checksum verification.

```bash
git clone -b develop/raiden-pd-overlap https://github.com/sgl-project/sglang-jax.git
cd sglang-jax
export SGL_JAX_ROOT="$PWD"
python3.12 -m venv .venv
source .venv/bin/activate
python -m pip install -U pip
python -m pip install torch torchvision 'torchcodec>=0.16.0' --index-url https://download.pytorch.org/whl/cpu
python -m pip install -e 'python[tpu]' 'jax==0.11.1' 'jaxlib==0.11.1' 'libtpu==0.0.46.1' 'flax==0.12.9'

export RAIDEN_SRC="$(dirname "$SGL_JAX_ROOT")/tpu-sync-pd"
export RAIDEN_CACHE_ROOT="$HOME/.cache/sglang-jax-raiden"
git clone https://github.com/google/tpu-sync.git "$RAIDEN_SRC"
git -C "$RAIDEN_SRC" checkout 6d43141191210b73359d22743be619532ad587dc
PYTHON_BIN=python HERMETIC_PYTHON_VERSION=3.12 bash scripts/disaggregation/build_raiden_wheel.sh
CACHE_DIR="$RAIDEN_CACHE_ROOT/6d43141191210b73359d22743be619532ad587dc-jax0.11.1-jaxlib0.11.1-libtpu0.0.46.1"
(cd "$CACHE_DIR" && sha256sum -c SHA256SUMS)
python -m pip install --no-deps "$CACHE_DIR/$(cat "$CACHE_DIR/READY")"
python -c 'from sgl_jax.raiden import preload_raiden; preload_raiden()'
```

The loader imports `tpu_sync` before JAX. A wheel built against other JAX/XLA headers is not interchangeable. The build cache's `READY` file identifies a completed wheel build, not serving/device validation.

</details>

<details>
<summary>2. Launch bootstrap, prefill, decode and router separately</summary>

Run these commands in the same Bash shell on **one v7x-8 host/pod** (four physical chips, eight JAX devices), using a local Qwen3-30B-A3B checkpoint. This layout reproduces the validated single-pod scope, not multi-host SPMD.

```bash
export MODEL_PATH=/path/to/Qwen3-30B-A3B
export PD_HOST_IP="$(hostname -I | awk '{print $1}')"
export ALLOW_MULTIPLE_LIBTPU_LOAD=true
export TPU_CHIPS_PER_PROCESS_BOUNDS=1,2,1 TPU_PROCESS_BOUNDS=1,1,1
unset TPU_CHIPS_PER_HOST_BOUNDS TPU_HOST_BOUNDS TPU_MESH_CONTROLLER_ADDRESS TPU_MESH_CONTROLLER_PORT ENABLE_MULTI_NUMA
export NO_PROXY='*' no_proxy='*'
mkdir -p /tmp/pd-overlap

COMMON=(--model-path "$MODEL_PATH" --host 0.0.0.0 --nnodes 1
  --tp-size 4 --dp-size 1 --ep-size 4 --moe-backend epmoe
  --dtype bfloat16 --kv-cache-dtype bf16 --attention-backend fa
  --page-size 128 --chunked-prefill-size 2048 --context-length 32768
  --max-running-requests 32 --max-total-tokens 262144 --mem-fraction-static 0.8
  --disable-radix-cache --skip-server-warmup
  --precompile-bs-paddings 1 4 16 32 --precompile-token-paddings 128 512 2048
  --disaggregation-use-raiden --disaggregation-host-ip "$PD_HOST_IP"
  --disaggregation-bootstrap-url "http://$PD_HOST_IP:8998"
  --disaggregation-max-inflight-transfers 4
  --disaggregation-enable-chunk-prefill-transfer)
OVERLAP=() # baseline first; set (--disaggregation-enable-overlap-schedule) for overlap

python -m sgl_jax.srt.disaggregation.run_bootstrap --host 0.0.0.0 --port 8998 \
  > /tmp/pd-overlap/bootstrap.log 2>&1 &
BOOTSTRAP_PID=$!
# Wait for bootstrap /health to return 200 before starting P/D.
curl -f "http://$PD_HOST_IP:8998/health"

TPU_VISIBLE_CHIPS=0,1 JAX_COMPILATION_CACHE_DIR=/tmp/pd-overlap/cache-p \
  python -m sgl_jax.launch_server "${COMMON[@]}" "${OVERLAP[@]}" \
  --disaggregation-mode prefill --port 30000 \
  --disaggregation-transfer-port 30001 --disaggregation-side-channel-port 9600 \
  > /tmp/pd-overlap/prefill.log 2>&1 &
P_PID=$!

TPU_VISIBLE_CHIPS=2,3 JAX_COMPILATION_CACHE_DIR=/tmp/pd-overlap/cache-d \
  python -m sgl_jax.launch_server "${COMMON[@]}" "${OVERLAP[@]}" \
  --disaggregation-mode decode --port 30010 \
  --disaggregation-transfer-port 30011 --disaggregation-side-channel-port 9700 \
  > /tmp/pd-overlap/decode.log 2>&1 &
D_PID=$!
# Wait for both /health endpoints to return 200; model load/compilation takes time.
curl -f http://127.0.0.1:30000/health
curl -f http://127.0.0.1:30010/health

python -m sgl_jax.srt.disaggregation.launch_router --pd-disaggregation --mini-lb \
  --prefill "http://$PD_HOST_IP:30000" 8998 --decode http://127.0.0.1:30010 \
  --prefill-bootstrap-host "$PD_HOST_IP" --max-concurrent-requests 32 \
  --host 0.0.0.0 --port 30020 > /tmp/pd-overlap/router.log 2>&1 &
ROUTER_PID=$!
curl -f http://127.0.0.1:30020/health
```

Overlap can be enabled on either role independently. Chunk transfer is a separate flag; keep chunk settings consistent between P/D. For another model, adjust TP/EP, attention backend and memory limits rather than treating these as generic defaults.

</details>

## CPU validation

- 32 core overlap UT cases; removed 42 redundant/parameterized cases. Full related CPU regression: **553 passed, 12 skipped, 12 subtests passed**. No TPU/e2e CI jobs added.
- Manual tools passed local HTTP token-window/truncation checks and mocked 30-window orchestration; no new TPU result is claimed.
- Logprob handling follows upstream `sglang`'s delta + advancing-offset + empty-row contract. The JAX output-only branch needs this explicitly because the tokenizer accumulates with `extend()`.

```bash
PYTHONPATH=python python -m pytest -q \
  test/srt/disaggregation/test_pd_{decode_overlap,prefill_overlap,overlap_args,logprob_stream}.py
```

For an already-running idle P/D pair, use `scripts/disaggregation/pd_overlap_check.py --model "$MODEL_PATH" --reference /tmp/pd-reference.json --record-reference` with overlap OFF; restart P/D with overlap ON and run the same command without `--record-reference`. Manual scripts require TPU serving and are intentionally outside `test/` and CI.


Yanko-7 (dismissed):

python/sgl_jax/srt/disaggregation/prefill.py
```diff
@@ -231,6 +253,120 @@ class SchedulerDisaggregationPrefillMixin:
...
+            self.process_input_requests(self.select_dp_for_request(recv_reqs))
+            if self._engine_paused:
+                self._drain_disagg_prefill_overlap_results()
+                continue
+
+            self._wait_donation_safe()
+            self.send_kv_chunk()
+            # PD producers never merge into the local decode running batch.
+            # Their chunk continuation is owned by chunked_reqs/ChunkCache.
+            self.last_batch = None
```
zorrofox, line 348: **retired producer can be resurrected by the next AbortReq / pause**

(Anchoring here because `_retire_chunk_producer_ownership` at line 741 is outside the diff.) It only clears `last_batch.reqs_info[dp].chunked_req`. In `event_loop_overlap_disagg_prefill`, `last_batch` is set to `None` at line 338 before scheduling and only becomes the launched batch at line 368, so at resolve time (`_resolve_disagg_prefill_result`, called between those two lines) `last_batch is None` and the branch is skipped. The just-launched batch is in `self.cur_batch` and its `reqs_info[dp].chunked_req` still points at the retired request. At the end of the tick it becomes `last_batch`; the next `AbortReq` (any rid, `abort_request` line 2928) or `pause_generation` (line 3041) runs `_sync_chunked_req_owners`, which re-installs it into `chunked_reqs`, and the request gets scheduled for one more chunk before `_raiden_handoff_chunk` retires it again. Compute count stays >0 across this, so the slot is not released early and nothing corrupts, but each such event costs a wasted chunk forward on a request whose sender already failed or was aborted.

Suggestion: also clear `self.cur_batch.reqs_info[dp].chunked_req` here (same guard as `last_batch`), or have `_sync_chunked_req_owners` skip requests whose sender is already terminal / `to_finish` is set.

python/sgl_jax/srt/disaggregation/prefill.py
```diff
@@ -231,6 +253,120 @@ class SchedulerDisaggregationPrefillMixin:
...
+        chunk_transfer = self.server_args.disaggregation_enable_chunk_prefill_transfer
+        handoffs = []
+        for req in _batch_reqs(batch):
+            if req.bootstrap_room is None:
+                continue
+            end = len(req.fill_ids)
+            start = end - req.extend_input_len
+            is_final = not any(req is r for r in chunked)
+            pages = (
+                self._extract_req_block_ids_range(req, start if chunk_transfer else 0, end)
```
zorrofox: **invariant failure escalates from per-request to process-wide**

`_extract_req_block_ids_range` raises `ValueError` on an invalid range or an unaligned `start`. In the normal loop that call sits inside the `try` in `_raiden_handoff_chunk`, so a violation fails one request (`sender.fail` / `_abort_prefill_req`). Here it is called from the snapshot, outside any `try`, directly in the event loop, so the same violation exits the P scheduler and drops every in-flight request on the role. Startup validation makes this unreachable under a sane config (`chunked_prefill_size % page_size == 0`), so it is a robustness point rather than a live bug, but the defensive checks exist for a reason. Wrapping the per-request snapshot in the same failure path (`_abort_prefill_req` with a `chunk_handoff`-style reason) would keep the blast radius per request.

python/sgl_jax/srt/disaggregation/prefill.py
```diff
@@ -231,6 +253,120 @@ class SchedulerDisaggregationPrefillMixin:
...
+            if self._engine_paused:
+                self._drain_disagg_prefill_overlap_results()
+                continue
+
+            self._wait_donation_safe()
+            self.send_kv_chunk()
+            # PD producers never merge into the local decode running batch.
+            # Their chunk continuation is owned by chunked_reqs/ChunkCache.
+            self.last_batch = None
+            batch = (
```
zorrofox, line 349: **global backpressure is over-broad for DP>1**

`_prefill_chunk_transfer_backpressured` is evaluated across all ranks and the result is `batch = None`, so a pending chunk on rank 0 stops scheduling on rank 1 as well. Under DP1 this is equivalent to skipping just the owner (`chunked_reqs[dp] is not None` already blocks that rank's waiting queue), so the validated configuration is unaffected. Since the PR states DP1-only scope this is fine to leave, but a per-rank gate (skip `add_chunked_req` for the backpressured rank inside `get_new_batch_prefill`) would be the right shape when DP>1 is validated. Worth a TODO so it does not get lost.

python/sgl_jax/srt/disaggregation/raiden_transfer/conn.py
```diff
@@ -663,6 +663,17 @@ def sender_done(self, req_id: str) -> bool:
         with self._poll_lock:
             return req_id in self._done_sending
 
+    def sender_failed(self, req_id: str, dp_rank: int) -> bool:
```
zorrofox, line 666: **multi-endpoint failed send never reaches a terminal state**

Confirmed the premise in tpu-sync@6d43141: expired sends land in `failed_recving` (`kv_cache_manager_with_transfer.cc` deadline path) and the NUMA-aware wrapper emits the first sub-manager failure before the others settle, so retaining ownership until all shards report is the safe default. The gap is that nothing ever ends the wait: with `sender_failed()` pinned to `False`, `poll()` never sees `sender_done or sender_failed` for the expired child, and `request_abort("timeout")` from the reaper only sets `_pending_failure_reason` because the state is still `TRANSFERRING`. The request's KV pages and slot are held until process restart. `main` behaves the same (it only checks `sender_done`), so this is not a regression, but `test_multi_endpoint_failure_does_not_release_other_shards` now asserts the hang as the intended contract.

Suggestion: keep the conservative retention, but add an escape hatch: once `_pending_failure_reason` is set and `now - transfer_started_at >= ack_timeout`, treat all sub-managers as settled and `_finish(FAILED)`. Or leave a TODO referencing this so multi-NUMA users know.

python/sgl_jax/srt/disaggregation/decode.py
```diff
@@ -420,6 +533,16 @@ def _drain_decode_transfer_terminals(self: Scheduler) -> None:
                         )
                     )
                     self._set_decode_bookkeeping(entry.req, entry.kv_indices)
+                    # D recomputes the last prompt token. If that token alone
```
zorrofox, line 535: **nit: trim before bookkeeping**

The fix itself is right (verified: `prefix_indices` excludes the page, the extend allocator opens a fresh page, `main` never frees the received one). Minor ordering nit: `_set_decode_bookkeeping` has already stored the untrimmed list in `req._pd_prealloc_kv_indices` and the commit callback captured `entry.kv_indices` before the trim. Nothing reads `_pd_prealloc_kv_indices` today, so no double free, but truncating `entry.kv_indices` before the bookkeeping call would leave no stale ownership record for a future release path to trip over.

python/sgl_jax/srt/disaggregation/prefill.py
```diff
@@ -231,6 +253,120 @@ class SchedulerDisaggregationPrefillMixin:
...
+            if had_pending:
+                self._resolve_disagg_prefill_result(self.result_queue.popleft())
+            elif batch is None:
+                self.new_token_ratio = self.init_new_token_ratio
+                if self._comm_backend is not None:
+                    self._comm_backend.wait_for_new_requests(0.001)
+            self.send_kv_chunk()
+            self.last_batch = batch
+
     def event_loop_normal_disagg_prefill(self: Scheduler) -> None:
```
zorrofox, line 379: **nit: `send_kv_chunk` runs twice per tick**

Line 333 (after `_wait_donation_safe`) and line 370 (end of tick) both poll every registered sender (`poll_engine` plus per-child `sender_done` / `sender_failed` under the lock), with only `recv_requests` between the bottom call and the next top call. One call per tick is enough; the bottom one already covers the idle path.

test/srt/disaggregation/test_pd_prefill_overlap.py
```diff
@@ -0,0 +1,309 @@
...
+    scheduler._resolve_disagg_prefill_result = events.append
+    scheduler.send_kv_chunk = lambda: events.append("reap")
+    scheduler.last_batch = scheduler.cur_batch = object()
+    scheduler._drain_disagg_prefill_overlap_results()
+    assert events == [first, second, "reap"]
+    assert scheduler.last_batch is scheduler.cur_batch is None
+    assert not scheduler.result_queue
+
+
+def test_cancelled_snapshot_aborts_existing_sender_without_republishing():
```
zorrofox, line 165: **coverage gaps worth closing**

The existing cases pin down the pieces individually, but the interactions that matter most are asserted only via pre-set state, not trigger sequences:

- **Abort in flight, non-chunk mode** (`prefill.py:476-480`, `sender is None`): no test. Expect `check_finished` → `_stream_prefill_req` once → `_release_prefill_req_resources` deferred → real release in the `finally` of `_resolve_disagg_prefill_result` exactly once.
- **Sender FAILED while a later chunk is in flight**: the scheduler side (`has_pending_failure` → `_retire_chunk_producer_ownership`, then no further chunk scheduled, then terminal callback after compute hits 0) has no test; the only failure test is at the sender level.
- **Full abort chain**: `AbortReq` → `_mark_pending_chunked_aborts` → resolve → terminal → release, with `_disagg_prefill_compute > 0` throughout. `test_cancelled_snapshot_aborts_existing_sender_without_republishing` starts from `to_finish` already set and never invokes the terminal callback.
- **One end-to-end chain** driving real `run_batch` (fake worker) → `PendingPrefillResult` → real `_resolve_disagg_prefill_result` → real `process_prefill_chunk` → real `_raiden_handoff_chunk` against a fake kv manager. Today each link is tested with its neighbour stubbed.

Same pattern as #1523: tests that assert the ordering of events (abort arrives at tick N while result of tick N-1 is pending) catch this class of bug; end-state tests with mocks pre-set do not.

zorrofox (dismissed):
Went through this with the same lens as #1523: abort / retract / sender-failure interleavings against the new in-flight state on both roles, plus the shared (non-PD) code paths this touches. Traced each claim against the head commit rather than the description. Summary, details inline:

**Verified as correct (no action needed):**

- D-side admission fence covers the only native-destination alloc site (`_admit_decode_prealloc`); chunk-transfer mode allocates the whole request up front, so later `start_read` calls never touch unfenced pages. Retract (`update_running_batch`) and `pause_generation` both drain before freeing. The recursion in `update_running_batch` is depth ≤ 1 and operates on the same `running_batch` object.
- The tail-page release in `_drain_decode_transfer_terminals` fixes a real leak on `main`: when `(len(prompt) - 1) % page_size == 0`, `prefix_indices` excludes the last received page, the extend allocator opens a fresh page for the recomputed token, and nothing ever frees the received one. Good that it is not gated on the overlap flag.
- P-side: abort while a chunk forward is in `result_queue` is handled (`to_finish` set via `cur_batch` scan or `_mark_pending_chunked_aborts`, consumed at resolve, release deferred until the compute count hits 0). A sender that fails while a later chunk is in flight costs at most one extra chunk forward before `_raiden_handoff_chunk` retires the producer. Snapshot page IDs come from the host-side `req_to_token` written synchronously in `prepare_for_extend`, so they are stable across the in-flight window. `_wait_pd_prefill_kv_ready` runs after `replace_all` on the forward thread, so it fences the right buffer.
- Shared-path changes are all fixes of real `main` bugs, and their blast radius is wider than the description suggests: `_handle_abort_req` on `main` raises `KeyError` on a repeated or late `AbortReq` and takes the process down via `print_exception_wrapper`; the output-logprob delta bug hits every native `stream=true` + `return_logprob=true` request (tokenizer rewrites those to `return_output_logprob_only` when no `logprob_start_len` / `top_logprobs_num` is given), not just PD; `_iter_stream_lines` removes aiohttp's 128 KiB `readline` ceiling.

**Non-blocking issues (inline):**

1. `_retire_chunk_producer_ownership` clears the batch-side owner on `last_batch`, but in the overlap loop `last_batch` is `None` at resolve time and the just-launched batch is still in `cur_batch`; any later `AbortReq` or `pause_generation` resurrects a retired producer for one more chunk. Waste only, no corruption, but it is the same class of hole `_retire_chunk_producer_ownership` was introduced to close in #1523.
2. `_snapshot_prefill_handoffs` calls `_extract_req_block_ids_range` outside any `try`, so a `ValueError` there (range / alignment invariant) kills the P scheduler instead of failing the one request as the normal loop does.
3. Backpressure gating skips `get_next_batch_to_run()` globally. Equivalent to per-request under DP1 (the chunked owner already monopolizes its rank), but over-broad for DP>1.
4. `sender_failed` returning `False` for multi-endpoint ranks leaves a failed multi-NUMA send in `TRANSFERRING` forever (the ack-timeout abort cannot finish either). This is the same as `main`, so not a regression, but the new test enshrines the hang as intended behaviour.

**Test coverage:** the two overlap-loop files test each link with its neighbours stubbed, so a few interactions are only argued, not asserted. Suggestions inline on `test_pd_prefill_overlap.py`.

**Benchmark question:** the 4K/256 C32 row shows TTFT p95 going 251 ms to 844 ms with overlap ON. The description attributes it to higher throughput, but the branch history had per-role ablations (P-only / D-only overlap) that were trimmed out. Could you put the C32 role ablation back in the description, plus p50 TTFT? That attributes the p95 shift to one role and tells operators which flag to leave off if TTFT is their SLO.

Also: the timeline image links to a blob at 5b352574, which a later commit deleted; after squash-merge that commit becomes unreachable and the image will 404.

Overall: no blocking correctness issue found after an adversarial pass. Approving; items 1 and 2 are small enough to fold into this PR if you prefer.

PR #1615, opened 2026-09-09
fix(mla): bind mesh explicitly to shard_map in MLA and DSA attention backends

aashishrampal-lab:
## Description
1. Explicitly pass `mesh=self.mesh` to `jax.shard_map` invocations in both `MLAAttentionBackend` and `DSASparseAttentionBackend`.
2. Update `apply_fused_mlp_sharded` in `fused_mlp.py` to support `P("data", None)` sharded inputs/outputs and replace `check_rep=False` with `check_vma=False`.

### Motivation
When running in data-parallel or unified mesh environments (e.g., DP=32, TP=1 or hybrid TP/DP), `jax.shard_map` requires the explicit mesh context and correct data partition specifications to resolve against the global mesh.

> **Note:** Code cleanup and unified mesh consolidation are currently in progress.

### Test Plan
- Verified shard_map partitioning across TP=32, DP=32, and EP=32 configurations.
- Tested inference and forward pass on TPU v7x.

### DIFF
--- a/python/sgl_jax/srt/kernels/fused_mlp.py
+++ b/python/sgl_jax/srt/kernels/fused_mlp.py
@@ -128,18 +128,18 @@ def apply_fused_mlp_sharded(
     b_inter: int = 128,
 ) -> jax.Array:
     in_specs = (
-        P(None, None),  # x
+        P("data", None),  # x
         P(None, "tensor"),  # w_gu (combined gate/up weight, sharded along tensor axis)
         P("tensor", None),  # wd (down weight, sharded along tensor axis)
     )
-    out_specs = P(None, None)
+    out_specs = P("data", None)
 
     @functools.partial(
         shard_map,
         mesh=mesh,
         in_specs=in_specs,
         out_specs=out_specs,
-        check_rep=False,
+        check_vma=False,
     )
     def local_fused_mlp(x_loc, w_gu_loc, wd_loc):
         seq_len, hidden_size = x_loc.shape
--- a/python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
+++ b/python/sgl_jax/srt/layers/attention/dsa_sparse_backend.py
@@ -420,7 +420,9 @@ class DSASparseAttentionBackend(MLAAttentionBackend):
                 vmem_limit_bytes=self.vmem_limit_bytes,
             )
 
-        return jax.shard_map(_run, in_specs=in_specs, out_specs=out_specs, check_vma=False)(
+        return jax.shard_map(
+            _run, mesh=self.mesh, in_specs=in_specs, out_specs=out_specs, check_vma=False
+        )(
             ql,
             qpe,
             kvc,
@@ -473,7 +475,9 @@ class DSASparseAttentionBackend(MLAAttentionBackend):
                 vmem_limit_bytes=self.vmem_limit_bytes,
             )
 
-        return jax.shard_map(_run, in_specs=in_specs, out_specs=out_specs, check_vma=False)(
+        return jax.shard_map(
+            _run, mesh=self.mesh, in_specs=in_specs, out_specs=out_specs, check_vma=False
+        )(
             ql,
             qpe,
             kvc,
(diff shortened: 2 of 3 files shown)


zorrofox:
Thanks for the fix. Overall direction looks right, a few things before this can go in:

**1. CI failure is the un-updated mock in `test_mla_attention.py`, not flakiness.** `unit-test-1-tpu (1)` fails with `TypeError: fake_shard_map() got an unexpected keyword argument 'mesh'`. `test_backend_reshards_new_k_pe_for_shard_map` patches `mla_backend.jax.shard_map` with `fake_shard_map(_fn, *, in_specs, out_specs, check_vma)`, which has no `mesh` kwarg. Adding `mesh=None` to the fake (and ideally asserting it is the backend mesh) fixes it.

**2. `P("data", None)` in `apply_fused_mlp_sharded` breaks the padding contract.** `apply_fused_mlp_with_padding` pads the global `seq_len` to a multiple of `b_seq`, but after this change each data shard only sees `seq_len / dp` rows and the kernel grid is `local_seq // b_seq`. With `b_seq=64` and DP=32, a decode batch of 8 gets padded to 64 globally, each shard gets 2 rows, grid becomes 0 and the output is never written. Any local length that is not a multiple of `b_seq` leaves the tail rows uncomputed, and the padded global length also has to be divisible by the data axis size or `shard_map` errors. Sharding along data is the right call to avoid redundant compute per DP rank, but the padding needs to be a multiple of `b_seq * mesh.shape["data"]` and `b_seq` should be chosen from the local length. This path is only hit by unquantized GLM5 dense MLP so CI will not catch it; a small unit test with `data > 1` would be good.

**3. `check_rep` -> `check_vma` is correct but is an independent bug.** `jax.shard_map` in the pinned jax has no `check_rep`, so the fused MLP path is currently broken on main regardless of mesh. Suggest splitting that into its own small PR with a test so it can land quickly.

**4. DSA backend is only partially covered.** The branch is ~54 commits behind main. Current main `dsa_sparse_backend.py` has five `jax.shard_map` calls without an explicit mesh (the indexer calls and the three `_run` calls). Please rebase and bind `mesh=self.mesh` on all of them, otherwise the unified-mesh motivation does not hold for DSA. The MLA change itself matches what `gdn_backend` / `kda_backend` already do.

**5. Test plan.** Could you add which model and sequence lengths were used for the TP=32 / DP=32 / EP=32 runs? Given (2), I would expect DP=32 with GLM5 fused MLP to produce wrong outputs, so it would help to know whether that path was actually exercised.

zorrofox:
Two follow-ups after a closer pass:

**6. The actual trigger is the `forward_and_sample` path, not DP size.** `model_runner.py:895` (current main) deliberately does not enter `use_mesh` for the fused forward+sample path used by Pathways PD decode, and the NOTE there says "Model shard_maps carry mesh explicitly". Any `jax.shard_map` without `mesh=` reached from that path raises `The context mesh cannot be empty`, on a single device as well as at DP=32. Under `_forward_raw`, which does run inside `set_mesh`, `mesh=None` resolves fine even for DP=32/TP=1. Suggest rewording the description to state that, so the next person does not look for a DP-specific issue. A more robust fix than patching leaf call sites would be to wrap only the `model(...)` call inside `jitted_run_and_sample` with `jax.sharding.use_abstract_mesh(self.mesh.abstract_mesh)`, leaving the sampler outside as the NOTE requires; that covers every backend at once.

**7. `flashattention_backend.py:770` has the same gap.** The `_ragged_paged_attention_with_fused_kv` `shard_map` has no `mesh=` either, while `self.mesh` is available. GQA/MHA models on the same PD fused path will hit the identical error, so if this PR keeps the leaf-site approach it should include that call plus the DSA indexer `shard_map` mentioned above. After those two, I could not find any other live `shard_map` under `srt/` without an explicit mesh; the eagle kernels are already wrapped in `set_mesh` by `eagle_util.py`.

On (2): I ran the PR's `fused_mlp.py` in Pallas interpret mode with a `(data=32, tensor=1)` mesh and bs=32 decode input. `b_seq` resolves to 256, each device gets 8 rows, `grid=(0,)`, `pallas_call` runs zero steps without error and the output is uninitialized memory (NaN in interpret mode). So the padding fix is required before this can be merged, not just a nice-to-have.

PR #1619, opened 2026-09-09
[Spec Decoding] Add explicit anchor sampling for DeepSpec DFlash checkpoints

lingebeng:
## Summary

Some DFlash checkpoints predict the next token from the anchor position. Skipping that position misaligns draft proposals and results in low acceptance.

Add `--speculative-sample-from-anchor` to select this layout explicitly:
- Sample from the anchor and use a draft query one token shorter than the target verification width.
- Adjust draft attention metadata and KV writes while preserving the full target verification layout.
- Infer verification width from checkpoint block size and the flag when no explicit width is supplied.

The flag defaults to false and currently requires `DFLASH`. Document both layouts and their configuration.

### Parameter naming and upstream alignment

`--speculative-sample-from-anchor` describes a checkpoint's prediction layout, independently of whether it has a Markov head. Upstream SGLang defaults `sample_from_anchor` to `True` in its DSpark path, but its ordinary DFlash path samples from output position 1. DeepSeek's `dflash_qwen3_8b_block7` has no Markov head (`markov_rank=0`) and does not declare `sample_from_anchor` in its config; its training labels nevertheless require sampling from output position 0. Applying the ordinary DFlash offset would misalign its proposals.

This PR therefore adds an explicit, generally named layout override while retaining the existing DFlash default. The name follows the `--speculative-*` convention; it is a new sglang-jax option, not an existing upstream CLI flag. Automatic checkpoint-field resolution and DSpark runtime support are outside this PR.

Reference: [upstream DSpark layout default](https://github.com/sgl-project/sglang/blob/2f7393f0d245bfaa8ebe0b9d0533432e6b3bd9ed/python/sglang/srt/speculative/dspark_components/dspark_config.py#L29) and [ordinary DFlash sampling](https://github.com/sgl-project/sglang/blob/2f7393f0d245bfaa8ebe0b9d0533432e6b3bd9ed/python/sglang/srt/speculative/dflash_worker_v2.py#L2325).

Example for the DeepSeek DFlash block7 checkpoint:

```bash
--speculative-algorithm DFLASH
--speculative-draft-model-path deepseek-ai/dflash_qwen3_8b_block7
--speculative-num-draft-tokens 8
--speculative-sample-from-anchor
```

## Validation

TPU v7x-8 with Qwen/Qwen3-8B as the target, using the same 20 GSM8K prompts, temperature 0, and thinking disabled:

| Draft checkpoint | Anchor sampling | Mean acceptance length, including bonus |
| --- | --- | ---: |
| deepseek-ai/dflash_qwen3_8b_block7 | Disabled, verify width 8 | 1.018 |
| deepseek-ai/dflash_qwen3_8b_block7 | Enabled, verify width 8 | 4.662 |
| z-lab/Qwen3-8B-DFlash-b16 | Disabled, verify width 16 | 6.466 |

These measurements were collected before the parameter rename, with identical decoding behavior. This is a 20-prompt greedy acceptance evaluation, not a GSM8K accuracy measurement or a benchmark run of upstream SGLang.

## Why anchor sampling is needed

The distinction is **which draft output position corresponds to the first candidate token**. Both layouts include the anchor token in the draft input; the flag controls where sampling starts in the output.

For the DeepSeek block7 checkpoint, the position alignment follows from the training code in DeepSpec commit [`005e03b`](https://github.com/deepseek-ai/DeepSpec/tree/005e03b81cec38b7da6399833d609ee89a2587f2). Although the reference code lives under `dspark`, this change keeps the serving algorithm as `DFLASH`.

### 1. Draft input positions start at the anchor

[`create_position_ids()`](https://github.com/deepseek-ai/DeepSpec/blob/005e03b81cec38b7da6399833d609ee89a2587f2/deepspec/modeling/dspark/common.py#L251) constructs position offsets starting at **0**:

```python
offsets = torch.arange(block_size, device=device).view(1, 1, -1)
return (anchor_positions.unsqueeze(-1) + offsets).view(
    bsz,
    num_blocks * block_size,
)
```

For an anchor at position `a` and a block size of 7, the draft input positions are:

```text
a, a+1, a+2, a+3, a+4, a+5, a+6
```

`create_noise_embed()` places the anchor token in the first input position and mask tokens in the remaining positions.

### 2. Training labels start one position after the anchor

The [label construction](https://github.com/deepseek-ai/DeepSpec/blob/005e03b81cec38b7da6399833d609ee89a2587f2/deepspec/modeling/dspark/qwen3/modeling.py#L432) uses offsets starting at **1**:

```python
label_offsets = torch.arange(1, self.block_size + 1, device=device).view(
    1, 1, -1
)
label_indices = anchor_positions.unsqueeze(-1) + label_offsets
```

The corresponding tokens are gathered from `input_ids` into `target_ids`. For block size 7, their positions are:

```text
a+1, a+2, a+3, a+4, a+5, a+6, a+7
```

### 3. The loss pairs outputs and labels without another shift

The [cross-entropy loss](https://github.com/deepseek-ai/DeepSpec/blob/005e03b81cec38b7da6399833d609ee89a2587f2/deepspec/modeling/dspark/loss.py#L109) directly pairs the flattened logits and labels:

```python
flat_logits = draft_logits.reshape(-1, vocab_size)
flat_targets = target_ids.reshape(-1)
loss_per_token = F.cross_entropy(flat_logits, flat_targets, reduction="none")
```

Therefore, draft output position `j` is trained against the token at position `a + j + 1`:

```text
Block index:          0        1        2        …        6
Input position:       a       a+1      a+2       …       a+6
Input token:        anchor    mask     mask      …       mask
Output hidden:        h₀       h₁       h₂       …        h₆
                      │        │        │                 │
Training label:     x[a+1]   x[a+2]   x[a+3]     …      x[a+7]
```

The first output, `h₀`, already predicts the first candidate after the anchor.

### Previous behavior

The existing sampling code always started at output position 1:

```python
proposal_hidden = draft_hidden[:, 1:, :]
```

For this checkpoint, that discards `h₀` and places the prediction from `h₁`—trained against `x[a+2]`—into the first candidate slot, where the target expects `x[a+1]`.

```text
Target candidate slot:       1              2              3
Expected token position:   a+1            a+2            a+3
Old sampled output:         h₁             h₂             h₃
Its training label:        a+2            a+3            a+4
```

This positional mismatch explains the low acceptance observed with the previous configuration.

### Updated behavior

`--speculative-sample-from-anchor` explicitly selects output position 0 as the sampling start:

```python
sample_start = 0 if sample_from_anchor else 1
proposal_hidden = draft_hidden[:, sample_start:, :]
```

To preserve the same number of candidates, the draft query length depends on the starting position:

| For 7 candidates | Default layout | Sample from anchor |
| --- | --- | --- |
| Sampled output positions | `h₁…h₇` | `h₀…h₆` |
| Draft query length | 8 | 7 |
| Candidate count | 7 | 7 |
| Target verification width, including anchor | 8 | 8 |

For the anchor layout:

```text
Draft input — 7 positions:
[anchor, mask, mask, mask, mask, mask, mask]
    ↓      ↓     ↓     ↓     ↓     ↓     ↓
   [x₁,   x₂,   x₃,   x₄,   x₅,   x₆,   x₇]   Seven candidates

Target verification — 8 positions:
[anchor,  x₁,   x₂,   x₃,   x₄,   x₅,   x₆,   x₇]
```

Accordingly, draft attention metadata and KV writes use the shorter query length, while target verification and accepted-context KV materialization retain the full verification width.

The flag defaults to false and is not inferred from the checkpoint name or architecture. An explicitly supplied `--speculative-num-draft-tokens` remains the target verification width. When omitted, that width is inferred from the checkpoint’s `block_size`, adding one when anchor sampling is enabled.


cjx0709 (commented):
It's a good fix, but here seems no e2e test throughtout benchmark or eval acc test here. Will these be tested in the incoming prs?

lingebeng:
> It's a good fix, but here seems no e2e test throughtout benchmark or eval acc test here. Will these be tested in the incoming prs?

It's a small modification and I'll add speed benchmark in #1632.

Here, accuracy benchmark is OK:
image


cc @cjx0709

lingebeng:
Upstream SGLang defaults `sample_from_anchor` to `True` only in the DSpark path. However, DeepSeek’s DFlash checkpoint uses the same anchor-aligned prediction layout without a Markov head, so applying the default DFlash sampling offset can misalign its proposals and reduce acceptance. Since the sampling start position describes a checkpoint’s prediction layout rather than a DSpark-specific feature, I propose renaming `--dspark-sample-from-anchor` to the more general `--speculative-sample-from-anchor`. This preserves upstream-compatible defaults while providing an explicit override for checkpoints such as `deepseek-ai/dflash_qwen3_8b_block7`.

cc @cjx0709

cjx0709 (approved):
LGTM

Issue #1613, opened 2026-09-08
[TPU v7x] jax 0.11.1 bump regresses bs=1 decode TPOT up to +57% (SparseCore gather offload); proposal: opt-in per-executable compiler_options workaround


zorrofox:
Since the jax 0.10.2 -> 0.11.1 bump (#1574), single-stream decode on TPU v7x regresses significantly for MoE models. On GLM-5.2 (tp16, 2 hosts, dsa_sparse backend, 135k context config) decode TPOT goes **13.7 -> 21.4 ms/token (+57%)** measured from a fresh boot on the same nodes with isolated compile caches. Root cause is in libtpu, not sglang-jax code: 0.11.1 started offloading small bs=1 decode gathers to SparseCore, adding ~7 ms/step of `gather_offload` + sort ops. Full attribution matrix, traces and flag ablations are in the upstream report: jax-ml/jax#40553.

Two facts that shape the fix:

1. `LIBTPU_INIT_ARGS="--xla_tpu_offload_gather_to_sparsecore=false --xla_tpu_offload_all_supported_gathers_to_sparsecore=false"` fully recovers decode (13.66 ms, exactly matching the 0.10.2 baseline), **but costs +21% on 110k-token chunked prefill** (24.3 s -> 29.3 s), because the offload is genuinely profitable for large prefill gathers. So a process-global env flag is a workload trade-off, not a clean fix.

2. The flag is honored **per-executable** via `lower(...).compile(compiler_options={"xla_tpu_offload_gather_to_sparsecore": "false"})` — verified on v7x/jax 0.11.1: the offload forms disappear from that executable's optimized HLO only. This allows disabling the offload for decode while keeping it for prefill, which recovers decode with no prefill cost.

Proposal (opt-in, default behavior unchanged): add a server flag (e.g. `--disable-sc-gather-offload-for-decode`) that injects the compiler option only for decode-shaped executables. Candidate wiring points: (a) the AOT dispatch compile path (`aot_dispatch.py`) where each shape is compiled separately, or (b) a dedicated decode jit wrapper. `model_runner.py` already passes `compiler_options` to `jax.jit` for the kernel log recorder, so the plumbing pattern exists. There is also precedent in vllm's tpu-inference, which ships a `compiler_options.py` that injects SparseCore offload thresholds into its step-function jit by default.

[...]

PR #1631, opened 2026-09-10
fix(dsa/moe): recover jax 0.11.1 long-context perf regressions

zorrofox:
Fixes the jax 0.10.2 -> 0.11.1 performance regressions tracked in #1613 (upstream report: jax-ml/jax#40553). On GLM-5.2 v7x tp16 at the 135k-context config, decode TPOT regressed 13.7 -> 21.4 ms/token and 110k TTFT 22.1 -> 24.4 s; with this PR both return to the 0.10.2 baseline: **decode 13.68 ms/token, 110k TTFT 22.29 s** (baseline 13.67 / 22.13, i.e. within 1%).

Three independent 0.11.1 interactions, three scoped fixes:

1. **Decode SparseCore gather offload (opt-in)** — 0.11.1's libtpu now offloads the DSA decode sparse-KV gather (2048 indices from an HBM-resident paged pool) to SparseCore, costing ~7 ms/step at bs=1, while the same offload is genuinely profitable for prefill's large MoE gathers (a process-global `LIBTPU_INIT_ARGS` disable costs +11.6% on 110k TTFT). `AotDispatcher` gains a `compiler_options_fn` hook, and `SGLANG_JAX_DECODE_DISABLE_SC_GATHER_OFFLOAD=1` compiles only decode-shaped executables with `xla_tpu_offload_gather_to_sparsecore=false`. Default off; no behavior change unless the env var is set. The JAX team has confirmed in jax-ml/jax#40553 that per-executable `compiler_options` is **officially supported and a viable workaround** for this, that the compiler currently lacks a cost model for gather offload profitability, and that a cost-model fix is tracked internally — per cjx0709 in #1613, this flag is a workaround to revert once that fix ships. (The JAX team also suggested `jax.experimental.compute_on` per-op annotations as a finer-grained alternative; that would require annotating the specific gather inside the model/backend code, so this PR takes the runtime-level per-executable route and leaves per-op pinning as a possible follow-up.)

2. **qblock prefill kernel: keep the pool's native layout in the graph** — on 0.11.1 the host-side flat view `kv.reshape(1, Pn*ps, D)` of the 4D word-packed KV pool materializes a full T(2,128)->T(8,128) retile per chunk (~1.9 s per 110k prefill), and the write-back overflow `lax.cond` pays a matching retile inside its scatter branch (~2.0 s). The kernel now takes the pool 4D and flattens the HBM ref in-kernel (free, same trick as `_write_back_kernel`), and the overflow fallback scatters 4D-natively. GLM/DSA-only code path; no change for other models. qblock canaries 19/19.

3. **EPMoE unpermute: pick the strategy from the token count** — the unrolled per-k gather from #1578 avoids building the einsum's `(tokens, top_k, hidden)` fp32 intermediate, a large win for prefill-shaped batches on 0.11.1 (-2.8 s per 110k prefill vs the einsum path), but costs +0.9 ms/token on decode-shaped batches where that intermediate is tiny. Branch on the static token count (<=256 -> einsum, else unrolled) so both shapes keep their fast path. Bit-identical to the einsum reference for the small-batch path; the large-batch path is unchanged from #1578.

## Measurements (GLM-5.2 753B, v7x tp16, 2 hosts, 135k-context config)

| metric | 0.10.2 baseline | 0.11.1 before | 0.11.1 + this PR |
|---|---|---|---|
| decode TPOT (1k ctx, ignore_eos, (t257-t1)/256, 2 rounds) | 13.67 ms | 21.4 ms (+57%) | **13.68 ms** |
| 110k TTFT (x3) | 22.13 s | 24.4 s (+10%) | **22.29 s** |
| gsm8k 200q (5-shot, temp 0) | 0.950-0.955 (einsum numerics) | 0.960 (#1578 numerics) | 0.945 (back on the 0.10.2-equivalent einsum numerics for decode; within the 200-question noise of its historical band, 0 invalid) |
| 49k config TPOT / cc=1/8/32 | 13.33 ms | 13.33 ms (unaffected) | 13.31 / 13.32-13.33 ms (no regression; cc points serialized by max_running=1 in this config) |
| 16k TTFT | 3.55 s | — | 2.31 s |
| qblock kernel canaries | — | — | 19/19 passed |

## Test plan

- [x] decode TPOT recovery on v7x 135k (13.665 ms, 2 rounds, spread <0.1%)
- [x] 110k TTFT recovery (22.31 s x3, zero spread)
- [x] qblock TPU canaries (test_qblock_kernel_tpu.py + parity, 19/19)
- [x] CPU parity of the unpermute branch vs einsum reference (top_k in {1,2,4,8}, bitwise)
- [x] gsm8k 200q on the final stack (0.945, 0 invalid; numerics-path attribution above)
- [x] 49k config regression (TPOT 13.31; cc=1/8/32 all 13.32-13.33)
- [x] 16k TTFT spot check (2.31 s)
- [x] pre-commit clean

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -835,12 +835,9 @@ class EPMoE(nnx.Module):
         )
         weights_fp32 = weights.astype(jnp.float32)
 
-        # Small (decode-shaped) batches: the einsum's (tokens, top_k, hidden)
-        # fp32 intermediate is tiny, while the unrolled per-k gather costs
-        # +0.9 ms/token on GLM-5.2 decode (top_k=8, 78L, TPU v7x, jax 0.11.1).
-        # Large (prefill-shaped) batches: that intermediate is ~0.5 GB/layer
-        # and the unrolled path from #1578 is a big win (-2.8 s on 110k TTFT).
-        # Branch on the static token count: both shapes get their fast path.
+        # Static token-count branch: small (decode) batches avoid the unrolled
+        # per-k gather overhead; large (prefill) batches avoid the fp32
+        # (tokens, top_k, hidden) intermediate. Benchmarks in the PR description.
         if weights.shape[0] <= 256:
             unsort_intermediate = jnp.take(intermediate, indices=argsort_indices, axis=0)
             reshaped_intermediate = jnp.reshape(
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -833,16 +833,36 @@ class EPMoE(nnx.Module):
             .at[sorted_selected_experts]
             .set(jnp.arange(expected_tokens, dtype=jnp.int32))
         )
-        grouped_indices = jnp.reshape(argsort_indices, (weights.shape[0], top_k))
         weights_fp32 = weights.astype(jnp.float32)
 
-        output = None
-        for k in range(top_k):
-            contribution = (
-                jnp.take(intermediate, indices=grouped_indices[:, k], axis=0).astype(jnp.float32)
-                * weights_fp32[:, k, None]
+        # Small (decode-shaped) batches: the einsum's (tokens, top_k, hidden)
+        # fp32 intermediate is tiny, while the unrolled per-k gather costs
+        # +0.9 ms/token on GLM-5.2 decode (top_k=8, 78L, TPU v7x, jax 0.11.1).
+        # Large (prefill-shaped) batches: that intermediate is ~0.5 GB/layer
+        # and the unrolled path from #1578 is a big win (-2.8 s on 110k TTFT).
+        # Branch on the static token count: both shapes get their fast path.
+        if weights.shape[0] <= 256:
+            unsort_intermediate = jnp.take(intermediate, indices=argsort_indices, axis=0)
+            reshaped_intermediate = jnp.reshape(
+                unsort_intermediate,
+                (weights.shape[0], top_k, -1),
             )
-            output = contribution if output is None else output + contribution
+            output = jnp.einsum(
+                "BKE,BK -> BE",
+                reshaped_intermediate.astype(jnp.float32),
+                weights_fp32,
+            )
+        else:
+            grouped_indices = jnp.reshape(argsort_indices, (weights.shape[0], top_k))
+            output = None
+            for k in range(top_k):
+                contribution = (
+                    jnp.take(intermediate, indices=grouped_indices[:, k], axis=0).astype(
+                        jnp.float32
+                    )
+                    * weights_fp32[:, k, None]
+                )
+                output = contribution if output is None else output + contribution
 
         final_output = output.astype(self.dtype)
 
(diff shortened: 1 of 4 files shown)


python/sgl_jax/srt/model_executor/model_runner.py
```diff
@@ -385,12 +385,41 @@ def jitted_compute_logprobs(mesh, logits, next_tokens):
...
+        # executables with the offload pass disabled while prefill keeps the
+        # default (the offload is profitable for large prefill gathers, and a
+        # process-global LIBTPU_INIT_ARGS disable costs ~+12% on 110k prefill).
+        # Only effective together with SGLANG_JAX_AOT_DISPATCH since it hooks
+        # the per-shape AOT compile path.
+        decode_no_sc_gather = jax.default_backend() == "tpu" and get_bool_env_var(
+            "SGLANG_JAX_DECODE_DISABLE_SC_GATHER_OFFLOAD"
+        )
+
+        def _run_model_compiler_options(dyn_args):
```
cjx0709: Could we move the env-var check and decode compiler-options helper into `aot_dispatch.py`, leaving only the callback wiring here? That would keep this temporary XLA workaround out of the common model runner and make it easier to remove once the upstream fix lands.

python/sgl_jax/srt/layers/moe.py
```diff
@@ -833,16 +833,36 @@ def _unpermute(self, intermediate, sorted_selected_experts, weights):
...
-        for k in range(top_k):
-            contribution = (
-                jnp.take(intermediate, indices=grouped_indices[:, k], axis=0).astype(jnp.float32)
-                * weights_fp32[:, k, None]
+        # Small (decode-shaped) batches: the einsum's (tokens, top_k, hidden)
+        # fp32 intermediate is tiny, while the unrolled per-k gather costs
+        # +0.9 ms/token on GLM-5.2 decode (top_k=8, 78L, TPU v7x, jax 0.11.1).
+        # Large (prefill-shaped) batches: that intermediate is ~0.5 GB/layer
+        # and the unrolled path from #1578 is a big win (-2.8 s on 110k TTFT).
+        # Branch on the static token count: both shapes get their fast path.
```
cjx0709: Could we shorten this to explain the small-batch overhead versus large-batch memory trade-off, and keep the benchmark details in the PR description? The branch is straightforward; two lines should be enough here.

python/sgl_jax/srt/model_executor/model_runner.py
zorrofox: Done in 9f517fa: the env check, the once-per-boot log, and the decode compiler-options callback now live in `aot_dispatch.decode_no_sc_gather_compiler_options_fn()`, with a docstring marking it as a temporary workaround to remove once the upstream fix lands. `model_runner` only passes the returned callback into `AotDispatcher`.

python/sgl_jax/srt/layers/moe.py
zorrofox: Done in 9f517fa: trimmed to three lines stating the small-batch overhead vs large-batch intermediate trade-off, benchmark numbers stay in the PR description.

cjx0709 (approved):

PR #1636, opened 2026-09-10
fix: allow bootstrap health check to connect via both IPv4 and IPv6 loopback

iwknow:
## Motivation

This is a bug fix for `python/sgl_jax/srt/disaggregation/bootstrap.py`. the previous health check only allow the self ip to be ipv4. The updated implementation extends the current health check to allow ipv6 address.

## Modifications

Extend the 

## Accuracy Tests

Unit tests added

## Benchmarking and Profiling
N/A

## Checklist

- [X] Please use English, otherwise it will be closed.
- [X] The purpose of the PR, or link existing issues this PR will resolve.
- [X] The test plan, such as providing test command.


cjx0709 (approved):
LGTM

PR #1334, opened 2026-06-09
Kimi K2.5 INT4 support

chaitanyapk:
## Motivation

The goal of this Pull Request is to support loading and running inference on the original **Kimi 2.5 INT4 quantized weights** in both single-host and multi-host TPU environments.

Previously, there was no pipeline to unpack packed 4-bit weights dynamically on TPUs, this PR introduces an optimized loader that performs dynamic bit-unpacking on TPU and defers transpositions to post-unpacking, enabling clean compilation and execution using the Megablox GMM TPU kernels.

---

## Modifications

*   **`python/sgl_jax/srt/utils/weight_utils.py`**:
    *   Added `unpack_4bit_jax` to unpack raw `int32`/`int8` weights to `int4` on TPU.
    *   Deferred transposition of `int4` weights to TPU post-unpacking to avoid bit corruption.
*   **`python/sgl_jax/srt/layers/moe.py`**:
    *   Casts unpacked `int4` weights to `int8` right before Megablox GMM kernel compilation/execution.
    *   Fixed expert-sharded scale placeholders for static model initialization (`is_static=True`).
*   **`python/sgl_jax/test/kernels/`**:
    *   Added `test_kimi_int4_loading.py` (multi-host TP=16/EP=16) and `test_kimi_int4_loading_single.py` (TP=8/EP=1) verification tests.
    *   Added wrapper execution script `run_multihost_test.sh`.

---

## Accuracy Tests

We verified that the JAX/XLA graph compiles and executes successfully on TPU. 

*   **Test Environment**: 16 TPU devices of v7x / Ironwood (2 nodes of 4 chips each, where each chip has 2 devices).
*   **Execution Verification**: Both Dense (Layer 0 MLP) and Sparse MoE (Layer 1 EPMoE) compiled and materialized outputs successfully on TPU:
    *   **Dense FFN output shape**: `(16, 7168)`, dtype: `bfloat16`.
    *   **EPMoE GMM output shape**: `(16, 7168)`, dtype: `bfloat16`.
    *   **GMM Compiled shapes**: LHS `(128, 7168)`, RHS `(24, 7168, 2048)` (EP=16 Expert-sharded layout), scale `(24, 224, 1, 2048)`.

### Weight Loader Datatype Lifecycle Logs

Below are the trace logs verifying the correct datatype transitions from host loading to TPU compilation:

1.  **JAX TPU Bit-Unpacking**: Unpacks raw `int32` elements loaded from safetensors into JAX `int4` weights:
    ```
[...]

PR #1637, opened 2026-09-10
Kimi k2.5 : Changes to support int4 checkpoint.

chaitanyapk:
## Motivation

The goal of this Pull Request is to support loading and running inference on the official Kimi K2.5 / DeepSeek-V3 compressed-tensors INT4 pack-quantized weights in both single-host and multi-host TPU environments.

Previously, sglang-jax lacked a pipeline to load and unpack packed 4-bit checkpoint formats dynamically on TPUs. This PR introduces:

Dynamic On-Device Bit-Unpacking: JIT-compiled unpack_4bit_jax on TPU that unpacks int32/uint32/int8 packed arrays with the -8 offset correction required by the compressed-tensors signed 4-bit format.
Deferred Transposition: Prevents premature CPU/packed transpositions to eliminate bit corruption, performing the transpose in JAX on-device post-unpacking.
Abstract Memory Initialization: Introduces abstract jax.ShapeDtypeStruct parameter instantiation under nnx.eval_shape to prevent host RAM / TPU HBM out-of-memory errors when initializing large MoE models (e.g. 61 layers, 384 experts).
Scale Sharding Alignment: Corrects wo_scale and wi_scale tensor parallel sharding in EPMoE and DeepSeek-V3 weight mappings to match Megablox GMM kernel contracts, making wo_scale sharding conditional on block quantization (k_blocks > 1).
KV Cache Data Mesh Alignment: Aligns KV token capacity with the data mesh dimension in ModelRunnerKVCacheMixin.

## Accuracy Tests

All unit tests and end-to-end forward/decode passes have been verified on TPU:

Test Environment:
Hardware: TPU v7x (8 devices single-host, 16 devices multi-host).
Model: Kimi K2.5 (/dsk/models/kimi_original_new, 61 layers, 384 experts, packed INT4 safetensors).
Execution Verification:
Abstract initialization under nnx.eval_shape passed without OOM.
Both Dense FFN (Layer 0) and Sparse MoE (Layer 1 EPMoE) compiled and executed on TPU.
Full 61-layer model prompt generation tested with greedy sampling producing accurate responses:
Prompt: "The capital of France is" 
→
→ Output: "Paris."
Prompt: "What is the boiling point of water in Celsius? Answer:" 
→
→ Output: "100"
Prompt: "Calculate: 25 * 4 = " 
→
→ Output: "100"
Weight Loader Datatype Lifecycle Logs
JAX TPU Bit-Unpacking: Unpacks raw int32 tensors from safetensors into JAX int4 weights with centering:
INFO:root:Unpacking 4-bit weights with -8 offset on TPU: int32 -> int4
Materialized Parameter Shape & Dtype:

INFO:test_kimi_int4_loading:Parameter type after loading: <class 'jaxlib._jax.ArrayImpl'>
INFO:test_kimi_int4_loading:Loaded Parameter shape: (384, 7168, 2048), dtype: int4
Megablox GMM Execution:
INFO:sgl_jax.srt.layers.moe:EPMoE GMM stage - LHS activation shape: (16, 7168), dtype: bfloat16 | RHS weight shape: (384, 7168, 2048), dtype: int4 | RHS scale shape: (384, 224, 1, 2048), dtype: bfloat16

In addition to this, MMMU benchmarks were run and 78.11% accuracy was obtained using 32k token budget. 
The changes included in this PR along with https://github.com/sgl-project/sglang-jax/pull/1314 were done to test the model accuracy using evalscope. 

https://gist.github.com/chaitanyapk/b07160398746c2c37818506f7993376d

This has the initial 16k token budget run log with 75.11% accuracy and the 6 underperforming subjects that were hitting token budget were re-run with 32k budget to get the reported 78.11% accuracy.

## Checklist

### DIFF
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -220,7 +220,7 @@ class EPMoE(nnx.Module):
                         f"Expected k_blocks dimension to be 1 or {expected_k_blocks}."
                     )
             final_scale_sharding = (
-                P("expert", None, None, None)
+                P("expert", "tensor", None, None)
                 if scale_name == "wo_scale"
                 else P("expert", None, None, "tensor")
             )
@@ -247,7 +247,7 @@ class EPMoE(nnx.Module):
 
                 if scale.shape == (num_experts, out_dim, expected_k_blocks):
                     final_scale_sharding = (
-                        P("expert", None, None, None)
+                        P("expert", "tensor", None, None)
                         if scale_name == "wo_scale"
                         else P("expert", None, None, "tensor")
                     )
@@ -261,7 +261,7 @@ class EPMoE(nnx.Module):
                         else P("expert", "tensor", None)
                     )
                     final_scale_sharding = (
-                        P("expert", None, None, None)
+                        P("expert", "tensor", None, None)
                         if scale_name == "wo_scale"
                         else P("expert", None, None, "tensor")
                     )
@@ -340,40 +340,119 @@ class EPMoE(nnx.Module):
                 k_blocks_wi = (hidden_size // block_size_k) if block_size_k else 1
                 k_blocks_wo = (intermediate_dim // block_size_k) if block_size_k else 1
                 wi_scale_sharding = P("expert", None, None, "tensor")
-                wo_scale_sharding = P("expert", None, None, None)
+                wo_scale_sharding = P("expert", "tensor", None, None)
+                wi_sharding = P("expert", None, "tensor")
+                wo_sharding = P("expert", "tensor", None)
 
-                if hasattr(self, "wi_0_scale"):
-                    del self.wi_0_scale
-                self.wi_0_scale = nnx.Param(
-                    jnp.zeros(
-                        (num_experts, k_blocks_wi, 1, intermediate_dim),
-                        dtype=jnp.float32,
+                is_abstract = isinstance(self.wi_0.value, jax.ShapeDtypeStruct)
+
+                if is_abstract:
+                    if hasattr(self, "wi_0_scale"):
+                        del self.wi_0_scale
+                    self.wi_0_scale = nnx.Param(
+                        jax.ShapeDtypeStruct(
+                            (num_experts, k_blocks_wi, 1, intermediate_dim),
+                            dtype=self.dtype,
+                            sharding=jax.sharding.NamedSharding(self.moe_mesh, wi_scale_sharding),
+                        )
+                    )
+
+                    if hasattr(self, "wi_1_scale"):
+                        del self.wi_1_scale
+                    self.wi_1_scale = nnx.Param(
+                        jax.ShapeDtypeStruct(
+                            (num_experts, k_blocks_wi, 1, intermediate_dim),
+                            dtype=self.dtype,
+                            sharding=jax.sharding.NamedSharding(self.moe_mesh, wi_scale_sharding),
+                        )
+                    )
+
+                    if hasattr(self, "wo_scale"):
+                        del self.wo_scale
+                    self.wo_scale = nnx.Param(
+                        jax.ShapeDtypeStruct(
+                            (num_experts, k_blocks_wo, 1, hidden_size),
+                            dtype=self.dtype,
+                            sharding=jax.sharding.NamedSharding(self.moe_mesh, wo_scale_sharding),
+                        )
+                    )
+
+                    self.wi_0 = nnx.Param(
+                        jax.ShapeDtypeStruct(
+                            (num_experts, hidden_size, intermediate_dim),
+                            dtype=self.quantized_dtype,
+                            sharding=jax.sharding.NamedSharding(self.moe_mesh, wi_sharding),
+                        )
+                    )
+                    self.wi_1 = nnx.Param(
+                        jax.ShapeDtypeStruct(
+                            (num_experts, hidden_size, intermediate_dim),
+                            dtype=self.quantized_dtype,
+                            sharding=jax.sharding.NamedSharding(self.moe_mesh, wi_sharding),
+                        )
+                    )
+                    self.wo = nnx.Param(
+                        jax.ShapeDtypeStruct(
+                            (num_experts, intermediate_dim, hidden_size),
+                            dtype=self.quantized_dtype,
+                            sharding=jax.sharding.NamedSharding(self.moe_mesh, wo_sharding),
+                        )
+                    )
+                else:
+                    if hasattr(self, "wi_0_scale"):
+                        del self.wi_0_scale
+                    self.wi_0_scale = nnx.Param(
+                        jnp.zeros(
+                            (num_experts, k_blocks_wi, 1, intermediate_dim),
+                            dtype=self.dtype,
+                            out_sharding=wi_scale_sharding,
+                        ),
                         out_sharding=wi_scale_sharding,
-                    ),
-                    out_sharding=wi_scale_sharding,
-                )
+                    )
 
-                if hasattr(self, "wi_1_scale"):
-                    del self.wi_1_scale
-                self.wi_1_scale = nnx.Param(
-                    jnp.zeros(
-                        (num_experts, k_blocks_wi, 1, intermediate_dim),
-                        dtype=jnp.float32,
+                    if hasattr(self, "wi_1_scale"):
+                        del self.wi_1_scale
+                    self.wi_1_scale = nnx.Param(
+                        jnp.zeros(
+                            (num_experts, k_blocks_wi, 1, intermediate_dim),
+                            dtype=self.dtype,
+                            out_sharding=wi_scale_sharding,
+                        ),
                         out_sharding=wi_scale_sharding,
-                    ),
-                    out_sharding=wi_scale_sharding,
-                )
+                    )
 
-                if hasattr(self, "wo_scale"):
-                    del self.wo_scale
-                self.wo_scale = nnx.Param(
-                    jnp.zeros(
-                        (num_experts, k_blocks_wo, 1, hidden_size),
-                        dtype=jnp.float32,
+                    if hasattr(self, "wo_scale"):
+                        del self.wo_scale
+                    self.wo_scale = nnx.Param(
+                        jnp.zeros(
+                            (num_experts, k_blocks_wo, 1, hidden_size),
+                            dtype=self.dtype,
+                            out_sharding=wo_scale_sharding,
+                        ),
                         out_sharding=wo_scale_sharding,
-                    ),
-                    out_sharding=wo_scale_sharding,
-                )
+                    )
+
+                    self.wi_0 = nnx.Param(
+                        jnp.zeros(
+                            (num_experts, hidden_size, intermediate_dim),
+                            dtype=self.quantized_dtype,
+                        ),
+                        out_sharding=wi_sharding,
+                    )
+                    self.wi_1 = nnx.Param(
+                        jnp.zeros(
+                            (num_experts, hidden_size, intermediate_dim),
+                            dtype=self.quantized_dtype,
+                        ),
+                        out_sharding=wi_sharding,
+                    )
+                    self.wo = nnx.Param(
+                        jnp.zeros(
+                            (num_experts, intermediate_dim, hidden_size),
+                            dtype=self.quantized_dtype,
+                        ),
+                        out_sharding=wo_sharding,
+                    )
                 return
 
             # Quantize weights along k-dim (axis=1 in [g, k, n] layout)
@@ -437,7 +516,7 @@ class EPMoE(nnx.Module):
                 del self.wo_scale
             self.wo_scale = nnx.Param(
                 wo_scale,
-                out_sharding=P("expert", None, None, None),
+                out_sharding=P("expert", "tensor", None, None),
             )
 
     @named_scope
@@ -514,7 +593,7 @@ class EPMoE(nnx.Module):
                     # scales [g, 1, 1, n]
                     P("expert", None, None, "tensor"),
                     P("expert", None, None, "tensor"),
-                    P("expert", None, None, None),
+                    P("expert", "tensor", None, None),
                     # biases [g, 1, n] (unused)
                     P("expert", None, "tensor"),
                     P("expert", None, "tensor"),
@@ -857,6 +936,7 @@ def create_moe_weights_mapping(
     moe_path: str = "mlp",
     source_expert_pattern: str = "experts.{i}",
     physical_to_logical_map=None,  # np.ndarray shape (num_physical,) or None
+    weight_suffix: str = "weight",
 ) -> dict:
     """Generate a unified mapping dictionary for MoE layer expert weights."""
     if moe_backend == "epmoe":
@@ -884,7 +964,7 @@ def create_moe_weights_mapping(
 
         # Source weight paths for logical experts only
         expert_keys = [
-            f"{prefix}.{moe_path}.{source_expert_pattern.format(i=i)}.{source_name}.weight"
+            f"{prefix}.{moe_path}.{source_expert_pattern.format(i=i)}.{source_name}.{weight_suffix}"
             for i in range(num_experts)
         ]
 
--- a/python/sgl_jax/srt/layers/moe.py
+++ b/python/sgl_jax/srt/layers/moe.py
@@ -220,7 +220,7 @@ class EPMoE(nnx.Module):
                         f"Expected k_blocks dimension to be 1 or {expected_k_blocks}."
                     )
             final_scale_sharding = (
-                P("expert", "tensor", None, None)
+                P("expert", "tensor" if scale.shape[1] > 1 else None, None, None)
                 if scale_name == "wo_scale"
                 else P("expert", None, None, "tensor")
             )
@@ -247,7 +247,7 @@ class EPMoE(nnx.Module):
 
                 if scale.shape == (num_experts, out_dim, expected_k_blocks):
                     final_scale_sharding = (
-                        P("expert", "tensor", None, None)
+                        P("expert", "tensor" if expected_k_blocks > 1 else None, None, None)
                         if scale_name == "wo_scale"
                         else P("expert", None, None, "tensor")
                     )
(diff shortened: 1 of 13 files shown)


chaitanyapk:
Previous [draft](https://github.com/sgl-project/sglang-jax/pull/1334) which got closed in favour of this

chaitanyapk:
@Bob-Chen222 @cjx0709 request you review these changes when you find time. The earlier dependency cited [here](https://github.com/sgl-project/sglang-jax/pull/1334#issuecomment-4665823409) has been merged and the current PR has been tested to achieve reported accuracy on MMMU benchmark.

python/sgl_jax/test/kernels/test_kimi_int4_loading.py
```diff
@@ -0,0 +1,198 @@
+import logging
+import os
+
+host0 = "10.128.0.2"
```
cjx0709, line 4: this test case should not be hardcode.

python/sgl_jax/srt/models/deepseek_v3.py
```diff
@@ -940,13 +955,21 @@ def add_linear(hf_prefix: str, target_prefix: str, sharding_std: tuple):
             physical_to_logical_map = np.array(jax.device_get(metadata.physical_to_logical_map))
             phy_to_log = physical_to_logical_map[layer_idx]
 
+        int4_types = [
```
cjx0709, line 958: the same int4_types been defined in 4 places, maybe we can make them into one.

python/sgl_jax/srt/utils/quantization/quantization_utils.py
```diff
@@ -173,9 +167,14 @@ def _replace_linear_recursive(obj, path: str = "", visited: set | None = None):
                 if isinstance(attr_value, LinearBase):
                     # Check if this path matches any rule
                     dot_path = child_path.replace("/", ".")
+                    path_parts = dot_path.replace("[", ".").replace("]", "").split(".")
                     if any(
-                        dot_path == ignored or dot_path.endswith(f".{ignored}")
-                        for ignored in ignored_layers
+                        ig in path_parts
```
cjx0709, line 172: `ig in path_parts` broadens the matching logic for ALL models — before it was suffix-only, now any path segment match. e.g. `ignored_layers=["mlp"]` would skip every layer containing "mlp". could you scope this to int4 only, or add a test proving existing models are not affected?

python/sgl_jax/srt/utils/quantization/quantization_utils.py
```diff
@@ -96,18 +96,12 @@ def apply_linear_quantization(
     from sgl_jax.srt.layers.linear import LinearBase, QuantizedLinear
 
     quant_config = model_config.quantization_config
-    if quant_config is None:
-        raise ValueError(
-            "apply_linear_quantization called but model_config.quantization_config is None. "
-            "Ensure --quantization-config-path is set."
-        )
+    if quant_config is None or not quant_config.has_linear_quantization():
+        return model
```
cjx0709, line 100: silently returning hides misconfigured quant setups — previously this was a hard error. for int4 (no linear rules) returning is correct, but a fp8 model with forgotten linear_rules would get silent no-quantization. maybe gate on `has_linear_quantization()` specifically?

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
```diff
@@ -502,8 +502,12 @@ def _apply_token_constraints(
                 )
             token_capacity = min(token_capacity, max_total_tokens)
 
-        # Page alignment
-        token_capacity = token_capacity // self.server_args.page_size * self.server_args.page_size
+        # Page and Data mesh alignment
+        data_mesh_dim = 1
+        if hasattr(self, "mesh") and self.mesh is not None and "data" in self.mesh.shape:
+            data_mesh_dim = self.mesh.shape["data"]
+        alignment = self.server_args.page_size * max(1, data_mesh_dim)
```
cjx0709, line 509: this changes token_capacity alignment for ALL models with data parallelism (e.g. data_mesh=4, page_size=64 → alignment 64→256, losing capacity). what concrete failure does this fix? if Kimi-specific, better handle it there.

python/sgl_jax/srt/layers/moe.py
```diff
@@ -340,40 +340,119 @@ def _get_block_size_k(
...
-                    del self.wi_0_scale
-                self.wi_0_scale = nnx.Param(
-                    jnp.zeros(
-                        (num_experts, k_blocks_wi, 1, intermediate_dim),
-                        dtype=jnp.float32,
+                wo_scale_sharding = P("expert", "tensor" if k_blocks_wo > 1 else None, None, None)
+                wi_sharding = P("expert", None, "tensor")
+                wo_sharding = P("expert", "tensor", None)
+
+                is_abstract = isinstance(self.wi_0.value, jax.ShapeDtypeStruct)
```
cjx0709, line 347: abstract vs concrete branch doubles ~80 lines of near-identical shape/sharding logic. since int4 static checkpoint always uses `nnx.eval_shape` (abstract path), does the concrete `jnp.zeros` path for wi_0/wi_1/wo ever get hit? consider extracting a helper or dropping the dead branch.

python/sgl_jax/srt/layers/moe.py
```diff
@@ -514,7 +593,7 @@ def __call__(
                     # scales [g, 1, 1, n]
                     P("expert", None, None, "tensor"),
                     P("expert", None, None, "tensor"),
-                    P("expert", None, None, None),
+                    P("expert", "tensor" if (wo_scale is not None and wo_scale.shape[1] > 1) else None, None, None),
```
cjx0709, line 596: the inline `"tensor" if (wo_scale is not None and wo_scale.shape[1] > 1) else None` is repeated 5 times across this file (`_normalize_scale_for_gmm` ×3, `quantize_weights`, `__call__`). tpu-inference handles this much cleaner — determines the sharding spec once upfront based on shape, then applies it everywhere. see `tpu-inference/layers/common/process_weights/moe_weights.py:574-607` for reference. suggest computing `wo_scale_sharding` once and reusing.

python/sgl_jax/test/kernels/test_kimi_int4_e2e_prompt.py
```diff
@@ -0,0 +1,233 @@
...
+        out = generate_for_prompt(p, model_runner, tokenizer, max_new_tokens=32)
+        print(f"RESPONSE: {out}")
+        print("=" * 50 + "\n")
+        assert len(out.strip()) > 0, f"Empty response generated for prompt: {p}"
+
+    logger.info("SUCCESS: All prompt generation tests passed with verified legible responses!")
+
+
+if __name__ == "__main__":
+    main()
```
cjx0709, line 233: a few test quality things:
- existing convention is `*_test.py`, these use `test_*.py`
- this and `test_kimi_int4_loading*.py` are `main()` scripts not pytest functions — CI won't pick them up
- hardcoded model paths (`/dsk/...`) should be env-gated with `pytest.mark.skipif`

`test_int4_quantization.py` is the only one in proper pytest format 👍

python/sgl_jax/srt/layers/moe.py
chaitanyapk: Done. extracted helper function.

python/sgl_jax/test/kernels/kimi_int4_e2e_prompt_test.py
```diff
@@ -0,0 +1,233 @@
...
+        out = generate_for_prompt(p, model_runner, tokenizer, max_new_tokens=32)
+        print(f"RESPONSE: {out}")
+        print("=" * 50 + "\n")
+        assert len(out.strip()) > 0, f"Empty response generated for prompt: {p}"
+
+    logger.info("SUCCESS: All prompt generation tests passed with verified legible responses!")
+
+
+if __name__ == "__main__":
+    main()
```
chaitanyapk: Done. made test file related fixes.

python/sgl_jax/test/kernels/test_kimi_int4_loading.py
chaitanyapk: Done.

python/sgl_jax/srt/models/deepseek_v3.py
chaitanyapk: Done.

python/sgl_jax/srt/utils/quantization/quantization_utils.py
chaitanyapk: Done, made this check specific.

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
chaitanyapk: This was the initial sharding divisibility error that was faced.
`  File "/usr/local/lib/python3.12/site-packages/jax/_src/sharding.py", line 72, in _common_shard_shape
    raise IndivisibleError(
jax._src.sharding.IndivisibleError: Sharding NamedSharding(
    mesh=Mesh('data': 2, 'tensor': 4), 
    spec=PartitionSpec('data', None, 'tensor', None, None), 
    memory_kind=device
) implies that array axis 0 is partitioned 2 times, but the dimension size is 781 
(full shape: (781, 16, 16, 1, 128), per-dimension tiling factors: [2, 1, 4, 1, 1] should evenly divide the shape)`

Updated the logic to retain the previous code and update only for cases where data_mesh_dim>1.
`total_pages = (total_pages // data_mesh_dim) * data_mesh_dim` ensuring divisbility.

python/sgl_jax/srt/layers/moe.py
chaitanyapk: Done.

python/sgl_jax/srt/utils/quantization/quantization_utils.py
chaitanyapk: Done.

chaitanyapk:
@cjx0709 I have incorporated feedback from initial review. Please take a look when you find time and suggest any changes.

python/sgl_jax/srt/models/deepseek_v3.py
```diff
@@ -918,7 +933,25 @@ def add_linear(hf_prefix: str, target_prefix: str, sharding_std: tuple):
                 add_linear(f"{prefix}.mlp.{proj}", f"{target}.mlp.{proj}", sharding)
             return mappings
 
-        # MoE gate (router) — NOT quantized in HF FP8 checkpoint.
+        # MoE Shared Experts
```
cjx0709, line 936: Could we remove this duplicate shared-expert block? The existing block below already handles both backends and uses the correct non-fused target path.

python/sgl_jax/srt/models/deepseek_v3.py
```diff
@@ -967,17 +1005,19 @@ def add_linear(hf_prefix: str, target_prefix: str, sharding_std: tuple):
...
-                # expands via jnp.take, which fails if the gathered axis is
-                # tensor-sharded (ambiguous output sharding). The converter reshards
-                # to model_param.value.sharding at the end.
+                # Stacked checkpoint scale is `[E, out_blocks, in_blocks]`.
+                # For wi_0/wi_1: out_dim (intermediate_dim) is dim 1, sharded on "tensor".
+                # For wo: in_dim (intermediate_dim // 32 = k_blocks_wo) is dim 2, sharded on "tensor".
+                if "wo" in target_base:
+                    scale_sharding = ("expert", None, "tensor")
+                else:
+                    scale_sharding = ("expert", "tensor", None)
```
cjx0709, line 1017: Could we keep the old scale sharding for non-INT4 checkpoints? On JAX 0.11.1, the FP8 loader's jnp.take raises ShardingTypeError with this tensor-sharded layout; the old replicated layout works.

python/sgl_jax/srt/layers/moe.py
```diff
@@ -340,39 +359,64 @@ def _get_block_size_k(
...
-                self.wi_0_scale = nnx.Param(
-                    jnp.zeros(
-                        (num_experts, k_blocks_wi, 1, intermediate_dim),
-                        dtype=jnp.float32,
-                        out_sharding=wi_scale_sharding,
-                    ),
-                    out_sharding=wi_scale_sharding,
+                self.wi_0_scale = _make_param(
+                    (num_experts, k_blocks_wi, 1, intermediate_dim),
+                    self.dtype,
```
cjx0709, line 386: Could we keep float32 scales for non-INT4 checkpoints? The loader casts scales to this dtype, so this also downcasts existing FP8 scales to BF16.

python/sgl_jax/srt/models/deepseek_v3.py
```diff
@@ -940,13 +973,18 @@ def add_linear(hf_prefix: str, target_prefix: str, sharding_std: tuple):
             physical_to_logical_map = np.array(jax.device_get(metadata.physical_to_logical_map))
             phy_to_log = physical_to_logical_map[layer_idx]
 
+        is_int4_moe = is_int4_dtype(getattr(quant_config, "moe_weight_dtype", None))
```
cjx0709, line 976: Could we also require is_static_quant before selecting weight_packed? Dynamic INT4 should load ordinary .weight tensors first, but currently fails looking for packed weights.

python/sgl_jax/srt/utils/quantization/quantization_utils.py
```diff
@@ -97,17 +106,31 @@ def apply_linear_quantization(
 
     quant_config = model_config.quantization_config
     if quant_config is None:
-        raise ValueError(
-            "apply_linear_quantization called but model_config.quantization_config is None. "
-            "Ensure --quantization-config-path is set."
-        )
+        return model
```
cjx0709, line 109: Could we retain the existing validation and QuantizationConfig interface? The loader already skips this call without linear rules; the INT4 tests can do the same without adding silent returns and fallback branches here.

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
cjx0709: The pool adds dp_size reserved pages afterward, so data=2 and dp_size=1 still produces an odd page count. Could we include those pages in the alignment and add a regression test, or split this fix out?

python/sgl_jax/srt/utils/quantization/quantization_utils.py
cjx0709: The new startswith check still affects all linear quantization configs, while this INT4 path has no linear rules and never reaches it. Could we leave the existing matching behavior unchanged?

python/sgl_jax/srt/utils/quantization/quantization_utils.py
```diff
@@ -174,8 +197,10 @@ def _replace_linear_recursive(obj, path: str = "", visited: set | None = None):
                     # Check if this path matches any rule
                     dot_path = child_path.replace("/", ".")
                     if any(
-                        dot_path == ignored or dot_path.endswith(f".{ignored}")
-                        for ignored in ignored_layers
+                        dot_path == ig
+                        or dot_path.startswith(f"{ig}.")
```
cjx0709, line 201: The new startswith check still affects all linear quantization configs, while this INT4 path has no linear rules and never reaches it. Could we leave the existing matching behavior unchanged?

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
```diff
@@ -503,10 +503,19 @@ def _apply_token_constraints(
...
 
         # DP scale
         token_capacity = token_capacity * dp_size
+
+        # Data mesh sharding alignment on total pages
+        if hasattr(self, "mesh") and self.mesh is not None and "data" in self.mesh.shape:
+            data_mesh_dim = self.mesh.shape["data"]
+            if data_mesh_dim > 1:
+                total_pages = token_capacity // self.server_args.page_size
+                total_pages = (total_pages // data_mesh_dim) * data_mesh_dim
```
cjx0709, line 516: The pool adds dp_size reserved pages afterward, so data=2 and dp_size=1 still produces an odd page count. Could we include those pages in the alignment and add a regression test, or split this fix out?

python/sgl_jax/srt/models/deepseek_v3.py
chaitanyapk: Done.

python/sgl_jax/srt/models/deepseek_v3.py
chaitanyapk: Done.

python/sgl_jax/srt/layers/moe.py
chaitanyapk: Done.

python/sgl_jax/srt/models/deepseek_v3.py
chaitanyapk: Done.

python/sgl_jax/srt/utils/quantization/quantization_utils.py
chaitanyapk: Done.

python/sgl_jax/srt/utils/quantization/quantization_utils.py
chaitanyapk: Done, retained old behaviour.

python/sgl_jax/srt/model_executor/model_runner_kv_cache_mixin.py
chaitanyapk: updated the logic to incorporate addition later and retain divisibility. Added tests toward that end too. Let me know if that looks good to you.

chaitanyapk:
@cjx0709 incorporated changes from the second review. I've added tests targeted towards the gaps identified. Please take a look when you find time and share your thoughts.

python/sgl_jax/srt/utils/weight_utils.py
```diff
@@ -2252,6 +2310,9 @@ def load_weights_from_safetensors(
                         _t_assign = time.monotonic() - _t_assign_start
                         if _pd_cache:
                             _PD_WEIGHT_CACHE[target_path] = model_param.value
+                        del stacked_weight
+                        import gc
+                        gc.collect()
```
cjx0709: is it necessary for all bf16/fp8 model?

chaitanyapk: I think this might not be needed for all the other models. with kimi, since the packed weights were used, host memory was of relevance. I see that this is outdated on the latest commit. Let me know if anything else needs to be done on this item.

cjx0709:
@chaitanyapk  hello, the back-and-forth process of reviewing comments and making revisions is too inefficient. I’ve pushed some commits based on the points I thought needed changing（hope that you will not mind）, please take a look at these updates.

PR #1639, opened 2026-09-11
Enabling Spec decode for GLM 5.2

amanseervi:
## Motivation

GLM-5.2 MoE checkpoints ship a native Multi-Token Prediction (MTP) module, but sglang-jax had no way to use it: the speculative decoding path only knew how to build draft models for a limited set of architectures, and the GLM decoder/model code had no MTP entrypoint. As a result, GLM MoE models could only run in plain autoregressive decode mode, leaving a large amount of decode throughput on the table.

This PR adds end-to-end support for NEXTN / MTP speculative decoding for GLM-5.x MoE models, so the native MTP layer that already exists in the released weights can be used as the draft model at serving time.

## Modifications

**Draft model architecture routing**
- Added draft-model architecture remapping so that when a GLM MoE checkpoint is loaded as a draft model, it is routed to the dedicated NextN architecture instead of the full target architecture. This mirrors the existing remapping done for other MTP-capable model families.

**GLM NextN (MTP) draft model**
- Implemented a new `GlmMoeDsaForCausalLMNextN` module and registered it as an entry class. It implements the standard MTP structure: token embedding, `enorm` / `hnorm` normalization of the embedding and the target hidden state, an `eh_proj` fusion projection, a single decoder block, a shared head norm, and the LM head + logits processor.
- The draft forward pass reshards the incoming target hidden states to match the embedding sharding before concatenation, so the draft step stays consistent with the target model's sharding under multi-device meshes.
- Added weight loading for the MTP layer: the MTP block's weights are discovered at the checkpoint's trailing layer index and remapped onto the local draft module paths, reusing the existing MoE layer mapping logic so quantized and unquantized checkpoints are handled the same way. Post-load hooks are invoked for attention, MLP and shared experts so the draft block goes through the same weight post-processing as the target model.
- The draft's KV output is normalized into the format expected by the KV pool, including the fused-cache variant.

**Shared embedding / LM head between target and draft**
- Added `get_embed_and_head` / `set_embed_and_head` / `set_embed` accessors on both the target and the draft model, so the draft worker can share the target model's embedding and LM head instead of holding a duplicate copy. This avoids redundant memory and keeps draft and target token spaces exactly aligned.

**Decoder layer reuse for the MTP block**
- Made the shared GLM decoder layer usable as an MTP block:
  - Added a `force_moe` flag so the MTP block always builds the MoE MLP, instead of falling back to a dense MLP when its layer index happens to be below `first_k_dense_replace`.
  - Made indexer-type lookup bounds-safe, since the MTP layer index can sit outside the target model's per-layer indexer configuration.

**KV cache and memory accounting for speculative decoding**
- The draft worker now reports a single KV-cache layer, matching the fact that the NextN module only has one decoder block. Previously it inherited the target model's layer count, which over-allocated the draft KV pool and could push allocations out of bounds.
- When speculative decoding is enabled, the target model runner now reserves headroom out of the profiled KV-cache budget for the draft worker's KV buffer and allocator fragmentation, preventing OOM during draft worker initialization.

**Speculative worker fixes**
- Fixed the draft output buffer shape selection so `NEXTN` with `topk == 1` uses the multi-layer/step-shaped buffers, matching what the NextN draft actually produces.

## Accuracy Tests

Tried with a simple curl request : 

Request : 
```bash
 curl -X POST http://${SERVER}:30271/v1/chat/completions \
  -d '{
    "model": "/weights",
    "messages": [
      {"role": "user", "content": "Hello! Give me a quick introduction to JAX."}
    ],
    "temperature": 0.0,
    "max_tokens": 128
  }'
```

Response : 

```bash
{"id":"d2b93f75dc7e43bb8de9eac288ab4dc7","object":"chat.completion","created":1785766167,"model":"/weights","choices":[{"index":0,"message":{"role":"assistant","content":"1.  **Understand the User's Request:**\n    *   **Topic:** JAX (a numerical computing library in Python).\n    *   **Goal:** Provide a quick introduction.\n    *   **Tone:** Friendly, informative, concise.\n\n2.  **Identify Key Concepts of JAX:**\n    *   What is it? (NumPy on steroids, developed by Google/DeepMind).\n    *   Core features/philosophy:\n        *   Autograd (automatic differentiation).\n        *   JIT (Just-In-Time compilation) via XLA.\n        *   Vectorization (vmap).\n       ","reasoning_content":null,"tool_calls":null},"logprobs":null,"finish_reason":"length","matched_stop":null,"hidden_states":null}],"usage":{"prompt_tokens":23,"total_tokens":151,"completion_tokens":128,"prompt_tokens_details":null}}

```

## Benchmarking and Profiling

Benchmarked with `sgl_jax.bench_one_batch_server`, input length 8192, output length 1024:

### Baseline (no speculative decoding)

| Batch size | Latency (s) | Input throughput (tok/s) | Output throughput (tok/s) | ITL (ms) | Input cost ($/1M) | Output cost ($/1M) |
| ---------- | ----------- | ------------------------ | ------------------------- | -------- | ----------------- | ------------------ |
| 1          | 36.21       | 8201.01                  | 29.09                     | 34.38    | 0.10              | 19.10              |
| 4          | 46.76       | 8324.03                  | 95.64                     | 41.82    | 0.10              | 5.81               |
| 8          | 59.15       | 8382.38                  | 159.60                    | 50.12    | 0.09              | 3.48               |
| 16         | 79.66       | 8363.95                  | 256.05                    | 62.49    | 0.09              | 2.17               |
| 32         | 118.38      | 8398.50                  | 375.93                    | 85.12    | 0.09              | 1.48               |
| 64         | 236.35      | 3127.98                  | 953.38                    | 67.13    | 0.25              | 0.58               |

### With NEXTN / MTP speculative decoding (num speculative tokens = 3)

| Batch size | Latency (s) | Input throughput (tok/s) | Output throughput (tok/s) | ITL (ms) | Input cost ($/1M) | Output cost ($/1M) |
| ---------- | ----------- | ------------------------ | ------------------------- | -------- | ----------------- | ------------------ |
| 1          | 10.08       | 9690.97                  | 110.93                    | 9.01     | 0.08              | 5.01               |
| 4          | 19.78       | 10975.13                 | 243.96                    | 16.40    | 0.07              | 2.28               |
| 8          | 27.30       | 11063.69                 | 383.27                    | 20.87    | 0.07              | 1.45               |
| 16         | 42.00       | 11016.48                 | 544.28                    | 29.40    | 0.07              | 1.02               |
| 32         | 68.25       | 11086.87                 | 734.57                    | 43.56    | 0.07              | 0.76               |
| 64         | 129.48      | 5411.68                  | 2010.05                   | 31.84    | 0.15              | 0.28               |

### Speedup summary

| Batch size | Latency (s) base → MTP | Speedup | Output tok/s base → MTP | Speedup | ITL (ms) base → MTP | Output cost ($/1M) base → MTP |
| ---------- | ---------------------- | ------- | ----------------------- | ------- | ------------------- | ----------------------------- |
| 1          | 36.21 → 10.08          | 3.59x   | 29.09 → 110.93          | 3.81x   | 34.38 → 9.01        | 19.10 → 5.01                  |
| 4          | 46.76 → 19.78          | 2.36x   | 95.64 → 243.96          | 2.55x   | 41.82 → 16.40       | 5.81 → 2.28                   |
| 8          | 59.15 → 27.30          | 2.17x   | 159.60 → 383.27         | 2.40x   | 50.12 → 20.87       | 3.48 → 1.45                   |
| 16         | 79.66 → 42.00          | 1.90x   | 256.05 → 544.28         | 2.13x   | 62.49 → 29.40       | 2.17 → 1.02                   |
| 32         | 118.38 → 68.25         | 1.73x   | 375.93 → 734.57         | 1.95x   | 85.12 → 43.56       | 1.48 → 0.76                   |
| 64         | 236.35 → 129.48        | 1.83x   | 953.38 → 2010.05        | 2.11x   | 67.13 → 31.84       | 0.58 → 0.28                   |

Across batch sizes 1–64, enabling native MTP speculative decoding gives roughly a **1.7x–3.6x** end-to-end latency reduction and **~2x–3.8x** higher output throughput, with a corresponding drop in per-token output cost. Prefill throughput is also slightly higher since the draft KV pool is now sized correctly and frees up headroom.

## Checklist

### DIFF
--- a/python/sgl_jax/bench_one_batch_server.py
+++ b/python/sgl_jax/bench_one_batch_server.py
@@ -21,11 +21,7 @@ import time
 
 import requests
 
-from sgl_jax.bench_serving import (
-    get_tokenizer,
-    sample_random_requests,
-    sample_generated_shared_prefix_requests,
-)
+from sgl_jax.bench_serving import get_tokenizer, sample_random_requests
 from sgl_jax.profiler import run_profile
 from sgl_jax.srt.entrypoints import http_server
 from sgl_jax.srt.server_args import ServerArgs
@@ -50,18 +46,6 @@ class BenchArgs:
     profile: bool = False
     profile_by_stage: bool = False
     api_type: str = "native"  # "native" or "openai"
-    dataset_name: str = "random"
-    gsp_num_groups: int = 1
-    gsp_prompts_per_group: int = 1
-    gsp_system_prompt_len: int = 1024
-    gsp_question_len: int = 100
-    gsp_output_len: int = 16
-    gsp_range_ratio: float = 1.0
-    seed: int = 1
-    gsp_send_routing_key: bool = False
-    gsp_num_turns: int = 1
-    gsp_fast_prepare: bool = False
-    gsp_ordered: bool = False
 
     @staticmethod
     def add_cli_args(parser: argparse.ArgumentParser):
@@ -94,25 +78,12 @@ class BenchArgs:
             choices=["native", "openai"],
             help="API type to use: 'native' for /generate or 'openai' for /v1/completions",
         )
-        parser.add_argument("--dataset-name", type=str, default=BenchArgs.dataset_name)
-        parser.add_argument("--gsp-num-groups", type=int, default=BenchArgs.gsp_num_groups)
-        parser.add_argument("--gsp-prompts-per-group", type=int, default=BenchArgs.gsp_prompts_per_group)
-        parser.add_argument("--gsp-system-prompt-len", type=int, default=BenchArgs.gsp_system_prompt_len)
-        parser.add_argument("--gsp-question-len", type=int, default=BenchArgs.gsp_question_len)
-        parser.add_argument("--gsp-output-len", type=int, default=BenchArgs.gsp_output_len)
-        parser.add_argument("--gsp-range-ratio", type=float, default=BenchArgs.gsp_range_ratio)
-        parser.add_argument("--seed", type=int, default=BenchArgs.seed)
-        parser.add_argument("--gsp-send-routing-key", action="store_true", default=BenchArgs.gsp_send_routing_key)
-        parser.add_argument("--gsp-num-turns", type=int, default=BenchArgs.gsp_num_turns)
-        parser.add_argument("--gsp-fast-prepare", action="store_true", default=BenchArgs.gsp_fast_prepare)
-        parser.add_argument("--gsp-ordered", action="store_true", default=BenchArgs.gsp_ordered)
 
     @classmethod
     def from_cli_args(cls, args: argparse.Namespace):
         # use the default value's type to cast the args into correct types.
         attrs = [(attr.name, type(attr.default)) for attr in dataclasses.fields(cls)]
-        # Use default from dataclass if argument isn't present in args
-        return cls(**{attr: attr_type(getattr(args, attr, getattr(cls, attr))) for attr, attr_type in attrs})
+        return cls(**{attr: attr_type(getattr(args, attr)) for attr, attr_type in attrs})
 
 
 def launch_server_internal(server_args):
@@ -160,42 +131,21 @@ def run_one_case(
     profile: bool = False,
     profile_by_stage: bool = False,
     api_type: str = "native",
-    bench_args: BenchArgs = None,
 ):
     requests.post(url + "/flush_cache")
 
     # Determine whether to use text or input_ids based on API type
     return_text = api_type == "openai"
-    
-    if bench_args is not None and getattr(bench_args, "dataset_name", "") == "generated-shared-prefix":
-        # Override prompts_per_group so we can test different batch sizes in one run
-        dynamic_prompts_per_group = max(1, batch_size // bench_args.gsp_num_groups)
-        bench_args.gsp_prompts_per_group = dynamic_prompts_per_group
-        
-        input_requests = sample_generated_shared_prefix_requests(
-            num_groups=bench_args.gsp_num_groups,
-            prompts_per_group=dynamic_prompts_per_group,
-            system_prompt_len=bench_args.gsp_system_prompt_len,
-            question_len=bench_args.gsp_question_len,
-            output_len=bench_args.gsp_output_len,
-            range_ratio=getattr(bench_args, "gsp_range_ratio", 1.0),
-            tokenizer=tokenizer,
-            args=bench_args,
-        )
-        batch_size = len(input_requests)
-        input_len = bench_args.gsp_system_prompt_len + bench_args.gsp_question_len
-        output_len = bench_args.gsp_output_len
-    else:
-        input_requests = sample_random_requests(
-            input_len=input_len,
-            output_len=output_len,
-            num_prompts=batch_size,
-            range_ratio=1.0,
-            tokenizer=tokenizer,
-            dataset_path="",
-            random_sample=True,
-            return_text=return_text,
-        )
+    input_requests = sample_random_requests(
+        input_len=input_len,
+        output_len=output_len,
+        num_prompts=batch_size,
+        range_ratio=1.0,
+        tokenizer=tokenizer,
+        dataset_path="",
+        random_sample=True,
+        return_text=return_text,
+    )
 
     use_structured_outputs = False
     if use_structured_outputs:
@@ -271,25 +221,20 @@ def run_one_case(
                     total_chunks += 1
     else:
         # Use native API
-        json_data = {
-            "sampling_params": {
-                "temperature": temperature,
-                "max_new_tokens": output_len,
-                "ignore_eos": True,
-                "json_schema": json_schema,
-                "stream_interval": stream_interval,
-            },
-            "return_logprob": return_logprob,
-            "stream": True,
-        }
-        if getattr(bench_args, "dataset_name", "") == "generated-shared-prefix":
-            json_data["text"] = [req.prompt for req in input_requests]
-        else:
-            json_data["input_ids"] = [req.prompt for req in input_requests]
-
         response = requests.post(
             url + "/generate",
-            json=json_data,
+            json={
+                "input_ids": [req.prompt for req in input_requests],
+                "sampling_params": {
+                    "temperature": temperature,
+                    "max_new_tokens": output_len,
+                    "ignore_eos": True,
+                    "json_schema": json_schema,
+                    "stream_interval": stream_interval,
+                },
+                "return_logprob": return_logprob,
+                "stream": True,
+            },
             stream=True,
         )
 
@@ -386,7 +331,6 @@ def run_benchmark(server_args: ServerArgs, bench_args: BenchArgs):
             result_filename="",
             tokenizer=tokenizer,
             api_type=bench_args.api_type,
-            bench_args=bench_args,
         )
         print("=" * 8 + " Warmup End   " + "=" * 8 + "\n")
 
@@ -411,7 +355,6 @@ def run_benchmark(server_args: ServerArgs, bench_args: BenchArgs):
                     result_filename=bench_args.result_filename,
                     tokenizer=tokenizer,
                     api_type=bench_args.api_type,
-                    bench_args=bench_args,
                 )
             )
 
@@ -437,7 +380,6 @@ def run_benchmark(server_args: ServerArgs, bench_args: BenchArgs):
                                 profile=bench_args.profile,
                                 profile_by_stage=bench_args.profile_by_stage,
                                 api_type=bench_args.api_type,
-                                bench_args=bench_args,
                             )[-1],
                         )
                     )
--- a/python/sgl_jax/bench_one_batch_server.py
+++ b/python/sgl_jax/bench_one_batch_server.py
@@ -21,7 +21,11 @@ import time
 
 import requests
 
-from sgl_jax.bench_serving import get_tokenizer, sample_random_requests
+from sgl_jax.bench_serving import (
+    get_tokenizer,
+    sample_random_requests,
+    sample_generated_shared_prefix_requests,
+)
 from sgl_jax.profiler import run_profile
 from sgl_jax.srt.entrypoints import http_server
 from sgl_jax.srt.server_args import ServerArgs
@@ -46,6 +50,18 @@ class BenchArgs:
     profile: bool = False
     profile_by_stage: bool = False
     api_type: str = "native"  # "native" or "openai"
+    dataset_name: str = "random"
+    gsp_num_groups: int = 1
+    gsp_prompts_per_group: int = 1
+    gsp_system_prompt_len: int = 1024
+    gsp_question_len: int = 100
+    gsp_output_len: int = 16
+    gsp_range_ratio: float = 1.0
+    seed: int = 1
+    gsp_send_routing_key: bool = False
+    gsp_num_turns: int = 1
+    gsp_fast_prepare: bool = False
+    gsp_ordered: bool = False
 
     @staticmethod
     def add_cli_args(parser: argparse.ArgumentParser):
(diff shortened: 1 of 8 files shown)


lingebeng (changes requested):
Thank you for your work!

The MTP indexer integration appears incomplete: the indexer’s top-k outputs are not propagated to subsequent draft steps for reuse through IndexShare.

Could you also add an accuracy benchmark, such as AIME 2026, comparing runs with and without MTP? Please report the accuracy scores for both configurations and the average acceptance length with MTP enabled.



image

zorrofox:
Thanks for taking this on — native MTP is the biggest missing piece for GLM decode throughput on TPU, and we are happy to benchmark this on v7x tp16 once the IndexShare propagation lands.

One design-stage heads-up from our v7x profiling, since it decides whether MTP shows a win at all on larger meshes: `model_runner` currently disables the AOT dispatch path whenever `speculative_algorithm` is set (the fallback at the top of the dispatcher setup), and this PR does not change that. On GLM-5.2 tp16 the stock pjit dispatch costs ~34 ms/step at ~2600 flat args (that overhead is why AOT dispatch exists; with it decode TPOT is ~13.7 ms). With MTP on, per-step time would be dominated by the Python dispatch, so even a healthy acceptance length is likely a net TPOT loss on multi-host meshes until spec and AOT dispatch can coexist. A second, smaller interaction: the jax 0.11.1 decode workaround from #1631 (`SGLANG_JAX_DECODE_DISABLE_SC_GATHER_OFFLOAD`) hooks the AOT compile path, so the fallback also re-exposes that regression on long-context configs.

Neither needs to block this PR, but it may be worth a note in the PR description that the initial target is single-host / small-mesh serving, and a follow-up issue for spec x AOT-dispatch compatibility so the v7x numbers have a path to positive.

amanseervi:
Hi, @zorrofox Thanks for the heads up! This is a valid concern to get good performance with spec decode enabled.
I wanted to see the dispatch overhead and so captured a [trace](https://xprof.corp.google.com/trace_viewer/amanseervi-8314063162583179822) for the speculative path  (TPU v7x TP32, bs=1, 4 draft tokens, 8k/1k workload)

Observations from the profile : 
In speculative decode, jit_fused_verify takes ~33.3 ms on TPU plus ~2.9 ms for jit_fused_draft_extend.
The TPU execution time (~33.3 ms) is longer than the Python host prep time (~25–30 ms). The CPU dispatch is completely hidden behind the TPU kernel execution, leaving no TPU idle bubbles in the trace.
So TPU stock pjit dispatch does not currently bottleneck the TP32 4-token speculative decode path (since TPU compute dominates). Also since it is not affecting the workload at bs=1 (where we expect it the most), it will definitely not affect the higher batch sizes as well.

Attaching few screenshots as well : 

The verify and draft model generation steps : 
2vefh8bhr0bs8

The CPU dispatches (overlapped with the previous prefill/decode step)
3ihb5989g8jio

zorrofox:
Thanks for the trace. I would read it differently: the host being hidden is a symptom of the device side being slower than it should be, not of dispatch being cheap. A non-spec decode step on v7x tp16 at bs=1 is ~12 ms of device time for one token; fused_verify processes 1+4 tokens with the same number of all-reduces and the same weight traffic, so it should land in the mid-teens, not 33 ms.

Op-level attribution of the median fused_verify window (33.1 ms, TensorCore 99.7% busy): all-reduce is 22.0 ms (66.5%) across 314 calls at ~70 us each, and the SparseCore plane shows 22.3 ms of activity in the same window, i.e. the small all-reduces are being offloaded to SparseCore. That is the jax 0.11.1 regression that #1631 handles on the non-spec path with per-executable compiler options. Two things in the current tree keep it alive on the spec path: `model_runner` still disables AOT dispatch whenever `speculative_algorithm` is set (the warning at the dispatcher setup), and none of `speculative/eagle_worker.py`, `dflash_worker.py`, `base_worker.py` pass compiler options into their compiles. Wiring the same per-exec options into the spec executables should bring verify from 33 ms to roughly 13-15 ms; at that point the host prep you measured becomes the binding constraint, which is why I would keep the AOT wiring on the critical path rather than behind it.

Why this is worth finishing: on tp16 the non-spec decode step is dominated by ~160 latency-bound all-reduces (about 37% of the step) and small-M weight reads, both of which are per-step costs that do not grow with the number of tokens verified. With verify at ~14 ms, draft at ~3 ms and an acceptance length around 4, per-token time lands near 4-5 ms against 12 ms today, a larger gain than any non-spec decode change we have on the table. So we would like to see this land and will run the v7x tp16 numbers once it is ready.

Two remaining items on the PR side: lingebeng's request to propagate the indexer top-k across draft steps (`index_share_for_mtp_iteration` in the GLM-5.2 config) plus an AIME-style accuracy check, and a rebase, since #1666 (merged 09-18) touched `eagle_draft_worker.py` and `flashattention_backend.py`.

zorrofox:
@amanseervi thanks again for starting this. Since the review from 09-11 (IndexShare propagation of the indexer top-k across draft steps, plus an accuracy comparison with and without MTP) is still open and the branch now conflicts with main, we have started implementing GLM-5.2 speculative decoding on current main in our tree so that the work does not stall. It follows the structure of this PR, adds the IndexShare propagation and the per-executable compiler options for the spec path discussed above, and we will run the accuracy comparison the reviewers asked for as part of it.

We would like to keep your authorship on the parts we carry over: the plan is to bring your commits forward with their author lines intact and add ours on top, then open a fresh PR against main and link it here. If you would rather keep driving this PR yourself, say so and we will contribute the pieces to your branch instead. Either way we will post the new PR link in this thread once it is up, and your review there would be very welcome.

amanseervi:
@zorrofox , Thanks for the taking on the work and adding the indexshare propagation and the per-executable compiler option on the spec path. Would be great if you could share the branch where these additions live.
Also, I'd like to continue driving this PR. Perhaps we can add the latest pieces to my branch.
Thank you!

zorrofox:
@amanseervi glad to hear you want to keep driving this, we will contribute onto your branch as offered. Summary of what we have and a proposal for how to merge it back.

**Where the changes live.** Our work is on top of your 37 commits rebased onto current `main` with your author lines intact, plus 19 commits of ours on top. It is pushed as `sgl-project/sglang-jax` branch `dev/glm52-mtp-on-1639` so you can pull, diff or cherry-pick. Everything is behind flags or auto-detected, nothing changes non-spec behaviour.

**What is in it, grouped.**
- Rebase fixes needed on today's `main` (explicit-sharding mesh, DSA sparse backend): `get_eagle_base_metadata` entry for the MLA/DSA backend, verify `positions`/`input_ids` placement, sparse-prefill self-write slots on spec batches, auto-sharded id rotation, and keeping the MTP layer's unquantized modules out of static fp8 quantization.
- Per-executable compiler options for the three decode-shaped spec executables (`fused_draft_extend`, `fused_eagle3_draft_extend`, `fused_verify`), same env/flags as the non-spec decode path.
- IndexShare: reuse the step-0 indexer selection across the GLM-5.2 draft steps (`index_share_for_mtp_iteration`, topk=1 only), with CPU tests for single/multi step, cross-request isolation and parity with the non-spec path. Acceptance length is bit-identical with and without it on our runs, as expected for prompts shorter than `index_topk`.
- Decode-form verify / draft-extend on the DSA backend (decode-shaped sparse attention plus a vectorised page-topk indexer instead of the extend-form prefill kernels), and a static Pallas dispatch in `paged_write_back` when the run table cannot overflow (removes a whole-pool copy per layer that only the spec path was paying).
- **Hidden-state relay across draft steps.** This is the one that matters for acceptance. In the fused draft loop each step j>0 was fed the target hidden states again while only the token window was rotated, so step j saw the pair (h_t, d_j) instead of (h^{(j-1)}, d_j). For GLM-5.2 with a single NextN block chained over `speculative_num_steps` this is the wrong input; the reference implementation feeds the previous step's output hidden state. With relay the per-position acceptance on gsm8k (200 questions, 5-shot, greedy, 3/1/4) went from 0.93 / 0.65 / 0.41 to 0.93 / 0.86 / 0.76 and the mean acceptance length (incl. bonus) from 2.87 to 3.33 on TPU v7x, which matches what we measure on H200 with sglang 0.5.17 and the same script (3.29 / 3.31, per-position 0.91 / 0.85 / 0.79). Enabled automatically when a single draft block is chained over more than one step, off for true multi-layer MTP models. CPU tests lock the three-step input sequence.
- A KVShare variant (draft steps j>0 read-only, no KV write) as opt-in. We measured it neutral on acceptance, so it stays off; mentioning it only so you do not need to redo that experiment.

Preliminary numbers on v7x (single run each, gsm8k p1/p8 200 questions): accuracy 0.960 / 0.960, acceptance length 3.33 / 3.33, bs1 TPOT 2.2x over non-spec. High concurrency (cc64) is currently a net loss on our side and we are still attributing that, so please treat the low-concurrency numbers as the validated ones.

**Plan.** Your branch is behind `main` and marked as conflicting, and most of our commits only apply on top of the rebased tree, so rather than asking you to rebase twice we will open a PR against `amanseervi/sglang-jax:spec-decode-glm5-2` that contains the rebase onto current `main` plus our commits, grouped by topic. You review one PR and #1639 stays yours; your commits and authorship are unchanged. If you would rather rebase yourself first, say so and we will re-send the pieces as separate PRs per topic instead.

zorrofox:
@amanseervi sure. Everything below is what our runs used, on a v7x 2x2x2 slice (8 chips on 2 hosts; tp16 because each v7x chip exposes 2 TensorCore devices) with the GLM-5.2 FP8 checkpoint; `$MODEL_PATH` is the local checkpoint directory (the same path is passed as the draft model, the NextN weights are inside the checkpoint). Tree: `spec-decode-glm5-2` at a42fcc7b, installed with `pip install -e ".[tpu]"`, jax 0.11.1 as pinned.

**Environment (identical for the MTP and the non-MTP arm):**

```
export DSA_PREFILL_SPARSE=1 DSA_TSA_PREFILL=1 DSA_TSA_PREFILL_QB=16 DSA_IDX_TSHARD=1 DSA_PAGE_TOPK=128 \
       DSA_INDEXER_KERNEL_PREFILL=1 DSA_INDEX_TOPK=4096 DSA_INDEXER_KERNEL=0 \
       SGLANG_JAX_DECODE_KV_LADDER="4096,16384,65536" SGLANG_JAX_AOT_DISPATCH=auto \
       SGLANG_JAX_DECODE_DISABLE_SC_GATHER_OFFLOAD=1 SGL_JAX_MOE_SC_PERMUTE=0 SGLANG_ROTARY_ILV_FIX=0
# MTP-specific:
export DSA_SPEC_AS_DECODE=1          # decode-form verify / draft-extend on the DSA backend (the 17.9 ms verify number)
export SGLANG_JAX_MTP_INDEX_SHARE=1  # IndexShare across draft steps (neutral on acceptance)
# SGLANG_JAX_MTP_HIDDEN_RELAY / SGLANG_JAX_MTP_RELAY_POS: leave unset (auto-on for the single-block chain); set =0 to A/B
# SGLANG_JAX_MTP_KVSHARE: leave unset (off)
```

`DSA_INDEX_TOPK=4096` deliberately overrides the checkpoint's `index_topk` 2048 (our tuned value on this backend; gsm8k acceptance is insensitive to it since those prompts fit either budget); drop it to run the checkpoint default. `DSA_PAGE_TOPK=128` only affects the sparse prefill path.

**Server (we ran it as 4 processes, 2 per host with 2 chips each, `$RANK` = 0..3, `$MASTER` = host 0; 2 processes with 4 chips each and `--nnodes 2` gives the same numbers; for a 32-device TP32 setup use `--tp-size 32 --ep-size 32` with `--nnodes` = number of processes):**

```
  --model-path $MODEL_PATH --trust-remote-code --device tpu \
  --tp-size 16 --dp-size 1 --ep-size 16 --moe-backend epmoe \
  --page-size 128 --context-length 135168 --chunked-prefill-size 8192 --max-prefill-tokens 8192 \
  --dtype bfloat16 --mem-fraction-static 0.80 --max-running-requests 64 --random-seed 3 \
  --attention-backend dsa_sparse --disable-radix-cache --disable-precompile --skip-server-warmup \
  --decode-log-interval 1 --log-level info --watchdog-timeout 3600 \
  --nnodes 4 --node-rank $RANK --dist-init-addr $MASTER:5000 --host 0.0.0.0 --port 30000 \
  --speculative-algorithm NEXTN --speculative-num-steps 3 --speculative-eagle-topk 1 \
  --speculative-num-draft-tokens 4 --speculative-draft-model-path $MODEL_PATH
```

The non-MTP control is the same command without the five `--speculative-*` flags. `--decode-log-interval 1` is only there so the server log has one `accept-len` value per verify step (that is where the per-position acceptance histogram comes from); use 40 for throughput runs.

**Benchmarks (on host 0 against `127.0.0.1:30000`; a 2-prompt concurrency-1 warmup first so the first measured run does not pay the cold compile):**

```
B="python3 -m sgl_jax.bench_serving --backend sgl-jax --host 127.0.0.1 --port 30000 --dataset-name random --random-range-ratio 1 --seed 42"
$B --random-input-len 1024 --random-output-len 256  --num-prompts 3   --max-concurrency 1    # bs1 TPOT (Mean/Median TPOT, run twice)
$B --random-input-len 1024 --random-output-len 1024 --num-prompts 16  --max-concurrency 8    # cc8
$B --random-input-len 1024 --random-output-len 1024 --num-prompts 128 --max-concurrency 64   # cc64 (run twice; the first run includes compiles)
$B --random-input-len 1200 --random-output-len 512  --num-prompts 16  --max-concurrency 8    # cc8 1200/512
```

**Acceptance length / accuracy:** `benchmark/gsm8k/bench_sglang_jax.py --base-url http://127.0.0.1:30000 --num-questions 200 --num-shots 5 --max-new-tokens 512 --temperature 0 --parallel 1` (p1) and `--parallel 8` (p8). Acceptance length = mean of the `accept-len` values in the server log over the run window, i.e. tokens emitted per verify per request including the bonus token; `GET /get_server_info` -> `internal_states[0].avg_spec_accept_length` gives the cumulative value.

**Step trace:** with one bs1 request streaming (a gsm8k prompt with `ignore_eos` and `max_new_tokens 512`), `POST /start_profile` with `{"output_dir": "...", "num_steps": 30, "host_tracer_level": 1, "python_tracer_level": 0}`, then read the `fused_verify` / `fused_draft_extend` module durations from the xplane.

If it helps I can push the phase scripts we use (warmup -> gsm8k p1 -> trace -> bs1 -> gsm8k p8 -> cc8 -> cc64) under `benchmark/` in a follow-up commit.

PR #1641, opened 2026-09-11
perf(rotary): compute GPT-J rotary in the interleaved domain and merge DSA indexer RoPE via concat (110k TTFT -5.0%, TPOT -3.9%)

zorrofox:
## Motivation

XProf attribution of a 110k-token GLM-5.2 prefill on TPU v7x shows ~4% of device time spent in rotary-embedding data movement. The hotspot is not the attention-path rotary (its per-device slice is only [T,4,64]) but the DSA indexer projection: `index_n_heads=32` gives q_rope [8192,32,64] (33.5 MB per layer-chunk), the indexer is not TP-sharded so every device computes the full block, and the result is written back with `query.at[:,:,:64].set(...)` which read-modify-writes the whole [8192,32,128] tensor.

The GPT-J (`interleave=True`) branch of `apply_rotary_emb` costs on three fronts: strided even/odd slices (`x[..., ::2]`), a `stack+reshape` re-interleave of the output, and the caller-side `at[].set` merge.

## Modifications

Two mechanical rewrites, both bit-identical to the current formulation (verified with `jnp.all(==)` on TPU and CPU):

1. `apply_rotary_emb`'s GPT-J branch computes in the interleaved domain: `o = x*repeat(cos,2) + swap_pairs(x)*(repeat(sin,2)*[-1,+1,...])`, where `swap_pairs` is a pair-local flip. No strided slice, no re-interleave; the output keeps the input layout.
2. `GlmDsaIndexer` (both `project` and `__call__`) merges the rotated 64 dims back via `concatenate((q_rope, query[:,:,64:]))` instead of `at[:,:,:64].set(q_rope)`.

Default on, `SGLANG_ROTARY_ILV_FIX=0` restores the previous formulation (same escape-hatch pattern as the QB256 default in #1585; this is a bit-exact rewrite, not a behavior change). The neox-style path is untouched.

## Microbenchmark

Indexer-shaped context on v7x (wq_b matmul -> slice -> rotary -> merge -> hadamard einsum, bf16, best of 5, two runs <0.5% apart):

| variant | us/step/layer-chunk |
|---|---|
| no-rotary floor | 262 |
| gptj + at.set (current) | 1476 |
| gptj + concat | 889 |
| interleaved + concat (this PR) | 575 (-61%) |

## End-to-end (GLM-5.2 FP8, v7x tp16, ctx 135168 chunk 8192, paired in-place env flip on the same server, 3 rounds each, zero variance)

| metric | before | after | delta |
|---|---|---|---|
| 110k TTFT | 22.27 s | 21.16 s | -5.0% |
| 16k TTFT | 2.41 s | 2.25 s | -6.8% |
| decode TPOT @1k | 13.70 ms | 13.17 ms | -3.9% |

The decode win comes from the same per-layer fixed overhead on the [1,32,64] decode-step rope.

## Correctness

The interleaved-domain formula produces the same multiplies and adds per element as the slice formulation (the sign is folded into the sin operand), so outputs are bit-identical, not merely close. `test_rotary_ilv_bitexact.py` asserts exact equality for the GPT-J and neox paths on both flag values, over bf16/fp32 and several shapes.

## Test plan

- [x] CPU/TPU bit-exact on both flag values (embeddings + indexer paths), new `test_rotary_ilv_bitexact.py` 7/7
- [x] paired e2e 110k/16k/TPOT, 3 rounds each, off-arm reproduces the prior anchor first
- [x] GSM8K 200q paired: fix 0.975 vs baseline 0.955, Invalid 0 on both (5-shot quick-test scale, temperature 0)
- [x] cc=1/8/32 compile face + TPOT paired: 13.12/35.53/98.45 ms vs 13.70/35.83/100.00 ms, throughput monotonic in concurrency
- [x] pre-commit clean

### DIFF
--- a/python/sgl_jax/srt/layers/embeddings.py
+++ b/python/sgl_jax/srt/layers/embeddings.py
@@ -568,7 +568,7 @@ def rotary_embedding_forward(
     return query, key
 
 
-_ILV_FIX = os.environ.get("SGLANG_ROTARY_ILV_FIX", "0") == "1"
+_ILV_FIX = os.environ.get("SGLANG_ROTARY_ILV_FIX", "1") == "1"
 
 
 # @partial(jax.jit, static_argnames=["is_neox_style"])
--- a/python/sgl_jax/srt/layers/embeddings.py
+++ b/python/sgl_jax/srt/layers/embeddings.py
@@ -15,6 +15,7 @@
 """Embedding Layers."""
 
 import math
+import os
 from typing import Any
 
 import jax
@@ -567,6 +568,9 @@ def rotary_embedding_forward(
     return query, key
 
 
+_ILV_FIX = os.environ.get("SGLANG_ROTARY_ILV_FIX", "0") == "1"
+
+
 # @partial(jax.jit, static_argnames=["is_neox_style"])
 def apply_rotary_emb(
     x: jax.Array,
@@ -587,6 +591,15 @@ def apply_rotary_emb(
     if is_neox_style:
         x1, x2 = jnp.split(x, 2, axis=-1)
     else:
+        if _ILV_FIX:
+            # GPT-J rotary computed directly in the interleaved domain:
+            # avoids the strided even/odd slices and the stack+reshape
+            # re-interleave; bit-identical to the slice formulation.
+            cos_il = jnp.repeat(cos, 2, axis=-1)
+            sin_il = jnp.repeat(sin, 2, axis=-1)
+            sign = jnp.tile(jnp.array([-1, 1], dtype=x.dtype), x.shape[-1] // 2)
+            x_swap = jnp.flip(x.reshape(*x.shape[:-1], -1, 2), axis=-1).reshape(x.shape)
+            return x * cos_il + x_swap * (sin_il * sign)
         x1 = x[..., ::2]
         x2 = x[..., 1::2]
     o1 = x1 * cos - x2 * sin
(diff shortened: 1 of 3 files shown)


cjx0709 (commented):
It's a good optimize. But now seems here are with two optimization controled by one environment, maybe we can decouple them. And for new environment, I think we can follow like: "https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/environ.py"(at the first version, maybe we can do it simply) instead of scattering across different files

zorrofox:
Done in b3336919, both points:

1. Decoupled: the two rewrites now have independent escape hatches — `SGLANG_JAX_ROTARY_INTERLEAVED` (interleaved-domain GPT-J rotary in `apply_rotary_emb`) and `SGLANG_JAX_INDEXER_ROPE_CONCAT` (concat merge in the DSA indexer). Both default on, `=0` restores the previous formulation independently.

2. Central registry: added a minimal `sgl_jax/srt/environ.py` following the sglang `environ.py` pattern (EnvField/EnvBool descriptors + an `Envs` class + `envs` singleton) — intentionally the simple first version you suggested. The two new variables are declared there; existing scattered env vars can migrate over incrementally.

Tests updated: the bit-exact suite now targets `SGLANG_JAX_ROTARY_INTERLEAVED`, plus a registry sanity test covering defaults and `=0` parsing for both fields (8/8 on CPU, pre-commit clean).

cjx0709 (approved):

PR #1642, opened 2026-09-11
[dsa] SparseCore radix-select for the streamindex_topk exit stage (exact; -43% decode / -66% prefill at DeepSeek-V4 scale)

zorrofox:
## Motivation

`streamindex_topk` ends with `jax.lax.approx_max_k(scores, k, recall_target=1.0)`. On TPU XLA short-circuits `recall_target=1.0` to `log2_reduction=0`, i.e. a full sort of every row (no PartialReduce). At DeepSeek-V4 scale (`compression_ratio=4`, 1M context → E = 262144 compressed entries per query, `k=512`) that exit stage is the dominant cost of the indexer kernel, not the Pallas scoring loop:

| shape (TPU7x, k=512, E=262144) | total | Pallas scores | approx_max_k sort | selection share |
|---|---|---|---|---|
| decode B=64 | 7.50 ms | 3.3 ms | 3.6 ms (+1.7 ms copies) | 56% |
| prefill T=2048 | 123.9 ms | 18.9 ms | 92.4 ms | 84% |

Lowering `recall_target` is not an option here: the indexer contract is exact top-k (see #1515), and the ApproxTopK partial reduce keeps only one candidate per bin, so it cannot be made exact in two stages.

## What this PR does

Vendors the openxla/tokamax SparseCore MSB radix-select top-k kernel (Apache-2.0, `tokamax/_src/ops/experimental/tpu/topk/pallas_mosaic_tpu_kernel.py` @ `6303c141386c`) as `sgl_jax/srt/kernels/dsa/sc_topk.py` and routes large rows through it. The kernel is exact and stable (deterministic post-fill by index), runs on the SparseCore so it does not compete with the TensorCore, and needs no changes to the Pallas scoring kernel.

- New `select_topk_indices(scores, k, backend="auto"|"sc"|"xla")` in `streamindex_topk.py`; `streamindex_topk` gains a static `topk_backend="auto"` kwarg. Output contract unchanged: exact top-k indices, descending by score, `-1` (from `-inf`) packed at the tail, shape `[T, k]`.
- `backend="xla"` is byte-for-byte the previous `approx_max_k(recall_target=1.0)` path.
- Routing policy (`should_use_sc_topk`): SparseCore when the chip reports one (`pltpu.get_tpu_info().generation >= 6` and `sparse_core` present), `E >= DSA_SC_TOPK_MIN_ENTRIES` (default 8192; very short rows are faster on XLA because of the fixed SparseCore launch cost) and the per-subcore key+value slice fits half of the subcore VMEM (E ≤ 262144 on v6e, ≤ 524288 on v7x). Everything else stays on XLA. `DSA_SC_TOPK=0` is a process-wide kill switch.
- Rows are padded with `-inf` to the kernel's layout constraint (a whole number of `num_lanes`-wide vectors per subcore tile, `num_subcores` tiles above 4096 entries), derived from `SparseCoreInfo` at trace time, not hard-coded.
- Only the kernel's single-window path (`num_seq_windows=1`) is used. The multi-window path in the vendored revision returns wrong results (the global histogram is only synchronized on the last window and earlier windows are post-filled locally); the header in `sc_topk.py` documents this and callers cannot enable it through `select_topk_indices`.
- The vendored file is unchanged apart from the provenance header and `absl.logging` → stdlib `logging` (absl is not a sglang-jax dependency). `pyproject.toml` gets a per-file ruff ignore (`B023`, `SIM212`) for it, same as the existing Pallas kernels.

## Benchmarks

Kernel-level: `streamindex_topk` wall time (median of 10, `jax.block_until_ready`), `k=512`, `H=64`, `D=128`, bf16 cache, `num_kv_pages_per_block=64`, `num_queries_per_block=(1,64,64)`; `xla` = previous path, `sc` = this PR with the kernel forced, `auto` = this PR default. Every `sc`/`auto` row matched the `xla` index set exactly on all rows.

TPU7x (jax 0.11.1):

| case | ctx | E | xla (ms) | sc (ms) | auto (ms) | Δ auto vs xla |
|---|---|---|---|---|---|---|
| decode B=64 | 1M | 262144 | 7.49 | 4.27 | 4.29 | -43% |
| prefill T=2048 | 1M | 262144 | 124.0 | 42.4 | 42.8 | -66% |
| decode B=64 | 100K | 25000 | 4.61 | 1.40 | 1.41 | -69% |
| decode B=8 | 1K | 256 (< k) | 0.66 | 0.63 | 0.56 | within noise (auto keeps XLA) |

TPU v6e (jax 0.11.1):

| case | ctx | E | xla (ms) | sc (ms) | auto (ms) | Δ auto vs xla |
|---|---|---|---|---|---|---|
| decode B=64 | 1M | 262144 | 7.94 | 4.95 | 4.93 | -38% |
| prefill T=2048 | 1M | 262144 | 171.3 | 64.5 | 64.5 | -62% |
| decode B=64 | 100K | 25000 | 5.16 | 2.20 | 2.21 | -57% |
| decode B=8 | 1K | 256 (< k) | 0.74 | 0.96 | 0.96 | +30% forced; auto keeps XLA (0.74) |

Selection stage only (same tensors, `approx_max_k` sort vs SparseCore kernel), TPU7x: decode `[64, 262144]` 4.12 → 0.86 ms; prefill `[2048, 262144]` 104 → 19.8 ms.

Kernel-level only, GLM-5.2 decode shapes (`index_topk=2048`, `compression_ratio=1`, 128K context → E=131072, `H=32`, `num_queries_per_block=1`): wall time of one `streamindex_topk` call, same protocol as above. Not a model-level TTFT/TPOT measurement. GLM-5.2 has 21 such full-indexer layers per decode step, hence the last column:

| chip | B | xla (ms) | auto (ms) | Δ | ×21 layers per step |
|---|---|---|---|---|---|
| TPU7x | 1 | 1.31 | 0.23 | -82% | 27.6 → 4.9 ms |
| TPU7x | 16 | 1.87 | 0.78 | -58% | 39.2 → 16.4 ms |
| TPU7x | 64 | 6.94 | 2.54 | -63% | 145.8 → 53.3 ms |
| v6e | 1 | 1.51 | 0.31 | -79% | 31.8 → 6.6 ms |
| v6e | 16 | 1.99 | 0.97 | -51% | 41.9 → 20.4 ms |
| v6e | 64 | 7.60 | 3.00 | -61% | 159.6 → 63.0 ms |

## Correctness

`test/srt/kernels/dsa/test_sc_topk_selector.py` (TPU only, skips the SparseCore cases on chips without one):

- `select_topk_indices(backend="sc")` vs `jax.lax.top_k` on random rows with a block of `-inf` columns, for `[64, 262144]`, `[3, 25000]` (not a multiple of 256, exercises padding) and `[1, 8192]`: same index set, descending, `-1` only at the tail, no `-inf` column selected.
- E < k (300 valid of 8192), E == k, all `-inf` rows, 600 tied maxima.
- Routing policy as a pure function with mocked v6e geometry, the VMEM boundary at exactly 262144 entries, off-TPU behaviour, `backend="sc"` without a SparseCore raises a clear `ValueError`, and `backend="auto"` observed to call the XLA selector for 256-entry rows and the SparseCore selector for 262144-entry rows.
- End to end `streamindex_topk` with `topk_backend` in `xla`/`sc`/`auto` for decode 1M ctx, prefill 1M ctx, decode 100K ctx and decode 1K ctx (256 entries < k): identical index sets and trailing `-1` contract.

## Test plan

- [x] `pre-commit run --all-files` clean (isort, ruff, black, codespell, mypy)
- [x] v6e-8: `test_sc_topk_selector.py` passed (14 passed)
- [x] v6e-8: `pytest test/srt/kernels/dsa/` passed (48 passed, 4 skipped)
- [x] TPU7x: `test_sc_topk_selector.py` passed (14 passed)
- [x] TPU7x: `pytest test/srt/kernels/dsa/` 48 passed, 4 skipped
- [x] Benchmarks above reproduced from the merged branch (fresh process per backend, 10 timed iterations after warm-up)

## Follow-ups (not in this PR)

- `DSA_SC_TOPK_MIN_ENTRIES` is set from two data points (256 and 25000 entries); a sweep between 4096 and 16384 would place it more precisely.
- `select_topk_indices` is the single swap point if a faster SparseCore top-k becomes available upstream (tokamax or jax); nothing else in the indexer depends on the kernel.
- For prefill at very long context a block-max coarse filter before selection would cut the HBM traffic of the score matrix; orthogonal to this change.

Related: #1607 (completed-groups visibility mask in the same kernel; independent, both apply cleanly to main).

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -149,6 +149,9 @@ ignore = [
 "sgl_jax/srt/kernels/fused_moe/v1/kernel.py" = ["B023"]
 "sgl_jax/srt/kernels/fused_moe/v2/kernel.py" = ["B023", "F841", "SIM102", "SIM108"]
 "sgl_jax/srt/kernels/mla/v2/kernel.py" = ["B023"]
+# Vendored openxla/tokamax SparseCore top-k kernel: kept byte-identical to upstream
+# apart from the logging import; same statically-unrolled-loop closures as above.
+"sgl_jax/srt/kernels/dsa/sc_topk.py" = ["B023", "SIM212"]
 # TPU benchmark/debug scripts intentionally initialize distributed JAX before
 # importing local kernels, and benchmark closures capture loop-state values.
 "sgl_jax/srt/kernels/fused_moe/v2/bench_compare.py" = ["B023", "E402"]
--- a/python/sgl_jax/srt/kernels/dsa/sc_topk.py
+++ b/python/sgl_jax/srt/kernels/dsa/sc_topk.py
@@ -1,0 +1,894 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+# Vendored from openxla/tokamax @ 6303c141386c (2026-09-10)
+#   tokamax/_src/ops/experimental/tpu/topk/pallas_mosaic_tpu_kernel.py
+# Local changes (kept minimal so upstream diffs stay reviewable):
+#   * ``from absl import logging`` -> stdlib ``logging`` (absl is not a sglang-jax dependency).
+# Kernel semantics are unchanged. Only ``num_seq_windows=1`` is exercised by sglang-jax:
+# the multi-window (``num_seq_windows>1``) cooperative path produces wrong results in the
+# vendored revision (global histogram is only synchronized on the last window and
+# ``local_postfill`` pads earlier windows); callers must not enable it.
+
+"""Top-K implementation supporting multiple Vector Subcores.
+
+High-level algorithm:
+The file implements a Top-K selection algorithm targeting Vector Subcores.
+
+Large K: Distributed MSB Radix Select (`topk_multitile`)
+   - Uses a radix-based selection algorithm, proceeding digit-by-digit from the
+     most significant bit (MSB) down to the least significant bit (LSB).
+   - Multiple vector subcores may collaborate. In this case, they will work
+     independently on all windows except for the final window, during which
+     they will collaborate to build a global histogram, identify a global
+     threshold, postfill remaining (tied) K elements from first subcore to last,
+     and DMA each subcore's results to the correct subset of locations in HBM.
+   - If subcores are not collaborating, each operates independently on
+     a strided slice of the batch elements. This avoids synchronization and
+     communication, so can be performance beneficial for large enough batches.
+   - The algorithm processes data in sequential windows and can cooperate across
+     multiple subcores.
+     a) Histogram: Computes a histogram of the current digit's values across all
+        active candidates.
+     b) Global Sync (Optional: only when cooperating across subcores): Combines
+        local histograms into a global histogram.
+     c) Threshold Discovery: Accumulates the histogram from the maximum digit
+        down to zero to find a `target_digit`. The `target_digit` is chosen such
+        that elements with digits > `target_digit` are guaranteed to be in the
+        top K.
+     d) Filtering:
+        - Elements with digit > `target_digit` are added to the final Top-K
+          output.
+        - Remaining elements with digit == `target_digit` survive (if needed, to
+          meet K) as candidates for the next digit iteration, or for postfilling
+          of ties.
+        - Elements with digit < `target_digit` are eliminated.
+   - Postfill: If K elements haven't been found after all digits are processed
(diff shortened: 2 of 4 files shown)


zorrofox:
@cjx0709 could you review this one? It is a change to the GLM-5.2 DSA indexer kernel from #1515 (`kernels/dsa/streamindex_topk.py`) rather than a DeepSeek-V4 feature: the exit-stage `approx_max_k(recall_target=1.0)` lowers to a full row sort on TPU, and this PR routes large rows through an exact SparseCore radix-select instead. Output contract, tests and the XLA path are unchanged; `DSA_SC_TOPK=0` restores the old behaviour.

On GLM-5.2 decode shapes (`index_topk=2048`, `compression_ratio=1`, 128K context, 32 heads) one `streamindex_topk` call on TPU7x goes from 1.31 / 1.87 / 6.94 ms to 0.23 / 0.78 / 2.54 ms at B=1 / 16 / 64, with identical index sets; v6e is 1.51 / 1.99 / 7.60 → 0.31 / 0.97 / 3.00 ms. This matters for the decode configurations we are evaluating that send batched decode through the token-level indexer path instead of `page_topk`. It is independent of #1607 and composes with #1641 (different files).

CI is green (PR Test 25/25). Full benchmark tables and the exactness tests are in the description.

JamesBrianD (approved):

PR #1644, opened 2026-09-11
perf(grouped-topk): f32 tie-break reduction and single-use winner masks

JianmingTONG:
## Motivation

In the grouped top-k final-select loop the tie-break "lowest expert id achieving the max" is an int32 minimum reduction. TPU7x has no integer vector min, so Mosaic lowers each `vector.multi_reduction<minsi>` to a compare plus a select on every vreg of the [E, BT] block, and the shared winner one-hot mask is materialised, spilled to VMEM and re-expanded at each consumer. On the profiled 2,048-token router this kernel is 66.5 µs of the 71.1 µs module.

## Modifications

- Reduce the expert iota in f32 instead of int32 for the tie-break (exact for E <= 2**24): Mosaic emits one `vmin` per vreg instead of compare+select.
- Give each consumer of the winner one-hot (the weight gather and the drop-the-winner update) its own single-use compare on the same winner id, so the mask is not materialised and spilled.
- Outputs are bit-identical to the baseline kernel, including NaN/-inf scores and exact ties. No interface, shape or tiling change. One file: `python/sgl_jax/srt/kernels/grouped_topk/v1/kernel.py`.

This is independent of #1634 (reversed FP32 priorities); it keeps the integer index domain and only changes the reduction dtype and mask materialisation.

## Accuracy Tests

```bash
JAX_PLATFORMS=cpu PYTHONPATH=python python -m pytest -q python/sgl_jax/test/kernels/grouped_topk_test.py
```
JAX 0.11.1: **21 passed, 13 skipped** (skips are TPU-only packed-path cases). On TPU7x the candidate's outputs matched the retained Phase 1 reference outputs exactly in all three hardware repetitions.

## Benchmarking and Profiling

Candidate-only runs on TPU7x (one host, 10 warmups, 30 measured dispatches, separate profiled dispatch), compared with the saved Phase 1 profile of the same workload (`biased_grouped_topk`, 2,048 tokens, 896 experts, top-16, shape [2048, 896, 16]):

| 2,048-token TPU7x router | Device span (µs) |
| --- | ---: |
| Saved baseline | 71.133 |
| Candidate repetition 1 | 56.291 |
| Candidate repetition 2 | 56.373 |
| Candidate repetition 3 | 56.656 |

**−20.7 %** device span. Compiled LLO: 119,797 → 92,133 bundles; vector-ALU instructions −103,289, vector load/store −72,797 (static counts from the actual compiler output). Host medians moved from 271.9 µs to 254.0 / 294.2 / 247.7 µs, so this is a device-span claim, not an end-to-end serving claim.

---
Produced by the HierEvo v2 Phase 2 loop (Lowering–Exposure Graph localization + Claude Opus 5 oracle); candidate-only measurements on TPU7x via GKE. Receipts, LLO diffs and post-modification traces are retained in the HierEvo campaign record (`docs/leg_campaign_20260910.md`).

https://claude.ai/code/session_015GXDwgzoD3AKyC5HGBUS1U

### DIFF
--- a/python/sgl_jax/srt/kernels/grouped_topk/v1/kernel.py
+++ b/python/sgl_jax/srt/kernels/grouped_topk/v1/kernel.py
@@ -19,7 +19,9 @@ Algorithm (matches `_biased_grouped_topk` exactly, ties included):
 Renormalize / routed_scaling_factor are applied by the caller (`TopK.__call__`).
 
 Tie-break: selection uses `max` + masked `min(iota)` (smallest index achieving the max) rather than
-`argmax`, because TPU Mosaic's reduction argmax does not break ties toward the lowest index.
+`argmax`, because TPU Mosaic's reduction argmax does not break ties toward the lowest index. The
+iota is carried in f32 (exact for E <= 2**24): the vector core has float min/max but no integer
+min/max, so an int32 masked-min costs a compare+select on every vreg of the [E,BT] block.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


Yanko-7:
LGTM

Yanko-7 (approved):

PR #1650, opened 2026-09-12
feat(moe): opt-in SparseCore permute/unpermute kernels for EPMoE (110k TTFT -5.4%)

zorrofox:
## Summary

Vendors tpu-inference's SparseCore `ragged_gather_v2` / `ragged_gather_reduce_v2` kernels (Apache-2.0) under `sgl_jax/srt/kernels/sparse_core/` and wires them into `EPMoE` behind an opt-in:

- `SGL_JAX_MOE_SC_PERMUTE=1` (or `EPMoE(use_sc_permute=True)`), default **off**.
- dispatch: `inputs_2d[token_indices]` -> `ragged_gather_v2` gathering only this expert shard's sorted-row range.
- combine: the SC path replaces the unpermute gather + weighted reduce with `ragged_gather_reduce` (fused gather + top-k weighted reduce, fp32 accumulate, non-local rows masked), inserted ahead of the stock fp32 paths from #1578/#1631 and following the 2-D routing-weights contract from #1577.
- Trace-time gate (`should_use_sparse_core`): below tpu-inference's VMEM heuristic (decode-sized batches) the layer lowers to a module byte-identical to the flag-off build (test enforced). No SparseCore -> plain XLA path.

## Why

On v7x, XLA already offloads the EPMoE gathers to SparseCore (`gather_offload_async_*`), but it materializes every `tokens * top_k` row for all experts (805 MB per layer-chunk at chunk 8192 / hidden 6144) and the TensorCore blocks on the async-done; the combine einsum stays on the TensorCore. In a GLM-5.2 110k prefill trace the permute/unpermute glue is ~2.1 s per device (7% of device time). The ragged kernels only move the local expert rows.

## Microbench (v7x single chip, bf16, hidden 6144, top_k 8, 16/256 local experts)

| T tokens | dispatch XLA | dispatch SC | combine XLA | combine SC |
|---|---|---|---|---|
| 8192 | 0.87 ms | 0.13 ms | 1.66 ms | 0.37 ms |
| 2048 | 0.15 ms | 0.06 ms | 0.44 ms | 0.13 ms |
| 1024 | 0.18 ms | 0.06 ms | 0.23 ms | 0.10 ms |
| 64 | fallback (identical HLO) | | | |

## E2E (GLM-5.2 FP8, v7x tp16/ep16, dsa_sparse stack, jax 0.11.1)

Current main (64dd38e5 image, ctx 135168 chunk 8192, same server pair, in-place flag toggle, 3 probe rounds x 2 independent runs, zero variance):

| metric | flag off | flag on | delta |
|---|---|---|---|
| TTFT 110k | 22.27 s (reproduces the pure-main anchor) | 21.07 s | -5.4% |
| TTFT 16k | 2.42 s | 2.25 s | -7.0% |
| decode TPOT cc=1/8/32 (1024/256) | 13.70 / 35.83 / 100.0 ms (same-image pure-main reference) | 13.74 / 36.01 / 99.06 ms | neutral, throughput monotonic in concurrency |
| gsm8k 200q 5-shot temp0 (quick regression, not an accuracy claim) | 0.945-0.955 (same-image pure-main runs) | 0.965, Invalid 0 | in band |

Earlier same-pod paired suite on the pre-#1631 base (archived for reference): TTFT 110k -5.9%, 16k -7.5%, output tok/s cc=1/8/32 +2%/+4%/+4.6%, gsm8k on/off 0.965/0.960 with 191/200 predictions identical.

Numerics: whole-layer parity SC vs XLA cos > 0.9999; a float64 emulation puts the SC path closer to exact than the XLA einsum (bf16-pass f32 dot).

## Test plan

- [x] `test/kernels/sc_moe_permute_test.py` on v7x (15 tests: kernel parity incl. masking + f32/bf16 weights, EPMoE parity, small-batch HLO identity, default off)
- [x] flag-off regression: `test_moe_weight_loader_sharding.py`, `test_moe_topk.py`
- [x] e2e TTFT 16k/110k A/B on current main, off-arm reproduces the pure-main anchor first
- [x] cc=1/8/32 sanity (throughput monotonic, no point below single-stream)
- [x] gsm8k 200q in band
- [x] pre-commit clean

## Notes / limits

- Performance claims are scoped to v7x tp16; on v6e only the unit tests run (CI). tpu-inference disabled one of these kernels on v6e for accuracy (#3036), so keep the opt-in off on v6e until tested there.
- `core_map_helper` uses `jax._src` internals (same as upstream tpu-inference); validated on jax 0.11.1.
- No tuned params for (65536, 6144, 8, bf16): default heuristic; tuning table is a follow-up.

### DIFF
--- a/python/sgl_jax/srt/kernels/sparse_core/__init__.py
+++ b/python/sgl_jax/srt/kernels/sparse_core/__init__.py
@@ -1,0 +1,1 @@
+"""SparseCore Pallas kernels (vendored from vllm-project/tpu-inference) and MoE permute wrappers."""
--- a/python/sgl_jax/srt/kernels/sparse_core/_logging.py
+++ b/python/sgl_jax/srt/kernels/sparse_core/_logging.py
@@ -1,0 +1,20 @@
+"""Minimal ``init_logger`` shim so the vendored tpu-inference kernels keep their logging calls."""
+
+import logging
+
+
+class _OnceLogger(logging.LoggerAdapter):
+    def __init__(self, logger):
+        super().__init__(logger, {})
+        self._seen: set[str] = set()
+
+    def warning_once(self, msg, *args, **kwargs):
+        key = msg % args if args else str(msg)
+        if key in self._seen:
+            return
+        self._seen.add(key)
+        self.warning(msg, *args, **kwargs)
+
+
+def init_logger(name: str) -> _OnceLogger:
+    return _OnceLogger(logging.getLogger(name))
--- a/python/sgl_jax/srt/kernels/sparse_core/core_map_helper.py
+++ b/python/sgl_jax/srt/kernels/sparse_core/core_map_helper.py
@@ -1,0 +1,77 @@
+# Adapted from https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/sparse_core/core_map_helper.py
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Run single-mesh SparseCore kernels through ``core_map``.
+
+``pl.kernel`` lowers through ``mpmd_map``, which is slower than the ``core_map``
+path for these single-mesh kernels on TPU. ``kernel`` here is a drop-in
+replacement for ``pl.kernel`` that uses ``core_map`` directly, so the SparseCore
+kernels keep that lowering without depending on the jax-side ``pl.kernel``
+implementation.
+"""
+
+from jax._src import api
+from jax._src import core as jax_core
+from jax._src import lax, tree_util
+from jax._src.pallas import core as pl_core
+
+
+def _empty_out_ref(out_type):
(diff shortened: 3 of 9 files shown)


JamesBrianD:
Could you also bring in the partitioning fix from tpu-inference https://github.com/vllm-project/tpu-inference/pull/3513?

At the current head (`3e94d0e3e6e`), `_calculate_num_column_partitions` still stops splitting when the preferred four pipeline stages cannot be maintained, before satisfying the hard `num_row_partitions <= num_simd_lanes` constraint. Upstream #3513 adds a second loop that continues splitting while alignment and exact divisibility permit it.

Using this PR's configuration functions with JAX 0.11.1 hardware metadata, both of these BF16 cases pass the SC gate but produce an invalid partition count:
- v7x: hidden=512, 32768 routed rows → 32 row partitions, limit 16.
- v6e: hidden=1024, 32768 routed rows → 16 row partitions, limit 8. This is also the shape used by the new whole-layer parity test (8192 tokens × top-k 4).

These are host-side configuration checks, not TPU execution results. Please include the upstream fix and regression coverage for these cases in this PR.

zorrofox:
Done in cd66f8cf: ported the tpu-inference #3513 fix — the preferred-pipeline-depth loop now extracts its divisibility check into `_can_split_further`, and a second loop keeps doubling the column partitions until `num_row_partitions <= num_simd_lanes` holds, even when that drops below the preferred four pipeline stages.

Both of your cases are covered by a new host-side regression class (`ColumnPartitionHardLimitTest`, runs on CPU): (hidden=512, 32 cores, simd 16) used to yield 32 row partitions and now yields 16 (2 column partitions), and (hidden=1024, 32 cores, simd 8) used to yield 16 and now yields 8 (4 column partitions). A second test sweeps a hidden x cores x simd x input grid and asserts the hard limit is met whenever divisibility permits reaching it.

JamesBrianD (approved):

PR #1651, opened 2026-09-13
feat(eval): add chat sglang_mmlu evaluation

pcmoritz:
## Motivation

I was running evaluations of the qwen 3.5 9b model and noticed that with the current `sglang_mmlu` evaluation it does not do very well and actually has worse performance than the qwen 3 8b model. When I switched the evaluation code to use chat completion format, it was doing a lot better and was more in line with what would be expected.

## Modifications

Add a new `sglang_mmlu_chat` evaluation that uses chat completion (with thinking turned off) instead of raw text.

## Accuracy Tests

```
   Model         Mode         Accuracy
  ━━━━━━━━━━━━  ━━━━━━  ━━━━━━━━━━━━━━━
   Qwen3-8B      Raw            74.86%
  ────────────  ──────  ───────────────
   Qwen3-8B      Chat           73.21%
  ────────────  ──────  ───────────────
   Qwen3.5-9B    Raw            69.84%
  ────────────  ──────  ───────────────
   Qwen3.5-9B    Chat           82.75%
```

## Checklist


Junyi-99 (changes requested):
I suggest adding a separate `sglang_mmlu_chat.py` instead of coupling chat mode into `sglang_mmlu.py`.

They are different evaluation protocols, and keeping them separate makes the behavior clearer and avoids changing the semantics of the existing `sglang_mmlu`.

pcmoritz:
Thanks a lot for your feedback, I updated the PR now, let me know if you have any further comments or feedback!

cjx0709 (approved):
LGTM

Junyi-99 (approved):
LGTM

pcmoritz:
@cjx0709 Can you merge the PR?

Issue #1500, opened 2026-07-31
perf: is device_put / TPU-launch overhead actually a scheduler bottleneck? (host-vs-forward masking analysis)


Yajingleo:
## Context

Following up on #293 (is the scheduler thread the throughput bottleneck?) and the discussion on #1484. On current `main`, the scheduler's per-step host work is **masked** under the TPU forward. In #1484 @JamesBrianD noted the current bottlenecks are instead:
1. `device_put` when converting `ScheduleBatch` → `ForwardBatch`, and
2. TPU JIT launch overhead.

This issue tracks a focused question: **in which regime do `device_put` / launch overhead actually bite** — i.e. cause measurable TPU idle / lost throughput — since where they're *masked* they cost ~0.

## What `device_put` is

`ForwardBatch.init_new` (`forward_batch_info.py`) H2D-transfers ~8 per-step batch-metadata arrays (`input_ids`, `seq_lens`, `cache_loc`, `positions`, …) via `device_array` → `jax.make_array_from_callback`, sharded on the `data` (DP) axis. It runs every forward step and allocates fresh device buffers each step (no reuse). Optimization lever (if it un-masks): preallocate persistent bucketed-shape device buffers + in-place update.

## Findings — nnodes=1 (single host)

Setup: Qwen3-8B, v6e-4, conc 16, 1024/1024, JAX profiler; **device busy/step = forward-module compute ÷ unprofiled step** (profiler-invariant).

| regime | forward | throughput | device idle |
|--------|---------|-----------|-------------|
| DP1/TP4 | 4.85 ms | 3296 tok/s | ~2% |
| DP2/TP2 | 7.35 ms | 2176 tok/s | ~2–6% |
| DP4/TP1 | 14.5 ms | 1104 tok/s | ~3–6% |

**The device is ~saturated in every DP regime → `device_put`/launch is masked under the forward → reducing it buys ~0 throughput here.** The 2–6% residual is within measurement noise (chip-to-chip spread 3% vs 6% for identical work).

Note the DP-scaling intuition (more DP → more `device_put` → un-mask) is *disproved* on a single host: with fixed chips, more DP means less TP-per-replica, which **lengthens the forward faster than it grows the host**, so the device only gets *more* saturated.

[...]

PR #1652, opened 2026-09-14
perf(scheduler): reduce decode host overhead and precompile token gather

Yanko-7:
## Motivation

While debugging Qwen3-8B, I noticed substantial idle bubbles between decode computations. XProf showed significant CPU host overhead that could delay submission of the next batch.

Decode device computation duration before optimization

Decode host execution before optimization

JAX host calls and decode bubbles before optimization

The profile highlights three issues:

1. A device forward takes **6.183 ms**, while the corresponding host processing span on the runner thread takes **6.621 ms**. These spans can overlap, but host work of this magnitude can limit how quickly subsequent batches are submitted. 
2. `PjitFunction(jitted_sampler)` takes **1.814 ms** and includes slow `cpp_pjit_shard_arg_fallback`, indicating additional input preparation. 
3. `PjitFunction(jitted_run_model)` also has substantial host overhead.

### Unused outputs from disabled precision tracing

Even with precision tracing disabled, each trace point returns a JAX device boolean. Qwen3-8B returns **109 such markers per forward**, which the host must process and release along with the model outputs.

The original behavior is equivalent to:

```python
if tracing_enabled:
    return record_tensor(...)
return jnp.bool_(True)
```

This PR changes the disabled path to return `None`:

```python
if tracing_enabled:
    return record_tensor(...)
return None
```

Releasing the unused marker list alone took **0.831 ms wall time / 0.828 ms calling-thread CPU time** in the earlier TPU profile. In the tracer-only investigation, host `_forward_raw` decreased from **3.435 ms to 0.929 ms**, a reduction of approximately **73%**. 

Returning `None` removes the unused device outputs from the compiled function. The observed improvement in `PjRtCApiLoadedExecutable::Execute`, as well as subsequent output cleanup, points to output management as an important source of host overhead. The cleanup measurement covers only part of that cost; it does not isolate the individual operations inside `Execute`.

[PR #129](https://github.com/sgl-project/sglang-jax/pull/129), introduced the disabled `jnp.bool_(True)` return. Enabled precision tracing remains unchanged.

### Repeated preparation of JIT inputs

Some JIT inputs are still Python or NumPy values and require conversion, device placement, or sharding preparation at invocation time. These inputs can enter `cpp_pjit_shard_arg_fallback` instead of using an already prepared device buffer. The event name alone does not identify the triggering argument.

This PR reduces that repeated work in three places:

- **Sampler flags and unused vocabulary masks:** prepare and reuse device arrays. Mutable grammar masks are still uploaded after their contents are updated.
- **Future-token slot indices:** compute slots inside the existing JIT using device-side `seq_lens` and `req_pool_indices`, instead of constructing a NumPy slot array on the CPU and passing it into JIT. The existing per-request addressing and padding behavior is preserved.
- **Sampler step counter:** replace the Python integer counter introduced in [PR #940](https://github.com/sgl-project/sglang-jax/pull/940) with a device scalar, incremented and returned inside the sampler JIT. This avoids preparing a changing host scalar on every step while retaining the RNG step sequence.

### Profiles after optimization

The following captures show reduced decode bubbles after the host optimizations. They illustrate the change in the timeline; the repeated, unprofiled benchmark results are reported separately below.

Decode host execution after optimization

JAX host calls after optimization

Decode timeline after optimization

### Additional changes and related work

AOT dispatch still flattens stable weight pytrees on cache hits. Its dynamic key also omits pytree metadata and Python scalar type distinctions, which can select an incompatible executable. This PR addresses both issues. AOT was disabled in the decode profiles and benchmark discussed here, so it is not an explanation for their observed bubbles.

The overlap path requires a token-ID replication gather. This PR retains it and covers it during precompilation to avoid compiling it during serving.

Related reports: [#293](https://github.com/sgl-project/sglang-jax/issues/293) and [#1500](https://github.com/sgl-project/sglang-jax/issues/1500) discuss scheduler/launch overhead and whether device execution masks it. Those reports do not establish that these specific causes explain every workload's host bubbles.

## Modifications

- Return `None` for disabled precision tracing, eliminating device-output leaves. Keep the enabled callback and flag.
- Prepare cached device scalar flags and unused masks when constructing sampling metadata, keyed by sharding and mask shape. Upload mutable grammar masks at the existing grammar-update point. Both sampler paths consume the prepared metadata directly, without a separate adapter or per-call metadata copy.
- Keep the RNG step counter on device and increment it inside the sampler JIT, including the fused path. Return the updated counter internally while preserving the caller-facing sampler result structure.
- Compute future-token slots inside the existing JIT from batch device arrays; update precompilation and slot regression tests.
- Precompile token-ID gather immediately after `set_future_token_ids` in the existing decode bucket loop, using actual sampler outputs. Runtime and precompilation obtain the same cached JIT function from `get_token_ids_gather(mesh)`, preserving replicated output sharding without callback plumbing or a separate bucket loop.
- Flatten only dynamic arguments on AOT cache hits, using relative kept-input indices. Include the dynamic tree structure and aval shape/dtype/weak type in the cache key. Preserve stable-weight replacement/invalidation checks.

No new runtime switches or profiling instrumentation. AOT and fusion retain main's defaults.

## Accuracy Tests

- **GSM8K: 93.0% (465/500)** using the Qwen3 cookbook recipe; 500 requests succeeded, 0 errors. Qwen3-8B BF16 on TPU v7x, TP8 / DP1, concurrency 8, thinking enabled.

### Stock main vs full host optimizations

Qwen3-8B, BF16, single-host TPU v7x, four active JAX devices (TP4, DP2), fixed 2,640 input / 128 output tokens, no radix cache. Both variants ran on the same pod in ABBAAB order, three independent service sessions each. Each session tested concurrency 4/16/32 with 128 warmup requests followed by 1,024 measured requests. 

| Concurrency | Metric | Before | After | Change |
|---:|---|---:|---:|---:|
| 4 | Output tokens/s | 513.52 | 711.93 | +38.6% |
| 4 | Median TPOT (ms) | 6.52 | 4.29 | -34.2% |
| 4 | P99 TTFT (ms) | 232.32 | 211.98 | -8.8% |
| 16 | Output tokens/s | 1511.41 | 1757.63 | +16.3% |
| 16 | Median TPOT (ms) | 7.95 | 6.49 | -18.4% |
| 16 | P99 TTFT (ms) | 535.44 | 532.89 | -0.5% |
| 32 | Output tokens/s | 2203.66 | 2320.91 | +5.3% |
| 32 | Median TPOT (ms) | 10.01 | 9.17 | -8.4% |
| 32 | P99 TTFT (ms) | 989.24 | 981.20 | -0.8% |

The results show that, as concurrency increases, each decode forward has more work to do, making it easier to hide host overhead. This is consistent with expectations.

## Discussion of other issues

Although these changes address the decode bubbles under light workloads for now, the latest XProf traces reveal remaining design issues in our overlap strategy:
image

1. The scheduler's `run_batch` and the forward thread execute serially on the host timeline.
2. The scheduler performs a series of blocking `DevicePutWithSharding` calls.
3. The scheduler blocks while waiting for `jax.Array` results from the previous iteration.

These blocking operations on the host are at odds with the scheduler's intended role in an overlapped execution pipeline

## Checklist

- [x] Problem, relevant introducing PRs, and related reports identified.
- [x] Focused regression tests pass.
- [x] Experimental annotations and unrelated changes excluded.
- [x] Collected repeated main/full-patch benchmark results and documented the final gather wiring validation boundary.


JamesBrianD (approved):
LGTM

Issue #1655, opened 2026-09-14
[RFC] Qwen3.8-Flash-Next


real3stone:
## \[RFC\] Support Qwen3.8-Flash-Next on SGL\_JAX

## Summary

This RFC proposes adding **text-only** inference support for **Qwen3.8-Flash-Next** (`Qwen/Qwen3.8-Flash-Next`).

Despite the name, this is not a size variant of the Qwen3.8 family. Its `architectures` is `Qwen4ExpForConditionalGeneration` (`model_type: qwen4_exp`), and Qwen's release post calls the Next architecture ["a prototype of the next-generation Qwen4 series"](https://qwen.ai/blog?id=qwen3.8-flash-next). It shares the GDN hybrid backbone with Qwen3.5 — which this repo already supports — but adds three mechanisms that have no counterpart here.

## 1. Model Architecture

### 1.1 Backbone

180B total / ~6B active, 48 layers, `max_position_embeddings=262144`. `full_attention_interval: 4` — 36 layers run Gated DeltaNet linear attention, 12 run Qwen Sparse Attention. The full-attention layers are **standard GQA** (24 query heads, 2 KV heads, head_dim 256), not MLA. All 48 layers are MoE: 512 routed experts top-10, one shared expert, pre-fused on disk.

As a hybrid-recurrent model it carries the same serving constraints as Qwen3.5: `--disable-radix-cache`, or `--enable-unified-radix-tree` with `--page-size 1`.

### 1.2 QSA — Qwen Sparse Attention

Each full-attention layer carries a **learned indexer** (an `index_qk_proj` plus two layernorms) that selects which past tokens the layer attends to. Selection budget is `indexer_budget: 2048` tokens over 4-token compressed blocks.

### 1.3 N-gram Embedding (`ple` in the config and weights)

Present on exactly one layer (`ple_layer_ids: [2]`, 1-based — `layers.1` in the checkpoint). A 20M-entry bigram/trigram table — 51B of the model's 180B parameters, sharded 128 ways — feeds a gate and a dilated short convolution (`ple_conv_kernel_size=4`, dilation `ngram_size=3`), alongside key/value projections and three norms.

vLLM implements this on its Mamba base class, so **the layer carries conv state**. That makes it a third stateful component alongside GDN's recurrent and conv state, with its own entry in the state pool and memory budget.

### 1.4 Hyper Connections

[...]

PR #1661, opened 2026-09-16
feat(layers): add Qwen4Exp gated-residual hyper connections

real3stone:
## Motivation

First step toward supporting Qwen3.8-Flash-Next (`qwen4_exp`). See #1655 for the full plan; this PR is its P2.

The model widens the inter-block residual path to `hc_count=4` parallel streams, so hidden states between blocks are `[..., HC*HS]` rather than `[..., HS]`, with a gate pair wrapping every block. Nothing in this repo has a counterpart.

It lands standalone, ahead of the config class and the model module, because the backbone shape convention it defines is what the config, the N-gram layer and the model all build on.

## Modifications

`python/sgl_jax/srt/layers/gated_residual.py`:

- `GatedResidual` — the `mix`/`combine` pair that wraps one block. `mix` collapses the streams into one `HS`-wide block input through a low-rank sigmoid gate; `combine` adds the block output back into every stream through a learned per-stream scalar. `combine` consumes the normalized hyper input that `mix` produced rather than recomputing it, so `mix` hands it back.
- `GatedResidualMixer` — the same collapse without the injection, used once at the end of the backbone. Its checkpoint prefix `hyper_connection_mixer` carries only three tensors; `hyper_connection` adds `block_inject_weight`.

Reuses `GemmaRMSNorm` (its `add_unit_offset=True` path is the `* (1.0 + weight)` form this needs) and `LinearBase`. The three projections are replicated: `input_mix_weight_up` emits `[..., HC*HS]`, so a row-parallel split would add an all-reduce at each of the model's `2 * num_layers` call sites to save a few hundred MB of weights.

### The shape convention this settles

For the config class, the N-gram layer and the model module:

- Inter-block hidden states are `[..., HC*HS]`, HC outer and HS inner, matching the checkpoint layout.
- Two instances per layer, one per sub-block: 384 `hyper_connection` tensors / 4 per instance = 96 = 48 layers x 2, plus 3 for the top-level mixer.
- The entry is a parameter-free broadcast. The checkpoint has no "expand one stream into HC" weight, so all HC streams start identical and diverge only through the per-stream injection scalars.

### Why not `kernels/mhc`

`kernels/mhc` (#1547) implements DeepSeek-V4's variant of the same paper, and is not reusable here:

|  | `kernels/mhc` | `qwen4_exp` |
| --- | --- | --- |
| cross-stream mixing | Sinkhorn-normalized doubly stochastic `comb` | none, i.e. the identity |
| gate projection | full-rank single dot, `rows` fixed at `hc` or `(2+hc)*hc` | low-rank, `hc_lowrank=320` |
| platform | `tune.py` has a v6e-only schedule table and raises otherwise | — |

The low-rank objection only rules out `mhc_pre_fused`. `mhc_post_fused` takes the gates as inputs, and this `combine` is its special case with `comb` set to the identity — but reaching it that way means building a `[T, HC, HC]` identity and running a bilinear contraction where a broadcast multiply-add does, so it costs more than it saves.

## Accuracy Tests

Nothing consumes this layer yet, so there is no end-to-end number. The numerics are covered two ways.

`python/sgl_jax/test/layers/test_gated_residual.py`, registered in `unit-test-cpu` — no TPU, no hub download:

```bash
python3 -m pytest python/sgl_jax/test/layers/test_gated_residual.py -q
```

```
9 passed
```

The reference is fp64 numpy that loops over the streams explicitly, so a layout mistake in the vectorized path does not reproduce in it. Two invariants the backbone depends on have their own cases: the streams are only ever added to (a zero block output leaves them unchanged), and every stream receives the same block output scaled by its own scalar in (0, 2).

Cross-checked offline against vLLM's `GatedResidual` (`vllm/models/qwen4_exp/common/hyperconnection.py`) run under torch with shared weights:

```
mix       max_abs_err=3.427e-07  max_rel=1.521e-07
combine   max_abs_err=4.768e-07  max_rel=7.787e-08
```

## Benchmarking and Profiling

Not applicable — the layer has no call sites yet.

## Checklist


python/sgl_jax/srt/layers/gated_residual.py
```diff
@@ -0,0 +1,213 @@
...
+                f"got {block_output.shape[-1]}"
+            )
+
+        # 2*sigmoid, not sigmoid: the per-stream scales live in (0, 2), so a block can
+        # write into a stream harder than one-for-one. The second /hc_count sits here.
+        inject, _ = self.block_inject_weight(normed)
+        inject = 2.0 * jax.nn.sigmoid(inject / self.hc_count)
+
+        out = self._unflatten(hyper_input) + block_output[..., None, :] * inject[..., :, None]
+        return out.reshape(hyper_input.shape).astype(hyper_input.dtype)
```
JamesBrianD: I didn't see any explicit type conversion during the calculation process; is it necessary to convert the type back? Or is `block_inject_weight` in FP32?

python/sgl_jax/test/layers/test_hyperconnection.py
```diff
@@ -0,0 +1,274 @@
...
+TOKENS = 6
+
+ATOL = 2e-5
+RTOL = 2e-5
+
+
+def _sigmoid(x):
+    return 1.0 / (1.0 + np.exp(-x))
+
+
```
JamesBrianD, line 32: The format and style of the test file do not match the sglang style; please refer to other test files and make the necessary adjustments.

python/sgl_jax/srt/layers/gated_residual.py
real3stone: You're right, there's no need to do this. Dropped both casts.

python/sgl_jax/test/layers/test_hyperconnection.py
real3stone: Converted to unittest / CustomTestCase with a `unittest.main()` entry point,
matching test/srt/. Dropped `runner="pytest"` from the suite registration.

Let me know if you meant something else in that range.

JamesBrianD:
Before diving into the implementation details, I have one high-level design suggestion.

I think the current implementation is a bit too specialized around the Qwen4Exp use case and could benefit from a more general HyperConnection abstraction. Could we first align the layer structure with the SGLang implementation in `python/sglang/srt/layers/hyperconnection.py`?

In SGLang, the implementation is organized around a common `HyperConnectionBase`, with `GatedResidual` implemented as a subclass. The same abstraction also handles different roles through `use_mix` / `use_combine`, instead of introducing a separate class for each usage such as the final mixer.

I think following a similar structure here would make the implementation more cohesive, easier to extend to other HyperConnection variants, and keep the JAX implementation closer to upstream. It would also make the later model integration and weight mapping more straightforward.

So before adding more implementation-specific logic, I would suggest restructuring this around a common HyperConnection base class and then implementing the gated residual behavior as a subclass.

real3stone:
> Before diving into the implementation details, I have one high-level design suggestion.
> 
> I think the current implementation is a bit too specialized around the Qwen4Exp use case and could benefit from a more general HyperConnection abstraction. Could we first align the layer structure with the SGLang implementation in `python/sglang/srt/layers/hyperconnection.py`?
> 
> In SGLang, the implementation is organized around a common `HyperConnectionBase`, with `GatedResidual` implemented as a subclass. The same abstraction also handles different roles through `use_mix` / `use_combine`, instead of introducing a separate class for each usage such as the final mixer.
> 
> I think following a similar structure here would make the implementation more cohesive, easier to extend to other HyperConnection variants, and keep the JAX implementation closer to upstream. It would also make the later model integration and weight mapping more straightforward.
> 
> So before adding more implementation-specific logic, I would suggest restructuring this around a common HyperConnection base class and then implementing the gated residual behavior as a subclass.

Agreed with your high-level design suggestion.

Restructured around HyperConnectionBase with GatedResidual as a subclass, and use_mix / use_combine in place of the separate mixer class.

python/sgl_jax/srt/layers/hyperconnection.py
```diff
@@ -0,0 +1,286 @@
...
+            )
+        self.hidden_size = hidden_size
+        self.epsilon = epsilon
+        self.group_size = group_size
+        self.weight = nnx.Param(
+            nnx.with_partitioning(nnx.initializers.zeros, kernel_axes)(
+                jax.random.PRNGKey(0), (hidden_size,)
+            )
+        )
+
```
JamesBrianD, line 92: One small mismatch with the vLLM Torch implementation: `GroupedGemmaRMSNorm` doesn't appear to honor `config.params_dtype`.

The linear weights are created with `params_dtype`, but `hc_norm.weight` is initialized independently without receiving that dtype. In vLLM, `GroupedGemmaRMSNorm` explicitly creates the norm weight with `dtype=config.params_dtype`.

Could we pass `params_dtype` into `GroupedGemmaRMSNorm` as well? It may also be worth asserting the norm parameter dtype in the bf16 test, since the current round-trip test only checks the output dtype.

python/sgl_jax/srt/layers/hyperconnection.py
```diff
@@ -0,0 +1,286 @@
...
+        self._check_hyper(hyper_input)
+        mixed = jnp.mean(self._unflatten(hyper_input), axis=-2)
+        # Second element is unused here; the payload stays a 2-tuple so the
+        # base and its subclasses share one contract.
+        return mixed, (hyper_input, hyper_input)
+
+    @named_scope
+    def combine(self, block_output: jax.Array, residuals: tuple[jax.Array, jax.Array]) -> jax.Array:
+        """Add the block output into every stream."""
+        hyper_input, _ = residuals
```
JamesBrianD: Could we keep this consistent with upstream and return `hyper_input` directly here?

`HyperConnectionBase` doesn't need the extra tuple element, while `GatedResidual` can keep returning `(hyper_input, hyper_input_normed)` since it actually needs both.

python/sgl_jax/srt/layers/hyperconnection.py
real3stone: Both applied. thanks.

`config.params_dtype` now reaches `GroupedGemmaRMSNorm` (default fp32, so other
norms in the repo are unchanged), and the bf16 test asserts each parameter's
dtype. 
It was written without one by following the repo's own `GemmaRMSNorm`,
which takes no dtype. Upstream SGLang doesn't take a `params_dtype` either — it
relies on PyTorch's module-level dtype handling, which JAX has no equivalent of
— so this is a JAX-specific gap rather than the vLLM signature.

JamesBrianD (approved):

PR #1662, opened 2026-09-16
feat(configs): add Qwen4Exp config and hybrid-recurrent runner seams

real3stone:
## Motivation

Next step toward supporting Qwen3.8-Flash-Next (`qwen4_exp`). See #1655 for the full plan; this PR is its P1. Independent of #1661 (P2) — either can land first.

After this, `get_config("Qwen/Qwen3.8-Flash-Next")` resolves to a local config exposing the hybrid/GDN interface the runner reads, and the runner routes the model's 36 linear-attention layers to the existing GDN backend. The model itself still does not load: `Qwen4ExpForConditionalGeneration` has no module yet, which is P5.

## Modifications

`python/sgl_jax/srt/configs/qwen4_exp.py`, defined from `PretrainedConfig` directly as the repo's other hybrid families are, rather than inheriting from Qwen3.5's text config.

The backbone matches Qwen3.5 — GDN linear attention on 36 of 48 layers, full attention on the rest, MoE FFN — so `layer_types`, the RoPE flattening and `linear_state_params` mirror `configs/qwen3_5.py` and are consumed by the same `model_runner_kv_cache_mixin` helpers via duck typing.

Three mechanisms have no Qwen3.5 counterpart. They are surfaced here, with the derived shapes the layers that consume them will size against:

| | config fields | derived |
| --- | --- | --- |
| Hyper connections (#1661) | `hc_count`, `hc_lowrank` | `hyper_hidden_size` = 10240 |
| N-gram embedding | `ple_*`, `ngram_*`, `split_ngram_parts` | `short_conv_layer_ids` = `[1]` (`ple_layer_ids` is 1-based), `short_conv_state_shape` = `(10240, 9)`, `ngram_context_len` = 2 |
| QSA | `indexer_*` | `indexer_block_topk` = 512 |

Two seams hook it up. `model_runner_kv_cache_mixin.py` gains a `qwen4_exp_config` property feeding `linear_recurrent_config`, which is where hybrid-recurrent dispatch is centralized — everything downstream of it is duck-typed and unchanged. **Missing that branch is silent**: the model would take the dense path with a plain KV pool and no recurrent state, with no error, so there is a test for it. `hybrid_linear_attn_backend.py` widens the GDN branch of `attn_backend_wrapper` to match, which is what routes each of the 48 layers to GDN or to full attention; `GDNAttnBackend` reads only `linear_*` fields, so the backend itself is unchanged.

No model module, no weight-mapping table, no GDN backend or kernel change.

## Accuracy Tests

No model outputs change — nothing constructs a Qwen4Exp model yet.

`python/sgl_jax/test/models/test_qwen4_exp.py`, registered in `unit-test-cpu`, hub-free so it does not skip on an offline runner:

```bash
python3 -m pytest python/sgl_jax/test/models/test_qwen4_exp.py -q
```

Covers RoPE flattening off the real nested block, the 36/12 layer split, every derived shape in the table above, `linear_state_params` against `LinearRecurrentStateParams`, and registry resolution.

The config's constraints are ported from the upstream vLLM config so a malformed checkpoint fails at construction, naming the field, rather than surfacing as a shape mismatch inside a kernel. 12 parametrized rejection cases cover them, plus the `swish` → `silu` normalization and the `layer_types` length assert.

Two cases are there for traps rather than coverage:

- `test_runner_routes_qwen4_exp_to_the_recurrent_path` — pins the silent dense-path fallthrough described above, and asserts Qwen3.5's branch is unaffected.
- `test_gdn_and_ngram_conv_widths_coincide_but_are_unrelated` — GDN's `conv_dim` (`2·16·128 + 48·128`) and the N-gram conv's channel count (`hidden_size · hc_count`) are both 10240 for this checkpoint by coincidence. Their state lengths are 3 and 9. Two separate conv states of the same width are easy to confuse.

Run together with the existing Qwen3.5 config tests as a regression check: `48 passed`.

## Benchmarking and Profiling

Not applicable — no runtime path changes for any currently supported model.

## Checklist


python/sgl_jax/test/models/test_qwen4_exp.py
```diff
@@ -0,0 +1,260 @@
...
+"""
+
+from __future__ import annotations
+
+import os
+from types import SimpleNamespace
+
+import pytest
+
+os.environ.setdefault("JAX_PLATFORMS", "cpu")
```
JamesBrianD: Please delete this test files.

JamesBrianD:
Aside from the test files being overly redundant, everything else looks good to me.

real3stone:
> Aside from the test files being overly redundant, everything else looks good to me.

Happy to trim them. Do you want that in this PR, or in the follow-up that builds on it? (4-chip jobs queue for hours here, so I'd rather not spin CI twice if it can wait.)

JamesBrianD:
> > Aside from the test files being overly redundant, everything else looks good to me.
> 
> Happy to trim them. Do you want that in this PR, or in the follow-up that builds on it? (4-chip jobs queue for hours here, so I'd rather not spin CI twice if it can wait.)

I understand your concerns. Please keep the changes to this PR simple; I won't wait for the full CI run to complete—I'll merge the PR directly once the linting finishes.

real3stone:
> > > Aside from the test files being overly redundant, everything else looks good to me.
> > 
> > 
> > Happy to trim them. Do you want that in this PR, or in the follow-up that builds on it? (4-chip jobs queue for hours here, so I'd rather not spin CI twice if it can wait.)
> 
> I understand your concerns. Please keep the changes to this PR simple; I won't wait for the full CI run to complete—I'll merge the PR directly once the linting finishes.

@JamesBrianD  Trimmed the tests.

Because of a conflict in the unit-test-cpu list, I had to rebase and force-push rather than stack a commit.

JamesBrianD:
> > > > Aside from the test files being overly redundant, everything else looks good to me.
> > > 
> > > 
> > > Happy to trim them. Do you want that in this PR, or in the follow-up that builds on it? (4-chip jobs queue for hours here, so I'd rather not spin CI twice if it can wait.)
> > 
> > 
> > I understand your concerns. Please keep the changes to this PR simple; I won't wait for the full CI run to complete—I'll merge the PR directly once the linting finishes.
> 
> @JamesBrianD Trimmed the tests.
> 
> Because of a conflict in the unit-test-cpu list, I had to rebase and force-push rather than stack a commit.

Thanks for trimming the tests. I think I wasn't clear in my previous comment — by keeping this PR simple, I meant removing `test_qwen4_exp.py` entirely rather than shortening it.

At this stage, I don't think we need a standalone test file for these config assertions. We can add focused tests later when the actual model/runtime path is implemented.

Could you please remove this test file from this PR? Thanks!

real3stone:
> > > > > Aside from the test files being overly redundant, everything else looks good to me.
> > > > 
> > > > 
> > > > Happy to trim them. Do you want that in this PR, or in the follow-up that builds on it? (4-chip jobs queue for hours here, so I'd rather not spin CI twice if it can wait.)
> > > 
> > > 
> > > I understand your concerns. Please keep the changes to this PR simple; I won't wait for the full CI run to complete—I'll merge the PR directly once the linting finishes.
> > 
> > 
> > @JamesBrianD Trimmed the tests.
> > Because of a conflict in the unit-test-cpu list, I had to rebase and force-push rather than stack a commit.
> 
> Thanks for trimming the tests. I think I wasn't clear in my previous comment — by keeping this PR simple, I meant removing `test_qwen4_exp.py` entirely rather than shortening it.
> 
> At this stage, I don't think we need a standalone test file for these config assertions. We can add focused tests later when the actual model/runtime path is implemented.
> 
> Could you please remove this test file from this PR? Thanks!

Done.

PR #1663, opened 2026-09-16
feat(csa): add Pallas compressor operator for DeepSeek-V4

KoalaYuFeng:
## Motivation

Add a standalone ratio-4 overlapping compressor for CSA's compressed KV and Lightning Indexer keys, supporting prefill, decode, and ragged inputs with caller-owned state and paged caches.

## Modifications

- Fuse projection, softmax pooling, RMSNorm, RoPE, and cache encoding in Pallas.
- Preserve native cache layouts, using bulk DMA for aligned contiguous blocks and scatter for irregular writes.
- Add an independent NumPy reference, correctness tests, and a configurable benchmark.

## Accuracy Tests

All tests passed. FP32 arithmetic uses `rtol=2e-2, atol=1e-2`; cache outputs use quantization-aware checks. Encoding identical FP32 values is checked byte-for-byte. Tests cover stateful execution, padding, quantization boundaries, and cache writes.

## Benchmarking and Profiling

Measured on TPU v6e-1 with JAX/JAXLIB 0.11.1, BF16 projection operands, FP32 accumulation, and a 224 MiB cache. All values are mean device latency in **milliseconds**, including computation and state/cache writes. Measurements use four rotating input buffers, 3 warmups, and 10 timed iterations; compilation and host-to-device resets are excluded.

- **Decode update:** updates state without emitting a compressed record; context is S+1.
- **Decode emit:** updates state and emits a compressed record at context S.
- **Prefill:** S input tokens per request.
- **Ragged:** query lengths repeat `[1, S/4, S/2, S]`; B=1 uses S/2 tokens. Remaining tokens form the prefix.

### D=4096

| Batch | S | Decode update | Decode emit | Prefill | Ragged |
|---:|---:|---:|---:|---:|---:|
| 1 | 512 | 0.06 | 0.06 | 0.08 | 0.05 |
| 1 | 2048 | 0.05 | 0.05 | 0.23 | 0.13 |
| 1 | 8192 | 0.06 | 0.06 | 0.89 | 0.44 |
| 8 | 512 | 0.04 | 0.04 | 0.44 | 0.33 |
| 8 | 2048 | 0.04 | 0.04 | 2.24 | 0.86 |
| 8 | 8192 | 0.05 | 0.05 | 8.87 | 3.88 |
| 32 | 512 | 0.07 | 0.08 | 2.24 | 1.13 |
| 32 | 2048 | 0.07 | 0.08 | 8.88 | 3.92 |
| 32 | 8192 | 0.07 | 0.08 | 35.84 | 15.80 |

### D=7168

| Batch | S | Decode update | Decode emit | Prefill | Ragged |
|---:|---:|---:|---:|---:|---:|
| 1 | 512 | 0.06 | 0.06 | 0.11 | 0.07 |
| 1 | 2048 | 0.07 | 0.07 | 0.32 | 0.17 |
| 1 | 8192 | 0.07 | 0.07 | 1.64 | 0.65 |
| 8 | 512 | 0.05 | 0.05 | 0.65 | 0.42 |
| 8 | 2048 | 0.06 | 0.06 | 3.20 | 1.44 |
| 8 | 8192 | 0.06 | 0.06 | 12.71 | 5.69 |
| 32 | 512 | 0.09 | 0.10 | 3.21 | 1.74 |
| 32 | 2048 | 0.09 | 0.10 | 12.72 | 5.73 |

### Reproduction

```bash
PYTHONPATH=python:. python -m pytest -q \
  test/srt/kernels/csa_compressor/test_compressor.py

PYTHONPATH=python:. python \
  benchmark/kernels/csa_compressor/benchmark_compressor.py \
  --batch 1 4 8 16 32 \
  --sequence 512 2048 8192 \
  --hidden 4096 \
  --pattern decode prefill ragged \
  --cache-pages 2048 \
  --output compressor-d4096.json
```

- Additional benchmark options are documented in `benchmark/kernels/csa_compressor/README.md`.

## Checklist


KoalaYuFeng:
Hi @zorrofox @JamesBrianD @cjx0709, following our earlier discussion, we’ve split out the Deepseek-v4-flash CSA compressor into a standalone operator with correctness tests and benchmarks. Could you please take a look when you have a chance? Thanks!

zorrofox:
Tried this kernel on TPU7x (v7x, jax 0.11.1) and wired a fork of `csa_dual_uniform_prefill_pallas` into a DeepSeek-V4-Flash prefill path. Three things worth sharing.

**1. v7x runs fine on the v6e schedule; `tune.py` could fall back instead of raising.** `get_compressor_schedule` raises for `TPU7x`. Monkeypatching it to the v6e values (`projection_k_tile=2048`, `query_tile=128`, which fit v7x's 64 MiB VMEM with a lot of margin) gives, with `benchmark_compressor.py --batch 1 --hidden 4096 --pattern prefill ragged`: S=8192 prefill 0.88 ms (your v6e-1 figure is 0.89), S=4096 0.37 ms, ragged 4096-query/4096-prefix 0.37 ms. Same time on a chip with ~2x the MXU throughput suggests the kernel is not compute-bound at these shapes, so a generic fallback schedule seems safe.

**2. The first group of a sequence pools over the carried state rows.** `_pool_uniform_groups` (and `_csa_state_step_kernel`) always read the "previous group" from the state, and `_pool_normalize` softmaxes over all 8 window rows with no in-sequence mask. At position 0 there is no previous group, so the first record's content depends entirely on what the caller put in the fresh request's state. If that is zeros, the first record pools 4 zero kv rows with softmax weight exp(0 - max) each, which differs from a reference that pools only real tokens. The decode path does set `-inf` scores for padded batch rows, so maybe the intended contract is that a fresh state also carries `-inf` scores? If so it would be worth stating in the docstring (or masking on `position < 4` inside the kernel). For our comparison we fed kv=0 / score=-1e30 for the first chunk and then matched our reference to half a bf16 ulp.

**3. Isolated speedup vs end-to-end.** With a bf16-record variant of the dual prefill kernel (we keep bf16 caches), one CSA layer at 8K on v7x went from 1.38 ms (two separate compressor calls) to 1.05 ms including the state-ring conversion. End-to-end on a 4-chip v7x slice with sequence-parallel prefill the TTFT moved only ~1% at 8K and ~2% at 32K, because the compressor mostly overlaps with the sequence-parallel collectives there; GSM8K-200 was 194 vs 195 (bf16-level differences, 3 answers flipped either way). Not a problem with the kernel, just a data point that the win depends on what else is on the critical path.

Happy to share the v7x benchmark JSON or the bf16-record fork if useful.

KoalaYuFeng:
> Tried this kernel on TPU7x (v7x, jax 0.11.1) and wired a fork of `csa_dual_uniform_prefill_pallas` into a DeepSeek-V4-Flash prefill path. Three things worth sharing.
> 
> **1. v7x runs fine on the v6e schedule; `tune.py` could fall back instead of raising.** `get_compressor_schedule` raises for `TPU7x`. Monkeypatching it to the v6e values (`projection_k_tile=2048`, `query_tile=128`, which fit v7x's 64 MiB VMEM with a lot of margin) gives, with `benchmark_compressor.py --batch 1 --hidden 4096 --pattern prefill ragged`: S=8192 prefill 0.88 ms (your v6e-1 figure is 0.89), S=4096 0.37 ms, ragged 4096-query/4096-prefix 0.37 ms. Same time on a chip with ~2x the MXU throughput suggests the kernel is not compute-bound at these shapes, so a generic fallback schedule seems safe.
> 
> **2. The first group of a sequence pools over the carried state rows.** `_pool_uniform_groups` (and `_csa_state_step_kernel`) always read the "previous group" from the state, and `_pool_normalize` softmaxes over all 8 window rows with no in-sequence mask. At position 0 there is no previous group, so the first record's content depends entirely on what the caller put in the fresh request's state. If that is zeros, the first record pools 4 zero kv rows with softmax weight exp(0 - max) each, which differs from a reference that pools only real tokens. The decode path does set `-inf` scores for padded batch rows, so maybe the intended contract is that a fresh state also carries `-inf` scores? If so it would be worth stating in the docstring (or masking on `position < 4` inside the kernel). For our comparison we fed kv=0 / score=-1e30 for the first chunk and then matched our reference to half a bf16 ulp.
> 
> **3. Isolated speedup vs end-to-end.** With a bf16-record variant of the dual prefill kernel (we keep bf16 caches), one CSA layer at 8K on v7x went from 1.38 ms (two separate compressor calls) to 1.05 ms including the state-ring conversion. End-to-end on a 4-chip v7x slice with sequence-parallel prefill the TTFT moved only ~1% at 8K and ~2% at 32K, because the compressor mostly overlaps with the sequence-parallel collectives there; GSM8K-200 was 194 vs 195 (bf16-level differences, 3 answers flipped either way). Not a problem with the kernel, just a data point that the win depends on what else is on the critical path.
> 
> Happy to share the v7x benchmark JSON or the bf16-record fork if useful.

Hi @zorrofox, thanks for the feedback! Added the fresh-state initialization and state-reuse requirements to the API docstring. Could you please take another look when convenient? Thanks!

PR #1664, opened 2026-09-17
feat(csa): add Pallas gather and joint attention operator for DeepSeek-V4.

KoalaYuFeng:
## Motivation

Add a Pallas CSA attention operator that gathers selected compressed KV entries and computes joint attention over the sliding window and compressed history, using a single online softmax state without materializing the gathered KV in HBM.

## Modifications

- Compact selected KV into on-chip tiles for single-query execution, with pipelined page DMA and attention computation.
- Share page reads across query blocks for prefill and ragged inputs, preserving per-query causal and Top-K masks.
- Accept caller-owned metadata, precomputed Top-K indices, and explicit cache-format parameters.
- Preserve the existing compressed-cache layout; all input caches remain read-only.
- Add an independent NumPy reference, correctness tests, and a standalone benchmark.

## Benchmarking and Profiling

Single TPU v6e (device reports TPU v6 lite), JAX/JAXLIB 0.11.1. BF16 Q/output, 64 heads, head dimension 512, default dot precision. Window size 128, compression ratio 4, Top-K width 512; FP8/E8M0 NoPE cache uses 64-value scale blocks and RoPE uses BF16 with four rows per storage group.

Each point uses four independently generated input buffers, three warmup calls, and the arithmetic mean of ten device-timed calls. Compilation and host input generation/loading are excluded. Timing includes route preparation, gather, joint attention, and output write-back. Decode uses resident HBM buffers; prefill/ragged stage each selected host-backed buffer into HBM before the measured call.

Decode has one query per request; prefill has S queries per request. For ragged, request r has floor(S / (1 + r % 4)) queries and prefix length S minus that query count. All requests end at context length S. At B=1, ragged and prefill therefore describe the same workload.

### Decode (μs)

| Batch / context length | 512 | 1024 | 2048 | 4096 | 8192 |
|---:|---:|---:|---:|---:|---:|
| 1 | 12.61 | 11.30 | 13.26 | 14.20 | 18.08 |
| 4 | 20.13 | 21.21 | 26.18 | 32.83 | 44.09 |
| 8 | 33.03 | 35.24 | 45.69 | 59.21 | 81.63 |
| 16 | 59.44 | 63.64 | 84.81 | 111.83 | 156.59 |
| 32 | 112.25 | 120.92 | 162.70 | 216.87 | 306.58 |

### Prefill (ms)

| Batch / context length | 512 | 1024 | 2048 | 4096 | 8192 |
|---:|---:|---:|---:|---:|---:|
| 1 | 0.18 | 0.38 | 0.70 | 1.84 | 5.40 |
| 4 | 0.56 | 1.18 | 2.82 | 7.39 | 24.26 |
| 8 | 1.11 | 2.36 | 5.64 | 17.33 | 43.56 |
| 16 | 2.21 | 4.72 | 13.81 | 29.70 | 97.53 |
| 32 | 4.42 | 11.95 | 22.68 | 69.81 | OOM |

### Ragged (ms)

| Batch / context length | 512 | 1024 | 2048 | 4096 | 8192 |
|---:|---:|---:|---:|---:|---:|
| 1 | 0.18 | 0.38 | 0.70 | 1.84 | 5.40 |
| 4 | 0.33 | 0.66 | 1.63 | 4.46 | 13.65 |
| 8 | 0.66 | 1.35 | 3.29 | 8.98 | 30.06 |
| 16 | 1.36 | 2.72 | 6.63 | 20.64 | 55.00 |
| 32 | 2.73 | 5.47 | 15.90 | 36.20 | 121.07 |

Reproduce a point (after running the correctness suite):

```bash
PYTHONPATH=python:. python benchmark/kernels/csa_attention/bench_attention.py \
  --mode prefill --batch 16 --sequence 8192 --heads 64 --host-buffers
```

## Correctness

All tests passed. Outputs are checked against an independent NumPy reference with `rtol=2e-2`, `atol=1e-2`, and finite-output checks. Coverage includes decode, prefill, ragged inputs, empty/padded requests, page boundaries, cache-format parameters, interpretation mode, read-only caches, and rejection of a deliberately dropped compressed-attention contribution.


KoalaYuFeng:
Hi @zorrofox @JamesBrianD @cjx0709, this PR adds CSA gather and joint attention. Together with our compressor PR and the existing Indexer/Top-K operators, it provides the core components for CSA. Could you please take a look? Thanks!

KoalaYuFeng:
Added SWA cache writeback so subsequent stateful CSA calls can reuse the updated cache. The attention computation is unchanged, and all tests passed.

KoalaYuFeng:
Added a SWA-writeback capacity fix in 2070a118: size the scratch table from the final ring rather than the prefill token count, avoiding SMEM overflow on large prefills. Added a regression test; all joint-attention tests and pre-commit checks passed.

KoalaYuFeng:
Added a packed SparseCore decode schedule and direct SWA row writeback in 3e706d07, preserving the cache ABI. All local tests and pre-commit checks passed. The updated benchmarks include intermediate HBM staging and SWA writeback.

KoalaYuFeng:
Hi @zorrofox @JamesBrianD @cjx0709, this PR adds CSA gather and joint attention. Together with our compressor PR and the existing Indexer/Top-K operators, it provides the core components for CSA. Updated the benchmarks with all 75 decode/prefill/ragged configurations on TPU v6e. Could you please take a look? Thanks!

PR #1665, opened 2026-09-18
perf(gate): keep router gate weights in checkpoint-native bf16

zorrofox:
### Motivation

RouterGate stores its weight in f32 while GLM-family checkpoints ship it as BF16, so the loader upcasts at assign time and every decode step then streams the full f32 copy from HBM (num_layers x hidden x num_experts x 4B; ~470MB/step for a 75-layer, 6144x256 configuration at bs=1). Since the stored f32 values are exact upcasts of BF16, storing BF16 and upcasting at use is bit-identical math at half the read traffic.

### Modifications

Two lines in `layers/gate.py`: the RouterGate kernel param dtype becomes `jnp.bfloat16`, and the router dot upcasts the weight via `astype(jnp.float32)` before the existing `precision=HIGHEST` matmul (XLA fuses the convert into the dot's operand read, so HBM traffic is BF16-wide). Numerics are unchanged by construction: the f32 operand bits are identical to the previous f32-stored values.

### Benchmarks (GLM-5.2-FP8, v7x TP16, 135k context cell, 1024/256 random, mean TPOT)

cc=1: 13.70 -> 13.28ms (-3.1%). cc=8: 36.28 -> 35.57ms (-2.0%). cc=32: 98.43 -> 98.10ms (-0.3%). GSM8K 200q 5-shot: 0.965 vs 0.960 baseline, Invalid 0.

### Test plan

- [x] TPOT cc=1/8/32 (all improved or flat)
- [x] GSM8K 200q paired (in-band)
- [x] pre-commit run --files python/sgl_jax/srt/layers/gate.py


JamesBrianD:
Regarding the router gate's `dtype`: I'm not sure if some models default to `fp32`, whereas our gate doesn't currently allow the parameter `dtype` to be configured. We should probably check this. I'll check qwen/mimo/ling/deepseek/kimi these mainstream models.

zorrofox:
Good catch — I checked the shipped checkpoints for the models you listed (read the safetensors headers directly): gate weight is BF16 in GLM-5.2, DeepSeek-V3, Ling-2.6, Kimi-K3 and Qwen3.5-397B; the one exception is MiMo (v2-flash and v2.5-pro), which ships the gate in F32, so a hardcoded bf16 param would indeed lose bits there. (For reference, e_score_correction_bias is F32 across all of them, and this PR doesn't touch the bias.)

Updated the PR accordingly: the kernel dtype is now a constructor arg `kernel_dtype` defaulting to `jnp.float32`, so every existing model keeps exactly its previous behavior, and GLM-5.2 opts into checkpoint-native bf16 where the benchmark gains were measured. Models whose checkpoints are bf16 (per the list above) can opt in later with a one-line change; MiMo should stay on the f32 default.

JamesBrianD:
I checked the HF configs and all router gate weight headers for these checkpoints, alongside SGLang modeling. Reference summary (SGLang parameter dtype assumes the default BF16 model load):

| Model / HF config reference | Config dtype / router override | Checkpoint gate weight | SGLang gate weight |
| --- | --- | --- | --- |
| Qwen3 30B-A3B / 235B-A22B, Qwen3-Next 80B-A3B, [Qwen3.5 397B-A17B](https://huggingface.co/Qwen/Qwen3.5-397B-A17B/blob/main/config.json) | BF16 / none | BF16 | BF16 |
| MiMo-V2-Flash, MiMo-V2.5, [MiMo-V2.5-Pro](https://huggingface.co/XiaomiMiMo/MiMo-V2.5-Pro/blob/main/config.json) | BF16 / none | **FP32** | **FP32** |
| Ling-2.6 flash / [1T](https://huggingface.co/inclusionAI/Ling-2.6-1T/blob/main/config.json) | BF16 / `router_dtype="fp32"` | BF16 | **FP32** |
| [Ling-3.0-tiny](https://huggingface.co/inclusionAI/Ling-3.0-tiny/blob/main/config.json) | BF16 / `router_dtype="fp32"` | BF16 | BF16 (explicit modeling override) |
| DeepSeek V3 / R1 / V3.2 / [V4-Flash](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash/blob/main/config.json) | BF16 / none | BF16 | BF16 |
| Kimi K2-Instruct / K2.5 / Linear-48B-A3B-Instruct / [K3](https://huggingface.co/moonshotai/Kimi-K3/blob/main/config.json) | BF16 / none | BF16 | BF16 |
| GLM-5.2 / [GLM-5.2-FP8](https://huggingface.co/zai-org/GLM-5.2-FP8/blob/main/config.json) | BF16 / none | BF16 | BF16 |

Modeling references: [Qwen](https://github.com/sgl-project/sglang/blob/6bd1a0af1d9e5756d115dbec6d02e0289c1b6950/python/sglang/srt/models/qwen3_moe.py#L281), [MiMo](https://github.com/sgl-project/sglang/blob/6bd1a0af1d9e5756d115dbec6d02e0289c1b6950/python/sglang/srt/models/mimo_v2.py#L343), [Ling-2.6](https://github.com/sgl-project/sglang/blob/6bd1a0af1d9e5756d115dbec6d02e0289c1b6950/python/sglang/srt/models/bailing_moe_linear.py#L260), [Ling-3](https://github.com/sgl-project/sglang/blob/6bd1a0af1d9e5756d115dbec6d02e0289c1b6950/python/sglang/srt/models/bailing_moe_v3.py#L494), [DeepSeek / Kimi / GLM shared gate](https://github.com/sgl-project/sglang/blob/6bd1a0af1d9e5756d115dbec6d02e0289c1b6950/python/sglang/srt/models/deepseek_v2.py#L458), [Kimi-Linear](https://github.com/sgl-project/sglang/blob/6bd1a0af1d9e5756d115dbec6d02e0289c1b6950/python/sglang/srt/models/kimi_linear.py#L97).

No mixed FP32/BF16 router weight dtypes were found across layers within these checkpoints, including available MTP gates. Bias is separate: Kimi-Linear's correction bias is BF16, while the checked MiMo/Ling/DeepSeek-V3-family/Kimi-K2/K2.5/K3/GLM checkpoints use FP32 correction bias.

Could we make the shared gate's weight-storage, compute, and bias dtypes explicit, then wire each model to the appropriate defaults? Please preserve existing router config semantics (especially Ling's FP32 computation), rather than deriving gate storage solely from global `torch_dtype`/`dtype`. Keep MiMo weights FP32; BF16-checkpoint models can opt into BF16 storage while retaining their existing compute/bias behavior. The remaining model integrations can be a follow-up to this GLM change.

zorrofox:
Done. `GateLogit` now takes the three dtypes as independent constructor args: `kernel_dtype` (weight storage), `compute_dtype` (the router dot, both operands cast before the `HIGHEST` dot) and `bias_dtype` (correction-bias storage). Every default reproduces the previous behaviour exactly: f32 kernel, f32 dot, bias stored in the legacy `weight_dtype` (which only ever governed the bias; the kernel was always f32 before). So Ling's `router_dtype="fp32"` path, MiMo's F32 gate and all other callers are byte-for-byte unchanged, and BF16-checkpoint models can opt into BF16 storage later by passing `kernel_dtype` without touching their compute or bias behaviour.

GLM-5.2 is the only model wired in this PR: `kernel_dtype=bf16` (checkpoint-native), `compute_dtype=f32`, `bias_dtype=f32` (the checkpoint's `e_score_correction_bias` is F32). The layers doc describes the three dtypes. CPU check: a BF16-stored kernel computed in f32 gives bit-identical logits to an f32-stored kernel holding the same values; `pre-commit` passes on the changed files. The per-model wiring from your table (DeepSeek-family, Kimi, Qwen3-MoE) can follow as a separate change once each is benchmarked.

JamesBrianD (approved):

PR #1666, opened 2026-09-18
perf: reduce speculative decoding host metadata overhead

JamesBrianD:
## Motivation

Speculative decoding prepares host-side sampling and attention metadata on every iteration. DP penalty merging creates intermediate dense vocabulary-sized matrices, while overlap relay builds a token-slot-sized cache buffer only to extract page IDs. Repeated page uploads and late result prefetch add further host work.

## Modifications

- Write each DP rank's penalties directly into a fresh merged buffer. Apply min-new-token stop penalties only at active stop-token indices, preserving presence/frequency/min-new-token addition order and batch ownership.
- Build allocated page IDs directly as part of relay batch preparation, automatically for matching FlashAttention backends and page sizes. No separate cache mode or caller-supplied optimization switch is introduced. Preserve token-location preparation for legacy consumers and other attention backends.
- Reuse immutable, batch-owned page uploads across compatible FlashAttention metadata preparations. Invalidate on source, sharding, or padded shape changes; bypass reuse for request-length repacking. Refresh SWA mappings on every preparation.
- Start asynchronous host prefetch for steady-state relay verify outputs before result processing.

This PR does not change which sampling constraints speculative verify applies.

## Accuracy Tests

No new test files are included. Before removing the temporary CPU tests, local validation reported 60 passed and 3 deselected. That historical result includes tests no longer present in this PR. TPU execution and end-to-end model accuracy remain unverified.

After simplifying page preparation, the existing CPU-compatible speculative tests reported **28 passed, 3 deselected**; these do not establish end-to-end equivalence of the changed relay path. MTP module import checks passed.

Static validation: Ruff passed for the changed production files; `git diff --check` passed after removing the test files.

## Benchmarking and Profiling

No TPU throughput or profiling results are claimed for this change. The allocated-page metadata contains one int32 per page instead of one per token slot; actual end-to-end gains still require TPU profiling.

## Checklist


lingebeng (approved):
Thanks a lot. 

Tested on **TPU v7x-8 (TP8/DP1), batch size 8**, using **Qwen3-8B + EAGLE3** with 8K input tokens, 1K requested output tokens, 3 draft steps / 4 draft tokens, and overlap enabled.

The experiment followed a **baseline → PR → baseline replay** sequence on the same TPU allocation. Each launch included two warm-up batches and three measured batches.

| Metric | Baseline | PR | Reduction |
|---|---:|---:|---:|
| Instrumented scheduler CPU time | 6.325 ms/step | 5.837 ms/step | 7.7% |
| Steady-state decode cycle time | 7.885 ms/step | 7.402 ms/step | 6.1% |
| Batch latency, including prefill | 3.683 s | 3.549 s | 3.6% |

The largest improvement came from draft page-table preparation, which decreased from **~0.23 to 0.021 ms/call**. TPU verify and draft execution times remained essentially unchanged. Generated text matched across all nine measured batches.

CPU timings were collected with identical thread-time instrumentation and XProf disabled. They cover the instrumented scheduler stages, not total process CPU usage. These results demonstrate a benefit for the tested configuration; DP-related penalty optimizations were not exercised because DP=1.

PR #1669, opened 2026-09-18
[Spec Decoding][Perf] Remove host synchronization from EAGLE3 decode

lingebeng:
## Summary

Remove a host synchronization point between EAGLE3 verify and recurrent draft dispatch, and fold the sampler-step increment into the existing verify JIT. In the TP8/DP8 overlap profile, the median verify→draft device gap drops from **1.471 ms to 7.69 µs**, and the blocking `Array._value` calls inside draft launch disappear.

## Changes

- **Produce sequence lengths in both consumer layouts inside verify.** The scheduler retains its replicated `P()` output, while recurrent draft consumes a separate `P("data")` output directly. Previously, converting the replicated output with host-side `device_put` could materialize it through `Array._value`, forcing the host to wait for verify before submitting draft.
- **Advance the sampler step inside verify.** Return the updated step instead of dispatching a standalone scalar `jit_add` each decode iteration. The increment-before-`fold_in` RNG ordering is preserved.

This fixes the existing EAGLE3 decode path and does not require the separate configurable DP LM-head feature.

image

## Profile validation

Fresh comparison of upstream base `0c74a5504` against this PR (`9d7794b65`), on the same TPU v7x host with 8 JAX devices. Both variants retain the upstream LM-head layout.

Setup: Qwen3-8B + thoughtworks/Qwen3-8B-Eagle3, TP8 with DP1 or DP8, scheduler overlap enabled, greedy sampling, EAGLE3 steps=5 / draft tokens=6, ISL/OSL=1/4096, concurrency=64. Profiles were captured after warmup.

| Profile metric | TP8/DP1: before → after | TP8/DP8: before → after |
| --- | ---: | ---: |
| Median verify→draft device gap | 17.13 → 17.12 µs | **1471.24 → 7.69 µs** |
| `Array._value` calls inside recurrent draft launch | 0 → 0 | **39 → 0** |
| Standalone sampler-step `jit_add` modules | **39 → 0** | **39 → 0** |

Counts are per captured window; gaps use complete verify/draft pairs. DP1 did not exhibit the layout-conversion host wait in this measurement, while both configurations eliminate the standalone add dispatch. These measurements concern the verify→draft handoff, not complete elimination of all decode-loop gaps or host transfers. Captures do not enforce identical sequence positions across variants.

Validation scope: these profiles demonstrate removal of the targeted host synchronization and standalone dispatch. No end-to-end throughput or output-equivalence claim is made: fixed-input serving checks showed output differences across versions and within same-version repeats, with the cause unresolved.

## Tests

- 16 focused CPU tests passed, covering DP1/8, greedy and stochastic sampling, consecutive RNG steps, output sharding, and device handoff without host materialization.
- Pre-commit checks passed.


python/sgl_jax/test/speculative/test_dp_page_indices.py
```diff
@@ -0,0 +1,31 @@
...
+    np.testing.assert_array_equal(padded[:, :17], pages)
+    np.testing.assert_array_equal(padded[:, 17:], 0)
+
+
+def test_empty_fixed_capacity_dp_page_table():
+    padded = _pad_page_indices(np.array([], dtype=np.int32), 8, 256, dp_size=8)
+    np.testing.assert_array_equal(padded, np.zeros(256, dtype=np.int32))
+
+
+@pytest.mark.parametrize("pages,capacity", [(np.arange(7), 16), (np.arange(8), 17)])
```
JamesBrianD: This new test file not add to test suites. Maybe could delete this.

JamesBrianD (approved):
Others LGTM!

python/sgl_jax/test/speculative/test_dp_page_indices.py
lingebeng: OK, thanks a lot.

JamesBrianD (approved):

PR #1670, opened 2026-09-18
[Feat] Add configurable DP LM-head parallelism

lingebeng:


JamesBrianD (commented):
Could we make LM-head parallelism explicit at model construction, following SGLang's [`ParallelLMHead(use_attn_tp_group=...)`](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/vocab_parallel_embedding.py#L618)?

Currently, modeling still declares the old `kernel_axes`, `WeightLoader` silently overrides `WeightMapping.sharding`, and `configure_lm_heads()` patches the head and `LogitsProcessor` after loading. This creates a hidden dependency on the outer loader: constructing the model and calling its own `load_weights()` does not fully apply the configured parallelism.

I'd suggest:
- Pass the policy explicitly from config to `ParallelLMHead` and `LogitsProcessor` in modeling.
- Let the common LM-head layer define its weight layout and vocabulary padding, while preserving tied-embedding storage semantics.
- Have weight mappings and loaders follow that declaration, rather than discover LM-head modules and override their configuration.

The normal and dummy loader paths are covered today; this suggestion is about making the configuration contract explicit and keeping model behavior independent of the loading path.

JamesBrianD (dismissed):
LGTM

PR #1671, opened 2026-09-18
perf(multimodal): use SHM feature transport and Pillow image decoding

Yanko-7:
## Motivation

Avoid pickling large NumPy multimodal features into tokenizer-to-scheduler ZMQ messages by passing shared-memory references, following SGLang's approach. Restore Pillow image decoding because TorchCodec has a palette PNG + tRNS transparency decoding bug that can corrupt RGB output and potentially write past the allocation ([upstream fix](https://github.com/meta-pytorch/torchcodec/pull/1732), merged). Remove the TorchCodec dependency rather than relying on an affected installed version.

## Modifications

- Wrap CPU features in SHM references in both tokenizer paths; restore owned arrays and release SHM before dispatch or cross-host broadcast.
- Preserve source requests, clean up failed transfers, and fall back to inline pickle when SHM allocation fails.
- Use `PIL.Image.open(...).convert("RGB")` for encoded images; remove the TorchCodec decode switch, optional dependency, and uv source.
- No model, EPD, scheduling, or background-thread changes.

## Validation

Cross-process round trips, exact array comparisons, and allocation/send/receive failure cleanup passed. Formatting, lint, and commit hooks passed.

## Performance

Shared-memory transport microbenchmark (excludes image decoding; no decoding speedup claimed): 3 rounds, 120 measured samples per case after warmup. Synthetic float32 feature arrays; all values are **P50 / P95 in ms**.

| Feature size | Pickle handoff | SHM handoff | Pickle receive | SHM receive |
|---|---:|---:|---:|---:|
| 1 MiB | 1.264 / 4.870 | 0.346 / 0.637 | 0.124 / 0.566 | 0.149 / 0.385 |
| 8 MiB | 8.122 / 10.176 | 1.693 / 4.531 | 0.522 / 0.906 | 0.772 / 2.140 |
| 32 MiB | 33.323 / 47.992 | 5.103 / 7.782 | 1.882 / 2.792 | 2.542 / 4.271 |

Handoff includes producer notification, serialization/SHM allocation and write, IPC delivery, receiver copying and cleanup. Receive time measures only `recv_requests()` after socket readiness.

At 32 MiB, handoff P50 fell **33.323 → 5.103 ms**, while receive P50 rose **1.882 → 2.542 ms**: this change reduces transport latency, not scheduler-thread occupancy. Small-payload results were noisy (1 MiB pickle handoff per-round medians: 0.915–3.368 ms). These are not inference throughput/TTFT measurements; Linux/TPU inference validation remains pending.


python/sgl_jax/srt/managers/mm_utils.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def unwrap_shm_features(request):
+    """Restore features before dispatch or broadcasting to another host."""
+    try:
+        return _map_features(
+            request,
+            lambda value: value.materialize() if isinstance(value, ShmPointerMMData) else value,
+        )
+    finally:
```
lianfang0320, line 136: Would it be worth considering calling discard_shm_features(request) only on failure? Each successful materialize() already unlinks and closes its segment, so the unconditional finally traverses the original request and attempts to reopen segments that have already been unlinked. An except BaseException block with cleanup and a re-raise would preserve failure cleanup while avoiding this work on the success path.

Yanko-7: finally has been changed to Base Exception

python/sgl_jax/srt/managers/mm_utils.py
```diff
@@ -0,0 +1,158 @@
...
+            return segment
+        return value
+
+    try:
+        return _map_features(request, wrap)
+    except BaseException as error:
+        for segment in segments:
+            segment.close_and_unlink()
+        if not isinstance(error, OSError):
+            raise
```
lianfang0320, line 113: LGTM

lianfang0320 (approved):

PR #1674, opened 2026-09-18
feat(layers): add Qwen4Exp QSA indexer

real3stone:
#### Motivation

Next step toward supporting Qwen3.8-Flash-Next (`qwen4_exp`). See #1655 for the full plan; this PR is its P4a and builds on #1662.

Flash-Next's 12 full-attention layers do not attend over the whole context. A lightweight indexer aggregates the sequence into micro-blocks of `compress_ratio` tokens, scores at block granularity, and selects the `indexer_budget / compress_ratio` highest-scoring blocks; the sparse attention kernel then attends over the tokens those blocks expand to. This PR is the selection half. Nothing constructs the layer yet.

#### Modifications

`python/sgl_jax/srt/layers/attention/qsa_indexer.py`.

The scoring is not new code. `kernels/dsa/streamindex_topk` already computes `sum_h relu(q_h . k) * w_h`, and its `compression_ratio` makes entry `e` visible only once `e < (q_pos + 1) // compression_ratio` — exactly QSA's rule. QSA has no `weights_proj`, so all-ones weights reduce one to the other. What is new is the write side — turning raw keys into compressed block keys — and the expansion of selected block ids back into token ids.

Compression takes the packed `[T, D]` stream rather than one request at a time, because a forward pass hands the backend requests of different lengths and `vmap` needs them equal. For every token it materialises the `compress_ratio` keys ending at that token, reading this step's own keys while the index is still inside the request and the request's ring once it is not. That one window serves both outputs — a group's compressed key is the window of the token that closes it, and a request's new ring is the window of its last token masked to the tokens still open — for `compress_ratio` static gathers, with no segment reduce and no dynamic loop.

Four orderings are silent if changed. Upstream SGLang's `layers/attention/qsa/qsa_indexer.py` and vLLM's `qwen4_exp` agree on all four:

- compression is pool -> norm -> RoPE, not norm -> pool;
- the pool is an fp32 mean over the whole group;
- the compressed key is rotated at its group's **first** position. Rotating each raw key at its own position and then averaging would average four different angles and lose the position;
- the norm is `GemmaRMSNorm`, `x * (1 + w)`; a plain RMSNorm is off by the unit offset.

The rotation position is derived as `group * compress_ratio` rather than read back from the tokens, which is why multimodal RoPE is out of reach here: an mrope position is three numbers and cannot be recovered from an index. Upstream carries a parallel position buffer for exactly that.

The expansion is `budget + compress_ratio - 1` wide plus a trailing count column. The extra slots are not padding: they hold the causal tail of the open group — tokens that have not completed a group, so they are absent from the compressed cache and never compete in the top-k, yet are causally visible and still have to be attended. The count column is the sparse attention kernel's tile-loop bound, never a token id.

`python/sgl_jax/srt/kernels/dsa/ref.py`: `streamindex_topk_ref` gains the same `compression_ratio`. It had none, so the kernel's compressed path — which no caller in this repo exercises, GLM passing `compression_ratio=1` — had no CPU reference to check against at all. `compression_ratio=1` is the default and reproduces the previous behaviour exactly.

#### Accuracy Tests

No model outputs change — nothing constructs a `QSAIndexer` yet.

```bash
python3 python/sgl_jax/test/layers/test_qsa_indexer.py       # registered in unit-test-cpu, 14 tests
python3 -m pytest test/srt/kernels/dsa/test_ref.py -q        # ratio=1 regression, 13 passed
```

Compression is checked against an fp64 oracle that rebuilds each request's logical key run — the live rows of its ring followed by this step's keys — and indexes that run by absolute position, so it reaches the answer by a different route than the implementation's backwards walk. The batch it runs on mixes group-aligned and mid-group starts and includes a request holding less than a full group, so most of the group that request closes has to come out of the ring.

Two more cases exist for traps rather than coverage: compression across a step boundary (six tokens fed as 3 + 3 must equal six at once, the bug the ring exists to prevent), and the causal tail at `pos = 10` versus `pos = 11`, where closing the group moves it from the tail into the top-k.

The compressed branch of `streamindex_topk` has had no caller until now — GLM passes `compression_ratio=1` — so it is checked on v6e against exact fp32 scoring + argsort at QSA's shape (4 heads, head_dim 128, ratio 4, ones weights). A from-scratch prefill matches exactly; a 24-query extend over a 2036-token sequence reaches recall 0.997 with no entry past the causal bound, the same `>= 0.99` bar `test_streamindex_topk.py` applies to its own bf16 parity test.

#### Benchmarking and Profiling

Not applicable — no runtime path changes for any currently supported model.

#### Checklist


real3stone:
Closing in favour of a single PR covering all of P4 — indexer, kernel, compressed cache, backend.

The indexer's only non-test consumer is the QSA backend, so on its own this adds no reachable code, and the invariants it has to uphold — the compressed cache's page geometry, and the visibility rule the kernel's tail step completes — live in the other half. #1478 landed DSA as one PR with the same shape.

